PR #94, opened 2019-04-11
wake: support [a,b,c] list syntax

terpstra:
```
[] = Nil
[a] = (a, Nil)
[a,b,c] = (a, b, c, Nil)
[[4],[]] = ((4, Nil), Nil, Nil)
```

Do we want this? I don't think it helps people who already know wake, but every time someone first learns wake I hear complaints about ```Nil```. If having dedicated syntax for lists helps people get started, maybe it's worth the added complexity.


mmjconolly:
>     1. It muddies the water by disguising the fact that "a, b, c" means prepend a to b to c.
I don't feel that a new person would think that unless they already did a lot of functional programming. Until yesterday I had thought that adding the suffix `, Nil` was the way to construct a list, and `++ (item, Nil)` the only way to append.

If you find the time, I do feel that `[ ]` is a great readability improvement

sistillson:
I agree with Mathew. It's well understood by most people who have studied programming languages that a list is usually implemented as syntactic sugar for cons pairs i.e
[1,2,3] => (cons 1 (cons 2 (cons 3 Nil)))

and, although 1,2,3,Nil is easier to write than (cons 1 (cons 2 (cons 3 Nil))) it's even easier
tow write [1,2,3] and you would stop having to wrap all your lists with () to make sure
everything bound correctly.

terpstra:
Just an update on this PR. I am leaning heavily towards making unary , a postfix operator. Furthermore, I want to make all bijective functions available in pattern matches.

Therefore, it will soon be possible to:
```
export def x, = x, Nil
```

Thereafter, people can construct and match lists like this:
```
def mylist = 1, 2, 3,
def somecall = somefn (1, 2, 3,)
def oneElementList = match _
  a,  = Some a
   _ = None
```

I realize this is not quite the syntax people were hoping for (it requires a trailing comma), but I think it will be a better holistic fit for the wake language.

mmjconolly:
Wanted to circle around on this and see if there was any way we could get a non `, Nil` list syntax

terpstra:
I'm closing this as it's super old and I'm not happy with it.
If we add a `[a, b, c]` syntax, I'd prefer it to be for `Vector` instead.
Also, we can now write `(a, b, c,)` for a List, anyway.

PR #113, opened 2019-05-04
Update MacOS with Homebrew instructions

jackkoenig:
We need like a clean laptop we can wipe and check install instructions


terpstra:
Did you try this on a clean system?

jackkoenig:
I tried it on a Berkeley colleague’s laptop. Now that we’re open source we could add Travis build regressions that could check this.

PR #114, opened 2019-05-06
fuse.cpp: relicense under Apache 2

terpstra:
With permission (via private email) of the authors, Miklos Szeredi and
Sebastian Pipping, we've agreed to relicense this file from GPL to Apache 2.


terpstra:
@hartwork @szmi Here's the PR. It seems I cannot set you guys as a reviewer, so just linking you via comment!

hartwork:
I approve :smiley:

PR #120, opened 2019-05-08
use execvp instead of execv

albertchen-sifive:
so that we don't need to absolute paths to executables in our cmdlines.


terpstra:
The original rationale was: use which in wake so that if the selected executable ever changes, wake will rerun the job.

However, now that wake runners might run the job remotely to gain access to differently installed tools, it becomes inappropriate to select the tool using only local information.

PR #126, opened 2019-05-09
Beginnings of a quick reference guide

sistillson:
Please comment, and add.


terpstra:
It would be nice to include this in v0.15. But I guess everyone is too busy to finish it at this point?

PR #127, opened 2019-05-12
No exceptions

### LINKED ISSUE #87: Exceptions from C++ could use more context
When functions like 'mkdir' or 'regexp' fail, they should include what was the argument in the exception string. @albertchen-sifive

terpstra:
This PR completely removes Exceptions from the wake language.

Instead, we use the type 'Result' for methods which might fail and include an exceptional case in the 'Path' type. This makes the common case of chaining jobs together behave like it did before, but makes it easy to write much more robust code in general.

Fixes #87.


terpstra:
The most invasive parts of this change are two-fold:
- Regular Expressions must now use back-tics
- A few less-used system calls now return a Result

terpstra:
It's somewhat surprising to me that this PR only removes 88 lines despite modifying over a thousand. I guess the cost of handling failures follows the law of conservation of complexity.

PR #131, opened 2019-05-16
JSON based runners

terpstra:
This PR switches FUSE to use the JSON API.
It also adds an alternative implementation using a preload shim for use in docker.

The changes are fairly enormous mostly due to moving code from wake to common and attenuate refactoring work.


terpstra:
I'm going to merge this PR, even though it breaks control-C of builds on mac OS. I'm going to try another refactor where the FUSE mount is shared between invocations to improve reliability.

PR #132, opened 2019-05-16
Add CircleCI build rules for several OSes.

richardxia:
I set up some CircleCI build rules so that every PR causes a build of wake on several OSes. This is my first time using CircleCI, but I have to say, this is a relatively nice experience compared to Travis CI because of the first-class Docker support.

I have builds running and passing on:

- CentOS 7.6 (I tried building on CentOS 6.x, but gave up after having to work around several oddities/bugs and finally after learning that CentOS 6.8's g++ does not have support for C++ 11)
- Debian 8 (Jessie)
- Debian 9 (Stretch)
- Ubuntu 16.04 (Xenial)
- Ubuntu 18.04 (Bionic)

I did check that this would have caught the bug fixed by https://github.com/sifive/wake/pull/128.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,65 @@
+# We pick an arbitrary docker image provided by CircleCI, since all we want is a
+# Docker image that has the Docker client installed within it, and CircleCI
+# preinstalls the Docker client all on their images. We don't need any actual
+# language tools since the entire build happens within Docker.
+version: 2.1
+jobs:
+  centos_7_0:
+    docker:
+      - image: circleci/buildpack-deps:jessie
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run: |
+          docker build -f .circleci/dockerfiles/Dockerfile.centos_7_0 .
+
+  centos_7_6:
+    docker:
+      - image: circleci/buildpack-deps:jessie
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run: |
+          docker build -f .circleci/dockerfiles/Dockerfile.centos_7_6 .
+
+  debian_10:
+    docker:
+      - image: circleci/buildpack-deps:jessie
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run: |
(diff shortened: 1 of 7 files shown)


terpstra:
Could we swap Debian/jessie for Debian/wheezy and Debian/stretch for Debian/testing ?

Support for wake in wheezy is the main issue I keep tripping over.

terpstra:
I have back-ported libre2-dev and libre2 package for wheezy if that helps.

richardxia:
I tried Wheezy but it looks like the apt package repository isn't maintained anymore, since that version of Debian is way past its EOL, so when I ran `apt-get update` it failed. Do you have a pointer to a package repository mirror somewhere?

terpstra:
deb http://archive.debian.org/debian wheezy main

terpstra:
Does CentOS 7.0 work? It was released in June 2014 vs. 7.6 in October 2018.

The lowest threshold I am aiming to support is RHEL 7.0 and debian/wheezy.

richardxia:
I've force-pushed some of my changes, but it's not in a working state.

Things that I have change (or decided not to change):
- Debian 8 (Jessie) replaced with Debian 7 (Wheezy)
  - This is broken because I need the backported re2 package. Do you have it on a public apt repository somewhere?
- Debian 9 (Stretch) replaced with Debian 10 preview (Buster)
  - I chose not to target `testing` because it could be kind of annoying from a CI perspective to have something fail because something changed upstream. I'm hoping that targeting a fixed future version would be all right as a close approximate.
- Adding CentOS 7.0, but it's failing due to an RPM checksum issue. I ran into this earlier with trying to get CentOS 6.6 running, and it's related to CentOS not understanding the overlay filesystem that Docker uses. There's a workaround but I have to put down this work for several hours, and I figured I could ask you about your Debian Wheezy re2 backport.

terpstra:
I build my package from https://packages.debian.org/source/jessie/re2

richardxia:
OK, I'm giving up on Debian 7 (wheezy) for now, since compiling `re2` from source as part of the CI job 1) takes like 30 minutes and 2) times out on CircleCI since it the tests take so long that it triggers the 10-minute "no output seen" threshold.

The right way of doing this is building a backported .deb package somewhere else, uploading the .deb somewhere, and just downloading and installing it in the Debian 7 Docker CI build.

PR #133, opened 2019-05-21
Add functions for turning List of Option|Result to Option|Result of List

jackkoenig:
h/t @terpstra for most of the work.

The motivating example was to turn `List (Result a b)` -> `Result (List a) b`.

This can be done with `findFail` which finds the first `Fail` in a `List` or returns `Pass (List a)`. There are related functions `findFailFn`, `findPass`, and `findPassFn`.

We already had `findSome` which is similar but for `List (Option a)`, so we also generalized that into `findSome`, `findSomeFn`, `findNone`, and `findNoneFn`.


jackkoenig:
@terpstra Done, although I didn't add examples

Issue #98, opened 2019-04-17
If you're in a fuse directory when the job ends it doesn't get deleted


jackkoenig:
Workaround:
```
$ sudo diskutil unmount [dir]
```

terpstra:
This seems impossible for wake to solve. Either the fuse runner (and by extension wake) would have to hang indefinitely, or the directory does not get unmounted (current behaviour). Both are awful. I marginally prefer the current behaviour since sending SIGKILL to the mount at least solves this sometimes on linux with lazy unmount semantics.

terpstra:
Since the mount is now managed by a daemon, it can safely outlive the build.

PR #134, opened 2019-05-22
FUSE: use a daemon

terpstra:
This PR makes all FUSE runners share a common mount point.

The advantages to this approach:
- more reliable (no need to carefully coordinate subprocess kill on every job)
- faster (less mounting and unmounting)
- it fixes osxfuse builds faced with Control-C


terpstra:
Fixes #98

PR #145, opened 2019-05-25
Debian packaging

terpstra:
Improve wake with a release methodology.
Add debian package files to the repository.
Use CI to build debian packages on PRs.


debian/README
```diff
@@ -0,0 +1,10 @@
+The Debian Package wake
+----------------------------
+
+Wake is a build orchestration tool and language.
+If you have a build whose steps cannot be adequately expressed in
+make/tup/basel/etc, you probably need wake.
```
richardxia: It's spelled "bazel", since it's an anagram of Google's internal blaze tool.

terpstra:
That's embarrassing. I think I was cooking when I wrote that part of the
README. :P Thanks.

PR #146, opened 2019-05-28
Generate and publish release tarball and Debian/Ubuntu/RedHat packages

terpstra:
When finished, this PR will make the CI system generate the images we need to make a release.


richardxia:
By the way, I took a look at the [Circle CI documentation on build artifacts](https://circleci.com/docs/2.0/artifacts/), and they explicitly mention:

> Artifacts are designed to be useful around the time of the build. It is best practice not to rely on artifacts as a software distribution mechanism with long term future guarantees.

We'll need to come up with a different system for actually hosting the built releases longer term, but it's at least good to get this automation up and running.

PR #148, opened 2019-05-30
a .ctags file to create tags file for wake

sistillson:
Not perfect, or minimal, but effective.

### DIFF
--- a/doc/ctags
+++ b/doc/ctags
@@ -1,0 +1,13 @@
+# move this file to ~/.ctags
+# then, remove the comment lines
+# Then ctags *.wake will creat a tags file for your wake files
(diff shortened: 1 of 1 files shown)


terpstra:
Sorry for my ignorance, but what does this do?

sistillson:
It creates a file that vi can read, tags, that you can use to jump to definitions.

terpstra:
What does this do for:
```
def x +++ y = x + y
```

It looks like it would consider this a definition of 'x', not '+++'?

sistillson:
That is correct. it's far from perfect. I wasn't sure

sistillson:
i'm trying to figure it out right now

PR #151, opened 2019-05-31
Update tutorial to v0.15

jackkoenig:
I think more is necessary, but this updates things to work in v0.15 including some additional explanation.

I think a new section at the end on how to deal with failure (ie. `Result` and `BadPath`) would be another good thing to have, I'll see about writing one.


doc/tutorial.md
```diff
@@ -435,31 +482,31 @@ To make inline functions even easier to define, wake also supports a syntax
...
-Nevertheless, sometimes this syntax can be convenient, too.
+Nevertheless, this syntax can be convenient.
 
 ## Downloading and parsing files
 
-    cat >>tutorial.wake <<EOF
+    cat >>tutorial.wake <<'EOF'
     def curl url =
-      def file = simplify "{here}/{head (extract '.*/(.*)' url)}"
+      def file = simplify "{here}/{extract `.*/(.*)` url | head | getOrElse "file.txt"}"
```
terpstra: A better fix would be ```{replace `.*/` '' url}``` -- no failure possible.

doc/tutorial.md
```diff
@@ -470,10 +517,16 @@ curl and then grabs all the mathematical symbols from the table, formats
...
-the `simplify` function which transforms paths into canonical form. In this,
+The `curl` function also makes use of `extract` with a regular expression to
+split the filename out of the URL. `extract` returns a `List String` containing
+each occurrence of `()` in the regular expression. We then take the `head` of
+the `List` which returns an `Option` since a `List` may be empty. `Options` can
+either be `None`, which means no value is available, or `Some x`, which is a
+value `x`. `Option` is a data type like any other in wake.  It is particularly
+useful in those situations where one would use a null pointer in a language
+like C. `getOrElse` returns the value in a `Some`, or if the `Option` is
+`None`, returns the provided argument, in this case `"file.txt"`.  We also use
```
terpstra: If you follow my suggested code fix, this text will need to change.

richardxia:
> I hate how your editor has unnecessarily touched lines by rewrapping the text.
> This is why I split text so sentences fragments are always at the start of a line.

This practice is also given the name of [semantic linefeeds/line breaks](https://rhodesmill.org/brandon/2012/one-sentence-per-line/), which might be a good practice for us to adopt universally at SiFive.

jackkoenig:
@terpstra I've incorporated your feedback and attempted to restore many of the line breaks, the diff looks much cleaner now. I agree with we should generally avoid rewrapping things for the sake of it, I just messed up while editing some of it.

PR #160, opened 2019-06-02
Target syntax

### LINKED ISSUE #124: Memoize is broken. Let's fix it.
Here's a straw-man:

```scala
def foo x y =
  memoize a b c ignore d e f
  <code with x+y+a+b+c+d+e+f in scope>

def box = foo 4 5 # creates a memoize map
def box2 = foo 5 6 # another independent memoize map

def v1 = box 1 2 3 4 5 6 # creates and saves key (1,2,3) into box
def v1again = box 1 2 3 7 8 9 # returns same result as v1.

def v2 = box 9 8 7 6 5 4 # create and saves key (9,8,7) into box

def v1box2 = box2 1 2 3 4 5 6 # creates and saves key (1,2,3) into box2
```

terpstra:
This PR adds the new 'target myfun x y = ...' syntax which will ultimately replace 'memoize'.
For now, the old memoize keyword will live in parallel to the new style.

Fixes #124.

Examples:

```
# Runs in linear time
global target fib z = match z
  0 = 0
  1 = 1
  x = fib (x-1) + fib (x-2)
```

```
# Runs in exponential time
global def fib z = match z
  0 = 0
  1 = 1
  x = fib (x-1) + fib (x-2)
```

```
# ALSO runs in exponential time (the '_' is not part of the memoized key)
global target fib = match _
  0 = 0
  1 = 1
  x = fib (x-1) + fib (x-2)
```

The implementation is value- and type-safe, ... up until you use the ```\``` feature:

```
global target hax Unit \ y = y
global def foo = hax Unit 55
global def bar = hax Unit "asd"
```

This will result in:
```
ERROR (this will be fatal in 0.16.0): Target subkey mismatch for 'hax Unit \ y' (build.wake:85:[15-26])
  from build.wake:85:[15-26]
  from build.wake:85:[15-26]
  from build.wake:86:[23-33]
  from build.wake:86:[12-19]
  from share/wake/lib/core/tuple.wake:17:[13-16]
  from share/wake/lib/core/tuple.wake:17:[13-16]
  from build.wake:86:[12-19]
```


terpstra:
Aside from making 'target' a keyword, this is a backwards compatible change. I am considering including it in 0.15.0, so that we can start migrating away from memoize. Then I remove memoize in 0.16.0.

jackkoenig:
> The implementation is value- and type-safe, ... up until you use the `\` feature:

In https://github.com/sifive/wake/issues/124#issuecomment-495511769 we discussed not ignoring things to the right of `\`, but rather checking them for correctness. Did you change your mind?

terpstra:
You're right. I just forgot. I am concerned that this could trigger spuriously, though, so I think we should make it just a warning for now.

jackkoenig:
I'm fine with a warning for now, syntax looks good!

PR #161, opened 2019-06-03
Add tap (aka Kestrel combinator) for debugging pipe chains

jackkoenig:
Requested by @richardxia

This allows you to insert debugging into chains of pipes:

eg.
```
global def tapExample =
  makePlan (which "echo", "$FOO", Nil) Nil
  | tap (getPlanEnvironment _ | (println "tap1: {format _}"))
  | editPlanEnvironment ("FOO=hi", _)
  | tap (getPlanEnvironment _ | (println "tap2: {format _}"))
  | runJob
  | getJobStdout
```

### DIFF
--- a/share/wake/lib/core/syntax.wake
+++ b/share/wake/lib/core/syntax.wake
@@ -23,3 +23,7 @@ global def argument | pipeFn = pipeFn argument # bash-style pipe operator
 global def dollarFn $ argument = dollarFn argument
 global def f ∘ g = \x f (g x)
 global def flip f x y = f y x
+# Insert a debugging function
+global def tap tapFn argument =
+  def _ = tapFn argument
+  argument


jackkoenig:
> Neat, I didn't know this thing had an FP name.

Me neither, but I did some Googling to see if Scala had an equivalent and the name came up 🙂

terpstra:
Since the most common use for this is to println a formatted string, why don't we include that?
```
global def tapFn tag fn x =
  val _ = printLevel logWarn "{tag}: {format (fn x)}\n"
  x
global def tap tag x = tapFn tag (_) x
```

... or alternatively, maybe do nothing since this is such a simple function that the person doing the debugging should write it on the spot.

Dunno.

hcook:
+1 this

Writing it on the spot is annoying

hcook:
writing tap by hand again

terpstra:
How about:
```
global def tapExample =
  makePlan (which "echo", "$FOO", Nil) Nil                                                                                                                                        
  |? "tap1: {format _}"
  | editPlanEnvironment ("FOO=hi", _)
  |! "tap2: {getPlanEnvironment _ | format}"
  | runJob
  | getJobStdout

global def x |? formatFn =
  def _ = printlnLevel logDebug (formatFn x)
  x
global def x |! formatFn =
  def _ = printlnLevel logWarning (formatFn x)
  x
```

terpstra:
It's somewhat irksome that you need the ()s around the ""s.

jackkoenig:
I think a special library function for logging makes sense, and one thing I like about `!` is that it implies side effects to me. I do think there is still value in just `tap` since there are other side-effecting things you may want to do, like write a file.

PR #162, opened 2019-06-03
Add tutorial section for dealing with failure

jackkoenig:
I think more is necessary, and I think a section on JSON to provide a more complete example is still needed but I wanted to PR this to get feedback rather than just continuing to add to it.

### DIFF
--- a/share/doc/wake/tutorial.md
+++ b/share/doc/wake/tutorial.md
@@ -567,6 +567,8 @@ Previously, we glossed over functions like `getOrElse` and `getWhenFail`.
 In this section we will explain those as well as the more general concept
 of dealing with operations that can fail in a build flow.
 
+### Option
+
 Many operations can fail like taking the first element (`head`) of a `List`
 (it may be empty) or reading a file (it may not exist).
 The simplest and most common way we deal with this in wake is with `Option` as we
@@ -583,6 +585,8 @@ A more concise way of accomplishing the same goal is to use `getOrElse` which wi
 return the value in a `Some` or a default if the `Option` is `None`.
 
     def myFunction l = head l | getOrElse 0
+
+### Result
 
 `Options` are a great way to deal with things that can fail in one obvious way
 (eg. `head` fails when the `List` is empty), but sometimes an operation may have
@@ -633,6 +637,8 @@ We can rewrite the above as simply:
 Or perhaps more clearly using `|`:
 
     global def safeRead filename = source filename | read | getWhenFail "Read failed!"
+
+### BadPath
 
 Besides `Option` and `Result`, there is a 3rd important piece to handling failure: `BadPath`.
 Recall that `Path` is the type used to represent paths on the file system.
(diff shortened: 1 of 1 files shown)


share/doc/wake/tutorial.md
```diff
@@ -560,3 +560,129 @@ As a final note, wake files must be UTF-8 encoded.
...
 identifiers, and even operators.
 In fact, all the operators we just printed are legal in wake.
+
+## Dealing with failure
+
+Previously, we glossed over functions like `getOrElse` and `getWhenFail`.
+In this section we will explain those as well as the more general concept
+of dealing with operations that can fail in a build flow.
+
+Many operations can fail like taking the first element (`head`) of a `List`
```
richardxia: Is there a different word than "fail" that we can use here? I'm a bit worried that the word "fail" may be conflated with the `Fail` from `Result` rather than the more colloquial "a function may not perform desired action". I think people who understand FP will "get it", but people who are totally new may have more trouble grasping why `Option`s are useful and that we have both `Option` and `Result` in `wake`.

Even if it sounds awkward, I think even "unsuccessful" or "not succeed" may be a better word to use here because there's no name conflict with a capital F Fail.

share/doc/wake/tutorial.md
```diff
@@ -560,3 +560,129 @@ As a final note, wake files must be UTF-8 encoded.
...
+(it may be empty) or reading a file (it may not exist).
+The simplest and most common way we deal with this in wake is with `Option` as we
+have seen before.
+
+    def myFunction l = match (head l)
```
richardxia: Perhaps it would help with comprehension for some of these code examples to have comments with the type signature of the function? For example, this one could be `# List Int => Int`

share/doc/wake/tutorial.md
jackkoenig: I'm open to this, but saying stuff like 
"Many operations may not succeed like taking the first element (`head`) of a `List`" seems less clear to me. I agree with the idea of avoiding the name aliasing though

share/doc/wake/tutorial.md
jackkoenig: This seems like the kind of thing that should be consistent throughout the tutorial, but it's not clear to me where it should be introduced. For example, early on we have:
```
    echo 'int helper() { return 42; }' > help.cpp
    git add help.cpp
    cat > tutorial.wake <<'EOF'
    def variant = "native-cpp11-release"
    global def build _ =
      def main = compileC variant ("-I.", Nil) Nil (source "main.cpp")
      def help = compileC variant ("-I.", Nil) Nil (source "help.cpp")
      linkO variant ("-lm", Nil) (main, help, Nil) "tutorial"
    EOF
    wake 'build 0'
```

But I think pointing out that linkO returns `Path` may not be that helpful this early...

jackkoenig:
@richardxia I've added subsections, I'm thinking about merging and leaving your other comments as future improvements. I don't disagree, I'm just not sure of the right way or amount of work to put into addressing them

richardxia:
> @richardxia I've added subsections, I'm thinking about merging and leaving your other comments as future improvements. I don't disagree, I'm just not sure of the right way or amount of work to put into addressing them

Yep, no problem, and I agree with the "not successful" wording being awkward; I just wanted to put down something as an example so that I didn't just dump the hard wordsmithing to you. Totally cool with merging this as is!

PR #165, opened 2019-06-04
Precedence

terpstra:
Add an operator precedence table to the quick ref.


sistillson:
docs look ok.  A lot of unicode, but ok.

PR #176, opened 2019-06-07
Change getPathResult to return Result Path Error

jackkoenig:
It originally returned Result String Error which is problematic for
propagating Paths

Weirdly,  @john-sifive doesn't show up for assigning reviewers

### DIFF
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -17,10 +17,10 @@
 global def read path =
   def imp p = prim "read"
   match (getPathResult path)
-    Pass f = match (imp f)
+    Pass f = match (imp f.getPathName)
       Pass body = Pass body
       Fail f = Fail (makeError f)
-    x = x
+    Fail e = Fail e
 
 # Simplify away the ..s and .s in a String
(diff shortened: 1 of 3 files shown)


share/wake/lib/system/path.wake
```diff
@@ -17,10 +17,10 @@
 global def read path =
   def imp p = prim "read"
   match (getPathResult path)
-    Pass f = match (imp f)
+    Pass f = match (imp f.getPathName)
       Pass body = Pass body
       Fail f = Fail (makeError f)
-    x = x
+    Fail e = Fail e
```
jackkoenig: This one was fun, I figured it out quickly but adding the `.getPathName` had no effect on the reported type error which was a bit surprising!

terpstra:
I already have getPathError. Is that not appropriate in the case where you are using this modified getPathResult?

jackkoenig:
@john-sifive needs `Result Path Error`, I guess he can do `(\x getPathError x | getOrPass x)`

jackkoenig:
That's the spirit!

PR #187, opened 2019-06-12
Fix target loops

terpstra:
There are some corner cases where a target can be captured by it's own output. This PR fixes those we can and detects and reports those ew can't.


terpstra:
This is the wrong solution. I am going to just have to bite the bullet and implement tracing garbage collection. That will improve performance, anyway.

PR #196, opened 2019-06-14
wake.spec: create a relocatable RPM

terpstra:
That was easy.


jackkoenig:
so now this works with `rpm --prefix`?

terpstra:
Yes. Although not in 0.15.1, obviously.

PR #202, opened 2019-06-25
C++ sources can also end in .c++ extension

nategraff-sifive:
compileC expects C++ sources to be named `.cpp`, but .`c++` is also a recognized extension


terpstra:
I would like to eventually replace `compileC` entirely. This PR does not go far enough; .cc and .cxx are also common extensions.

PR #211, opened 2019-07-03
Emit type-annotated and cross-referenced HTML

terpstra:
When run with `--html`, wake now creates source code reference guide.

Getting all the necessary information to where it is needed required refactoring a few of the intermediate forms.


terpstra:
@drom I've de-npm'd the javascript from wake-html and plopped it into this repository. `wake --html` now generates completely self-contained html.

terpstra:
Merging this to start using CI to auto-build documentation. Work remains to improve pub/sub linkage and types of overly general match expressions.

PR #213, opened 2019-07-11
init port to FreeBSD

sistillson:
This is what you need to work on freebsd. still rough, not sure if fuse is working....

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,27 @@
 
 VERSION	:= $(shell if test -f manifest.wake; then sed -n "/publish releaseAs/ s/^[^']*'\([^']*\)'.*/\1/p" manifest.wake; else git describe --tags --dirty; fi)
 
+ifeq ($(shell uname -s),FreeBSD)
+CC	:= cc -std=c99
+CXX	:= c++ -std=c++11
+else
 CC	:= gcc -std=c99
 CXX	:= g++ -std=c++11
+endif
 CFLAGS	:= -Wall -O2 -flto -DVERSION=$(VERSION)
-LDFLAGS	:=
 
+ifeq ($(shell uname -s),FreeBSD)
+LDFLAGS	:= -L/usr/local/lib
+else
+LDFLAGS	:= 
+endif
+
+ifeq ($(shell uname -s),FreeBSD)
+LOCAL_CFLAGS :=	-Iutf8proc -Igopt -Icommon -I/usr/local/include
+else
 LOCAL_CFLAGS :=	-Iutf8proc -Igopt -Icommon
+endif
+
 FUSE_CFLAGS  :=	$(shell pkg-config --silence-errors --cflags fuse)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -2,12 +2,27 @@
...
+CXX	:= c++ -std=c++11
+else
 CC	:= gcc -std=c99
 CXX	:= g++ -std=c++11
+endif
 CFLAGS	:= -Wall -O2 -flto -DVERSION=$(VERSION)
-LDFLAGS	:=
 
+ifeq ($(shell uname -s),FreeBSD)
+LDFLAGS	:= -L/usr/local/lib
```
terpstra: Can you tell me which library was not found without this -L/usr/local/lib?

Makefile
```diff
@@ -2,12 +2,27 @@
...
+else
+LDFLAGS	:= 
+endif
+
+ifeq ($(shell uname -s),FreeBSD)
+LOCAL_CFLAGS :=	-Iutf8proc -Igopt -Icommon -I/usr/local/include
```
terpstra: Similarly, which headers were not found?

Makefile
```diff
@@ -20,13 +35,22 @@ CORE_LDFLAGS :=	$(shell pkg-config --silence-errors --libs sqlite3 || echo -lsql
 		$(shell pkg-config --silence-errors --libs ncurses tinfo || pkg-config --silence-errors --libs ncurses || echo -lncurses)
 
 COMMON := common/jlexer.o $(patsubst %.cpp,%.o,$(wildcard common/*.cpp))
+ifeq ($(shell uname -s),FreeBSD)
+WAKE_ENV := WAKE_PATH=$(shell dirname $(shell which $(firstword $(CC)))):/usr/local/bin
```
terpstra: ... and which programs/executables?

fuse/fuse.cpp
```diff
@@ -73,7 +73,7 @@ int main(int argc, char *argv[])
 		pid_t pid = fork();
 		if (pid == 0) {
 			ofs.close();
-			execl(daemon.c_str(), "fuse-waked", mpath.c_str(), 0);
+			execl(daemon.c_str(), "fuse-waked", mpath.c_str(), (void *)0);
```
terpstra: What was the error that required (void *)?

sistillson:
I think this version is more to show that it can easily be done and that it can be done without gcc. I wish it were straightforward to get rid of gmake (I always feel it's best to use as little gnu code as possible on a BSD system)

As to a system to use, I just use virtual box and download an image from https://www.freebsd.org/where.html (just one of the regular install dvd isos) and go from there.

I did the build on 12.0, but 11.3 should be pretty similar.

as root (or install sudo)
pkg install gmake pkgconf re2 sqlite fusefs-libs

If you aren't used to freebsd, you will probably be very frustrated. It's not so hand-holdy as linux but it's super consistent.

terpstra:
I've merged an update for 12.0.

Issue #106, opened 2019-05-01
Infinite recursion across memoize returns a Future


jackkoenig:
As the title states. It results in a future sneaking through into user code.

terpstra:
Invoking
```
global target foo Unit = foo Unit
```
now hangs as it should. If you use '-v' it will give you some warning as well.

terpstra:
Fix was bad. Issue reopened.

terpstra:
```
global target bad Unit = bad Unit
```

When invoked now reports:
```
terpstra@pumpkin:~/wake$ ./bin/wake bad Unit
<future>
Infinite recursion detected across 'bad Unit' (build.wake:90:[15-22])
```

PR #219, opened 2019-07-16
Undo loop detect

terpstra:
The cure was worse than the disease.

Reopen #106.

### DIFF
--- a/src/job.cpp
+++ b/src/job.cpp
@@ -767,33 +767,6 @@ bool JobTable::wait(WorkQueue &queue) {
   return compute;
 }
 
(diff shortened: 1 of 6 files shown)


terpstra:
This approach is too fragile. It caused wake to appear to hang even when there was no recursive target loop.

There are several ways we can end up with a cycle in memory due to 'target' allowing mutation. I am going to have to switch to a tracing collector to solve those. This will have the side benefit of making larger DefBindings possible. It will also mean that a recursive target loop will get collected by the GC! So the fix we will need instead is to check the Target table upon destruction for future values.

Issue #184, opened 2019-06-12
Wake leaks memory when a target captures an input function in its output


terpstra:
Targets with no functions in either input or output are fine.
```
target foo x y = x + y
```

Targets which take a function and return a simple type are also fine.
```
target foo x \ fn  = fn x + 1
```

Targets which return a function defined inside the target are fine.
```
target foo x = \y x + y
target bar x = scmp x # curried
```

Targets that capture an input function in their output function are **NOT** fine.
```
target foo x \ fn = \y fn (x + y)
target bar x \ fn = fn, Nil
```

The problem here is that `target` internally builds a table from input hashes to output values. A function passed to a `target` may be created later than the `target` and may thus have a reference to it. If the function is then captured by the value entry in the `target`, there is now a cycle in memory.

Like GHC, wake without `target` is a pure language where memory cycles are impossible. This allows for simplified garbage collection. Unfortunately, to fix this bug, wake will either need to forbid complex `target` return types (ie: those containing functions) or implement tracing collection.

For now, I choose to leak memory rather than make a decision.

terpstra:
I bit the bullet and implemented a tracing collector. Valgrind confirms there are no more leaks.

PR #220, opened 2019-07-18
C++ implementation of a moving collector

terpstra:
This should be pretty fast, since all allocations consist of a
simple comparison and increment. Collection itself dispatches
via C++ virtual methods, also fairly fast. The collector itself
is a single-generation(for now) two-space algorithm. Only live
objects are traced, so this should work well in our environment
where even the 'stack' gets heap allocated and creates garbage.


terpstra:
This is obviously nowhere close to ready for merging, but I want to confirm that the C++ 11 features I'm using compile on all target platforms.

terpstra:
Closes #184.
Closes #106.

terpstra:
Compared to master, this branch also seems to cause a 3x performance increase on computations in wake.

PR #228, opened 2019-08-01
lexer: add support for multiline strings

terpstra:
Multiline strings in wake are introduced by following a " with a line feed:
```
def foo = "
  this line defines the indentation and begins the string (no prior newline)
  this line has matching indentation so starts immediately after the newline
   this line is indented by one character
  "   # this line terminates the string with a newline immediately after "character"

def foo = "
  \ this line is indented by one charater (backslash-space equals a space)
  this one is not indented {str (55+1)} = 56
  \tthis line starts with a tab and ends without a trailing line feed"
```

It is a syntax error if a multiline string has less indentation than the surrounding wake code. It is also a syntax error if lines later in a multiline string have less indentation than the first line.

A delimiter can be placed between the opening " and the line feed:
```
def foo = "XYZ
   This string can include "double quotes" safely.
   That's because the string will not end until we match the token XYZ.
   Like this:
   XYZ"
```

When a delimiter has been defined, it also affects {}s and \s:
```
def foo = "%
  This \n is not interpreted.
  However, this %\n will have a line feed between "this" and "will".
  The expression {abcdef} is not interpreted by wake.
  However, %{str (55+5)} == 60, like you'd expect.
  %"
```


terpstra:
My personal favorite is:
```
def foo = "Ω
  for i in $(seq 100); do echo "${i}"; done
  echo "Wake says 46 == Ω{str (42+4)}"
  Ω"
```

Unicode Ω is just alt-Z on my keyboard and is sure not to conflict with most shell scripts. However, this PR does not force you to use Unicode if you are some sort of 1980s programmer.

terpstra:
The downside to this PR is that the lexer is no longer a regular language. It's context sensitive. However, any approach that strips indentation from a string literal would have to be context sensitive, so going all in with a customizable prefix does not seem to be any worse.

sistillson:
How much does making the parser context sensitive affect other parts of the language?

terpstra:
It does not really affect wake, the build tool. The increase in parsing complexity affects any other tools  that want to parse wake, like syntax highlighting. I doubt I'll be able to make a correct syntax highlighter for multiline strings in joe, for example.

terpstra:
It looks like any parser using perl-style regular expressions would still be able to identify the end of a string using 'back references'. However, that's a perl extension many tools don't support. Python does, though, do pygments would still work. Presumably any tool that can parse Rust strings would also be able to parse wake strings, since they also use matching open/close tokens.

richardxia:
I think there are a few things that are still a bit unclear to me, even with the examples.

1. What happens if you have an indented multiline string where a line is indented _less_ than the first line? E.g.

    ```wake
    def foo = "EOF
        This line is indented by 4 spaces
      This line is indented by 2 spaces
        EOF"
    ```

2. How exactly is the whitespace at the end of the string handled when you use delimiter? In the following example, what does the byte representation of this string look like?
    ```wake
    def foo = "EOF
      Hello
      EOF"
    ```
    Using C notation here, is it `"Hello"`, `"Hello\n"`, ```"Hello\n  "```? (There's supposed to be 2 trailing spaces in that last one, but GitHub's code block seems to be collapsing it into a single one.)

3. The change in behavior of `\` and `{}` when you use a delimiter is quite confusing to me, since I haven't seen any language do anything similar to that. Could I suggest that you don't change the behavior of either of those just based on presence of a multiline string delimiter? In Python, they use prefixes to the quote character to change the ["type" of string literal](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals), which affects interpolation/escaping. E.g. `f"This is a string that supports interpolation: {foo}"` and `r"This is a raw string, where escape characters change their meaning, so this is a literal \n"`. Scala has something similar, where an [s-string](https://docs.scala-lang.org/overviews/core/string-interpolation.html#the-s-string-interpolator) is similar to Python's f-string.


Final request: Could we have this information present in the docs for wake? It'd help us keep the documentation up to date if we add documentation at the same time we add new features.

terpstra:
> 1. What happens if you have an indented multiline string where a line is indented _less_ than the first line? E.g.
>    ```
>    def foo = "EOF
>        This line is indented by 4 spaces
>      This line is indented by 2 spaces
>        EOF"
>    ```

That's a syntax error:
Insufficient whitespace indentation at build.wake:91:0

> 2. How exactly is the whitespace at the end of the string handled when you use delimiter? In the following example, what does the byte representation of this string look like?
>    ```
>    def foo = "EOF
>      Hello
>      EOF"
>    ```

That's the same as "Hello\n"

> 3. The change in behavior of `\` and `{}` when you use a delimiter is quite confusing to me, since I haven't seen any language do anything similar to that. Could I suggest that you don't change the behavior of either of those just based on presence of a multiline string delimiter?

The whole point of using the delimiter *IS* to prefix the escapes. This way you can copy-paste shell code verbatim without change into a wake file, without having to escape the \s or {}s or "s in the shell code.

I am aware of the python approach and I don't think it's suitable here. We WANT to be able to put wake expressions into the shell script, but not have to escape the shell scripts code. If you *DO* want to escape the shell scripts code and not have to escape wake's interpolation, then don't use a prefix in the multiline string.

> Final request: Could we have this information present in the docs for wake? It'd help us keep the documentation up to date if we add documentation at the same time we add new features.

Yes. My intent was to use the examples in the PR as the examples in the docs.

richardxia:
> The whole point of using the delimiter IS to prefix the escapes. This way you can copy-paste shell code verbatim without change into a wake file, without having to escape the \s or {}s or "s in the shell code.

I would view the current definition of the delimiter as serving two, potentially orthogonal purposes: 1) to prefix the escapes 2) to indicate the end of the multiline string. I have a feeling that people thinking of the multiline string delimiter as being the same as a heredoc delimiter (like in Bash, Ruby, or Perl) will find it unintuitive that it is also used as the escape prefix character.

I guess we'll see how things work here using the same delimiter. Since Wake is not meant to be a general purpose language, and the main intended use case of the multiline strings is to write shell programs, maybe what I'm saying is not that common of a use case.

terpstra:
> I would view the current definition of the delimiter as serving two, potentially orthogonal purposes: 1) to prefix the escapes 2) to indicate the end of the multiline string.

Yes. I agree.

> I have a feeling that people thinking of the multiline string delimiter as being the same as a heredoc delimiter (like in Bash, Ruby, or Perl) will find it unintuitive that it is also used as the escape prefix character.

Since AFAIK no other language has thought to escape it's own escapes (other languages always force you to escape the guest languages escape), I agree it will be unintuitive to people first learning wake. However, given that \s and {}s are both heavily used in shell scripts, I think it makes sense to let the programmer decide if it is the host language (wake) or the guest language (shell) whose escapes need to be escaped.

PR #229, opened 2019-08-01
respect USE_FUSE_WAKE on mac os

albertchen-sifive:
Fixes bug where `fuseRunner` is always used on mac os even if `USE_FUSE_WAKE` is set to `0`.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -387,14 +387,13 @@ def wOK = 1
 def xOK = 2
 def access file mode = prim "access"
 
-global def defaultRunner = match sysname
-  "Darwin" = fuseRunner
-  _ = match (getenv "USE_FUSE_WAKE")
-    Some "0" = preloadRunner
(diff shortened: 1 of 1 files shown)


terpstra:
Oh. Right. The preload runner does not work on macOS Mojave, because "security". So you actually need FUSE on that platform.

Did you actually get it to work with this change?

albertchen-sifive:
yeah https://travis-ci.com/sifive/block-inclusivecache-sifive/jobs/221761373

albertchen-sifive:
also preload runner seems to work fine on my laptop running macOS Mojave?

terpstra:
Are you sure it is actually working? I tried running it on my mac, and it builds successfully the first time, but the database shows no inputs are detected.

terpstra:
Try running 'wake -o' on some output file and compare the fuseRunner to the preloadRunner on osx.

terpstra:
Confirmed: preloadRunner does not work on Mojave.

PR #239, opened 2019-08-07
Constant space lists

terpstra:
I've found an evaluation order which causes many expressions to run in constant space. This PR needs to be a bit more clever about when it runs the recursive calls. Ideally, we would like to run them in demand-order, like lazy languages.


terpstra:
Closing in favor of #244

PR #243, opened 2019-08-11
Add documentation for runners and environment packages

LeytonHo:
Credit to @albertchen-sifive for providing the base content of this.

### DIFF
--- a/share/doc/wake/quickref.md
+++ b/share/doc/wake/quickref.md
@@ -128,7 +128,7 @@ Wake is generally functional.
 
 ### Data Types
 
-####primitives
+#### Primitives
 * String -> "A String" : a string interpreted as unicode
     * a "" string also allows string interpolation using {}
         * for example  `"{foo}{str (4 + 3)}"`
@@ -200,11 +200,11 @@ global data JValue =
 Wake has many useful function for extracting data from a json structure.
 
 
-#### jobs
+#### Jobs
 
 Jobs are how wake runs external programs. They are created by executing a Plan with 
 a Runner. The Plan is a tuple describing how to run the program, and the Runner is 
-a function which reads the Plan and runs the program. 
+a function which reads the Plan and runs the program accordingly. 
 
 Jobs provide wake's central useful functionality. Wake can be viewed as a functional
 programming language for running jobs.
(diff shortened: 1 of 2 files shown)


share/doc/wake/quickref.md
```diff
@@ -220,8 +220,31 @@ global def runit _ =
...
+* The last argument is the base runner that the current runner is built on top of. This is because all runners must be built on top of a preexisting runner. For example, in environment-example-sifive, localRISCVRunner and defaultRISCVRunner are built on top of localRunner and defaultRunner respectively.
+* `publish runner` must be executed for Wake to recognize a runner
+
+Runners are chosen for a given Plan based on the value returned by their `score` function and the plan's `RunnerFilter` field.
+
+* In `runJob`,  `subscribe runner` is called to retrieve the list of all published runners. The `RunnerFilter` function inside the Plan tuple can be set to filter out runners that the plan wants to exclude from consideration. Then the `score` function of each runner is called and runners that return Fail or a score <= 0.0 are excluded. Of the remaining runners, the one with the highest score is picked. 
+* Local runners can only run when the Plan tuple has `LocalOnly` set to true
+* Default runners can only run when the Plan tuple has `LocalOnly` set to false
```
richardxia: Can we add a description of what a local runner is or what the `Plan.LocalOnly` property means? It's probably also worth defining what `localRunner` and `defaultRunner` are, since they are the two builtin ones.

share/doc/wake/quickref.md
```diff
@@ -220,8 +220,31 @@ global def runit _ =
...
+
+#### Using environment packages 
+
+Environment packages can be added to a workspace to provide tools (using runners) for running the workflow. You can check out [environment-example-sifive](https://github.com/sifive/environment-example-sifive) as an example of an environment package that contains runners for Wake-based workflows.
```
richardxia: It might be nice to explicitly mention that an environment package is in fact a collection of Wake files which define and publish runners. It's also probably a good idea to mention that environment packages are meant to be defined for each unique running environment, which decouples the Wake rules from how tools are specifically installed in a specific environment.

share/doc/wake/quickref.md
```diff
@@ -220,8 +220,31 @@ global def runit _ =
...
+* The `post` argument is of type `Pair (Result RunnerOutput Error) a → Result RunnerOutput Error` and is similar to the `pre` function but is called after the job has run. `post` is for editing the reported outputs/inputs/usage of the job. For example, the fuse runner uses the `post` hook to set the output and input files reported by the FUSE filesystem. 
+* The last argument is the base runner that the current runner is built on top of. This is because all runners must be built on top of a preexisting runner. For example, in environment-example-sifive, localRISCVRunner and defaultRISCVRunner are built on top of localRunner and defaultRunner respectively.
+* `publish runner` must be executed for Wake to recognize a runner
```
albertchen-sifive: I think it would be good to put this stuff as comments next to their respective definitions in the wake code. now that we have the sphinx stuff we should be able to generate documentation from code comments if they are in this format https://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions

share/wake/lib/system/job.wake
```diff
@@ -44,6 +44,11 @@ tuple Runner =
   Fn:           Job => Result RunnerInput Error => Result RunnerOutput Error
 
 # Create new Runner given pre- and post-hooks around an existing Runner
+# The `score` argument is of type `Plan → Result Double String` and is called by `runJob` to produce a score representing the priority of a runner with respect to the given Plan. For example, if `plan.getPlanResources` returns list `("python/3.7.1", Nil)` then the RISC-V runner probably cannot provide that resource and should return something like `Fail RISCVRunner: cannot provide resource: python/3.7.1". If the plan resources is `("riscv-tools/2019.02.0", Nil)` then the `score` function should return `Pass 1.0` or some other positive number.
```
richardxia: As we talked offline, let's:

1. Wrap the lines so that they're not super long. My personal preference is 80 chars, but I feel that that's increasingly less popular. Let's at least keep it under 120, which seems to be the max that GitHub shows without scrolling
2. Trim some of the content here. The Quickref can keep more of the more detailed info. This can focus on more of the specifics about what this argument is about.
3. Can we format this in a way such that Sphinx is able to understand that these are arguments of the function? @terpstra, does the Sphinx plugin include Javadoc-like comment syntax like the `@param` in this example: https://en.wikipedia.org/wiki/Javadoc#Examples

PR #244, opened 2019-08-14
Lazy recursion

terpstra:
This PR changes the evaluation order of wake from strict to a hybrid of lazy+strict. This gives us infinite lists and constant-space chaining of list operations. However, it does result in a slight reduction of visible parallelism. It's also possible to slightly tweak this PR to gain the constant-space list operations while maintaining strict evaluation. What's best is TBD. I'll make it a runtime knob for now.

### DIFF
--- a/share/wake/lib/core/string.wake
+++ b/share/wake/lib/core/string.wake
@@ -17,7 +17,9 @@
 
 global def strlen string = prim "strlen"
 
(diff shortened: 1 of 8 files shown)


terpstra:
Way out of date and while an interesting experiment, it's not going to happen.

PR #247, opened 2019-08-14
fuse: support rename of directories with contents

terpstra:
Previously this just broke on hashing


sistillson:
are you going to merge this into v0.15.1?

terpstra:
I've already pulled it into the v0.15 branch.

PR #252, opened 2019-08-21
User-controlled Plan Usage

terpstra:
This is now possible:
```
editPlanUsage (setUsageMemBytes XXX _)
```

wake -p50 will limit wake to using only 50% of memory+cpu+etc, guided by:
1. the database-statistics-based predicted usage
2. the user supplied predict usage
3. the default guess (1 thread, 2MiB)


terpstra:
Percentage to use comes from -pXX, then WAKE_PERCENT environment variable, then 90%.

PR #260, opened 2019-08-31
Add documentation describing Wake from a Scala perspective.

richardxia:
I took a stab at writing a document aimed at describing Wake from the perspective of a Scala programmer. I am neither a Wake nor a Scala expert, so feel free to make any changes that you want to this.

This is not comprehensive, but I thought it at least covered some of the most common data types used in Wake. I wish this talked a bit more about other, non-data-type similarities and differences between Scala and Wake, but I haven't figured out how to organize that kind of information yet.

Note that I tried out writing this in AsciiDoc, since I thought it'd be a richer markup language for writing documentation. However, if you prefer a different language like Markdown, I could port this over (although the codes blocks within the table may be tricky to do in Markdown without relying on HTML).

### DIFF
--- a/share/doc/wake/from-scala.adoc
+++ b/share/doc/wake/from-scala.adoc
@@ -1,0 +1,401 @@
+= Learning Wake as a Scala programmer
+:toc:
+
+This document describes Wake from the perspective of a programmer who is already familiar with Scala.
+Wake is a functional programming language, and it is very similar to the functional aspects of Scala.
+
+Wake is, however, "purer" than Scala,
+so while Scala allows for a blend of functional, imperative, and object-oriented styles,
+Wake more strictly enforces a functional style.
+In fact, the only side effects in Wake are printing, running subprocesses, and panicking.
+
+Since Wake does not support function overloading, the same function name cannot be defined more than once with different arguments.
+Therefore many of the built-in collection types define functions that are prefixed with the first initial of the type name.
+For example, while the `List` type has a `map` function, the `Option` type has an `omap` function, and the `Result` type has an `rmap` function.
+
+See the following sections to learn about how various concepts in Scala map to Wake.
+
+
+== Primitive types
+
+Wake generally has familiar syntax for primitive types.
+Syntax for more advanced functionality of strings -- such as interpolation and multi-line strings -- differs between Wake and Scala.
+See <<table-primitives>> for more information.
+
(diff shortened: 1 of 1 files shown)


share/doc/wake/from-scala.adoc
```diff
@@ -0,0 +1,396 @@
...
+"Plain string"
+
+s"${Interpolated} string"
+
+"""
+Multi-
+line
+string
+"""
+----
```
jackkoenig: Scala has multiline strings with interpolation:

```scala
s"""
Multi-line string
with ${interpolation}
"""
```

It might be worth talking about Scala strip prefix vs. Wake's multi-line string start, eg.:
```scala
    val x = s"""
      |Each line
      |starts at the pipe
      |""".stripPrefix
```
vs.
```
    def x = "%
      Each line
      starts lined up with the first line
      %"
```
I might have Wake's syntax slightly wrong, but it's relevant, possibly worth just putting in a topic-document though rather than here.

richardxia: Thanks, I added one more example with the Scala multi-line string with interpolation, but I think the remainder of the string features belong in a document specific to string handling in Wake.

share/doc/wake/from-scala.adoc
```diff
@@ -0,0 +1,401 @@
...
+val newResult = result.flatMap(reciprocal)
+----
+|
+[source,wake]
+----
+def reciprocal d =
+  if (d == 0.0)
+    makeError("Divide by zero")
+  else
+    1.0 / d
```
terpstra: You presumably need a Pass + Fail here. You also need to decide if it's a Double or an Integer. How about?:
```
  def reciprocal = match _
    0.0 = Fail (makeError "Divide by Zero")
    d = Pass (1.0 /. d)
```

share/doc/wake/from-scala.adoc
```diff
@@ -0,0 +1,401 @@
...
+|
+[source,wake]
+----
+tuple Foo =
+  global MyString: Strings
+  global MyInt: Integer
+
+def foo = Foo "foo" 0
+def myFooString = foo \| getFooMyString
+def myFooCopy = foo \| editFooMyInt 1
```
terpstra: You mean setFooMyInt 1.

terpstra:
Oh. This was merged.

PR #271, opened 2019-09-06
Tuple Documentation

jackkoenig:


share/doc/wake/tour/tuples.adoc
```diff
@@ -0,0 +1,183 @@
...
+----
+global def makeCat name age = Cat name age None
+----
+
+Now, all existing usages of `Cat` will continue to work,
+while users can take advantage of the generated accessors to use the new field:
+
+----
+def myCat =
+  Cat "Bill" 3
```
terpstra: This should be makeCat, right?

jackkoenig: Yep

PR #278, opened 2019-09-16
Provide install method for Arch Linux

sequencer:
official package is still in review.
AUR version is ready in [wake-git](https://aur.archlinux.org/packages/wake-git/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,16 @@ On Redhat (6.6 or later):
     # On RHEL6: sudo yum install devtoolset-6-gcc devtoolset-6-gcc-c++
     sudo yum install makedev fuse fuse-devel sqlite-devel gmp-devel ncurses-devel pkgconfig git gcc gcc-c++ re2-devel dash
 
+On Arch Linux:
+ 
+ Install from official repo(stable version):
+ 
+    sudo pacman -S wake
+ 
+ Install from AUR(git master version), use `pakku` or some other AUR helpers:
+ 
+    sudo pakku -S wake-git
+
 On FreeBSD (12 or later):
 
     pkg install gmake pkgconf gmp re2 sqlite3 fusefs-libs dash


terpstra:
FYI, the reason I have not merged this is because the README is describing how to BUILD wake, not install it. Could you please revise this PR to describe how to install the dependencies on arch instead?

sequencer:
It just like deb to describe how to build a program for Debian.
It provide a [build and install script](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wake-git) in Arch Linux, This [line](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wake-git#n11) describe the dependency and the script are following the official readme to build wake.
So just run `pacman -S wake` can install the dependency and install `wake` from Arch Linux repository, while `yaourt -S wake` can install the dependency and build from source and finally install it.

terpstra:
I could as well have said `apt-get build-dep wake` in the debian section to get build dependencies for wake and `apt-get install wake` to install it. That would be the debian equivalent to the instructions you posted. However, that is not the point of that section of the README.

The part of the README is describing how someone can build wake from the copy of source they have checked out. Your arch commands (and the matching debian commands I quoted) are doing something very different: they are fetching dependencies for and building the version of wake that is packaged by arch/debian. Suppose wake has a new dependency added in a new version; your "pacman -S" will not know about it, but the README will accurately describe what is required. Similarly, if older versions used to require a package and someone needs to build an older version of wake, "pacmac -S" will not do the right thing.

I am not opposed to adding a section that describes how to install packages from a given distribution (since the README is also what you see in the github repository), but those would be very different instructions than what we have now. I'm also uncertain of how useful those instructions would be, since "pacman -S" is the same command for every package just like "apt-get build-dep". People probably don't need the README to tell them how to install packages on their system; they already know this. In contrast, they may NOT know which particular packages they need to build wake from a given source version.

sequencer:
Got it! I'll close this PR, and maintain the PKGBUILD for Arch.

terpstra:
To be clear, I would love to merge a PR that describes how to build on arch from source.

PR #313, opened 2019-10-02
Increase MAX_JSON from 1 MiB to 1 GiB

jackkoenig:


jackkoenig:
Note this is against v0.15

terpstra:
Doesn't 1GB seem a bit excessive?

jackkoenig:
It’s a max so does it really matter? We can do 64 or 128 MiB if you prefer

terpstra:
I'm going to fix this via #316 on master and then a cherry-pick to v0.15

PR #329, opened 2019-10-22
Stop scan at wake db

terpstra:
This affects which .wake-files are processed.
As a consequence, we can now have tests of wake itself.


terpstra:
#359 Is a better approach.

PR #331, opened 2019-11-05
Bump MAX_JSON to 128 MiB.

john-sifive:
Increase the size of the JSON string. The old value, 1Mib, results in I/O error messages from wake. The increase in MAX_JSON is already part of master - this PR back ports it to v0.17.

Note we will also need to create and distribute a 0.17.1 release of wake.

Update: Will need the same change in the 0.15 release.

### DIFF
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -24,7 +24,7 @@
 #define _XOPEN_SOURCE 700
 #endif
 
-#define MAX_JSON (1024*1024)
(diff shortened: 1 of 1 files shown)


richardxia:
I'm going to squash/replace this with a cherry-pick of https://github.com/sifive/wake/pull/316, since the title and commit messages are misleading due to using a lowercase "b", which refers to bits, instead of a capital "B", which refers to bytes.

PR #332, opened 2019-11-08
Increase MAX_JSON in v0.15.2

john-sifive:
Increase the size of the JSON buffer to deal with more "visible" files.  

This is already part of the master branch, but requesting a back port.
Expecting a new release 0.15.3.

### DIFF
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -24,7 +24,7 @@
 #define _XOPEN_SOURCE 700
 #endif
 
-#define MAX_JSON (1024*1024)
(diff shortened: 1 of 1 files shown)


richardxia:
Cherry-picked https://github.com/sifive/wake/pull/316 so that it's more obvious where this change came from.

richardxia:
Actually, it looks like this change is already merged in to the `v0.15` branch (https://github.com/sifive/wake/commit/26aab29f85b3d760e86630bbf66fd9254641ba7a). We just haven't actually created a tagged release from the tip of the `v0.15` branch.

I'm going to close this PR.

PR #334, opened 2019-11-11
Set committer and release notes for v0.17.1.

richardxia:
Same as https://github.com/sifive/wake/pull/333 except against the `v0.17` branch.

Also note that we had not created the placeholder section for the next release in the changelog, so I cloned the previous one as `0.17.0-1`.


jackkoenig:
Going to redo this

PR #357, opened 2020-01-23
write: use target to avoid multiple output errors

albertchen-sifive:
make `writeImp` a target so that calling `write` with the same arguments does not result in an error.
```
$ wake 'write "foo" "bar", write "foo" "bar", Nil'
File output by multiple Jobs: foo
```

### DIFF
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -74,7 +74,7 @@ global def mkdir path =
     Nil      = panic "impossible"
   path | simplify | tokenize `/` | root
 
-def writeImp inputs mode path content =
+target writeImp inputs mode path content =
   def writeRunner =
     def imp m p c = prim "write"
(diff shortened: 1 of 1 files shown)


sistillson:
what happens if you write the same file name with wake being in different directories?
(PWD as hidden state)

albertchen-sifive:
> what happens if you write the same file name with wake being in different directories?
> (PWD as hidden state)

I don't think wake has any hidden pwd state? wake always runs as if it's from the wake.db directory

albertchen-sifive:
I forgot that you could can always just do `target memoizedWrite = write`. I think `write` was originally left as a `def` to give people the choice of using `target` or not? In that case I think there should be an additional `memoizedWrite` in the wake standard library then?

sistillson:
Makes sense to me. a write and memoizedWrite would work.

Issue #355, opened 2020-01-22
Ignore some wake files


terpstra:
In some cases, you can end up with wake files in a workspace that should be ignored.

Use cases:
- unit tests for wake itself
- submodules in repositories only half transitioned to wit

There are some related improvements that will reduce this need:
- use namespaces (avoids symbol conflicts)
- multiple concurrent wake processes sharing the same database

However, the above examples illustrate this feature should probably exist.

mmjconolly:
Hi @terpstra 
How do you feel if we had a `.wakeignore` file with regex patterns much like how git has `.gitignore`?
The use case I'm looking at is trying to transition some submodules of a wit dependency into wit, and am running into duplicate symbol definitions.
```
workspace
├── repo1
│   └── foo.wake
└── repo2
    └── repo1
        └── foo.wake
```
I want to be able to run `wake` from `workspace`, but also run `make` from `repo2`, with it's submodules initialized.
In this example the `.wakeignore` would be in `repo2` with the contents `repo1/*`

PR #359, opened 2020-02-04
Search for .wakeignore files

mmjconolly:
Files called `.wakeignore` can contain globs to
exclude wake source files from usage.
The exclusions are globs and are relative
to the root of the workspace.

```
workspace
├── wake.db
├── repo1
│   └── foo.wake
└── repo2
    └── repo1
        └── foo.wake
```
If `repo1` is checkout out twice, like above, then if a `.wakeignore` file at path`workspace/repo2` contained `repo2/repo1` then `foo.wake` would not be read twice.

Issue #355

### DIFF
--- a/share/doc/wake/tutorial.md
+++ b/share/doc/wake/tutorial.md
@@ -699,10 +699,19 @@ Examples include testing, where you could exclude some files from regular usage,
 or a directory structure that includes duplicate repository checkouts, where
 duplicate symbol definitions would raise an error.
 
-Wake can look for files named `.wakeignore` containing patterns.
-The pattern language is shell filename globbing with leading directory match.
+Wake looks for files named `.wakeignore` containing patterns.
+The pattern language is shell filename globbing.
 One pattern per line, each relative to the path of the `.wakeignore` file.
-Patterns will only match against files with the `.wake` filename extension.
+
+The concrete syntax is:
+- empty lines are ignored
+- lines starting with a `#` are comments and are ignored
+- `?` matches a single non-slash character
+- `*` matches any number (including zero) of non-slash characters
+- `[a-z]` matches a single lower-case character
+- `/**/` in an expression like `foo/**/bar` stands in for any number of directories (including zero)
+- `foo/**` recursively matches all contents of the directory `foo`
+- `**/bar` matches all files `bar` contained in this directory or any subdirectory
 
 An example:
 
@@ -716,12 +725,13 @@ An example:
                 └── foo.wake
 
 If `repo1` is checked-out out twice like above, then if a `.wakeignore` file
-at path `workspace/repo2` contained `repo1` then `foo.wake` would not
+at path `workspace/repo2` contained `repo1/**` then `foo.wake` would not
 be read twice.
 
 The following patterns in `workspace/repo2/.wakeignore` would all match
-the above `repo2/repo1/foo.wake`:
+the above `foo.wake`:
 
-    repo1
     repo1/foo.wake
-    */*.wake
+    repo1/**
+    repo1/**/foo.wake
+    **/[a-z]?[o].wake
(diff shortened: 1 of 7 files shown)


mmjconolly:
>     3. I'm having some slight second thoughts about using regular expressions for the .wakeignore file. I compiled and ran this branch locally to test it out, and it felt kind of unintuitive to use regular expressions. .gitignore actually uses glob patterns, which have the nice property of being able to just literally write down a file's name and having it be ignored. (PRCE) Regular expressions have more special characters, including the `.` character. Also, a .gitignore pattern that doesn't include a `/` character will match a file at any depth of the tree. Thoughts on this?
>        Also, to save you time, here's the Git documentation on .gitignore: https://git-scm.com/docs/gitignore

Happy to change this to use Glob-style [1], but I'm pretty hesitant to try replicating the quirks of the gitignore format, as it has a lot more than just glob.

Another option I'm hesitant about is to fork out to `git check-ignore --no-index` [2] rather than implementing this ourselves.
We do seem to inherently rely[3] on a git executable being available to run wake already.

How do you feel about forking out to git? It would be a bit slower than RE2, but would give a more familiar experience to those that expect gitignore's format.

[1] http://man7.org/linux/man-pages/man3/glob.3.html
[2] https://git-scm.com/docs/git-check-ignore
[3] https://github.com/sifive/wake/blob/master/src/sources.cpp#L87

richardxia:
> Happy to change this to use Glob-style [1], but I'm pretty hesitant to try replicating the quirks of the gitignore format, as it has a lot more than just glob.

Yeah, I agree that there's no need to replicate all the quirks of .gitignore. The high-order bit here is glob vs regex, where I think glob will feel much more natural to people.

> Another option I'm hesitant about is to fork out to `git check-ignore --no-index` [2] rather than implementing this ourselves.
> We do seem to inherently rely[3] on a git executable being available to run wake already.
> 
> How do you feel about forking out to git? It would be a bit slower than RE2, but would give a more familiar experience to those that expect gitignore's format.

I had never heard about `git check-ignore` before, so this was interesting for me to learn about, but I don't think we need to go this far. I think using glob instead of regex is sufficient.

mmjconolly:
re -> globs
I haven't added any tests/docs/ci, looking to get feedback on the code itself first.

These globs are still relative to the root of the workspace, let me know if relative is really wanted instead.

mmjconolly:
Added 1 test, in the same style as wit.
Added a small section to tutorial.md doc.

terpstra:
Thanks for taking a stab at implementing this!

I have a couple observations:
1. It crawls the workspace tree twice. Once for the wake files and once for the .wakeignore files. Perhaps we could improve push_files to find both at the same time, and then filter after. 
2. This will accept files like myfile.wakeignore - is this intentional?

terpstra:
Should .wakeignore also affect results from 'sources'?

I think the ignore content should definitely be relative to the directory which contains the wakeignore. Richard is probably right that glob is better than regexp for this situation. 'fnmatch' appears to be a POSIX system call.

mmjconolly:
Changes,
* patterns are now relative to the `.wakeignore` file
* crawls filesystem once
* tried again to write a regex for matching '.wakeignore'

As for the `sources` function, I think I'd be surprised as a user if the ignore file silently affected that.

terpstra:
In reviewing this code, I found a significant number of problems. I am going to substantially rework this PR before it is merged. Please come by my desk sometime and I will provide you direct feedback. There is too much to post here.

terpstra:
I pushed a bunch of commits to show the superficial issues, but I am going to start from scratch.

terpstra:
Can those of you using this feature please confirm that it still works for the cases where you were using it?

richardxia:
> Can those of you using this feature please confirm that it still works for the cases where you were using it?

Seems to work with the toy example test I had in my local checkout. Thanks for taking a look at this, and for adding documentation!

PR #375, opened 2020-03-08
Wake command-line subcommands now controlled by package exports

terpstra:
Invoking wake now lists subcommands exported by a package.

For example:
```
terpstra@peach:~/wake$ ./bin/wake
Available wake subcommands:
  build
  install
```

Subcommands are functions that accept List String and return Result.

Running 'wake build default' from the command-line will result in evaluation of 'build ("default", Nil)'.

This PR also adds two new functions to make coping with files provided by the command-line:
  'in dir path' - combines dir and path
  'cwd' - the directory where wake was invoked

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,7 @@ jobs:
           debuild -us -uc
           cd ..
           apt-get install -y ./wake_*.deb
-          wake --no-workspace 'shellJob "ls /" Nil | getJobStdout'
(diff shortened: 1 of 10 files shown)


terpstra:
After implementing it, I am not sure I like this. The original plan of a single 'main' function per package may have been better.

terpstra:
I'm not 100% sure this is the right approach, but merging it as the starting point for discussion.

PR #383, opened 2020-03-24
wake: implement refutable pattern matches with 'when'

### LINKED ISSUE #382: Support '?' early error return operator
Something to allow early return of errors could be nice

https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html

terpstra:
Fixes #382 .

Working with Result and Option can be a pain.
Wake currently allows you to do this:
```
def (a,, b,, c) = someFnThatReturnsATuple 42
def Pair x y = someFnThatReturnsAPair 43
```

However, you cannot do this:
```
def Pass ok = someFnThatReturnsAResult 42
```

The above fails, because it is an in-exhaustive pattern match.

However, a lot of things in a wake build can fail! It becomes super awkward to have to continually indent when progressively matching all the ways things can fail.

This PR introduces a new keyword `when`, which can be used like this:
```
def myFun _ =
  def foo = Pass (Pair 45 "x")
  when Pass (Pair x y) = foo
  def s = str x, y, Nil
  Pass s
```

The above code gets de-sugared to:
```
def myFun _ =
  def foo = Pass (Pair 45 "x")
  match foo
    Fail f = Fail f
    Pass (Pair x y) =
       def s = str x , y, Nil
       Pass s
```

There is nothing special about `Result` in this syntax. It works with all types. However, mixing types in a single function will lead to a type-check failure, since the `when` keyword acts like an early return with the type of the un-matched case. For example,

```
def myFun2 str1 str2 = 
  when Some x = int str1
  when Some y = int str2
  Some (x + y)
```

The above function conditionally "adds" two strings. The de-sugared form is:
```
def myFun2 str1 str2 = 
  match (int str1)
    None = None
    Some x = match (int str2)
       None = None
       Some y = Some (x + y)
```

The keyword `when` was selected because using it has two consequences:
- The expression performs a pattern match and `when` selects one of the cases
- Everything after the `when` keyword must WAIT for the pattern to be fulfilled.

Concretely,
```
def a = action1 # action1-3 execute concurrently
def b = action2
def c = action3
when Pass x = b # everything below here happens only once action2 passes/fails
def d = action4 x # action4-5 execute concurrently
when Pass z = action5 x
when Pass w = a # everything below here happens only once action1, 2, and 5 pass/fail
```


mmjconolly:
What's the one sentence version of what `when` means?
I read the examples a few times and I don't think I could explain it to someone

mmjconolly:
For this example I think the 'when' to define a function/variable to be unexpected to a casual reader. We also end up writing `Some` and `Pass` all over
```
def myFun2 str1 str2 = 
  when Some x = int str1
  when Some y = int str2
  Some (x + y)
```
I can see that you're trying to go another path than `?` but I do think it looks pretty clean,
`?` could just take presume that the first type in the Option/Result datatype is good, and the other type is returned under `?` ?
```
def myFun2 str1 str2 = 
  def x = int str1? 
  def y = int str2?
  Some (x + y)
```

terpstra:
The one sentence version of `when`:
```
when a = b
stuff
```
means: do `stuff` when you can turn `b` into `a`.

terpstra:
The problem with `?` is that it is an expression. Expressions in wake are evaluated in parallel. In rust, there is a defined order for everything. Therefore, you can have `?` return from the function and have a well-defined language. That would not work here.

While I agree that it is a bit annoying to have to write `when Pass` instead of just `when`, it makes it possible for us to use this same sugar for all types, not just Result or Option. It also means we can do a full pattern match extraction, which I am sure will be quite handy. One could also perhaps argue that saying 'when Pass x = thing1' also more naturally says what is happening: when thing 1 passes, run stuff.

terpstra:
To elaborate on the problem with `?` a bit:
Consider this:
```
def myFun2 str1 str2 = 
  Some (int str1? + int str2?)
```
Both parts of the expression run at the same time. Both, neither, or either could fail. If both fail, whose failure gets returned? If 'int' has side effects, will people be surprised that 'int str?' ran even though 'int str1?' failed?

PR #384, opened 2020-03-25
V0.17

davidmlw:
`wake --ctags` and `wake --etags` will generate ctags and etags files for vim and emacs.


terpstra:
I just learned that my preferred editor also supports ctags! So you can be certain I will merge a variant of this PR. Unfortunately, the proposed implementation relies on globals. There are almost no globals after wake 0.18. :-(

terpstra:
Closing in favor of #445

PR #388, opened 2020-03-27
wake: start to define the v1 API

terpstra:
When wake v1.0 is released, the `v1_wake` package will be locked down. There will be no symbols added, changed, or removed (other than bugfixes). This PR is intended to start a discussion of what goes into the v1 public and stable API. The `wake` package will always be a grab-bag of all the newest experimental features of wake. I imagine most files will look like this:
```
from vX_wake import _ # defines the wake compatibility level of this file
from wake import extra1 extra2 # extra non-standard items needed from experimental wake
```

I've included only those functions I consider essential and non-controversial. If you have any objections to things already in this list, please raise them!

There is obviously a ton of stuff I did not include yet. In particular, none of the Job API. I would like to tackle that after a potential llbuild-inspired refactor of how build targets work in wake.

### DIFF
--- a/share/wake/lib/versions/v1.wake
+++ b/share/wake/lib/versions/v1.wake
@@ -1,0 +1,91 @@
+# Copyright 2019 SiFive, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You should have received a copy of LICENSE.Apache2 along with
+# this software. If not, you may obtain a copy at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


share/wake/lib/versions/v1.wake
```diff
@@ -0,0 +1,91 @@
...
+# unicodeCanonical unicodeIdentifier unicodeLowercase unicodeToInteger
+
+# Integer
+from wake export def unary + -
+from wake export def binary + - * / % ^
+from wake export def binary <=> < > >= <= == !=
+# ~ << >> abs and gcd icmp lcm min max or powm prod root sqrt sum xor ∏ ∑
+
+# Double - not included in v1
+# JSON   - not included in v1
```
albertchen-sifive: why not JSON? what about just the AST nodes and `parseJSONBody`/`parseJSONFile`?

terpstra: Sorry, the comment is misleading.

I definitely think we should include JSON in v1, but we should be sure are willing to commit to those function names.

PR #401, opened 2020-04-02
vim: Update syntax highlighting for new keywords.

richardxia:
Highlight topic names as Identifiers.
Highlight package/import/export keywords and Includes.
Highlight panic as Exception.
Correctly highlight unary/binary as keywords and not identifiers.

Screen Shot 2020-04-01 at 11 21 19 PM

I don't really know what I'm doing with vim syntax highlighting, but I think this makes it look better and closer to what I'm used to, at least for my vim color scheme. Yes, I know that any identifier that follows a `def` or `topic` in the import/export list gets incorrectly colored, and, yes, I know that operators are kind of broken, but I don't think I can fix that without spending a lot more time learning how vim syntax highlighting works.

### DIFF
--- a/share/doc/wake/syntax/vim/syntax/wake.vim
+++ b/share/doc/wake/syntax/vim/syntax/wake.vim
@@ -17,11 +17,15 @@ syn match lineComment "#.*"
 
 " Keywords
 " TODO Should important globals from prim.wake we marked keywords?
-syn keyword wakeKeyword if then else here global export import package from type topic binary unary subscribe match require data tuple prim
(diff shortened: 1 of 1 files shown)


terpstra:
Is neovim = vim? We have a tree-sitter parser and highlighter for wake. I wonder if vim can use that?

richardxia:
> Is neovim = vim? We have a tree-sitter parser and highlighter for wake. I wonder if vim can use that?

No. Neovim is a popular fork of vim, but it is not the same thing as vim.

PR #407, opened 2020-04-06
List append item function

mmjconolly:
Explicit list append and prepend functions

I've seen (and written) lines that look like,
`def newList = someList ++ (someItem, Nil)`
I think it's easier to read as
`def newList = append someList someItem`


terpstra:
To make these chain well with `|`:
```
def prepend item list = item, list
def append item list = list ++ (item, Nil)
```

Issue #434, opened 2020-05-18
Command line option "--strict" to remove /bin & /usr/bin from PATH


mmjconolly:
We've found that people can choose to not be explicit about the resources that their `Job` depends on when they install programs into `/bin` and `/usr/bin`.
This can make portability and reproducibility of wake rules that use `Job` difficult.

Please provide a strictness option that sets `PATH=""` for a wake invocation, for example
```
$ wake -x 'makePlan ("/bin/sh", "-c", "echo $PATH", Nil) Nil | runJob | getJobStdout'
Pass "/usr/bin:/bin\n"
$ wake --strict -x 'makePlan ("/bin/sh", "-c", "echo $PATH", Nil) Nil | runJob | getJobStdout'
Pass "\n"
````
This would allow us to detect and migrate our jobs away from `PATH=/bin:/usr/bin`.

Once we're satisfied that all our rules are migrated we can set the option in our own regression tests

terpstra:
Fixed by #436

PR #436, opened 2020-05-19
Add "--strict" option to wake cli to set PATH=""

mmjconolly:
Motivated by: #434

### DIFF
--- a/share/wake/lib/system/environment.wake
+++ b/share/wake/lib/system/environment.wake
@@ -31,8 +31,13 @@ publish path = match (getenv "WAKE_PATH")
   None = Nil
 
 # Setup a default path
-publish path = "/usr/bin", Nil
-publish path = "/bin", Nil
+publish path =
+  def default = "/usr/bin", "/bin", Nil
+  match (getenv "WAKE_STRICT_PATH")
+    Some "1" = Nil
+    Some _ = default
+    None = default
+
 
 # Default environment for new Plans
 export topic environment: String
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -59,6 +59,7 @@ void print_help(const char *argv0) {
     << "    --no-tty         Surpress interactive build progress interface"              << std::endl
     << "    --no-wait        Do not wait to obtain database lock; fail immediately"      << std::endl
     << "    --no-workspace   Do not open a database or scan for sources files"           << std::endl
+    << "    --strict-path    Set PATH shell variable as empty (default is /usr/bin:/bin)"<< std::endl
     << "    --heap-factor X  Heap-size is X * live data after the last GC (default 4.0)" << std::endl
     << "    --profile-heap   Report memory consumption on every garbage collection"      << std::endl
     << "    --profile  FILE  Report runtime breakdown by stack trace to HTML/JSON file"  << std::endl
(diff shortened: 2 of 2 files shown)


src/main.cpp
```diff
@@ -204,6 +207,10 @@ int main(int argc, char **argv) {
     percents = getenv("WAKE_PERCENT");
   }
 
+  if (strict) {
+    setenv("WAKE_STRICT_PATH", "y", 1);
```
richardxia, line 211: I think elsewhere Wake uses `"1"` and `"0"` to indicate boolean values for environment variables (`USE_FUSE_WAKE` is the one I'm thinking of), so we'd probably want to be consistent with that.

share/wake/lib/system/environment.wake
```diff
@@ -31,8 +31,10 @@ publish path = match (getenv "WAKE_PATH")
   None = Nil
 
 # Setup a default path
-publish path = "/usr/bin", Nil
-publish path = "/bin", Nil
+publish path = match (getenv "WAKE_STRICT_PATH")
+  Some _ = Nil
+  None = "/usr/bin", "/bin", Nil
```
richardxia, line 36: The way this is written, any value that this is set to will be treated as "true", I think including the empty string. Since it's kind of annoying to actually unset environment variables on the command line (I'm not sure if you can do it in a single state for a single command), I think it'd be better to pattern match on a single "truthy" value and treat all other values as "falsey".

```suggestion
publish path = match (getenv "WAKE_STRICT_PATH")
  Some "1" = Nil
  Some _ = "/usr/bin", "/bin", Nil
  None = "/usr/bin", "/bin", Nil
```

src/main.cpp
```diff
@@ -59,6 +59,7 @@ void print_help(const char *argv0) {
     << "    --no-tty         Surpress interactive build progress interface"              << std::endl
     << "    --no-wait        Do not wait to obtain database lock; fail immediately"      << std::endl
     << "    --no-workspace   Do not open a database or scan for sources files"           << std::endl
+    << "    --strict         Set PATH shell variable as empty (default is /usr/bin:/bin)"<< std::endl
```
richardxia, line 62: Do you envision this flag being used for other types of strictness that we may want to optionally enable in the future, or just the PATH variable? If so, I think this is fine as is, but if you think this would only ever be used for the PATH, then I'd probably prefer a more detailed option name like `--strict-path`.

mmjconolly:
Modified per comments

terpstra:
Is the intent that you can control this setting both via an environment variable and a command-line directive?

My only reservations about this PR is that it affects all jobs in the build. It is also possible to just modify Plans to use a different environment. Is this just a patch you guys are using for testing, or are you proposing to merge it as an actual feature?

For 0.19, I had plans, as you know, to rework the Job/Plan APIs to be more container and environment friendly. My hope is that new API would alleviate these issues, since the environment selected would define what's available and in PATH.

mmjconolly:
Actual feature, that is just having *some* way to control the path.
As it is currently we're stuck as we can't 'un-publish'.

I want to migrate flows away from presuming they have `/usr/bin:/bin` and unless I have a way actually remove them from the path I don't have to way to enforce it if they try to use something like `/usr/bin/java`.
If there are situations where we really must have `/usr/bin:/bin` we can manually add it to the job via `editPlanEnvironment` but I don't think that will be common once we specify resources in other ways.

My intent was that it was controllable by both command line and env var, though it's not actually necessary that we have both.

mmjconolly:
Also, I'd like this put into a point release for 0.18 as we're about to have a migration from 0.17.2 to whatever the next version is and I'd like a mechanism to control the PATH for that migration

mmjconolly:
If we didn't have to care about breaking user expectations, here is my preferred implementation:
```
diff --git a/share/wake/lib/system/environment.wake b/share/wake/lib/system/environment.wake
index fab2c80..d446ca1 100644
--- a/share/wake/lib/system/environment.wake
+++ b/share/wake/lib/system/environment.wake
@@ -30,10 +30,6 @@ publish path = match (getenv "WAKE_PATH")
   Some x = tokenize `:` x
   None = Nil

-# Setup a default path
-publish path = "/usr/bin", Nil
-publish path = "/bin", Nil
-
 # Default environment for new Plans
 export topic environment: String
 export def environment = subscribe environment
```

terpstra:
I see. So should this be a PR against the v0.18 branch?

I think I agree that /usr/bin and /bin should never have been in the default PATH.

If we did go with just removing them completely, that would be v0.19. I could then punt the bigger things I have planned for a v0.20.

mmjconolly:
> I see. So should this be a PR against the v0.18 branch?
Yes good point, rebased and retargeted now

PR #438, opened 2020-05-29
fuse: allow the .fuse directory to already exist

terpstra:
On linux you previously got this error:
  fuse: mountpoint is not empty
  fuse: if you are sure this is safe, use the 'nonempty' mount option

### DIFF
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -1421,7 +1421,16 @@ int main(int argc, char *argv[])
 	sigaction(SIGALRM, &sa, 0);
 
 	args = FUSE_ARGS_INIT(0, NULL);
-	if (fuse_opt_add_arg(&args, "wake") != 0) {
+#if FUSE_VERSION >= 24 && FUSE_VERSION < 30 && !defined(__APPLE__)
+	/* Allow mounting on non-empty .fuse directories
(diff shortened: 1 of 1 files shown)


terpstra:
Argh. This 'nonempty' option is recent. 'osx' does not have it so it complains when it is set.

mmjconolly:
Is the model that there is 1 fuse-waked and many fuse-wake?
If yes, perhaps we could just delete `.fuse` unconditionally once we get the lock on wake.db before we start fuse-waked?

terpstra:
Yes, there is one wake-fused. Using 'nonempty' is less destructive than your 'erase it all' proposal, because when wake is done, the original files reappear untouched.

terpstra:
Now that we have CI testing FUSE, rebase this branch against master to detect if the problems with it are macfuse-only.

terpstra:
Argh. I have it backwards. This option was REMOVED on October 16, 2016 and made the default. I pretty much agree with that decision. Unfortunately, important versions of linux are older than this. Pretty asinine to make it error out when the now obsolete flag gets set, though.

terpstra:
Awesome. The problem exists for versions >= 2.4.0 and < 3.0.0. Unfortunately, fuse.h does not include the version of the library.

Issue #443, opened 2020-07-10
.fuse/ directory being discovered by tools like VCS and causing failures within a single wake (fuse-waked) process


richardxia:
This is slightly different than https://github.com/sifive/wake/issues/439, since this is causing a problem while the top-level wake process is still running.

Let me jump straight into the error and the steps to reproduce.

## The Error
I am getting this error when running VCS while trying to enable coverage collection:

```
  Notice: timing checks disabled with +notimingcheck at compile-time
  Chronologic VCS simulator copyright 1991-2017
  Contains Synopsys proprietary information.
  Compiler version M-2017.03-SP2_Full64; Runtime version M-2017.03-SP2_Full64;  Jul 10 08:13 2020
  UVM_INFO /sifive/tools/synopsys/vcs/M-2017.03-SP2/etc/uvm-1.2/base/uvm_root.svh(402) @ 0: reporter [UVM/RELNOTES] 
  ----------------------------------------------------------------
  UVM-1.2.Synopsys
  (C) 2007-2014 Mentor Graphics Corporation
  (C) 2007-2014 Cadence Design Systems, Inc.
  (C) 2006-2014 Synopsys, Inc.
  (C) 2011-2013 Cypress Semiconductor Corp.
  (C) 2013-2014 NVIDIA Corporation
  ----------------------------------------------------------------
  
    ***********       IMPORTANT RELEASE NOTES         ************
  
    You are using a version of the UVM library that has been compiled
    with `UVM_NO_DEPRECATED undefined.
    See http://www.eda.org/svdb/view.php?id=3313 for more details.
  
    You are using a version of the UVM library that has been compiled
    with `UVM_OBJECT_DO_NOT_NEED_CONSTRUCTOR undefined.
    See http://www.eda.org/svdb/view.php?id=3770 for more details.
  
        (Specify +UVM_NO_RELNOTES to turn off this notice)
  
  
   VCS Coverage Metrics Release M-2017.03-SP2_Full64 Copyright (c) 1991-2017 by Synopsys Inc.
  
  Error-[MON-ICDF] Incompatible Code Coverage Directory
    Simulation could not read the database from the directory 
    '/scratch/rxia/block-pio-sifive-ws2/.fuse/32487/build/api-generator-sifive/pioDUT/sim/vcs/execute/demo/./../../compile/simv.vdb'.
    Please check if the database directory 
[...]

Issue #439, opened 2020-06-06
Symlinks to cover missing .fuse/ paths?


mmjconolly:
When we have a build that embeds a `workspace/.fuse` path into a build artifact, that path doesn't resolve post-build.

A program trying to resolve a path like `/home/user/workspace/.fuse/5555/build/foo` is really trying to resolve `/home/user/workspace/build/foo` post-build.

Could we leave behind a directory-symlink as `fuse-wake` is exiting? 
Example: `/home/user/workspace/.fuse/5555 -> /home/user/workspace`

While it doesn't solve the underlying issue, perhaps it could paper-over some of the pain points of the situation

terpstra:
The 5555 here is the process ID. So, unfortunately, those values get reused during a build. ie: two jobs might use the same ID.

Also: I would rather we solve this by using unshare. I was originally planning on bind mounting the fuse jail to /build and then running the build there. Do you think we could somehow play tricks so that for the user /build points back at their workspace? I'd like to use /build so it is the same output regardless of who built something (shared caching). However, as you note, it is helpful to be able to introspect the artifacts post-build.

richardxia:
Even if the process ID is reused, all of the directories within the .fuse/ directory should point to the workspace root, so it shouldn't matter if there's a duplicate, right? This suggested workaround is mostly meant to be a quick to implement, easy to remove temporary measure until we have more time to implement a more complete solution such as the bind mounting idea you described.

mmjconolly:
I agree that unshare/namespacing is a good way forward generally, but I do feel that we need to consider the short term too.

We have support load helping people who don't understand why some programs they run that consume build artifacts containing paths that don't exist.

I have some questions (where I presume we won't have unshare for enough time that it makes sense to implement a stopgap):
* Can you describe what the PIDs are used for in the fuse-wake/fuse-waked/wake interaction?
* What was the reasoning behind having workspace/.fuse/ as a mountpoint, what sorts of problems to we get if we just mount on top of workspace/ itself (thus keeping the paths the same)?
* If we must have a PID, could we suffix it something to take worries about pid_max rollover off the table, perhaps the process start time ticks (from /proc/$pid/stat)
* If we allow mounts on top of a pre-existing .fuse directory and we stick with only using PID for the id, can we pre-populate the directory with all possible pids from 0 to /proc/sys/kernel/pid_max then mount over the top of it thus hiding the symlinks from the sandboxed process

mmjconolly:
[...]

PR #444, opened 2020-07-10
fuse: use unshare to fully redact the workspace

terpstra:
Previously, fuse-wake provided a redacted view of the workspace on
.fuse/pid/ and ran Jobs there. This has two downsides:

- we trust the Job not to access the workspace directly
- the Jobs can notice that .fuse is in the build paths
  - this can result in non-deterministic output
  - this can confuse follow-on Jobs that expect .fuse filenames

This change only affects linux

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -31,6 +31,38 @@
 #include "json5.h"
 #include "execpath.h"
 
+#ifdef __linux__
+// For unshare and bind mount
+#include <sched.h>
+#include <sys/mount.h>
+
+static void write_file(const char *file, const char *content, int len)
+{
+	int fd;
+
+	fd = open(file, O_WRONLY);
+	if (fd < 0) {
+		std::cerr << "open " << file << ": " << strerror(errno) << std::endl;
+		exit(1);
+	}
+
+	if (write(fd, content, len) != len) {
+		std::cerr << "write " << file << ": " << strerror(errno) << std::endl;
+		exit(1);
+	}
+
+	close(fd);
+}
+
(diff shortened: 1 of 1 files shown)


terpstra:
Fixes #439 
Fixes #443

mmjconolly:
> I was going to ask whether unshare works on CentOS 7, but I remember that the CI tests for this repo test CentOS 6 and 7, so this LGTM! FWIW, CentOS 7 runs kernel version 3.10.

I don't think the CI tests here are representative of what a real OS would do, as they're using the container-runtime's kernel

terpstra:
Also, they are using the preload runner, not fuse.

terpstra:
Let me turn off the preload runner and see what happens.

terpstra:
My guess is this will just fail because FUSE is not enabled on these servers...

mmjconolly:
> My guess is this will just fail because FUSE is not enabled on these servers...

I think the FUSE wont have permission as it's already in an unshare being in a container

richardxia:
> I don't think the CI tests here are representative of what a real OS would do, as they're using the container-runtime's kernel

Yes, good point; I forgot about that.

I slightly regret bringing this up at all, since it might be more work than is worth it to try to make this work on CircleCI. I think testing it locally on Ubuntu 16.04 is sufficient.

bsousi5:
Wes, if you could get this in sooner, it would be great. We are hitting this issue when we need to rebuild wake test | ipdelivery after failure on gamma servers

terpstra:
Rebased this against the new CI. I'll add network and directory isolation in follow-up PRs.

richardxia:
Thanks! This is going to help with a lot of stuff!

PR #445, opened 2020-07-20
add ctags and etags support;

davidmlw:
1. adjust to master branch and it works.
2. for some reason the local maked version is ok, but the installing waked version will receive "address boundary error".

### DIFF
--- a/src/markup.cpp
+++ b/src/markup.cpp
@@ -20,8 +20,10 @@
 #include "execpath.h"
 #include "json5.h"
 #include <set>
+#include <map>
 #include <vector>
 #include <fstream>
+#include <sstream>
 #include <assert.h>
 
 struct ParanOrder {
@@ -194,4 +196,71 @@ void markup_html(std::ostream &os, Expr *root) {
   os << "<script type=\"wake\">";
   JSONRender(os).render(root);
   os << "</script>" << std::endl;
+}
+
+void markup_ctags(std::ostream &os, Expr *root, const std::vector<std::pair<std::string, std::string>>& globals) {
+  for (auto &g : globals) {
+    Expr *e = root;
+    while (e && e->type == &DefBinding::type) {
+      DefBinding *d = static_cast<DefBinding*>(e);
+      e = d->body.get();
+      auto i = d->order.find(g.second);
+      if (i != d->order.end()) {
+        int idx = i->second.index;
+        Expr *v = idx < (int)d->val.size() ? d->val[idx].get() : d->fun[idx-d->val.size()].get();
+
+        os << g.first << "\t" << v->location.filename << "\t" << v->location.start.row << std::endl;
+      }
+    }
+  }
+}
+
+void markup_etags(std::ostream &os, Expr *root, const std::vector<std::pair<std::string, std::string>>& globals) {
+
+  typedef std::vector<std::pair<std::string, Location>> Symbols;
+  std::map<std::string, Symbols> files;
+
+  for (auto &g : globals) {
+    Expr *e = root;
+    while (e && e->type == &DefBinding::type) {
+      DefBinding *d = static_cast<DefBinding*>(e);
+      e = d->body.get();
+      auto i = d->order.find(g.second);
+      if (i != d->order.end()) {
+        int idx = i->second.index;
+        Expr *v = idx < (int)d->val.size() ? d->val[idx].get() : d->fun[idx-d->val.size()].get();
+        std::string filename = v->location.filename;
+        std::map<std::string, Symbols>::iterator it;
+        if ((it = files.find(filename)) != files.end()) {
+          (*it).second.push_back(std::make_pair(g.first, v->location));
+        }
+        else {
+          Symbols symbols;
+          symbols.push_back(std::make_pair(g.first, v->location));
+          files.insert(std::make_pair(filename,symbols));
+        }
+      }
+    }
+  }
+
+  for(auto file : files) {
+    std::ifstream ifs(file.first.c_str(), std::ios_base::in);
+    std::vector<std::string> lines;
+    std::string line;
+    while (std::getline(ifs, line)) {
+      lines.push_back(line);
+    }
+
+    std::stringbuf buffer;
+    std::ostream bufos (&buffer);
+    for(auto symbol : file.second) {
+      auto row = symbol.second.start.row;
+      auto bytes = symbol.second.start.bytes;
+      assert(unsigned(row) <= lines.size());
+      bufos << lines[row-1] << char(0x7f) << symbol.first<< char(0x01) << row << "," << bytes << std::endl;
+    }
+
+    os << char(0x0c) << std::endl << file.first << "," << buffer.str().size() << std::endl;
+    os << buffer.str();
+  }
 }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -80,6 +80,8 @@ void print_help(const char *argv0) {
     << "    --version        Print the version of wake on standard output"               << std::endl
     << "    --html           Print all wake source files as cross-referenced HTML"       << std::endl
     << "    --globals -g     Print global symbols made available to all wake files"      << std::endl
+    << "    --ctags          Generate ctags file"                                        << std::endl
+    << "    --etags          Generate etags file"                                        << std::endl
     << "    --exports -e     Print symbols exported by the selected package (see --in)"  << std::endl
     << "    --help    -h     Print this help message and exit"                           << std::endl
     << std::endl;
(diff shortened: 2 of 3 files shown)


terpstra:
I don't know how to review this, as I don't know how to use it. If you can find someone who uses whatever editor this supports and knows wake, I'd be happy to merge this PR.

terpstra:
Although, it appears to only emit information about global variables, of which there should be few post 0.18...?

richardxia:
When I tried running this on a relatively large Wake codebase, I got a segfault. When I ran it with gdb, it looked like there were many, many stack frames, so I'm guessing that it was a stack overflow. I'm not very familiar with the Wake C++ codebase, but is there a chance that we are infinite recursing somewhere here?

davidmlw:
Many thanks @richardxia . I also noticed this segfault. While the local bin/wake is right. The wake installed to the system is not. I suspect that the Wake building rules build different things to Makefile rules.

`make` generated local bin/wake has no segfault, while the bin/wake-native-cpp11-release has got segfault. What is the difference of these two `wake`?

davidmlw:
I got it. When `wake --ctags` get segfault, `wake --global` get segfault, even without this patch.

terpstra:
The segfault should be fixed now by #455.

src/markup.cpp
```diff
@@ -195,3 +197,70 @@ void markup_html(std::ostream &os, Expr *root) {
...
+  for (auto &g : globals) {
+    Expr *e = root;
+    while (e && e->type == &DefBinding::type) {
+      DefBinding *d = static_cast<DefBinding*>(e);
+      e = d->body.get();
+      auto i = d->order.find(g.second);
+      if (i != d->order.end()) {
+        int idx = i->second.index;
+        Expr *v = idx < (int)d->val.size() ? d->val[idx].get() : d->fun[idx-d->val.size()].get();
+
```
terpstra, line 211: This algorithm is now copy-pasted into three places. It would probably be more DRY to transform 'globals' into a suitable form before handing it to these functions and/or the normal exports dumping algorithm.

terpstra:
I am concerned with how these flags interact with the package system.

The patch appears to have originally been written with a focus on the now deprecated globals.

I think there are three ways forward:
- --ctags and --etags just modify the output format of -e and -g
- --ctags and --etags imply -e (not -g as currently)
- --ctags and --etags dump symbols for all exports and all globals from all packages

Do these tools have a way to indicate which package provide a symbol, so the user can add a suitable `import` to their code?

richardxia:
> Do these tools have a way to indicate which package provided a symbol, so the user can add a suitable import to their code?

I'm only going to speak for exuberant ctags and vim. I'm used to ctags being created for all symbols, not just publicly exposed ones. In vim, when you want to jump to a definition, if there is a name conflict, it will just jump to the first one, I think based on the order in the tags file. There is a separate command for listing all definitions with the same name, which you can use to jump to a definition based on the file name.

A few more things I noticed while trying this patch out:

1. The tags file is not sorted, which causes the following vim error when trying to jump to a definition:

    ```
    E432: Tags file not sorted: tags
    ```

2. `topic`s do not seem to be handled correctly, and in the tags file they show up as `topic topicName ./path/to/file.wake 123`. They should only show up as `topicName ./path/to/file.wake 123`.
3. It looks like we are the missing constructors of `tuple`s in the tags files.

davidmlw:
> I am concerned with how these flags interact with the package system.
> 
> The patch appears to have originally been written with a focus on the now deprecated globals.
> 
> I think there are three ways forward:
> 
> * --ctags and --etags just modify the output format of -e and -g
> * --ctags and --etags imply -e (not -g as currently)
> * --ctags and --etags dump symbols for all exports and all globals from all packages
> 
> Do these tools have a way to indicate which package provided a symbol, so the user can add a suitable `import` to their code?

Of course the best way is jumping according to the current wake file "import" list. But this should be beyond TAGS/tags functionality. 

TAGS and tags are not modern style of jumping and are not as smart as current tools. I would suggest the third way "dump symbols for all exports and all globals from all packages". Usually vim/emacs will provide a way to select when there are multiple matches.

davidmlw:
> > Do these tools have a way to indicate which package provided a symbol, so the user can add a suitable import to their code?
> 
> I'm only going to speak for exuberant ctags and vim. I'm used to ctags being created for all symbols, not just publicly exposed ones. In vim, when you want to jump to a definition, if there is a name conflict, it will just jump to the first one, I think based on the order in the tags file. There is a separate command for listing all definitions with the same name, which you can use to jump to a definition based on the file name.
> 
> A few more things I noticed while trying this patch out:
> 
> 1. The tags file is not sorted, which causes the following vim error when trying to jump to a definition:
>    ```
>    E432: Tags file not sorted: tags
>    ```
> 2. `topic`s do not seem to be handled correctly, and in the tags file they show up as `topic topicName ./path/to/file.wake 123`. They should only show up as `topicName ./path/to/file.wake 123`.
> 3. It looks like we are the missing constructors of `tuple`s in the tags files.

I will try to print sorted tags, revise the right topic format. For the tuple definitions, I didn't see examples in `main.cpp`. As @terpstra said, I just re-formatted the `-e, -g` information.

davidmlw:
@richardxia I didn't met `E432` with vim version 7.4.6. And I noticed that even TAGS file works for Vim. I can use the tags or TAGS to do tag jumping in Vim. Maybe my workspace is not complex enough. Is `E432` needs dumplicate tags in different packages, then they need some sorting?

I just add tuple definition in the tags files.

While I noticed that `topic` is missing now, even in `global` printing. @terpstra .

richardxia:
> @richardxia I didn't met `E432` with vim version 7.4.6. And I noticed that even TAGS file works for Vim. I can use the tags or TAGS to do tag jumping in Vim. Maybe my workspace is not complex enough. Is `E432` needs dumplicate tags in different packages, then they need some sorting?

I'm also running vim 7.4. The tags file does indeed work for me for most terms, but only some hit that issue.

I took a few more steps to try to narrow down the issue, and it seems that it only affects terms that are alphabetically almost at the end. In my `tags` file, the `topic`s end up at the very bottom, regardless of sorting order, and it seems that if I attempt to look up any non-topic names at the bottom (for me, starting with the letter `w`), it hits the error. My guess is that vim must be attempting to do a binary search in the file, and if you search for a term that causes the binary search to hit the `topic`s, then it will notice that your terms are not in alphabetical order and then abort.

davidmlw:
I see, Maybe it is because the current implemention seperates defs, topics and types and does not sort them in one line. I will try to solve this.

> > @richardxia I didn't met `E432` with vim version 7.4.6. And I noticed that even TAGS file works for Vim. I can use the tags or TAGS to do tag jumping in Vim. Maybe my workspace is not complex enough. Is `E432` needs dumplicate tags in different packages, then they need some sorting?
> 
> I'm also running vim 7.4. The tags file does indeed work for me for most terms, but only some hit that issue.
> 
> I took a few more steps to try to narrow down the issue, and it seems that it only affects terms that are alphabetically almost at the end. In my `tags` file, the `topic`s end up at the very bottom, regardless of sorting order, and it seems that if I attempt to look up any non-topic names at the bottom (for me, starting with the letter `w`), it hits the error. My guess is that vim must be attempting to do a binary search in the file, and if you search for a term that causes the binary search to hit the `topic`s, then it will notice that your terms are not in alphabetical order and then abort.

davidmlw:
I just add line to sort all the collected symbols before dumping to tags. Hope this will solve `E432`.

richardxia:
I can confirm the latest changes seem to work well for me. I think @terpstra's comments about the code duplication are still valid and should be addressed, though.

davidmlw:
> I can confirm the latest changes seem to work well for me. I think @terpstra's comments about the code duplication are still valid and should be addressed, though.

I also noticed that `topic` symbols are missing after the segfault error is resolved. They are also missing in `--global`. @terpstra

terpstra:
> While I noticed that `topic` is missing now, even in `global` printing. @terpstra .
Works for me?

davidmlw:
> > While I noticed that `topic` is missing now, even in `global` printing. @terpstra .
> 
> Works for me?

I know the reason. `global` is right. I touched topic dumping in the patch. I have revised. 

I think it is ok to merge now.

terpstra:
You have not addressed any of my concerns regarding how this works with multiple packages.

davidmlw:
> I am concerned with how these flags interact with the package system.
> 
> The patch appears to have originally been written with a focus on the now deprecated globals.
> 
> I think there are three ways forward:
> 
> * --ctags and --etags just modify the output format of -e and -g
I think the current implementation is this way.

> * --ctags and --etags imply -e (not -g as currently)
And I think `wake --in package --ctags` works as user expected.

> * --ctags and --etags dump symbols for all exports and all globals from all packages
I dont know how to implement this, which needs elaborating all the available packages.
> 
> Do these tools have a way to indicate which package provided a symbol, so the user can add a suitable `import` to their code?

Is this resolve your question? I just re-format your -g and -e outputs. Others complex need further considerations. While I think the current version is usable.

PR #452, opened 2020-07-25
CircleCI: Enable fuse

terpstra:
This PR switches Circle CI to using VMs for builds. Those VMs can then run docker with the appropriate flags to enable FUSE.

This is helpful, because we're about to start making a bunch of improvements to FUSE.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   tarball:
     machine:
       image: ubuntu-1604:202007-01
-#    resource_class: large
+#    resource_class: large # paywalled feature: 4 cores
     steps:
       - checkout
       - run: sudo apt-get update && sudo apt-get install -y build-essential libfuse-dev libsqlite3-dev libgmp-dev libncurses5-dev pkg-config git g++ gcc libre2-dev python3-sphinx
@@ -21,7 +21,7 @@ jobs:
   alpine:
     machine:
       image: ubuntu-1604:202007-01
-#      docker_layer_caching: true
+#      docker_layer_caching: true # paywalled feature: retain docker build results between runs
 #    resource_class: large
     steps:
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,9 @@
 version: 2.1
 jobs:
   tarball:
-    docker:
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -1,211 +1,180 @@
 version: 2.1
 jobs:
   tarball:
-    docker:
-      - image: circleci/buildpack-deps:stretch
+    machine:
+      image: ubuntu-1604:202007-01
+#    resource_class: large
```
richardxia, line 6: You have several of these commented out keys. Do you want to remove them all before merging?

.circleci/config.yml
```diff
@@ -1,211 +1,180 @@
...
-          cp lib/.libs/libfuse.a /usr/lib
-          cd ..
           tar xJf /tmp/workspace/wake_*.tar.xz
-          cd wake-*
-          USE_FUSE_WAKE=0 make static
-          cd ..
-          mkdir -p release/alpine
-          mv wake-*/wake-static_* release/alpine
+          docker run --rm --mount type=bind,source=$PWD,target=/build --device /dev/fuse --cap-add SYS_ADMIN --security-opt apparmor:unconfined wake-alpine make -C wake-* static
+          docker run --rm --mount type=bind,source=$PWD,target=/build wake-alpine /bin/sh -c "tar xJf wake-*/wake-static_*; ./wake-*/bin/wake --no-workspace -x 'shellJob \"ls /\" Nil | getJobStdout'"
```
richardxia, line 33: I'm going to assume that you figured out all of these flags from trial and error? Hope that wasn't too painful!

.circleci/config.yml
terpstra: They are features I would like to use, but cost money.
It took me time to find the settings, so I wanted to keep them in case we ever pay circle CI.
I should probably add a comment to that effect.

.circleci/config.yml
terpstra: google helped a lot. it was painful.

.circleci/config.yml
terpstra: The only reason I didn't wholesale switch to github actions: at least circle CI *can* support more RAM and cores. It seems the only way to get more with github actions is to host the runners yourself!

PR #454, opened 2020-07-30
Provide resource overrides within runJob

mmjconolly:
* Adds resource overrides to `runJob`. 
* Adds `Label` concept to `Plan`.

The labels are included to allow the user to be very specific about which job they want to have control of.

The user would publish a function that matches a label and a resource string, which returns a new (or original) resource string
```
def matchAA10 label orig =
  if orig ==~ "a/a/1.0" && label ==~ "compile-rocketchip" then
    def _ = println "Using 'a/a/1.1' instead of 'a/a/1.0'"
    "a/a/1.1"
  else
    orig

# The user could ignore labels too
def matchBB _ orig = if orig ==~ "b/b/1.0" then "b/b/2.0" else orig

publish resourceOverrides = matchAA10, matchBB, Nil
```

```
$ wake -x 'makePlan ("/bin/true", Nil) Nil | setPlanLabel "compile-rocketchip" | setPlanResources ("a/a/1.0", Nil) | runJob'
Using 'a/a/1.1' instead of 'a/a/1.0'
Job 1
```

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -91,6 +91,7 @@ tuple Plan =
   export Usage:        Usage        # User-supplied usage prediction; overruled by database statistics (if any)
   export FnInputs:     (List String => List String) # Modify the Runner's reported inputs  (files read)
   export FnOutputs:    (List String => List String) # Modify the Runner's reported outputs (files created)
+  export Label:        String       # Human-readable label chosen by Plan author to describe the job intent
 
 def isOnce = match _
   ReRun = False
@@ -143,7 +144,7 @@ export def editPlanShare f =
 # Get a unique hash-code for the job
 export def getPlanHash plan =
   def signature cmd env dir stdin = prim "hash"
-  def Plan cmd _ env dir stdin _ _ _ _ _ _ _ _ _ _ = plan
+  def Plan cmd _ env dir stdin _ _ _ _ _ _ _ _ _ _ _ = plan
   signature cmd env dir stdin
 
 # The criteria which determine if Job execution can be skipped:
@@ -161,7 +162,7 @@ export def getPlanHash plan =
 # Set reasonable defaults for all Plan arguments
 def id x = x
 export def makePlan cmd visible =
-  Plan cmd visible environment "." "" logVerbose logWarn Normal Share False Nil (\_ True) defaultUsage id id
+  Plan cmd visible environment "." "" logVerbose logWarn Normal Share False Nil (\_ True) defaultUsage id id ""
 
 export def makeShellPlan script visible =
   makePlan (which "dash", "-c", script, Nil) visible
@@ -279,7 +280,7 @@ def jobLog stdout stderr echo =
   def bit = if thresh <= logLevelRaw then 1 else 0
   (bit << 4) + (num (stderr logLevel) << 2) + num (stdout logLevel)
 
-export def runJobWith (Runner _ _ run) (Plan cmd vis env dir stdin stdout stderr echo pers _ res _ usage finputs foutputs) =
+export def runJobWith (Runner _ _ run) (Plan cmd vis env dir stdin stdout stderr echo pers _ res _ usage finputs foutputs _) =
   runJobImp cmd env dir stdin res usage finputs foutputs vis pers run (jobLog stdout stderr echo)
 
 data RunnerOption =
@@ -288,7 +289,8 @@ data RunnerOption =
 
 # Run the job!
 export def runJob p = match p
-  Plan cmd vis env dir stdin stdout stderr echo pers lo res rf usage finputs foutputs =
+  Plan cmd vis env dir stdin stdout stderr echo pers lo res rf usage finputs foutputs label =
+    def resources = override label res
     # Transform the 'List Runner' into 'List RunnerOption'
     def qualify runner = match runner
       Runner name _ _ if ! rf runner = Reject "{name}: rejected by Plan"
@@ -303,7 +305,7 @@ export def runJob p = match p
         if score >. bests then Pair score (Some fn) else acc
     def log = jobLog stdout stderr echo
     match (opts | foldl best (Pair 0.0 None) | getPairSecond)
-      Some r = runJobImp cmd env dir stdin res usage finputs foutputs vis pers r log
+      Some r = runJobImp cmd env dir stdin resources usage finputs foutputs vis pers r log
       None =
         def create dir stdin env cmd signature visible keep log = prim "job_create"
         def badfinish job e = prim "job_fail_finish"
@@ -317,6 +319,15 @@ export def runJob p = match p
         # Make sure badlaunch completes before badfinish
         def _ = wait (\_ badfinish job error) (badlaunch job error)
         job
+
+# Provide a matching function and a replacement resource to override a resource
+# A non matching function should return the original resource
+global topic resourceOverrides: ((label: String) => (origResource: String) => (newResource: String))
+
+def override label defaultResources =
+  def fns = subscribe resourceOverrides
+  def fnsWithLabel resource = map (_ label resource) fns
+  map (\r fnsWithLabel r) defaultResources | flatten
 
 def toUsage (Pair (Pair status runtime) (Pair (Pair cputime membytes) (Pair ibytes obytes))) =
   Usage status runtime cputime membytes ibytes obytes
@@ -551,7 +562,7 @@ target hashcode f =
   def reuse = get f
   if reuse !=* "" then reuse else
     def hashPlan cmd  =
-      Plan cmd Nil Nil "." "" logNever logError Verbose ReRun True Nil (\_ True) hashUsage id id
+      Plan cmd Nil Nil "." "" logNever logError Verbose ReRun True Nil (\_ True) hashUsage id id ""
     def job = hashPlan ("<hash>", f, Nil) | runJobWith localRunner
     def hash =
       job.getJobStdout


share/wake/lib/system/job.wake
```diff
@@ -318,6 +320,15 @@ export def runJob p = match p
...
         job
 
+# Provide a matching function and a replacement resource to override a resource
+# A non matching function should return the original resource
+global topic resourceOverrides: ((label: String) => (origResource: String) => (newResource: String))
+
+def override label defaultResources =
+  def fns = subscribe resourceOverrides
+  def fnsWithLabel resource = map (_ label resource) fns
+  map (\r fnsWithLabel r) defaultResources | flatten
```
richardxia, line 330:
```suggestion
  def compose f g = f ∘ g  # Really? No ASCII name for the compose function?
  def id x = x
  def composeAll functions = foldl compose id functions

  def fnsWithLabel resource = map (_ label resource) fns
  def overrideResource = composeAll fns
  
  map overrideResource originalResources
```

If I'm reading the code correctly, I think your code will end up duplicating the resource string for any resource that isn't matched by the current override function. For example, if I have two override functions that map `foo/1.0` -> `foo/2.0` and `bar/1.0` -> `bar/2.0`, then if I apply it to a job that has `resources = "foo/1.0", "bar/1.0", Nil`, I'll get `"foo/2.0", "foo/1.0", "bar/1.0", "bar/2.0", Nil` as the final set of resources.

I think what you want to do is iterate over each resource and then apply each override function in a chain.

share/wake/lib/system/job.wake
```diff
@@ -91,6 +91,7 @@ tuple Plan =
   export Usage:        Usage        # User-supplied usage prediction; overruled by database statistics (if any)
   export FnInputs:     (List String => List String) # Modify the Runner's reported inputs  (files read)
   export FnOutputs:    (List String => List String) # Modify the Runner's reported outputs (files created)
+  export Label:        String       # Human-readable label chosen by Plan author to describe the job intent
```
richardxia, line 94:
```suggestion
  export Label:        String       # Machine-readable label chosen by Plan author to describe the job intent
```

I feel that "human-readable" might be misleading here, since it's actually being used to programmatically filter for specific plans. It's less important that it is human-readable and more important that it is a machine-compariable label.

terpstra:
Why is this limited to only modifying resources? It seems it would be more useful if it could completely rewrite the entire Plan. I'm also unsure that this belongs in the Plan API itself. This seems like a potentially separable problem.

I could well imagine a completely orthogonal API along these lines:
```
topic planOverrides: (label: String) => (plan: Plan) => Plan
def applyOverrides label plan =
  ...
def myJob =
  makePlan ...
  | applyOverrides "compile-rocket"
  | runJob
```

terpstra:
I'm not sure which argument order is best for the override function. You could also go:
```
topic planOverrides: (plan: Plan) => (label: String) => Plan
def applyOverrides label plan =
  def overrides = subscribe planOverrides
  foldr (_ _ label) plan overrides
def myOverride plan = match _
   "compile-rocket" = setPlanResources ("x", "y", Nil) plan
  _ = plan
```

share/wake/lib/system/job.wake
```diff
@@ -318,6 +320,15 @@ export def runJob p = match p
         def _ = wait (\_ badfinish job error) (badlaunch job error)
         job
 
+# Provide a matching function and a replacement resource to override a resource
+# A non matching function should return the original resource
+global topic resourceOverrides: ((label: String) => (origResource: String) => (newResource: String))
```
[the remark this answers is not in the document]
terpstra: FYI, if this is to be applied to wake/master, please do not use `global` any more.

terpstra:
If you plan to truly use the label as a key, it probably makes sense to pay once to build a binary search tree out of the keys:
```
# Build a binary search tree with all the overrides in it
def planTree = listToTreeMulti (_.getPairFirst <=>* _.getPairFirst) (subscribe planOverrides)
export topic planOverrides: Pair (label: String) (override: Plan => Plan)
export def applyOverrides label plan =
  def key = Pair label (_)
  def Pair matches _ = tequal key planTree
  foldr (_.getPairSecond _) plan matches

# Something doing rocket-chip
def myJob =
  makePlan ...
  | applyOverrides "rocket-chip"
  | applyOverrides "something-else"
  | runJob

# Apply changes to rocket-chip
publish planOverrides = Pair "rocket-chip" myOverride, Nil
def myOverride plan =
  def upgrade = match _
    "vcs/1.0" = "vcs/1.1"
    "whatever/2.3" = "something-else/0.0"
    x = x
  plan
  | editPlanResources (map upgrade _)
  | setPlanEnvironment ("PATH=/tmp", Nil)
```

terpstra:
I'm having second thoughts about using the label as a tree search key. You might want to match the label with a RegExp in your override.

BTW, in case it's not obvious: I think being able to modify a Plan defined in one project from another project is a great idea.

terpstra:
I'm torn. The binary tree approach obviously scales better if we use this API heavily. On the other hand, I could easily envision a desire to apply a resource remap function to all Plans. 

Perhaps these are two different APIs?
- remap the Plan for a selected label
- remap resource x to y for all labels

I don't know.

mmjconolly:
> Why is this limited to only modifying resources? It seems it would be more useful if it could completely rewrite the entire Plan.

I initially drafted it with `Plan` but then I couldn't come up with a use case for fields other than `Resources`, what did you have in mind as the other use-cases?

richardxia:
> I could well imagine a completely orthogonal API along these lines:
> 
> ```
> topic planOverrides: (label: String) => (plan: Plan) => Plan
> def applyOverrides label plan =
>   def overrides = subscribe planOverrides
>   foldr (_ label _) plan overrides
> def myOverride label plan = match label
>    "compile-rocket" = setPlanResources ("x", "y", Nil) plan
>   _ = plan
> def myJob =
>   makePlan ...
>   | applyOverrides "compile-rocket"
>   | runJob
> ```

@mmjconolly had suggested something similar earlier (offline), and I had pushed for doing it within `runJob` and with a topic, since the problem with the quoted approach above is that it requires the plan author to explicitly write out the `applyOverrides` line.

Doing it with a pub/sub topic and under the hood of `runJob` allows a consumer of a set of libraries to do things like "override every resource requirement of python 3.7 to python 3.8" without having to explicitly enumerate each plan label.

Re: the two APIs, I think the most generic API would be an arbitrary transform of any Plan based on any property of that Plan, but I could see utility in restricting the API to specific cases.

terpstra:
Why a Plan? I could imagine that we need to modify the command-line for a specifically upgraded tool version, for example.

"it requires the plan author to explicitly write out the applyOverrides line" -- This is not materially different than a `setPlanLabel "xyz"` line. I find the `applyPlanOverrides` much more explicit in what is being done.

If you just want to transform all Plans, that was the intent of the Runner API. I agree that the current API for that needs a refactor. I want Runners to be able to transform plans and then resubmit them for the next Runner.

terpstra:
Thinking about it more, I think it's actually a bad thing to allow every Plan to be manipulated by these overrides. This definitely feels like it should be an opt-in API. If you use a Runner to fulfill a resource, that's fine. But applying a transformation behind the curtains to every Plan? Seems fraught.

mmjconolly:
> But applying a transformation behind the curtains to every Plan? Seems fraught.

Think of it like LD_PRELOAD, its an escape hatch, not for regular day-to-day

mmjconolly:
> If you just want to transform all Plans, that was the intent of the Runner API. I agree that the current API for that needs a refactor. I want Runners to be able to transform plans and then resubmit them for the next Runner.

Is there a way to have the runner to swap out a resource string and then have it re-scored to find the real runner to provide the new resource?

terpstra:
That is how I want the Runner API to work. It is not how it currently works. That said, it would not be too difficult to fix, I think.

terpstra:
Switching perspectives for a moment, giving Plans a label seems like a good idea irrespective of how the `applyOverrides` behaves. This would work well for a GUI view of the Jobs. If you don't have a label, you default to the command-line, but if you do have one, record it in the database. Seeing 'scala-compile <project>' would be a lot more useful than seeing the java command-line.

richardxia:
> Why a Plan? I could imagine that we need to modify the command-line for a specifically upgraded tool version, for example.

I agree that this is a case that may need to happen, but my real concern is about whether the control is in the hands of the API provider or the API consumer.

> "it requires the plan author to explicitly write out the applyOverrides line" -- This is not materially different than a setPlanLabel "xyz" line. I find the applyPlanOverrides much more explicit in what is being done.

I did consider that these are almost syntactically the same, but I think semantically it is about _when_ the transformations are applied. Once you call `applyOverrides`, then you cannot make further modifications to the Plan, or rather, any further modifications cannot be overridden anymore. This makes it feel much less "declarative" because now there is an ordering statements.


> If you just want to transform all Plans, that was the intent of the Runner API. I agree that the current API for that needs a refactor. I want Runners to be able to transform plans and then resubmit them for the next Runner.

Yes, I agree that this is what we want, and I had mistakenly thought that this was already the case until @mmjconolly and I inspected the `runJob` code more carefully yesterday.

---

I think I more or less agree with all of the remaining comments both of you made since I last posted. Is there a quick, but forwards-looking path forward to having chainable Runners?

terpstra:
> I think semantically it is about _when_ the transformations are applied. Once you call `applyOverrides`, then ... any further modifications cannot be overridden anymore.

This sounds like a very useful feature to me. It lets the Plan author decide which parts can be overridden and which not. For Jobs used by wake itself or other very low-level Plans, I can well imagine wanting to allow some fields to be overridden, but not all.
 
> Is there a quick, but forwards-looking path forward to having chainable Runners?

I think I could fix this for 0.19. It should be possible to do in a mostly backwards compatible way.

richardxia:
> This sounds like a very useful feature to me. It lets the Plan author decide which parts can be overridden and which not. For Jobs used by wake itself or other very low-level Plans, I can well imagine wanting to allow some fields to be overridden, but not all.

I think my problem with this is that it is not syntactically obvious that there is an ordering constraint here, and it's also not clear how a Plan author should indicate to a consumer that wants to override a Plan which properties can be overridden. An overrider may write code that sets some field on the Plan only for that to be silently thrown away. The only way to find out is by carefully inspecting the code.

I will acknowledge that I also have a bit of a biased perspective as a consumer/integrator, where I am often running into issues where the Plans are not sufficiently "open" for me to configure what I need without going in and changing the Plan. Maybe we need to more carefully examine the things I run into and distill them into best practices for Plans in terms of what is often good to expose and what is good to keep private.

terpstra:
You don't find there is an order to this?:
```
def job =
  makePlan ...
  | setPlanEnvironment stuff
  | setPlanPersistence more-stuff
  | applyOverrides "compile-scala {project}"
  | setPlanEcho Normal # not allowed to make this job silent!
  | runJob
```

mmjconolly:
> You don't find there is an order to this?:   .....


Sorry I accidentally edited your comment instead of quoting it

For the
```
  | applyOverrides "compile-scala {project}"
```
This looks like the Plan itself has to consent to have overrides, I feel like this misses the intention of having an override (ie, an unconsented hack) at all

terpstra:
> most of the things in the pipeline are `setPlan*` calls, which are normally order independent

Oh?
```
  def plan =
    makePlan ...
    | editPlanEnvironement fn
    | ...
    | setPlanEnvironment xyz
```

richardxia:
> > most of the things in the pipeline are `setPlan*` calls, which are normally order independent
> 
> Oh?
> 
> ```
>   def plan =
>     makePlan ...
>     | editPlanEnvironement fn
>     | ...
>     | setPlanEnvironment xyz
> ```

Yes, I see that that's a contrived example that you will not find in practice. My point is that people treat the `makePlan |` pipelines as a way of updating a record type. This is basically equivalent to the following Scala code with a case class:

```scala
val plan = Plan(
  environment = env,
  localOnly = true,
  command = command,
)
```

terpstra:
I would think most people expect a pipeline to be executed in order. I agree we happen to use pipelines to also update records, but I do not think anyone is under the illusion that those updates are unordered.

richardxia:
> I would think most people expect a pipeline to be executed in order. I agree we happen to use pipelines to also update records, but I do not think anyone is under the illusion that those updates are unordered.

I'm not saying that people think that the pipelines are executed out of order, but I am saying that it is really easy to miss out on one line that _is_ order dependent _because_ we have no order-independent way of setting properties. We have a functional programming language where we are forced back into an imperative programming paradigm.

richardxia:
But anyway, let's talk less about `applyOverrides` and the ordering thing and more about what we want to do in the short term.

I'm willing to punt on that approach if we can get the recursive runner stuff working, and I think separately it is useful to declare a label on each Plan.

terpstra:
I certainly  agree with both of your last two points:
- labels are useful (should they be a required argument to makePlan in v1_wake ?)
- runner API needs to be able to re-dispatch

It's also true that you can get the "behind the scenes fixup based on label" you want entirely using these two things. I do still think this sort of pattern (injecting a point where overrides can be applied) is useful in general, but less useful for Plans if we had the other two things.

PR #458, opened 2020-08-01
Detect and report any unused definitions, topics, or arguments

### LINKED ISSUE #432: Warn on unused variables
It'd be useful to warn on any unused variables, perhaps controlled by a command line flag, since sometimes when refactoring code you may forget to remove some things that are unused. I'd certainly turn this on in a CI lint check if it existed!

terpstra:
This PR detects and warns on any unused definitions, topics, or arguments.
This PR adds a `--fatal-warnings` option, which prevents execution if there are any warnings.

Fixes #432.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -406,10 +406,10 @@ export def preloadRunner =
   makeJSONRunnerPlan preload score
   | makeJSONRunner
 
-def rOK = 0
-def wOK = 1
-def xOK = 2
-def access file mode = prim "access"
+export def rOK = 0
+export def wOK = 1
+export def xOK = 2
+export def access file mode = prim "access"
 
 export def defaultRunner = match sysname
   "Darwin" = fuseRunner
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -179,7 +179,7 @@ export def localRunner =
     Fail e =
       def _ = badlaunch job e
       Fail e
-    Pass (RunnerInput label cmd vis env dir stdin _ _ predict) = match (findSomeFn getPathError vis)
+    Pass (RunnerInput _ cmd vis env dir stdin _ _ predict) = match (findSomeFn getPathError vis)
       Some e =
         def _ = badlaunch job e
         Fail e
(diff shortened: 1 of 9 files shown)


share/wake/lib/system/job.wake
```diff
@@ -404,7 +404,7 @@ export def preloadRunner =
   makeJSONRunnerPlan preload score
   | makeJSONRunner
 
-def rOK = 0
+def _rOK = 0 # unused; remove _ if you use it
 def wOK = 1
 def xOK = 2
 def access file mode = prim "access"
```
mmjconolly, line 410: This should be export/global anyway I think, we have 2 places where it's copy/pasted already

mmjconolly:
> controlled by a command-line flag?

I'm hoping that this check only applies to the code that the user writes, and not the standard library code

If it's just a warning I'm less bothered but if it errors out it would be nicer as a flag since that could be annoying while trying to develop a rule, we can use the flag in CI tests

share/wake/lib/system/job.wake
terpstra: You're probably right. I'll export them.

terpstra:
> I'm hoping that this check only applies to the code that the user writes, and not the standard library code

Well, it applies to the standard library too, but it's easy to make sure the standard library has no warnings.

> If it's just a warning I'm less bothered

It is just a warning. There are surprisingly few violations, even in PCS.

terpstra:
This PR currently detects neither unused types nor unused arguments.

Another thing I was thinking to detect is exporting a definition which uses an unexported type.

terpstra:
The guard / pattern restructuring is necessary, because of expressions like:
```
    def example = match _
      Triple a b c if a = b
      _ = 66
```
The guard `a` and function body `b` are actually different functions after the lowering which reduces pattern matches. So without the last commit, the first function complained that `b` and `c` are unused while the second complained that `a` and `c` are unused. The change moves those two functions into a common function that looks notionally like: `\a\b\c Pair (\_ a) (\_ b)`. That way it only complains that `c` is unused.

There are actually a few other warnings wake prints currently. Ambiguous import, for example. I think there are perhaps three changes that should be made:
- prefix all warnings with something like 'warning: '
- accumulate whether or not a warning was produced and offer a '--stop-on-warnings' option
- suggest how the user should fix the warnings in the error message (add an `_` prefix, etc)

terpstra:
I've addressed all the three outstanding issues I had with this PR and will merge it as soon as it passes.

PR #460, opened 2020-08-06
export Tree type

albertchen-sifive:
export `Tree` so we can do
```
tuple ASDF =
   LKJ: Tree String
```


terpstra:
I've been meaning to add a warning to detect if you export a definition with types that are not exported. That would have caught this...

PR #462, opened 2020-08-06
Define the v0.19 API

terpstra:
By creating a package with a defined version, we make it possible for users to declare the version of the wake API they expect. Then, when we make breaking API changes, only those who opt-in see the new API.


terpstra:
In other words, people can write this at the top of their files:
```
from v0_19_wake import _
```
That will get you a stable API that does not change when the API evolves.

By contrast,
```
from wake import _
```
will get you the newest API that wake supports.

mmjconolly:
What does it do if someone doesn't write a `from ...` line at the top?
How does it interact if one source file wants one api and a second wants a different api?

terpstra:
If you don't have ANY imports, wake automatically inserts a `from wake import _` into your file. This was for backwards compatibility with pre-0.18 wake files.

It is totally safe to mix uses of different versions of the wake API between files. That was one of the major design goals of the package system. It requires some discipline on the part of the API designer to keep this interoperability working, but this is not a burden born by the API user.

Re:README; You're right. I forgot to update the package docs with the new behaviour.

PR #463, opened 2020-08-06
fuse: support unshare isolation for user, host, net, and workspace

terpstra:
These isolation levels are supported by requesting resources:
- isolate/workspace; makes the workspace appear to be /var/cache/wake
- isolate/user; makes the build user appear to be root
- isolate/host; makes the hostname appear to be build.local
- isolate/net; disables network access

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -167,22 +167,39 @@ int main(int argc, char *argv[])
 		if (stdin.empty()) stdin = "/dev/null";
 
 #ifdef __linux__
-		// Allow overriding this to; e.g /build
-		std::string workspace = cwd;
-
 		uid_t real_euid = geteuid();
 		gid_t real_egid = getegid();
 
+		// Allow overriding this to; e.g /mnt
+		std::string workspace = cwd;
+		uid_t euid = real_euid;
+		gid_t egid = real_egid;
+		int flags = CLONE_NEWNS|CLONE_NEWUSER;
+
+		for (auto &res : jast.get("resources").children) {
+		  std::string &key = res.second.value;
+		  if (key == "isolate/workspace") workspace = "/mnt";
+		  if (key == "isolate/user") { euid = egid = 0; }
+		  if (key == "isolate/host") flags |= CLONE_NEWUTS;
+		  if (key == "isolate/net") flags |= CLONE_NEWNET;
+		}
+
 		// Enter a new mount namespace we can control
-		if (0 != unshare(CLONE_NEWNS|CLONE_NEWUSER)) {
+		if (0 != unshare(flags)) {
 			std::cerr << "unshare: " << strerror(errno) << std::endl;
 			exit(1);
 		}
 
-		// Return to our original UID (otherwise we are nobody)
+		// Wipe out our hostname
+		if (0 != (flags & CLONE_NEWUTS)) {
+			sethostname("build", 5);
+			setdomainname("local", 5);
+		}
+
+		// Map our UID to either our original UID or root
 		write_file("/proc/self/setgroups", "deny", 4);
-		map_id("/proc/self/uid_map", real_euid, real_euid);
-		map_id("/proc/self/gid_map", real_egid, real_egid);
+		map_id("/proc/self/uid_map", euid, real_euid);
+		map_id("/proc/self/gid_map", egid, real_egid);
 
 		// Mount the fuse-visibility-protected view onto the workspace
 		if (0 != mount(rpath.c_str(), workspace.c_str(), NULL, MS_BIND, NULL)) {
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -90,8 +90,9 @@ int main(int argc, char *argv[])
 	if (!JAST::parse(json, std::cerr, jast))
 		return 1;
 
-	std::string name = std::to_string(getpid());
-	std::string daemon = find_execpath() + "/fuse-waked";
+	std::string exedir = find_execpath();
+	std::string daemon = exedir + "/fuse-waked";
+	std::string name  = std::to_string(getpid());
 	std::string cwd   = get_cwd();
 	std::string mpath = cwd + "/.fuse";
 	std::string fpath = mpath + "/.f.fuse-waked";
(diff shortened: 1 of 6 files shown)


terpstra:
Going forward, we should probably enable all of these by default and require you to opt-out.

fuse/fuse.cpp
```diff
@@ -167,22 +167,39 @@ int main(int argc, char *argv[])
...
 
+		// Allow overriding this to; e.g /mnt
+		std::string workspace = cwd;
+		uid_t euid = real_euid;
+		gid_t egid = real_egid;
+		int flags = CLONE_NEWNS|CLONE_NEWUSER;
+
+		for (auto &res : jast.get("resources").children) {
+		  std::string &key = res.second.value;
+		  if (key == "isolate/workspace") workspace = "/mnt";
```
mmjconolly, line 181: I think the path shouldn't be directly on top of a common mount point.
Perhaps something like `/mnt/wake` or `/tmp/wake`? 

Not that I think it's likely, but someone might have some toolchain (via a Resource) that was in `/mnt` that is now hidden

mmjconolly:
> Going forward, we should probably enable all of these by default and require you to opt-out.

I we do this it's likely a good time to flip the default on `PATH=/usr/bin:/bin` too, since there'd be plenty of Plans to manually edit

fuse/fuse.cpp
terpstra: > If we do this it's likely a good time to flip the default on `PATH=/usr/bin:/bin` too, since there'd be plenty of Plans to manually edit

Definitely.

Yeah. Your proposal to deal with /tmp/wake existing was to mount tmpfs over /tmp before doing a mkdir(/tmp/wake). That then hides everything in /tmp, similar to how this hides everything in /mnt.

What I would really like to do is mkdir(/build), but AFAIK there is no way to fake that.

mmjconolly: Hmm, hiding /tmp could be an advantage

We could also do an "if /build exists use it instead"

mmjconolly: If you are allowed to mount an overlay on top of `/` you could create `/build`

terpstra: Looking through the FHS, the best defined candidates to hide are:
- /boot (probably ok to hide)
- /mnt (Temporarily mounted filesystems)

Here is another option:
/var/cache/wake
... and make that directory as part of wake installation.

What's a bit unfortunate is that /var is not under /usr, so it's always a pain to work with when you have a relocatable package like wake that can live in /usr/local/...

I would propose to use /var/cache/wake (regardless of where wake is installed).

terpstra: Using BINPATH/../lib/wake/build might be a better choice, but it's not exactly the intended FHS for /usr/lib.

terpstra: Building on your (if it exists idea), how about this policy:
Use /var/cache/wake if it exists
Otherwise, use BINPATH/../lib/wake/build and create that as part of the wake install.

mmjconolly: > ... and make that directory as part of wake installation.

We could do that tactic with `/build` (or perhaps `/wake`, much like Nix does `/nix`), if they're using deb/rpm to install then they have root to make that directory.

mmjconolly: I don't think `/var/cache/wake` handles the case where wake was compiled from source by a non-root user, who hasn't previously installed wake before as root

terpstra: Ubuntu packages are not allowed to create directories in the top of the filesystem (/build), so /var/cache/wake is a better option since that is FHS compliant and will work on all distros.

I agree that you may not have permission to create /var/cache/wake (or /build or /wake), so we need a fallback. Since a non-root user can pick where they install wake, I've used a path relative to the users selected installation location as the fallback.

fuse/fuse.cpp
```diff
@@ -167,22 +168,45 @@ int main(int argc, char *argv[])
...
+		for (auto &res : jast.get("resources").children) {
+			std::string &key = res.second.value;
+			if (key == "isolate/user") euid = egid = 0;
+			if (key == "isolate/host") flags |= CLONE_NEWUTS;
+			if (key == "isolate/net") flags |= CLONE_NEWNET;
+			if (key == "isolate/workspace") {
+				if (access("/var/cache/wake", R_OK) == 0) {
+					workspace = "/var/cache/wake";
+				} else {
+					workspace = exedir + "/../../build/wake";
```
richardxia, line 189: Maybe this should be configurable via environment variable? You could still have this as the final fallback, but it could go in the order of 1) env var, 2) /var/cache, 3) relative to exedir?

share/wake/lib/system/job.wake
```diff
@@ -391,6 +391,13 @@ export def getJobStatus job = match (getJobReport job)
     else Signaled (-status)
 
 # Implement FUSE-based Runner
+# The FUSE runner on linux supports a few isolation options via resources:
+# - "isolate/user": the Job will appear to run as root
+# - "isolate/host": make the hostname appear to be "build.local"
+# - "isolate/net": disables network access
+# - "isolate/workspace": makes the build appear run in /var/cache/wake
+#    ... if /var/cache/wake does not exist, a directory 'build/wake' is
+#        used relative to where wake has been installed
```
richardxia, line 400: I'd request that this be a bit more precisely worded (and also can we add this all to a README?). I think we should make it clear _what_ appears to be in that dir and when. Here's a suggested rewording to capture what I think is missing:

> Make jobs appear to run in /var/cache/wake. Both the perceived workspace dir and the current working directory will be set to that directory. If /var/cache/wake does not exist, ... <remainder of your sentence>

terpstra: What README? This does not belong in the README.md of the wake project.
Do you mean in a file in the directory, saying why the directory exists?

fuse/fuse.cpp
terpstra: I would rather not make this any more complicated than it already is.

share/wake/lib/system/job.wake
richardxia: No, I think this needs to be in the README. Aren't the `isolate/*` resource strings meant to be used by Plan authors?

terpstra: What README are you talking about?
Do you mean the tutorial?

richardxia: I am using the term README a bit too loosely and incorrectly. What I mean is something documenting the Plan/Job/resource system.

I guess we don't have on that already.

terpstra: I see. I agree we should have documentation on the Plan/Job/resource system. It's on my TODO list.
However, this change sounds like a follow-up PR: write documentation.

PR #464, opened 2020-08-06
Plan labels

terpstra:
Add labels to Plans.

These are recorded in the database and shown when inspecting Jobs.
Progress bars use the label name for better readability.

### DIFF
--- a/re2c/re2c.wake
+++ b/re2c/re2c.wake
@@ -36,7 +36,11 @@ def re2cOk Unit =
 def re2cReal file =
   def cpp = replace `\.re$` ".cpp" file.getPathName
   def cmdline = which "re2c", "-8", "--tags", "--no-generation-date", "--input-encoding", "utf8", file.getPathName, "-o", cpp, Nil
-  def result = job cmdline (file, Nil) | getJobOutput
+  def result =
+    makePlan cmdline (file, Nil)
+    | setPlanLabel "re2c {cpp}"
+    | runJob
+    | getJobOutput
   def compress = which "gzip", "-nk9", result.getPathName, Nil
   def zip = job compress (result, Nil)
   match zip.getJobStatus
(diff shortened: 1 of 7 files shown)


re2c/re2c.wake
```diff
@@ -36,7 +36,11 @@ def re2cOk Unit =
 def re2cReal file =
   def cpp = replace `\.re$` ".cpp" file.getPathName
   def cmdline = which "re2c", "-8", "--tags", "--no-generation-date", "--input-encoding", "utf8", file.getPathName, "-o", cpp, Nil
-  def result = job cmdline (file, Nil) | getJobOutput
+  def result =
+    makePlan cmdline (file, Nil)
+    | setPlanLabel "re2c {cpp}"
```
richardxia, line 41: This looks pretty different than the Label in https://github.com/sifive/wake/pull/454, so I just wanted to check that this basically has no bearing to the original use case introduced there.

The main practical difference I see is that in the other PR, the label was intended to mean something that code could match on. Here, by embedding the file paths of the inputs of the command into the label, you are defeating the ability to programmatically match based on string, except with substring matching.

The intent of the label in https://github.com/sifive/wake/pull/454 was that many instances of a rule would all share a label, so the label was not necessarily unique in a given run of Wake.

share/wake/lib/system/gcc.wake
```diff
@@ -27,13 +27,17 @@ def staticLFlags  = ("-flto", "-static", "-s", Nil)
 def doCompileC variant gcc flags headers cfile =
   def obj = replace `\.c(pp)?$` ".{variant}.o" cfile.getPathName
   def cmdline = gcc, distinctBy scmp flags ++ ("-c", cfile.getPathName, "-frandom-seed={obj}", "-o", obj, Nil)
-  def result = job cmdline (cfile, headers)
-  result.getJobOutput
+  makePlan cmdline (cfile, headers)
+  | setPlanLabel "compile {replace `^.*/` '' gcc} {obj}"
```
richardxia, line 31: I think I would prefer this to be named `compileC` or `compile C` just to differentiate it from other compile steps that could be invoked in a build graph.

terpstra: It is called 'compile c++ foo.o' or 'compile c foo.o' already.

re2c/re2c.wake
terpstra: I would say it depends how you use the label. Within the context of the wake build "re2c symbol.cpp" is a pretty useful key. As you note, you can still look them up using a RegExp or whatever.

terpstra: However, I do agree the intended use of this label is mostly aimed at improving the progress view of the build, as compared to the other PR.

share/wake/lib/system/gcc.wake
richardxia: Ah, I didn't carefully read the code to understand what `gcc` was.

PR #465, opened 2020-08-06
database: add a schema version number

### LINKED ISSUE #405: Display clearer error message when encountering errors due to database schema changes
A common error for people juggling multiple versions of Wake at the same time is using a different version of Wake than was used to initialize the SQLite3 database. For example, you encounter this error when trying to run Wake 0.17.2 on a database initialized with Wake 0.18.0:

```
Could not insert a job; sqlite3_step: NOT NULL constraint failed: jobs.signature
The failing statement was: insert into jobs(run_id, use_id, directory, commandline, environment, stack, stdin) values(?, ?1, ?, ?, ?, ?, ?)
```

terpstra:
This way we can detect incompatible databases gracefully.
Fixes #405.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -22,6 +22,10 @@
 #include <sstream>
 #include <sqlite3.h>
 #include <unistd.h>
+#include <string.h>
+
+// Increment every time the database schema changes
+#define SCHEMA_VERSION "1"
 
 #define VISIBLE 0
 #define INPUT 1
@@ -76,6 +80,26 @@ struct Database::detail {
 Database::Database(bool debugdb) : imp(new detail(debugdb)) { }
 Database::~Database() { close(); }
 
+static int schema_cb(void *data, int columns, char **values, char **labels) {
+  // values[0] = 0 if a fresh DB
+  // values[1] = schema version
+
+  // Returning non-zero causes SQLITE_ABORT
+  if (columns != 2) return -1;
+
+  // New DB? Ok to use it
+  if (!strcmp(values[0], "0")) return 0;
+
+  // No version set? This must be a pre-0.19 database.
+  if (!values[1]) return -1;
+
+  // Matching version? Ok to use it
+  if (!strcmp(values[1], SCHEMA_VERSION)) return 0;
+
+  // Versions do not match
+  return -1;
+}
+
 std::string Database::open(bool wait, bool memory) {
   if (imp->db) return "";
   int ret;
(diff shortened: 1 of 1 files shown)


src/database.cpp
```diff
@@ -156,7 +183,17 @@ std::string Database::open(bool wait, bool memory) {
   while (true) {
     char *fail;
     ret = sqlite3_exec(imp->db, schema_sql, 0, 0, &fail);
-    if (ret == SQLITE_OK) break;
+    if (ret == SQLITE_OK) {
+      const char *get_version = "select (select count(row_id) from entropy), (select max(version) from schema);";
```
richardxia, line 187: What is this `entropy` thing about?

src/database.cpp
```diff
@@ -156,7 +183,17 @@ std::string Database::open(bool wait, bool memory) {
...
+      const char *set_version = "insert or ignore into schema(version) values(" SCHEMA_VERSION ");";
+      ret = sqlite3_exec(imp->db, get_version, &schema_cb, 0, 0);
+      if (ret == SQLITE_OK) {
+        sqlite3_exec(imp->db, set_version, 0, 0, 0);
+        break;
+      } else {
+        return "produced by an incompatible verison of wake; remove it.";
```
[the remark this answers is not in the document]
terpstra: You do not need to run `wake --init` in v0.18. If it finds a wit-lock.json, it will autoinit a wake.db.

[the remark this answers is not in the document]
terpstra: The error FYI is: 
```
Failed to open wake.db: produced by an incompatible verison of wake; remove it.
```

src/database.cpp
terpstra: Wake uses hashes to create signatures for some functions in build rules. If those build rules change, they get re-executed. The hash function I use needs a seed to prevent intentional collision attacks. The first time the database is created, a persistent random seed is chosen which will be used for all hashes. That way the hashes in the database do not change values between runs.

In this context, I am using the entropy as a means to detect that the database is "fresh"/unused. That's needed because earlier versions do not have a schema(version) to check.

src/database.cpp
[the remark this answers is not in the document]
richardxia: > You do not need to run wake --init in v0.18. If it finds a wit-lock.json, it will autoinit a wake.db.

Neat, I remember this now that you mention it, but totally forgot.

And thanks, yes, I agree that error message is pretty clear.

src/database.cpp
richardxia: OK, thanks for the additional context. I think it's worth adding a comment to describe that the choice of table here is not really important, and that it's just to detect a fresh DB. As an outside reader of the code, I was reading more into the entropy than was actually relevant for this specific use case.

terpstra: Comment added.

richardxia: Thanks!

PR #478, opened 2020-08-20
Reuse objects rather than reconstructing them

terpstra:
For example,
```
export def orElse alt = match _
  Some a = Some a
  None   = alt
```

is optimized to
```
export def orElse alt x = match x
  Some _ = x
  None   = alt
```

This also makes it possible to completely elide some destructuring.

For example,
```
export def orElse = match _ _
  _ (Some a) = Some a
  (Some a) _ = Some a
  None None  = None
```
gets optimized to the exact same thing as the other two examples.

This is because the inner case-analysis of the first argument ends up with
two cases, both the identity function.

### DIFF
--- a/src/inline.cpp
+++ b/src/inline.cpp
@@ -223,7 +223,7 @@ void RDes::pass_inline(PassInline &p, std::unique_ptr<Term> self) {
           cargs.back() = fnid+2;
           f->terms.emplace_back(new RArg());
           f->terms.emplace_back(new RApp(des->args[i], fnid+1));
-          f->terms.emplace_back(new RDes(std::move(cargs)));
+          f->terms.emplace_back(new RDes(sum, std::move(cargs)));
           PassInline q(p.common, fnid); // refs up to fun are unmodified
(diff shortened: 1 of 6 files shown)


terpstra:
To review, compare the output of `wake --stop-after-ssa -x orElse' and/or similar functions before and after.

src/inline.cpp
```diff
@@ -223,7 +223,7 @@ void RDes::pass_inline(PassInline &p, std::unique_ptr<Term> self) {
           cargs.back() = fnid+2;
           f->terms.emplace_back(new RArg());
           f->terms.emplace_back(new RApp(des->args[i], fnid+1));
-          f->terms.emplace_back(new RDes(std::move(cargs)));
+          f->terms.emplace_back(new RDes(sum, std::move(cargs)));
```
terpstra, line 226: The sum type of this must be adjusted to des->sum.

terpstra:
Way out of date and probably not worth doing.

PR #479, opened 2020-08-20
database: detect if a user supplied bad inputs

### LINKED ISSUE #358: SQLite error: NOT NULL constraint failed: filetree.file_id
While running a Wake build on Wake 0.17.1, I encountered the following error:

```
Could not save job inputs and outputs; sqlite3_step: NOT NULL constraint failed: filetree.file_id
The failing statement was: insert into filetree(access, job_id, file_id) values(?, ?, (select file_id from files where path=?))
```

I don't have a small example for reproducing it, but offline I can give you the steps to reproduce from a branch of an internal repository.

Without revealing anything internal, I can say that the failing Job involves invoking Verilator to compile the simulator binary and Verilator aborting due to what it considers a syntax error from parsing a Verilog file. In fact, this particular file is protected/encrypted, so we can't run it on Verilator anyway, so I don't actually need this fixed. However, I thought it was worth reporting this error since Wake should never encounter a SQL constraint error.

terpstra:
When an input file is specified which was not a visible file, it currently generates an sqlite3 exception.

Ever since https://github.com/sifive/wake/pull/139 (v0.15), we record the visible files. This makes it possible to test whether or not this condition occurs. Previously, when #358 was reported, that was not the case.

Therefore, this PR:
- detects when an input is not a visible file
- prints a warning to that effect
- omits the erroneous input from the database
- proceeds with the build

Fixes #358 .

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -20,6 +20,7 @@
 #include <unordered_set>
 #include <iostream>
 #include <sstream>
+#include <set>
 #include <sqlite3.h>
 #include <unistd.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


terpstra:
Note: there is also some sort of bug in the fuse-wake runner whereby input paths are getting corrupted.
I'm attempting to triage this as well.

terpstra:
The error message blames the user, because users can cause this. In the specific case we are running into now, I think there is something else that's to blame.

PR #485, opened 2020-08-26
FreeBSD limits

terpstra:
I recently reran wake/master on freeBSD and found a regression when there were more cores on the virtual machine.
These three commits fix the issue.

The symptom was that child processes of wake had a limit of 64 file descriptors, which can cause even reasonable software to break.

### DIFF
--- a/src/job.cpp
+++ b/src/job.cpp
@@ -410,10 +410,8 @@ JobTable::JobTable(Database *db, double percent, bool verbose, bool quiet, bool 
   if (maximum > OPEN_MAX) maximum = OPEN_MAX;
   if (maximum > FD_SETSIZE) maximum = FD_SETSIZE;
 
-  if (maximum >= requested) {
-    limit.rlim_cur = requested;
-  } else {
-    limit.rlim_cur = maximum;
+  if (maximum < requested) {
+    requested = maximum;
     std::cerr << "wake wanted a limit of " << imp->max_children;
     imp->max_children = (maximum - MAX_SELF_FDS) / 2;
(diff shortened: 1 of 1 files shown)


mmjconolly:
are there users of wake on freebsd?

terpstra:
I got a bug report and PR filed against me some time ago wrt. FreeBSD, which is why I setup a VM to test it.
Whether or not someone is actively using it, I couldn't say.

However, using it as another platform to test the C++ code quality is useful, either way.

mmjconolly:
I think I'm missing the initial context that motivates the PR, why does wake control the max num of file descriptors?

terpstra:
For every child process, wake needs two file descriptors: one for stdout and one for stderr. Therefore, the number of child processes wake can run is limited by the number of file descriptors wake has available. This part of the code is calculating a cap on how many children it might be desirable and feasible to run at once. To keep that cap high (for 64-core machines, remote runners, etc) wake *increases* the soft limit on the number of available file descriptors.

What I had not realized is:
- that cap is inherited by child processes
- wake was in some cases decreasing the cap
- FreeBSD CHILD_MAX and OPEN_MAX are well below the actual upper limits

So this PR does three things:
- prefer sysconf over CHILD_MAX
- prefer getrlimit over OPEN_MAX
- never decrease the soft limit on file descriptors (which could limit children)

One could argue that there is potential for a follow-up commit which explicitly sets the soft limit for child processes to some value, but it's not clear what that value should be. Really, child processes that need a lot of file descriptors should be doing what wake did: getrlimit + setrlimit.

PR #504, opened 2020-12-09
More quiet builds

terpstra:
Change the default values for command echo'ing. Plans now default to echo only on Verbose and hashing is not only on Debug.


terpstra:
Ok. I have a better idea on how to address this stuff.

PR #506, opened 2020-12-09
claimFileAsPath: now create output directory

### LINKED ISSUE #493: claimFileAsPath fails if destination directory doesn't exist
Because `claimFileAsPath` just tries to do a straight `ln -f` from the source to the destination, it will fail if the parent directory of the destination doesn't exist. This is pretty problematic because there isn't a way to pre-create the directory and pass it in, because `claimFileAsPath` does not accept any arguments for visible files. You have to trick Wake into thinking that there is a data dependency between creating the destination directory and calling `claimFileAsPath` or else there is no guarantee that your `mkdir` will run before `claimFileAsPath`.

terpstra:
There is now claimFileAsPathIn to put a file in a directory not created
using wake's mkdir. Otherwise, claimFileAsPath will auto-create the directory.

Fixes #493.

### DIFF
--- a/share/wake/lib/system/incremental.wake
+++ b/share/wake/lib/system/incremental.wake
@@ -19,12 +19,20 @@ package wake
 # Generally, one should obtain Paths from sources or as the output of Jobs.
 # This API is useful for things like a comannd-line-supplied input file.
 export def claimFileAsPath existingFile desiredWorkspacePath =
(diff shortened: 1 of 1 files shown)


terpstra:
This change does not attempt to avoid hard links. I suppose this could be changed to using, `cp` and it would still work.

richardxia (approved):
LGTM!

PR #507, opened 2020-12-09
main: allow any expression as the first argument to wake

terpstra:
This makes it possible to do stuff like:
```
   wake defaultScalaContext.getScalaCompiler a b c d
```
... and have the expression work as expected.

As discussed on slack.


richardxia (approved):
I'm still a bit iffy about this, but it's a small, backwards compatible change, and we can always change our minds later.

jackkoenig (approved):
Sounds like a fun experiment to me

PR #508, opened 2020-12-12
BSP support for wake

terpstra:
This PR adds build server protocol support to wake.
With this, your wake-based project can easily be driven from an IDE.
This also enables scala-metals to have jump to definition/etc.


terpstra:
Closing this in favor of a different, less intrusive, approach.

PR #510, opened 2020-12-17
JSON5 generation

terpstra:
Rather than trying to format JSON output directly, use an escape-safe C++ API.


terpstra:
I've split this out of the bsp PR, since it is really a separate change.
I'm not planning on merging the bsp PR that exists currently, but I'm going to keep parts of it.

PR #511, opened 2020-12-17
Database tagging API

terpstra:
Jobs can now have key-value annotations added to them in the database.
This is useful for tools which will introspect the database post-build.

For example, CI or a build server protocol implementation.

The new functionality works like this:
- setJobTag job "key" "value"
- wake -v --job &lt;job-id&gt; # Dump all tags for the specified job
- wake --tag-dag &lt;regexp&gt; # Dump a dependency DAG including only the tagged jobs
- wake --get-tag &lt;job-id&gt; &lt;tag-name&gt;

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -306,10 +306,13 @@ export def runJob p = match p
         def _ = wait (\_ badfinish job error) (badlaunch job error)
         job
 
-# Job => String => String => Unit
+# SString => String => Job => Job
 # Set the value of a tag on a Job
 # This is useful for post-build reflection into the database
-export def setJobTag job key value = prim "job_tag"
+export def setJobTag key value job =
+  def p job key value = prim "job_tag"
+  def _ = p job key value
+  job
 
 def toUsage (Pair (Pair status runtime) (Pair (Pair cputime membytes) (Pair ibytes obytes))) =
   Usage status runtime cputime membytes ibytes obytes
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -306,6 +306,11 @@ export def runJob p = match p
         def _ = wait (\_ badfinish job error) (badlaunch job error)
         job
 
+# Job => String => Unit
+# Add a tag to a Job
+# This is useful for post-build reflection into the database
+export def tagJob job tag = prim "job_tag"
+
 def toUsage (Pair (Pair status runtime) (Pair (Pair cputime membytes) (Pair ibytes obytes))) =
   Usage status runtime cputime membytes ibytes obytes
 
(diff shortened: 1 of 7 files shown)


share/wake/lib/system/job.wake
```diff
@@ -320,6 +320,11 @@ export def runJob p = match p
         def _ = wait (\_ badfinish job error) (badlaunch job error)
         job
 
+# Job => String => String => Unit
+# Set the value of a tag on a Job
+# This is useful for post-build reflection into the database
+export def setJobTag job key value = prim "job_tag"
+
```
jackkoenig, line 327: So this is defined on `Job` right? What happens if you set the same key multiple times, overwrite?

jackkoenig:
`wake --tag-dag` I'm guessing this could be used to generate a pretty graph of jobs like Github Actions now does?

mmjconolly:
Is a tag separate from a Plan's Label?

share/wake/lib/system/job.wake
terpstra: No. First time the tag is set, that's what sticks.

Hmm. That's probably bad, actually, in case you modify the wake code but the job is not re-executed. I'll change the behaviour to overwrite.

terpstra:
> `wake --tag-dag` I'm guessing this could be used to generate a pretty graph of jobs like Github Actions now does? (simple example: https://github.com/chipsalliance/chisel3/actions/runs/428972889)
> 
> Can it be used to get _all_ jobs, or only tagged ones? Would a regex of `.*` get ones without tags?

Not currently. The `.*` would still only match Jobs that have any tag at all.
However, I could certainly special-case that.
Should I?

terpstra:
> Is a tag separate from a Plan's Label?

Yes. The Label is just what gets printed in the progress bar during compilation. It also shows up when you look at a Job with -i or -o.

Tags are arbitrary key-value pairs that can be associated to any Job. These can only be generated from wake scripts, but can only be inspected outside of wake, from the command-line. The idea is that they provide a communication channel between your wake scripts and any tooling that you want to use to inspect the database, later.

jackkoenig:
> However, I could certainly special-case that.
Should I?

Not sure if it should be a special case or if there could also just be `wake --dag` or `wake --job-dag`, alternatively, CI or whatever could just go through and take every job if it really wants a full dag 🤷

terpstra:
> Not sure if it should be a special case or if there could also just be `wake --dag` or `wake --job-dag`, alternatively, CI or whatever could just go through and take every job if it really wants a full dag 🤷

What might be useful is for a --job-dag to also include all the non-tag meta-data in the output JSON. Right now you just get the tags you asked for and the job identifier. Simply because that's what I need for the scala BSP.

terpstra:
Anyhow, I do not think a `--job-dag` feature belongs in this PR. This PR is about tags.
The `--tag-dag` is necessary to get a "view" of the jobs which have that tag.

share/wake/lib/system/job.wake
terpstra: Fixed.

jackkoenig (approved):

terpstra:
Rebased against master in preparation for merge.

PR #512, opened 2020-12-17
exception: don't report C++ line numbers on panic

terpstra:
Printing out a failing REQUIRE line number for the c++ code on calls to `panic` was confusing.

This does not yet change panic to unreachable and then add back a "real" panic. That may be a follow-up PR down the road.


terpstra:
Rebased against master

Issue #489, opened 2020-09-17
Conventions for logging lines on error


mmjconolly:
It would be nice to handle logging in a regulated way.
To set the motivation, I feel that a successful build should not print anything at all (perhaps a progress bar), but just return an exit code 0 on success.

When a job runs, all the stdout/stderr could be captured and not displayed,
but with an option to pipe it through to the terminal if running verbose.

If a failure happens, dump the stdout/stderr output to the terminal with any other information we have about the job that failed (the command, wake source location, visible files etc)

mmjconolly:
Here's something inspired by https://github.com/sifive/api-scala-sifive/pull/49#discussion_r490476471 that I havent tried to compile

```
export def runJobQuiet plan =
  plan 
  | setPlanStdout logVerbose
  | setPlanStderr logVerbose
  | runJob
  match (job.getJobOutputs | findFailFn getPathResult)
  Pass a = Pass job
  Fail e =
    def _ = print "Job failed:\n STDOUT: {job.getJobStdout}\n STDERR: {job.getJobStderr}"
    Fail e
```

terpstra:
I was going to implement this when I realized that the default values for stdout already ARE logVerbose. In other words, unless you specify '-v' you will not get standard output. Standard error is set to logWarn, which is probably correct also.

So I think all we really want to change is that we set Echo default from Normal to Verbose... and make all the source / hash jobs only echo on Debug.

terpstra:
Dumping failed jobs is already supported via `wake --failed` after the build completes, so we have that too.

So I think the actionable item here is to modify the default for Echo and implement #488. Then we need to clean up the warnings in the source code that are spamming stderr. If a job really should silence stderr, because it is not printing warnings, that can still be done on a per-Job basis.

Issue #488, opened 2020-09-16
Silence the '<hash>' messages when using verbose (-v)


mmjconolly:
The `<hash>` log lines are very noisy, probably too noisy for the basic verbose level.

PR #513, opened 2020-12-17
wake: re-design how stdout/stderr work

terpstra:
Wake now has configurable log level streams.

Out of the box these streams are:
```
  -        logNever    -
  debug    logDebug    dark blue
  info     logInfo     -
  echo     logEcho     -
  warning  logWarning  yellow
  error    logError    red
```
A given Job configures three streams of output:
  Echo   - where the Job command-line is printed    (default: logEcho)
  Stdout - where the Job standard output is printed (default: logInfo)
  Stderr - where the Job standard error is printed  (default: logWarning)

Wake can now be directed to send log level streams to file descriptors 1-5:
  --stdout=stream,stream   controls which streams are sent to stdout (fd:1)
  --stderr=stream,stream   controls which streams are sent to stderr (fd:2)
  --fd:3=stream,stream     controls which streams are sent to fd:3
  --fd:4 and --fd:5

By default --stderr=error and fd:3-5 are unused.
Note: if used, file descriptors 3-5 must be redirected prior to running wake

Wake routes the log level streams to standard output / error as follows:
```
  quiet   (-q):  --stdout=nothing
  normal:        --stdout=warning
  verbose (-v):  --stdout=info,echo,warning,error
  debug   (-d):  --stdout=debug,info,echo,warning,error
```

Fixes: #488
Fixes: #489


jackkoenig:
This sounds awesome. 1 minor question is about `-v` vs. normal. Printing of commands being run is `echo` right? I feel like that might be nice to have so long as `hash` and other "utility" jobs are overriding their command echoing to be `info` or `debug`.

terpstra:
Hash jobs are on stream debug, so not included with `-v` (you need `-d`).
The new default is to not echo any jobs, unless you add `-v`.

So right now a "normal" build with no warnings or errors will print nothing other than the progress bars which it cleans up upon complete.

jackkoenig (approved):

PR #516, opened 2020-12-17
job: add isPathDir function

### LINKED ISSUE #425: How to test a path being a file or a directory?
lib/system/path.wake seems have no relative API for this.

terpstra:
Fixes #425

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -543,6 +543,10 @@ export def getPathParent path = match path
   Path name = Path (simplify "{name}/..")
   BadPath e = BadPath e
 
+export def isPathDir = match _
+  BadPath _ = False
+  Path str = matches `0+` (hashcode str)
+
 export def sortPaths paths =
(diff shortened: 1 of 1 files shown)


terpstra:
Whether or not the String inside a Path has a trailing '/' or not, we can hide this all behind the API accessors like getPathName and isPathDir.

share/wake/lib/system/job.wake
```diff
@@ -543,6 +543,10 @@ export def getPathParent path = match path
   Path name = Path (simplify "{name}/..")
   BadPath e = BadPath e
 
+export def isPathDir = match _
+  BadPath _ = False
+  Path str = matches `0+` (hashcode str)
```
mmjconolly, line 548: Is `hashcode` expensive for large files?

terpstra: It's free. This is already in the database from when the Path was created, so it's cheaper than stat, and avoids the race condition.

PR #521, opened 2020-12-18
FUSE debug

terpstra:
This PR adds some diagnostics around an area where fuse has been known to fail quasi-randomly.

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -212,10 +212,24 @@ int main(int argc, char *argv[])
 		map_id("/proc/self/uid_map", euid, real_euid);
 		map_id("/proc/self/gid_map", egid, real_egid);
 
+		// Detect if there is a problem with access() before mount
+		if (access(rpath.c_str(), X_OK) != 0)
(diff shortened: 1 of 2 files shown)


fuse/fuse.cpp
```diff
@@ -212,10 +212,24 @@ int main(int argc, char *argv[])
...
-		if (0 != mount(rpath.c_str(), workspace.c_str(), NULL, MS_BIND, NULL)) {
+		int attempt = 0;
+		while (0 != mount(rpath.c_str(), workspace.c_str(), NULL, MS_BIND, NULL)) {
 			std::cerr << "mount " << rpath << ": " << strerror(errno) << std::endl;
-			exit(1);
+			if (attempt != 3) {
+				sleep(1);
+				++attempt;
+			} else {
+				exit(1);
```
mmjconolly, line 231: Maybe also a message that we're bailing out

mmjconolly (approved):

fuse/fuse.cpp
mmjconolly: and something other than `1` as the error code

terpstra: Done

PR #523, opened 2020-12-18
bsp: harden pipe descriptor setup

terpstra:
Depending on where the bsp is invoked, you could totally end up with a pipe
descriptor in the range of 1-5 before dup2.

### DIFF
--- a/bsp/bsp.cpp
+++ b/bsp/bsp.cpp
@@ -168,20 +168,22 @@ void ExecuteWakeProcess::execute() {
   if (child == 0) {
     // Close the reading side of the pipe
     for (int i = 0; i < PIPES; ++i)
-      close(pipefds[i][0]);
+      if (close(pipefds[i][0]) != 0)
+        exit(42);
 
     // We need to dup2 pipefds[i][0,PIPES) to [1,PIPES]
     // Unfortunately, some of the these might be in the way
     // dup() any of them that are in the way, out of the way
     for (int i = 0; i < PIPES; ++i)
       while (pipefds[i][1] >= 1 && pipefds[i][1] <= PIPES)
-        pipefds[i][1] = dup(pipefds[i][1]);
+        if ((pipefds[i][1] = dup(pipefds[i][1])) == -1)
+          exit(43);
 
     // Put the pipes in their proper final position
     // This also closes any transient descriptors from dup() above
     for (int i = 0; i < PIPES; ++i) {
-      dup2(pipefds[i][1], i+1);
-      close(pipefds[i][1]);
+      if (dup2(pipefds[i][1], i+1) == -1) exit(44);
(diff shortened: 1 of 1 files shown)


bsp/bsp.cpp
```diff
@@ -166,14 +166,22 @@ void ExecuteWakeProcess::execute() {
...
+    // Close the reading side of the pipe
+    for (int i = 0; i < PIPES; ++i)
       close(pipefds[i][0]);
-      // !!!
-      if (pipefds[i][1] != i+1) {
-        dup2(pipefds[i][1], i+1);
-        close(pipefds[i][1]);
-      }
+
+    // We need to dup2 pipefds[i][0,PIPES) to [1,PIPES]
```
mmjconolly, line 173: `// We need to dup2 pipefds[i][0,PIPES) to [1,PIPES]` what motivates this?

terpstra: File descriptor 1=stdout, 2=stderr. So you need to put the file descriptors there.
Wake additionally allows for descriptors 3, 4, and 5 to be called out as output for particular output streams.

The BSP leverages this to keep the log/info/warning/error streams distinct so it can route the messages on these descriptors to the appropriate severity in the GUI.

terpstra: (although, as you can tell by the new 'quiet' option, the GUI does not appear to do a very good job of this routing!)

bsp/bsp.cpp
```diff
@@ -166,14 +166,22 @@ void ExecuteWakeProcess::execute() {
...
+    // Unfortunately, some of the these might be in the way
+    // dup() any of them that are in the way, out of the way
+    for (int i = 0; i < PIPES; ++i)
+      while (pipefds[i][1] >= 1 && pipefds[i][1] <= PIPES)
+        pipefds[i][1] = dup(pipefds[i][1]);
+
+    // Put the pipes in their proper final position
+    // This also closes any transient descriptors from dup() above
+    for (int i = 0; i < PIPES; ++i) {
+      dup2(pipefds[i][1], i+1);
```
mmjconolly, line 183: dup and dup2 have errors that can be checked

mmjconolly: and close

since it's debugging we probably want all the info we can get

mmjconolly (approved):

bsp/bsp.cpp
terpstra: While true, it's too late (post fork) to really do anything useful with them.
I guess I can print an error and not launch the subprocess.

terpstra: Actually, where would I even print it? This is what sets up stderr. :-/

terpstra: I'll just return rainbow-coloured exit codes.

PR #532, opened 2021-01-12
regexp: support string interpolation inside regular expressions

### LINKED ISSUE #528: Better RegExp syntactic sugar/UX
Currently it is very unwieldy to try to combine RegExps and Strings, especially when the Strings may have different runtime values. Consider the following:

```wake
def myFunc outputDirString filesToMatch =
    def pattern = regExpCat (outputDirString.quote, `/(include|lib)/.*`, Nil)
    ...
```

Having sugar for better handling this would be nicer, such as adding an interpolation syntax that can automatically quote strings or making it easier to just concatenate two RegExp objects together. Here's some example syntaxes:

```wake
`{outputDirString}/(include|lib)/.*`
outputDirString.quote + `/(include|lib)/.*`
regExpCat2 outputDirString.quote `/(include|lib)/.*`
```

terpstra:
Fixes #528.

This PR adds syntax for interpolation inside a RegExp using `${}` notation, which should be safe since `$` in a regular expression in wake means end-of-string, and would thus never be meaningfully followed by a `{` previously.

As an example:

```
export def foo x = `abc(${x.quote})*def`
```

The above defined a `String => RegExp` function which returns `abc(z\+)*def` when supplied with `"z+"`.


richardxia (approved):
Didn't read the code, but the described behavior LGTM.

PR #533, opened 2021-01-13
Better interpolation

terpstra:
While working on regular expression string interpolation, I noticed two things to improve in string interpolation:
- we did not make the earlier portions of an interpolated string canonical (oops)
- the error message for incorrect interpolation reports a location in C++ instead of the wake source file


richardxia (approved):
I don't know much about how the re2 library works, but this otherwise LGTM.

terpstra:
This is not the regular expression interpolation PR. That one is still WIP. This is bugfixes to normal String interpolation.

richardxia:
Yes, I understood that, and my comment about not understanding re2 was more about the stuff in the magic comments (`  /*!re2c`).

PR #535, opened 2021-01-18
Fix FUSE exit race

terpstra:
If a client tried to connect to the FUSE daemon between the time when the FUSE daemon started to exit and when it succeeded in exiting, there was a race condition, whereby the open of '.f.fuse-waked' succeeds, but the daemon exits anyway, leaving the client unable to make forward progress.

The solution taken is to recognize that once an exit attempt had been made, it cannot be stopped.  Instead, any future attempts to open '.f.fuse-waked' must fail.  Careful synchonization is needed, because canceling the SIGALRM does not guarantee the absence of invocations of handle_exit.

Revert #521 since that debug code was/is not actually helpful... and should now be unreachable, anyway.

### DIFF
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -1496,14 +1496,10 @@ int main(int argc, char *argv[])
 		close(null);
 	}
 
-	fprintf(stderr, "Starting main FUSE event loop\n");
-	fflush(stderr);
 	if (fuse_loop(fh) != 0) {
 		fprintf(stderr, "fuse_loop failed");
 		goto unmount;
 	}
-	fprintf(stderr, "Exited FUSE event loop cleanly\n");
-	fflush(stderr);
 
 	status = 0;
 
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -85,6 +85,7 @@ struct Job {
 	bool is_writeable(const std::string &path);
 	bool is_readable(const std::string &path);
 	bool is_visible(const std::string &path);
+	bool should_erase() const;
 };
 
 void Job::parse() {
@@ -140,7 +141,12 @@ struct Context {
 	std::map<std::string, Job> jobs;
 	int rootfd, uses;
 	Context() : jobs(), rootfd(-1), uses(0) { }
+	bool should_exit() const;
 };
+
+bool Context::should_exit() const {
+	return 0 == uses && jobs.empty();
+}
 
 static Context context;
 
@@ -160,6 +166,10 @@ bool Job::is_writeable(const std::string &path) {
 
 bool Job::is_readable(const std::string &path) {
 	return is_visible(path) || is_writeable(path);
+}
+
+bool Job::should_erase() const {
+	return 0 == uses && 0 == json_in_uses && 0 == json_out_uses;
 }
 
 static std::pair<std::string, std::string> split_key(const char *path) {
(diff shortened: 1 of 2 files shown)


fuse/daemon.cpp
```diff
@@ -207,14 +217,23 @@ static Special is_special(const char *path) {
...
 
-static int exit_attempts = 0;
+// If exit_attempts is > 0, we are in the impossible-to-stop process of exiting.
+// On a clean shutdown, exit_attempts will only ever be increased if context.should_exit() is true.
+static volatile int exit_attempts = 0;
 
-static void cancel_exit()
+// You must make context.should_exit() false BEFORE calling cancel_exit.
+// Return of 'true' guarantees the process will not exit
+static bool cancel_exit()
```
richardxia, line 226: I'm getting kind of confused by these comments, possibly due to me not knowing the bigger picture of how this code is run. If the process of exiting is "impossible to stop," then what is the purpose of a `cancel_exit()` function?

Is there a brief period of time where you may be "thinking of exiting," initiating the beginning of an exit, but not get far enough to cause the `exit_attempts` counter to increment, which indicates that it is still safe to abort the exit? In other words, you can start to exit, but you can always back out of it until you've hit the point of no return, which is indicated by incrementing the the `exit_attempts` variable?

This interpretation is the only one that I can come up with that makes your code make sense, but in general, I think you have some specific semantics in your head that aren't reflected in the names of the functions/variable and/or don't have any documentation. Perhaps you could slightly alter the wording of the comment you added to elaborate on this, and that clearly indicate that exits can still be canceled up until the point that the `exit_attempts` variable has been incremented?

fuse/daemon.cpp
```diff
@@ -1262,6 +1294,10 @@ static void *wakefuse_init(struct fuse_conn_info *conn)
 
 static void handle_exit(int sig)
 {
+	// It is possible that SIGLARM still gets delivered after a successful call to cancel_exit
```
richardxia, line 1297:
```suggestion
	// It is possible that SIGALRM still gets delivered after a successful call to cancel_exit
```

richardxia (approved):
This PR has made me realize I haven't thought about single-threaded, concurrent programming since college. I'm too used to modern languages providing higher level synchronization primitives or having multithreading, where you can just block until the thread in the critical section has exited the critical section.

As far as I can tell, I think this is correct, but I'm not used to having to think about a single-threaded program get interrupted at any point.

fuse/daemon.cpp
terpstra: Yes, there is a not-so-brief period (2s) where the server is not in-use, but exit has not started. The server stays alive a little bit assuming you might start another job soon. When all uses are done, we call `schedule_exit()` to trigger exit in 2 seconds. If another use appears, we `cancel_exit()`, but have to be careful to distinguish whether or not we've already begun to exit.

terpstra: The server can be three states: in-use, idle, exit-started. There is a 2s delay between idle and exit started. Actually exiting also takes time, depending on how busy your system is, maybe also on the order of seconds (normally a few ms, tho).

terpstra: Rereading the comments and function names, it seems to me the code correctly reflects this... and you also deduce the correct big picture from those comments, without reading how the rest of it works, so I'd say all is well.

PR #536, opened 2021-01-18
getrusage: add platform-dependent maxrss accessor

### LINKED ISSUE #530: Mem bytes field in Wake database incorrect
It seems like the "Mem bytes" field in the Wake database, as reported by commands such as `wake -o`, `wake --failed`, etc. is incorrect. I've been debugging an out-of-memory issue, where I'm running a Java process (FIRRTL) which has high memory usage requirements, and although I am pretty sure it is using between 5-7 GiB of memory, the Wake database is reporting around 3 <i>mega</i>bytes of memory.

terpstra:
Fixes #530.

It turns out that the units of ru_maxrss is platform dependent.

Put all the platforms I could easily find documentation for into a new
shared header file.  Use this header macro instead of ru_maxrss.

### DIFF
--- a/common/membytes.h
+++ b/common/membytes.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2019 SiFive, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You should have received a copy of LICENSE.Apache2 along with
+ * this software. If not, you may obtain a copy at
+ *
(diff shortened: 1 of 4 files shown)


terpstra:
Under linux, this makes the g++ usage for compiling wake go from:
```
  Mem bytes: 111444
```
to:
```
  Mem bytes: 114098176
```

Which seems in line with clang on osx:
```
  Mem bytes: 90243072
```

richardxia (approved):

PR #539, opened 2021-01-19
Fix installIn (and a relative corner case)

### LINKED ISSUE #490: path relative copy in stdlib
It could be nice to have an installer function that can copy (filtered) directory trees, where the base is specified instead of using the workspace root

Motivating example:
https://github.com/sifive/openocdcfg-generator/pull/19

terpstra:
Fixes #490.

This is a breaking API change, but the previous API *was* the bug.

These succeed:
installIn "/to" "from"  "from/subdir/file"  => "/to/subdir/file"
installIn "/to" "."     "from/subdir/file"  => "/to/from/subdir/file"
installIn "/to" "/from" "/from/subdir/file" => "/to/subdir/file"
installIn "/to" "/"     "/from/subdir/file" => "/to/from/subdir/file"

These fail:
installIn "/to" "from"  "subdir/file"
installIn "/to" "/from" "/subdir/file"
installIn "/to" "from"  "/from/subdir/file"
installIn "/to" "/from" "from/subdir/file"

This also fixes a bug when using `relative` with a path of `"/"`.


share/wake/lib/system/cp.wake
```diff
@@ -29,5 +29,21 @@ export def installAs dest file = match file.getPathResult
...
+# installIn "/to" "/from" "/subdir/file"
+# installIn "/to" "from"  "/from/subdir/file"
+# installIn "/to" "/from" "from/subdir/file"
+export def installIn toRoot fromRoot sourcePath =
+  require Path sourceStr = sourcePath
+  def rel = relative fromRoot sourceStr
+  def isAbs = matches `/.*`
+  def uneq a b = (a && !b) || (!a && b)
+  if matches `[.][.]/.*` rel || uneq fromRoot.isAbs sourceStr.isAbs
+  then makeBadPath (makeError "{sourceStr} is not contained in {fromRoot}")
```
mmjconolly, line 48: If we're going to change the API in this PR, could we also return a `Result`?

terpstra: Good point.

terpstra: Actually, I'd rather make this change when we change installAs. The grand elimination of BadPath will hopefully happen shortly, anyhow.

PR #540, opened 2021-01-19
sources: omit submodules from sources list

### LINKED ISSUE #497: Wake considers submodule directories to be source files
`git ls-files` includes submodule attachment directories in its output, but not normal directories.

As a consequence, `source "submodule"` and `sources` both return directories! This is super bad, because the modification timestamp of directories is not stable (adding a file to a directory bumps its modification time). That can lead to the same directory being treated as a source file twice.

terpstra:
Fixes #497.

`git ls-files` lists submodule directories.
This is a problem, because wake uses that to find source FILES.

We now subtract from the results of ls-files the contents of .gitmodules, if
any, thus pruning any unwanted directories from the source files list.

### DIFF
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -119,7 +119,7 @@ bool chdir_workspace(const char *chdirto, std::string &wake_cwd, std::string &sr
   return true;
 }
 
-static std::string slurp(int dirfd, bool &fail) {
+static std::string slurp(int dirfd, bool submodules, bool &fail) {
   std::stringstream str;
   char buf[4096];
   int got, status, pipefd[2];
@@ -138,7 +138,11 @@ static std::string slurp(int dirfd, bool &fail) {
       dup2(pipefd[1], 1);
       close(pipefd[1]);
     }
-    execlp("git", "git", "ls-files", "-z", nullptr);
+    if (submodules) {
+      execlp("git", "git", "config", "-f", ".gitmodules", "-z", "--get-regexp", "^submodule[.].*[.]path$", nullptr);
+    } else {
+      execlp("git", "git", "ls-files", "-z", nullptr);
+    }
     fprintf(stderr, "Failed to execlp(git): %s\n", strerror(errno));
     exit(1);
   } else {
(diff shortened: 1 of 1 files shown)


src/sources.cpp
```diff
@@ -119,7 +119,7 @@ bool chdir_workspace(const char *chdirto, std::string &wake_cwd, std::string &sr
   return true;
 }
 
-static std::string slurp(int dirfd, bool &fail) {
+static std::string slurp(int dirfd, bool submodules, bool &fail) {
```
richardxia, line 122: Can you give this parameter a more specific name? `bool submodules` could be interpreted as (in the order that I actually tried interpreting it):

1. Recursively slurp the submodules
2. There exist submodules, so therefore slurp the normal files but subtract the submodules
3. Just return the list of submodules

I didn't even consider 3) as a possibility from just reading the diff on this line, even with the context of your PR description/ticket.

Here are some suggestions for alternative names:
- `list_submodules`
- `only_submodule_paths`

Here's a suggestion for an even larger change, but one that I think would improve readability: Instead of having a boolean parameter that toggles between "list all files" mode and "list just submodules" mode, you replace this with an argument that is a pointer/reference to an array/vector of arguments that is the command itself. `slurp` then changes from being a function that happens to do one of two very specific things to a generic function that forks and slurps up any arbitrary output from running a (git) command.

e.g.

```c++
static char const * const GIT_LSFILES = { "git", "ls-files", "-z", nullptr };
static std::string slurp(int dirfd, char *const command[], bool &fail) { ... }

...
  slurp(dirfd, GIT_LSFILES, failed);
```

terpstra: Agreed.

src/sources.cpp
```diff
@@ -119,7 +119,7 @@ bool chdir_workspace(const char *chdirto, std::string &wake_cwd, std::string &sr
   return true;
 }
 
-static std::string slurp(int dirfd, bool &fail) {
+static std::string slurp(int dirfd, const char * const * argv, bool &fail) {
```
richardxia, line 122: BTW, I think I will never learn to read a type signature that looks like that without having to consult the internet (or something like https://cdecl.org/).

PR #541, opened 2021-01-20
symbol: regular expression `foo \${xyz}` should not expand

terpstra:
The parser did not handle the case of an escaped ${


mmjconolly:
Can wake have a directory of input files to unit test parsing

terpstra:
Yes. I've been meaning to setup a unit test directory for wake itself for some time. I need to build out the infrastructure for that still.

terpstra:
Whatever the scheme, it has to support wake files that are supposed fail to parse and/or type check.

PR #552, opened 2021-02-03
Add bind mounts to fuse-wake

mmjconolly:
# Add bind mount support to fuse-wake

The input json has fields added for the mounts.
They must be correctly ordered if there is any overlap and the mount types must be supported by the platform.
The mounts are non-recursive and are optionally read-only when supported by the underlying filesystem.
If at some point we add more mount types, we can add a 'mount-type' field to the json entries

This does not add anything to Plan or Job to use this from within wake.
Jobs that do not use bind mounts are unaffected

This was tested by running `fuse-wake in.json out.json`
This performs a copy of `/bin/bash` into `/home/matthewc/bind-mounts/test/ws-mount2`, except that the copy command uses neither of those two paths directly

`in.json`:
```
{
  "command": [
        "bash", "-c", "ls -l ws-mount*; cp -v ws-mount1/bash ws-mount3/; ls ws-mount2; findmnt"
  ],
  "environment": [
    "PATH=/usr/bin:/bin"
  ],
  "visible": ["ws-mount1", "ws-mount2", "ws-mount3"],
  "directory": ".",
  "mounts": [
    {
      "source": "/bin",
      "destination": "/home/matthewc/bind-mounts/test/ws-mount1",
      "read-only": true
    },
    {
      "source": "/home/matthewc/bind-mounts/test/ws-mount2",
      "destination": "/home/matthewc/bind-mounts/test/ws-mount3"
    }
  ],
  "stdin": "",
  "resources": [],
  "version": "v0.21.0-alpha-4-gde85a32-dirty",
  "usage": {
    "status": 1,
    "runtime": 5.40870000000000031e-02,
    "cputime": 4.00000000000000008e-03,
    "membytes": 3289088,
    "inbytes": 0,
    "outbytes": 0
  }
}

```

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -62,6 +62,45 @@ static void map_id(const char *file, uint32_t from, uint32_t to)
 	write_file(file, buf, strlen(buf));
 }
 
+static bool bind_mount(const std::string& source, const std::string& destination, bool readonly)
+{
+	const char* src = source.c_str();
+	const char* dest = destination.c_str();
+
+	if (0 != mount(src, dest, NULL, MS_BIND, NULL)) {
+		std::cerr << "bind mount (" << source << " -> " << destination << "): "
+			<< strerror(errno) << std::endl;
+		return false;
+	}
+	// Re-mount to set destination as read-only.
+	// Source filesystem must not have 'MS_NODEV' (a.k.a. 'nodev') set.
+	if (readonly) {
+		if (0 != mount(src, dest, NULL, MS_BIND | MS_RDONLY | MS_REMOUNT, NULL)) {
+			std::cerr << "bind mount (" << source << " -> " << destination << "): "
+				<< strerror(errno) << std::endl;
+			return false;
+		}
+	}
+	return true;
+}
+
+// Do the mounts specified in the parsed input json.
+// The input/caller responsibility to ensure that the mountpoint exists,
+// that the platform supports the mount type/options, and to correctly order
+// the layered mounts.
+static bool do_mounts_from_json(const JAST& jast)
+{
+	for (auto &x : jast.get("mounts").children) {
+		const std::string& src = x.second.get("source").value;
+		const std::string& dest = x.second.get("destination").value;
+		bool readonly = x.second.get("read-only").kind == JSON_TRUE;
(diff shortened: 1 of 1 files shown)


mmjconolly:
Actually, just noticed I wasn't handling errno correctly

fuse/fuse.cpp
```diff
@@ -62,6 +62,45 @@ static void map_id(const char *file, uint32_t from, uint32_t to)
...
+		}
+	}
+	return true;
+}
+
+// Do the mounts specified in the parsed input json.
+// The input/caller responsibility to ensure that the mountpoint exists,
+// that the platform supports the mount type/options, and to correctly order
+// the layered mounts.
+static bool json_mounts(JAST jast)
```
mmjconolly, line 91: const JAST& jast

fuse/fuse.cpp
```diff
@@ -62,6 +62,45 @@ static void map_id(const char *file, uint32_t from, uint32_t to)
...
+}
+
+// Do the mounts specified in the parsed input json.
+// The input/caller responsibility to ensure that the mountpoint exists,
+// that the platform supports the mount type/options, and to correctly order
+// the layered mounts.
+static bool json_mounts(const JAST& jast)
+{
+	for (auto &x : jast.get("mounts").children) {
+		std::string src = x.second.get("source").value;
```
mmjconolly, line 94: const std::string src

terpstra:
Long term filesystem plan?:

```json
"mounts": [
    {
      "source": "images/busybox.squashfs",
      "destination": "/"
    },
    {
      "source": "/sys",
      "destination": "/sys",
      "writeable": true
    },
    {
      "source": "/proc",
      "destination": "/proc",
      "writeable": true
    },
    {
      "source": "/dev",
      "destination": "/dev",
      "writeable": true
    },
    {
      "source": "tmpfs",
      "destination": "/tools",
      "writeable": true
    {
      "source": "./images/java9.squashfs",
      "destination": "/tools/java",
      "mkdir": true
     },
     {
       "source": ".",
       "destination": "/build",
       "writeable": true
      } 
]   
```

terpstra (approved):
Looks good!

PR #553, opened 2021-02-04
Add 'workspace' mount type

mmjconolly:
The workspace location is no longer hardcoded in fuse-wake, but instead hardcoded via wakelang (where it could become flexible later)

* Add `type` field to fuse-wake mount fields, define `bind` and `workspace` as fields.
* Remove the automatic 'hide fuse' mount of `workspace/.fuse/$PID -> workspace`.
* In wakelang's `makeJSONRunner`, add `workspace` mount type to replace the above 'hide fuse'.
* Consolidate logic for choosing the directory to run the job in (on the fuse-wake side).

The `workspace` mount type is just a `bind` mount type, except that the `source` is not provided by the json, instead it is the `.fuse/$PID` directory.

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -25,6 +25,7 @@
 #include <sys/time.h>
 #include <sys/resource.h>
 #include <sys/wait.h>
+#include <algorithm>
 #include <iostream>
 #include <fstream>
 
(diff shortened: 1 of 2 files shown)


fuse/fuse.cpp
```diff
@@ -252,19 +282,11 @@ int main(int argc, char *argv[])
...
-
 		// Mounts from the parsed input json.
-		if (!do_mounts_from_json(jast))
+		if (!do_mounts_from_json(jast, rpath))
 			exit(1);
-
-#else
-		std::string dir = rpath + "/" + subdir;
 #endif
+		std::string dir = get_running_dir(jast, rpath, subdir);
```
terpstra, line 289: Move subdir to + "/" + subdir here, instead?

fuse/fuse.cpp
```diff
@@ -84,25 +85,61 @@ static bool bind_mount(const std::string& source, const std::string& destination
...
+static std::string get_running_dir(const JAST& jast,
+		const std::string& host_workspace_dir, const std::string& subdir)
+{
+	for (auto &x : jast.get("mounts").children) {
+		const std::string& op = x.second.get("type").value;
+		if (op == "workspace")
+			return x.second.get("destination").value + "/" + subdir;
+	}
+
+	return host_workspace_dir + "/" + subdir;
```
terpstra, line 140: This should be an error, I think.

fuse/fuse.cpp
```diff
@@ -84,25 +85,61 @@ static bool bind_mount(const std::string& source, const std::string& destination
...
 
 #endif
 
+static std::string get_running_dir(const JAST& jast,
+		const std::string& host_workspace_dir, const std::string& subdir)
+{
+	for (auto &x : jast.get("mounts").children) {
+		const std::string& op = x.second.get("type").value;
+		if (op == "workspace")
+			return x.second.get("destination").value + "/" + subdir;
```
terpstra, line 137: This value might be a relative path, in which case, you need to qualify it with the host_workspace_dir.

share/wake/lib/system/job.wake
```diff
@@ -465,6 +465,14 @@ export def makeJSONRunner plan =
           "stdin"       → JString stdin,
           "resources"   → res | map JString | JArray,
           "version"     → JString version,
+          "mounts"      → JArray (
+            JObject (
+              "type"        → JString "workspace",
+              "destination" → JString workspace,
```
terpstra, line 471: I do not want to see /scratch/xyz/federation appear in any generated files.

fuse/fuse.cpp
```diff
@@ -27,6 +27,7 @@
 #include <sys/wait.h>
 #include <iostream>
 #include <fstream>
+#include <set>
```
terpstra, line 30: Set is an expensive API. => make fuse-wake probably a lot larger.

fuse/fuse.cpp
```diff
@@ -84,25 +85,61 @@ static bool bind_mount(const std::string& source, const std::string& destination
 	return true;
 }
 
+static bool validate_mount(const std::string& op, const std::string& source)
+{
+	const std::set<std::string> mount_ops {"bind", "workspace"};
+
+	if (mount_ops.find(op) == mount_ops.end()) {
```
terpstra, line 92:
```
static const char *validOptions[] = { // must be sorted
  "bind",
  "cats",
  "dogs",
  "workspace"
};
```

This your 'find' can be a binary search. (std::binary_search)

terpstra (approved):

PR #555, opened 2021-02-10
Move 'ifdef linux' to remove warning on MacOS, add MacOS CI test

mmjconolly:
This builds `wake` for MacOS.

I suspect that it's using the preload runner to do the build of wake, as `osxfuse` claims that it needs a reboot when it is being installed, but it should at least build-check the fuse codepath

```
osxfuse requires a kernel extension to work.
If the installation fails, retry after you enable it in:
  System Preferences → Security & Privacy → General

For more information, refer to vendor documentation or this Apple Technical Note:
  https://developer.apple.com/library/content/technotes/tn2459/_index.html

You must reboot for the installation of osxfuse to take effect.
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,23 @@
+name: "Test"
+
+on:
+    pull_request:
+
+jobs:
+    build:
+        name: "MacOS build"
+        runs-on: macos-latest
+
+        steps:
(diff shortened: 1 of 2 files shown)


terpstra:
I wonder if I should move everything over to github actions at some point...

terpstra (approved):

PR #557, opened 2021-02-12
Always use 'recursive' for bind mounts

mmjconolly:
Inside a user namespace, you are not allowed to separate mounts that you inherited from another mount namespace from their child mounts.
(though you're allowed to remount over the children in your own namespace)

So whenever a mount has subdirectories with mounts in them, we _must_ be recursive when we bind mount.
Since there is no downside to requesting recursive when the mount has no children, we apply `MS_REC` to all bind mounts. We could make this configurable in the future if we found a need to.

I ran into this when being denied trying to bind mount `/proc` which has the child `/proc/sys/fs/binfmt_misc`

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -68,7 +68,7 @@ static bool bind_mount(const std::string& source, const std::string& destination
 	const char* src = source.c_str();
 	const char* dest = destination.c_str();
 
-	if (0 != mount(src, dest, NULL, MS_BIND, NULL)) {
+	if (0 != mount(src, dest, NULL, MS_BIND | MS_REC, NULL)) {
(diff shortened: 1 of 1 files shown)


terpstra (approved):
A very important change!

This needs a comment in the code that matches what you put in the PR.

I can totally understand this requirement in hindsight (it would let users expose files that should have been hidden), but without a comment I would have had no idea why we need this. It's not so much that we WANT recursive, but we need it for this to work.

terpstra:
re: configurable, I don't think so. As you rightly mention, if there are no submounts, it is harmless. If there ARE submounts, it's required. The only scenario an option would make sense is if you ran fuse-wake as root, which I think we agree is an anti-design-goal.

PR #558, opened 2021-02-15
Report enumeration failure

terpstra:
When wake cannot walk the workspace, say why.

### DIFF
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -255,7 +255,7 @@ static bool push_files(std::vector<std::string> &out, const std::string &path, i
 
   struct dirent *f;
   bool failed = false;
-  for (errno = 0; !failed && (f = readdir(dir)); errno = 0) {
+  for (errno = 0; 0 != (f = readdir(dir)); errno = 0) {
     if (f->d_name[0] == '.' && (f->d_name[1] == 0 || (f->d_name[1] == '.' && f->d_name[2] == 0))) continue;
     bool recurse;
     struct stat sbuf;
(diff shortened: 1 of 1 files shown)


src/sources.cpp
```diff
@@ -249,12 +249,13 @@ static bool push_files(std::vector<std::string> &out, const std::string &path, i
...
   if (!dir) {
     close(dirfd);
+    fprintf(stderr, "Failed to fdopendirt %s: %s\n", path.c_str(), strerror(errno));
     return true;
   }
 
   struct dirent *f;
   bool failed = false;
-  for (errno = 0; !failed && (f = readdir(dir)); errno = 0) {
+  for (errno = 0; 0 != (f = readdir(dir)); errno = 0) {
```
mmjconolly, line 258: When I look at `0 != (f = readdir(dir))`, I start thinking this could be a while loop instead

terpstra: Well, the `0 != ` part is necessary to prevent a warning.

I guess it might be more readable as a while loop, though?:
```
errno = 0;
while (0 != (f = readdir(dir)) {
  ...
  errno = 0;
}
```
... maybe not.

mmjconolly:
```
while (true) {
  errno = 0
  f = readdir(dir);
  if (f != 0)
    break;
  
  // other stuff
}
```

mmjconolly: It's not really important for this PR though, it's more a style-opinion

terpstra: I dunno. I think using 'break' instead of a proper loop guard is actually worse.
The syntax of `if (0 != (x = foo()))` is pretty common in C++.

terpstra: I prefer to functions exit at the last line with a single return and have loops eschew break/continue.

Only if there is a significant reduction in overall complexity do I violate that.

PR #559, opened 2021-02-16
Add pivot-root mount option for fuse-wake

mmjconolly:
Add a pivot-root operation to fuse-wake so that we can change the base OS for each individual wake job.
This does not modify the wake rules to use the new functionality, just provides mechanism.

This PR makes the assumption that we want to build up our mount point, and then switch onto it as the last operation.
This isn't the only possible model, at a later date we could choose to expose the pivot_root's 'oldroot' as a json option if we wanted to make a model available where parts of oldroot can be mounted in post-pivot. It doesn't feel required just now

~The pivot expects that the new root has a `tmp` directory available as a subdir which we can pivot the old root onto. It's immediately unmounted.~ From reading docs it's looks like its possible to do a pivot without having that extra mountpoint, but it's _just unreadable_ to anyone who hasn't researched it.

The previous 'workspace' mount type now has an `after-pivot` option to change the working directory as otherwise the expected directory of the fuse mount is at the wrong path after pivoting.

Many systems have an _ancient_ manpage entry for pivot_root, instead see 2019-era docs at:
https://lwn.net/Articles/800381/

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <sys/time.h>
 #include <sys/resource.h>
+#include <sys/syscall.h>
 #include <sys/wait.h>
 #include <algorithm>
 #include <iostream>
@@ -88,11 +89,15 @@ static bool bind_mount(const std::string& source, const std::string& destination
 	return true;
 }
 
-static bool validate_mount(const std::string& op, const std::string& source)
+static bool validate_mount(
+	const std::string& op,
+	const std::string& source,
+	const std::string& after_pivot)
 {
 	static const std::vector<std::string> mount_ops {
 		// must be sorted
 		"bind",
+		"pivot-root",
 		"workspace"
 	};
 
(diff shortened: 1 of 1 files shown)


fuse/fuse.cpp
```diff
@@ -88,11 +89,13 @@ static bool bind_mount(const std::string& source, const std::string& destination
 	return true;
 }
 
-static bool validate_mount(const std::string& op, const std::string& source)
+static bool validate_mount(const std::string& op, const std::string& source,
+		const std::string& after_pivot)
```
terpstra, line 93:
```
static bool validate_mount(
\t  const std::string& op,
\t  const std::string& source,
\t  const std::string& after_pivot)
{
```

That's the style used everywhere else.

terpstra (approved):
Looks good. Please don't expect /tmp.

fuse/fuse.cpp
```diff
@@ -101,8 +104,55 @@ static bool validate_mount(const std::string& op, const std::string& source)
...
+static bool do_pivot(const std::string& newroot) {
+	// The pivot_root syscall requires that the new root location is already a mountpoint.
+	// Bind mount the new root onto itself to ensure this.
+	if (!bind_mount(newroot, newroot, false))
+		return false;
+
+	// We need somewhere to put the previous root as a subdir of newroot.
+	// It feels safe to assume that each rootfs will have a 'tmp' dir,
+	// it will be removed later in this function.
+	std::string tmp = "/tmp";
```
mmjconolly, line 138: Remove, use strange `., .` thing

PR #565, opened 2021-02-21
Add 'squashfs' mount type to fuse-wake

mmjconolly:
Adds a `squashfs` mount operation to fuse-wake

The input json to fuse-wake can have a entry that looks like
```
"mounts": [
  { "type": "squashfs", "source": "your-os-image.squashfs", "destination": "/mnt" }
]
```

Notes:
* This `fork()`s out to `squashfuse`
* `prctl` is used to kill `squashfuse` when `fuse-wake` terminates
* We poll to find out when the mount is ready before continuing\
* Requires `isolate/user` resource

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -38,6 +38,8 @@
 // For unshare and bind mount
 #include <sched.h>
 #include <sys/mount.h>
+#include <sys/prctl.h>
+#include <sys/sysmacros.h>
 
 static void write_file(const char *file, const char *content, int len)
 {
@@ -98,6 +100,7 @@ static bool validate_mount(
 		// must be sorted
 		"bind",
 		"pivot-root",
+		"squashfs",
 		"tmpfs",
 		"workspace"
 	};
@@ -107,7 +110,7 @@ static bool validate_mount(
 		return false;
 	}
 
-	if (op != "bind" && source.length() != 0) {
+	if ((op != "bind" && op != "squashfs") && source.length() != 0) {
 		std::cerr << "mount: " << op << " can not have 'source' option" << std::endl;
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


fuse/fuse.cpp
```diff
@@ -175,6 +181,69 @@ static bool mount_tmpfs(const std::string& destination) {
...
+			std::cerr << "squashfuse prctl: " << strerror(errno) << std::endl;
+			exit(1);
+		}
+		execlp("squashfuse", "squashfuse", "-f", source.c_str(), mountpoint.c_str(), NULL);
+		std::cerr << "execle squashfuse: " << strerror(errno) << std::endl;
+		exit(1);
+	}
+
+	// Wait up to 1 second for the mount to exist before we continue.
+	for(int i = 1; i < 100; i++) {
```
mmjconolly, line 236: exp backoff

fuse/fuse.cpp
```diff
@@ -175,6 +181,69 @@ static bool mount_tmpfs(const std::string& destination) {
...
+		if (is_fuse_filesystem(mountpoint) && is_squashfuse_mount(mountpoint))
```
mmjconolly, line 237: Use stat's st_dev

fuse/fuse.cpp
```diff
@@ -38,6 +38,8 @@
 // For unshare and bind mount
 #include <sched.h>
 #include <sys/mount.h>
+#include <sys/prctl.h>
+#include <sys/types.h>
```
mmjconolly, line 42: It looks like some of the build targets (like debian) want `#define _BSD_SOURCE` for using major/minor with sys/types.h.
I'm suprised about this, as on a ubuntu 16.04 it doesn't require it

mmjconolly: an musl on alpine wants something else too

terpstra (approved):
This looks great!

PR #567, opened 2021-02-23
Use exponential backoff waiting for squashfs mounts to become ready

mmjconolly:


terpstra (approved):
Less retries, yet supports longer timeouts on a busy system. 👍

PR #568, opened 2021-02-25
Namespace file

mmjconolly:
As a precursor step to having a (linux-only) sandbox executable as an alternative to 'preload' and a fuse-only 'fuse-wake', 
this moves all the linux-specific namespace and mount code into it's own file

* The first commit moves the `unshare` step into a function
* The second commit moves code into a separate file

The #includes have been trimmed


fuse/namespace.h
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef NAMESPACE_H
+#define NAMESPACE_H
+
+#ifdef __linux__
+
+#include "json5.h"
```
mmjconolly, line 23: replace with struct def
add include for std::string

terpstra (approved):

PR #571, opened 2021-03-08
fuse-wake: add create-file and create-dir mount operations

mmjconolly:
It can be useful to be able to create directories inside tmpfs mounts to act as further mount points.
Additionally, if you want to bind a file in you may want to create a file as a destination mount point.

Example use-case: run an X11 application using the .Xauthority file to authenticate with the X server
```
{
  "command": [
    "xeyes"
  ],
  "environment": [
    "PATH=/usr/bin:/bin",
    "HOME=/home/user",
    "DISPLAY=":0"
  ],
  "mount-ops": [
    { "type": "bind", "source":"/tmp/rootfs", "destination": "/mnt"},       
    { "type": "tmpfs", "destination": "/mnt/home"},
    { "type": "create-dir", "destination": "/mnt/home/user" },
    { "type": "create-file", "destination": "/mnt/home/user/.Xauthority" },
    { "type": "bind", source: "/home/myusername/.Xauthority", "destination": "/mnt/home/user/.Xauthority" },
...
```

### DIFF
--- a/fuse/namespace.cpp
+++ b/fuse/namespace.cpp
@@ -91,6 +91,8 @@ static bool validate_mount(
 	static const std::vector<std::string> mount_ops {
 		// must be sorted
 		"bind",
+		"create-dir",
+		"create-file",
 		"pivot-root",
 		"squashfs",
 		"tmpfs",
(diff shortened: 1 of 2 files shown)


terpstra (approved):
I did not fully understand your example as to why we need to create an empty file, but it seems like a reasonable thing to do. Approved.

mmjconolly:
If you want to bind mount a file, the destination needs to be a file too (rather than dir)

mmjconolly:
The example is running a X11 application (xeyes), using the user's real `.Xauthority` file to auth with the X server.
The destination rootfs has a /home mountpoint with nothing in it, so it needs
* a user directory created
* a .Xauthority file created, as a destination for the real .Xauthority
(and a DISPLAY env var)

PR #572, opened 2021-03-09
Add map-user-id mount operation

mmjconolly:
Create a `/etc/passwd` file containing correct user id with a configurable username, so that user sees a valid username instead of `I have no name!` when they run bash.
It does not set `$USER` (or anything usually set by [login(1)](https://man7.org/linux/man-pages/man1/login.1.html))

The implementation uses a temporary file and a bind mount, as we cant assume we have access to an overlayfs

```
$ cat example
...
   { "type": "pivot-root", "destination": "/mnt" },
   { "type": "map-user-id", "username": "user" }
  ],
  "resources": ["isolate/host"],
...
$ ./fuse-wake example out
user@build:/$ id
uid=1375(user) gid=1375 groups=1375,65534(nogroup)
user@build:/$ whoami
user
```
and with `isolate/user` resource we see the configured username (rather than root):
```
$ ./fuse-wake example.wb out.wb
user@build:/opt# id  
uid=0(user) gid=0(root) groups=0(root),65534(nogroup)
```

Some applications use the expected home directory rather than $HOME, for example
```
$ echo $USER
root
$ export HOME=/home/user
$ echo ~root
/root
$ nano
Unable to create directory /root/.local/share/nano/: No such file or directory
It is required for saving/loading search history or cursor positions.
```

This does not map the groups (can be added later if we want it)

### DIFF
--- a/fuse/namespace.cpp
+++ b/fuse/namespace.cpp
@@ -86,13 +86,15 @@ static bool bind_mount(const std::string& source, const std::string& destination
 static bool validate_mount(
 	const std::string& op,
 	const std::string& source,
-	const std::string& after_pivot)
+	const std::string& after_pivot,
+	const std::string& username)
 {
 	static const std::vector<std::string> mount_ops {
 		// must be sorted
 		"bind",
 		"create-dir",
 		"create-file",
+		"map-user-id",
 		"pivot-root",
 		"squashfs",
 		"tmpfs",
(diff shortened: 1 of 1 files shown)


terpstra:
I think that when we remap user-ids there are three interesting cases:
- you want to appear to be root (uid=0)
- you want to appear to be some standardized build user (uid=1000)
- you want to retain your own user id

These three options intersect with a different choice: replace / (and thus /etc/passwd and PAM in general) or transparently keep /.

- you want to appear as root works in both cases (presumably 0:root is in /etc/passwd in both for replaced[0] and transparent[1])
- switching to uid=1000 only really makes sense if you replace / (otherwise you became someone else at random) [2]
- retaining your own uid only really makes sense if you RETAIN /, because you want to have a view of files as owned by other people distinct from yourself. [3]

This PR is only useful in a hypothetical fourth case where you retain your own uid AND want to replace /. This might be useful, but it's probably not as as useful as the [01,2,3] cases above

fuse/namespace.cpp
```diff
@@ -235,6 +241,44 @@ bool create_file(const std::string& dest) {
...
+	int fd = mkostemp(name, O_CLOEXEC);
+	if (fd < 0) {
+		std::cerr << "mkstemp: " << strerror(errno) << std::endl;
+		return false;
+	}
+	// We rely on our userid having already changed if it is going to.
+	auto id = std::to_string(getuid());
+	std::string users =
+		username + ":x:" + id + ":" + id +":" + username + ":/home/" + username + ":/bin/sh\n";
+	if (-1 == write(fd, users.c_str(), users.length())) {
```
terpstra, line 258: Probably should check != users.length() in case of a partial write.

PR #573, opened 2021-03-11
fuse-wake: configurable user/group id

mmjconolly:
This makes the user and group id mapping configurable from the json
You could set it to 0/0, or 1000/1000 or leave out those json keys and get your real id

```
{
  "command": ["bash"],
  "environment": [
    "PATH=/usr/bin:/bin",
  ],
  "visible": [],
  "directory": ".",
  "user-id": 0,
  "group-id": 0,
  "mount-ops": [
...
```


terpstra (approved):
👍

PR #574, opened 2021-03-12
wakebox: a CLI separate from fuse-wake

mmjconolly:
This creates a separate executable for wakebox, so that it can evolve its CLI independently of fuse wake.
Longer term, we may move all the linux-only code to be wakebox only, and leave fuse-wake for non-linux platforms.

The actual arguments for the CLI are not final and will likely change in the near future.

Some refactoring happened too
* `arg` for gopt has moved into a shared location
* an argument to ignore the `stdin` json value is passed through to fuse.cpp
* Added squashfuse dependency to deb/rpm packages

I suspect that debian wheezy, ubuntu 14.04 and centos 6.6 are simply too old to use _any_ of the user namespace functionality, and they also don't have squashfuse packages. So they're likely broken as we haven't been feature testing the kernel to check if we should mount .fuse/$PID over the workspace

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,11 +47,11 @@ jobs:
           at: /tmp/workspace
       - run: docker build -f /tmp/workspace/.circleci/dockerfiles/debian-wheezy -t wake-debian-wheezy .
       - run: |
-          x=(/tmp/workspace/wake_*.tar.xz); y=${x%.tar.xz}; cp "$x" "${y##*/}.orig.tar.xz"
-          tar xJf wake_*.orig.tar.xz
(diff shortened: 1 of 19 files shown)


terpstra (approved):
Switch to using Recommends: instead of Depends: for debian and it should work.

gopt/gopt.h
```diff
@@ -136,6 +136,11 @@ void gopt_errors (const char *argv0, const struct option *options);
   terminates the program.
 */
 
+/*
+  Search for an option by long name.
+*/
+struct option *arg(struct option opts[], const char *name);
```
mmjconolly, line 142: Put in separate header

PR #578, opened 2021-03-18
wakebox: Use pivot-relative destination mountpoints

mmjconolly:
We have had to manually specify the destination mount prefix, for example the `/mnt`s in:
```
{
  "command": ["bash", "-c", "echo hello"],
  "environment": [
    "PATH=/usr/bin:/bin",
    "HOME=/root",
    "USER=root"
  ],
  "user-id": 0,
  "group-id": 0,
  "visible": [""],
  "directory": ".",
  "mount-ops": [
    { "type": "squashfs", "source": "centos8.wbi", "destination": "/mnt" },
    { "type": "bind", "source":"/proc", "destination": "/mnt/proc"},
    { "type": "bind", "source":"/sys", "destination": "/mnt/sys"},
    { "type": "bind", "source":"/dev", "destination": "/mnt/dev"},
    { "type": "tmpfs", "destination": "/mnt/tmp"},
    { "type": "tmpfs", "destination": "/mnt/run"},
    { "type": "tmpfs", "destination": "/mnt/var"},
    { "type": "tmpfs", "destination": "/mnt/home"},
    { "type": "workspace", "destination": "/mnt/workspace" },
    { "type": "pivot-root", "destination": "/mnt" }
  ]
}

```
This PR notices that we need a pivot-root, so we can write:
```
{
  "command": ["bash", "-c", "echo hello"],
  "environment": [
    "PATH=/usr/bin:/bin",
    "HOME=/root",
    "USER=root"
  ],
  "user-id": 0,
  "group-id": 0,
  "visible": [""],
  "directory": ".",
  "mount-ops": [
    { "type": "squashfs", "source": "centos8.wbi", "destination": "/" },
    { "type": "bind", "source":"/proc", "destination": "/proc"},
    { "type": "bind", "source":"/sys", "destination": "/sys"},
    { "type": "bind", "source":"/dev", "destination": "/dev"},
    { "type": "tmpfs", "destination": "/tmp"},
    { "type": "tmpfs", "destination": "/run"},
    { "type": "tmpfs", "destination": "/var"},
    { "type": "tmpfs", "destination": "/home"},
    { "type": "workspace", "destination": "/workspace" } 
  ]
}

```
Wakebox creates a directory (if it doesnt exist) as the prefix.
If the mount-ops list contains a destination of `/` then all mounts from then on have their destination prefixed.

This has the follow-on effect that we no longer need an 'after-pivot' field in the workspace mount, as it's now the same as it's 'destination', and that we don't have a 'pivot-root' mount op at all

### DIFF
--- a/fuse/namespace.cpp
+++ b/fuse/namespace.cpp
@@ -85,15 +85,13 @@ static bool bind_mount(const std::string& source, const std::string& destination
 
 static bool validate_mount(
 	const std::string& op,
-	const std::string& source,
-	const std::string& after_pivot)
+	const std::string& source)
 {
 	static const std::vector<std::string> mount_ops {
 		// must be sorted
 		"bind",
 		"create-dir",
 		"create-file",
-		"pivot-root",
 		"squashfs",
 		"tmpfs",
 		"workspace"
@@ -106,10 +104,6 @@ static bool validate_mount(
 
 	if ((op != "bind" && op != "squashfs") && source.length() != 0) {
 		std::cerr << "mount: " << op << " can not have 'source' option" << std::endl;
-		return false;
-	}
-	if (op != "workspace" && after_pivot.length() != 0) {
-		std::cerr << "mount: " << op << " can not have 'after-pivot' option" << std::endl;
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


terpstra (changes requested):
Since there is no advantage to having pivot-root not at the end, I'd argue it should always be at the end. At that point, you don't even need it, because if you see someone mount "/", then you know a pivot root will be necessary.

source="xyz" is always taken relative to the parent namespace.
destination="xyz" before a mount of destination="/" is relative to the parent namespace, while destinations after a mount of "/" are relative the new namespace.

fuse/namespace.cpp
```diff
@@ -241,38 +235,58 @@ bool create_file(const std::string& dest) {
 // the layered mounts.
 bool do_mounts_from_json(const JAST& jast, const std::string& fuse_mount_path)
 {
+	// Create a known mount-able dir in the parent mount namespace to base the rest of the mounts on.
+	const std::string mount_prefix = "/tmp/.wakebox-mount";
+	if (0 != mkdir(mount_prefix.c_str(), 0777)) {
```
terpstra, line 240: I wonder if we should perhaps only do the mkdir if there is a destination=="/".

An easy way to achieve this would be:
- eliminate replacedRoot
- leave mount_prefix initially unset (ie: "")
- always prepend mount_prefix to target
- whenever you find a destination=="/", run `mount_prefix += "/tmp/.wakebox-mount" && mkdir(mount_prefix)`
- if (!mount_prefix.empty()) run pivot_root

This even makes the tool behave somewhat sensibly if there are two mounts at destination=="/".

terpstra (approved):
This PR is ok as it is, but it is somewhat unfortunate that it always needs to create /tmp/.wakebox-mount, even if you won't need it in the end.

fuse/namespace.cpp
```diff
@@ -241,38 +235,58 @@ bool create_file(const std::string& dest) {
...
 		bool readonly = x.second.get("read-only").kind == JSON_TRUE;
 
-		if (!validate_mount(op, src, after_pivot))
-			return false;
+		if (dest == "/")
+			// All mount ops from here onward will have a prefixed destination.
+			// The prefix will be pivoted to after the final mount op.
+			replacedRoot = true;
+
+		const std::string& target = replacedRoot ? (mount_prefix + dest) : dest;
```
terpstra, line 262: What happens if the user supplies a path not starting with a "/", like "XYZ"? Looks like this would break and use /tmp/.wakebox-mountXYZ

terpstra: What do we want the behaviour to be in this case? Maybe we should require absolute paths for destination?

mmjconolly: For destinations before `/` is used, relative works, afterwards they must be absolute within the new mount namespace

mmjconolly:
> This PR is ok as it is, but it is somewhat unfortunate that it always needs to create /tmp/.wakebox-mount, even if you won't need it in the end.

Updated

PR #581, opened 2021-03-22
wakebox: add --force-shell arg

mmjconolly:
In wakebox it's useful to directly enter a shell environment, disregarding the `command` value from the params file.

This adds an option `-s/--force-shell` which changes the run command to `/bin/sh` and adds the originally intended command to an environment variable called `WAKEBOX_CMD` which can be eval'd

```
matthew@computer:/home/matthew/work$ wakebox -s -p example.wb
user@wakebox:/workspace$ env | grep WAKE
WAKEBOX_CMD="bash" "-c" "echo \"hello world\""
user@wakebox:/workspace$ eval $WAKEBOX_CMD
hello world
```

~There's also some refactoring, moving the JAST parse() into main()~

### DIFF
--- a/wakebox/main.cpp
+++ b/wakebox/main.cpp
@@ -19,6 +19,7 @@
 #include <cstring>
 #include <fstream>
 #include <iostream>
+#include <sstream>
 #include <string>
 #include <fcntl.h>
 #include <unistd.h>
@@ -27,6 +28,7 @@
 #include "fuse.h"
 #include "gopt.h"
 #include "gopt-arg.h"
+#include "shell.h"
 
 void print_help() {
 	std::cout <<
@@ -35,6 +37,9 @@ void print_help() {
 		"    -p --params FILE         Json file specifying input parameters\n"
 		"    -o --output-stats FILE   Json file written to with output results\n"
 		"    -i --allow-interactive   Ignore params json file's stdin value\n"
+		"    -s --force-shell         Run shell instead of command from params.\n"
+		"                             Implies --allow-interactive.\n"
+		"                             Use 'eval $WAKEBOX_CMD' to run the command from params.\n"
 		"    -h --help                Print usage\n";
 }
 
@@ -44,6 +49,7 @@ int main(int argc, char *argv[])
 		{'p', "params",       GOPT_ARGUMENT_REQUIRED},
 		{'o', "output-stats", GOPT_ARGUMENT_REQUIRED},
 		{'i', "interactive",  GOPT_ARGUMENT_FORBIDDEN},
+		{'s', "force-shell",  GOPT_ARGUMENT_FORBIDDEN},
 		{'h', "help",         GOPT_ARGUMENT_FORBIDDEN},
 		{0,   0,              GOPT_LAST }
 	};
@@ -53,6 +59,7 @@ int main(int argc, char *argv[])
 
 	bool has_output = arg(options, "output-stats")->count > 0;
 	bool use_stdin_file = arg(options, "interactive")->count == 0;
+	bool use_shell = arg(options, "force-shell")->count > 0;
 
 	const char* params  = arg(options, "params" )->argument;
 	const char* result_path = arg(options, "output-stats")->argument;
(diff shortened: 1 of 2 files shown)


terpstra:
If you run with "-s" and have a non-empty shell command which you put into WAKEBOX_CMD, print (to stderr) a message telling the user about this variable, like:
```
Entering interactive mode.
You can run the saved shell command using: eval ...
```

wakebox/main.cpp
```diff
@@ -73,12 +83,41 @@ int main(int argc, char *argv[])
...
+		env.push_back(const_cast<char*>(x.second.value.c_str()));
+
+	// Handle --force-shell command line arg.
+	if (use_shell) {
+		use_stdin_file = false;
+		env.push_back(escaped_command_env(jast));
+
+		command.clear();
+		char sh[] = "/bin/sh";
+		command.push_back(sh);
```
terpstra, line 108: This pushes a pointer to a stack allocated variable into the list, and then deallocates at "}".

wakebox/main.cpp
```diff
@@ -73,12 +83,41 @@ int main(int argc, char *argv[])
...
 
+	// Parse the params file as JSON.
+	JAST jast;
+	if (!JAST::parse(json, std::cerr, jast))
+		return 1;
+
+	// Extract the 'command' from the input json.
+	std::vector<char *> command;
+	for (auto &x : jast.get("command").children)
+		command.push_back(const_cast<char*>(x.second.value.c_str()));
```
terpstra, line 94: The above leaks the bug from execv all the way to the user.
Let's use a std::vector<std::string> instead.

Then, at execve call site, we convert it to std::vector<const char *> (which does not need a const_cast to convert).
At call, do: execve(... const_cast<char * const *>(vec.data()), ...)
And a comment indicating the const_cast is to work around the buggy POSIX execve API definition.

wakebox/main.cpp
```diff
@@ -73,12 +83,41 @@ int main(int argc, char *argv[])
...
+		env.push_back(escaped_command_env(jast));
+
+		command.clear();
+		char sh[] = "/bin/sh";
+		command.push_back(sh);
+	}
+
+	// run_in_fuse() will pass the vector contents to exec as a null-terminated array.
+	command.push_back(0);
+	env.push_back(0);
```
terpstra, line 113: These should be moved to where we convert std::vector<std::string> to std::vector<const char*> right before the execve call.

wakebox/main.cpp
```diff
@@ -98,3 +137,22 @@ int main(int argc, char *argv[])
 
 	return res;
 }
+
+char* escaped_command_env(const JAST& jast) {
```
terpstra, line 141: This messes up \\s in the arguments.
Just use: shell_escape from common/shell.h, it handles all the odd cases correctly.

wakebox/main.cpp
```diff
@@ -98,3 +137,22 @@ int main(int argc, char *argv[])
...
+			if (s[i] == '"')
+				ss << '\\';
+			ss << s[i];
+		}
+		ss << '"' << " ";
+	}
+	// Stack allocated buffers shouldn't be used for exec's env arg.
+	char* var_on_heap = (char*)malloc(ss.str().length()+1);
+	strcpy(var_on_heap, ss.str().c_str());
+	return var_on_heap;
```
terpstra, line 157: `ss` is stack-allocated, and freed on return.
So, you had to copy the value because otherwise you had dangling pointer.

Return a std::string instead and you won't have problems.

terpstra (changes requested):
Please fix-up these issues and I'll re-review.

terpstra:
Also, please make run_in_fuse take a `struct` for its arguments.
Then, create a common (JSON-filename or JAST) to run_in_fuse-struct conversion function.

That way, both clients can call the "process file" step, do some work manipulating the arguments as they see fit, and then pass the struct onwards to the run_in_fuse call.

wakebox/main.cpp
terpstra: Also, by using malloc() to work-around the stack-allocation issue, you are leaking memory.

wakebox/main.cpp
terpstra: ie: it's just luck this did not segfault

wakebox/main.cpp
```diff
@@ -81,8 +88,19 @@ int main(int argc, char *argv[])
 	args.working_dir = get_cwd();
 	args.use_stdin_file = use_stdin_file;
 
-	std::string result;
+	if (use_shell) {
+		std::stringstream ss;
+		ss << "WAKEBOX_CMD=";
+		for (auto &s : args.command)
+			ss << shell_escape(s) << " ";
+		args.environment.push_back(ss.str());
```
terpstra, line 96: Maybe a substr to trim last space away?

terpstra (approved):
LGTM

PR #582, opened 2021-03-24
Refactor to no longer pass around a JAST object

mmjconolly:
Mostly a refactor to not pass around the jast object so much

Some real changes happened:
* ~CLONE_NEWUTS is always set for unshare()~
* hostname and domain name are configurable~, defaults to build.local~
* the visible files json that is provided to fuse-waked is generated specifically, rather than passing the input json through


fuse/fuse.h
```diff
@@ -20,12 +20,39 @@
...
+	bool isolate_network;
+
+	int userid;
+	int groupid;
+
+	std::vector<mount_op> mount_ops;
+#endif
+};
+
+struct fuse_args : json_args {
```
terpstra, line 48: Please add `public` before json_args.

terpstra: Or alternatively, make json_args a member variable.

fuse/namespace.cpp
```diff
@@ -301,44 +296,35 @@ bool get_workspace_dir(
...
-		}
-		if (setdomainname("local", 5) != 0) {
-			std::cerr << "setdomainname(local): " << strerror(errno) << std::endl;
-		}
-	}
+	if (sethostname(hostname.c_str(), hostname.length()) != 0)
+		std::cerr << "sethostname(" << hostname << "): " << strerror(errno) << std::endl;
+
+	if (setdomainname(domainname.c_str(), domainname.length()) != 0)
+		std::cerr << "setdomainname(" << domainname << "): " << strerror(errno) << std::endl;
```
terpstra, line 327: Must allow "real" hostname to be unmodified, since this can break networking access for some tools.

fuse/namespace.cpp
```diff
@@ -284,14 +280,13 @@ bool do_mounts_from_json(const JAST& jast, const std::string& fuse_mount_path)
 }
 
 bool get_workspace_dir(
-	const JAST& jast,
+	const std::vector<mount_op> mount_ops,
```
terpstra, line 283: Please add an `&`.
`const std::vector<mount_ops> &mount_ops`

fuse/namespace.cpp
```diff
@@ -233,17 +235,11 @@ bool create_file(const std::string& dest) {
 // The input/caller responsibility to ensure that the mountpoint exists,
 // that the platform supports the mount type/options, and to correctly order
 // the layered mounts.
-bool do_mounts_from_json(const JAST& jast, const std::string& fuse_mount_path)
+bool do_mounts(const std::vector<mount_op> mount_ops, const std::string& fuse_mount_path)
```
terpstra, line 238: Missing `&`.

fuse/fuse.h
```diff
@@ -20,12 +20,39 @@
...
+#endif
+
+struct json_args {
+	std::vector<std::string> command;
+	std::vector<std::string> environment;
+	std::vector<std::string> visible;
+	std::string directory;
+	std::string stdin_file;
+
+#ifdef __linux__
```
terpstra, line 36: Please make these unconditionally parsed.
We can still choose to ignore them later.

fuse/fuse.cpp
```diff
@@ -37,25 +38,78 @@
...
+	for (auto &x : jast.get("environment").children)
+		result.environment.push_back(x.second.value);
+
+	for (auto &x : jast.get("visible").children)
+		result.visible.push_back(x.second.value);
+
+	result.directory = jast.get("directory").value;
+	result.stdin_file = jast.get("stdin").value;
+
+#ifdef __linux__
```
terpstra, line 58: zap ifdef

fuse/fuse.cpp
```diff
@@ -37,25 +38,78 @@
...
+	result.isolate_network = jast.get("isolate-network").kind == JSON_TRUE;
+
+	std::string hostname = jast.get("hostname").value;
+	result.hostname = !hostname.empty() ? hostname : "build";
+
+	std::string domainname = jast.get("domainname").value;
+	result.domainname = !domainname.empty() ? domainname : "local";
```
terpstra, line 65: You probably need to leave this empty to detect when the domainname should be unmodified.

fuse/fuse.cpp
```diff
@@ -37,25 +38,78 @@
...
+
+	std::string userid = jast.get("user-id").value;
+	result.userid = !userid.empty() ? std::stoi(userid) : geteuid();
+
+	std::string groupid = jast.get("group-id").value;
+	result.groupid = !groupid.empty() ? std::stoi(groupid) : getegid();
+
+	for (auto &x : jast.get("mount-ops").children) {
+		result.mount_ops.push_back({
```
terpstra, line 74: What happens if you use `emplace_back` instead? Does it work? If so, use that.

fuse/fuse.cpp
```diff
@@ -37,25 +38,78 @@
...
+			x.second.get("source").value,
+			x.second.get("destination").value,
+			x.second.get("read_only").kind == JSON_TRUE
+		});
+	}
+#endif
+	return true;
+}
+
+void execve_wrapper(std::vector<std::string> command, std::vector<std::string> environment) {
```
terpstra, line 85: static

fuse/fuse.cpp
```diff
@@ -37,25 +38,78 @@
...
+	for (auto &e : environment)
+		env.push_back(e.c_str());
+	env.push_back(0);
+
+	std::string command_path = find_in_path(cmd_args[0], find_path(env.data()));
+	execve(command_path.c_str(),
+		const_cast<char * const *>(cmd_args.data()),
+		const_cast<char * const *>(env.data()));
+
+	std::cerr << "execve " << command_path << ": " << strerror(errno) << std::endl;
```
terpstra, line 101: move the cerr out to the call site.

terpstra: also return what execv returns

fuse/fuse.cpp
```diff
@@ -101,8 +155,16 @@ int run_in_fuse(
...
 	(void)close(ffd);
 
+	// The fuse-waked process takes an input file containing visible files, json formatted.
 	std::ofstream ijson(ipath);
-	ijson.write(json.data(), json.size());
+	bool first = true;
+	ijson << "{ \"visible\": [";
+	for (auto &s : args.visible) {
+		ijson << (first ? "\"" : ", \"") ;
+		ijson << s << "\"";
```
terpstra, line 164: You did not escape `s`... Call either json_escape, or use the JSON construction APIs ala bsp/bsp.cpp

terpstra (approved):
Once fixes made, LGTM

fuse/fuse.cpp
mmjconolly:
```
++ -std=c++11 -Wall -O2 -DVERSION=v0.21.0-alpha-52-g4ea3d9a-dirty -Iutf8proc -Igopt -Icommon fuse/client.cpp fuse/fuse.cpp fuse/namespace.cpp common/jlexer.o common/execpath.o common/jparser.o common/json5.o common/lexint.o common/location.o common/shell.o common/unlink.o common/utf8.o common/whereami.o -o lib/wake/fuse-wake 
fuse/fuse.cpp: In function ‘bool json_as_struct(std::string, json_args&)’:
fuse/fuse.cpp:76:4: error: no matching function for call to ‘std::vector<mount_op>::emplace_back(<brace-enclosed initializer list>)’
   76 |   });
      |    ^
In file included from /usr/include/c++/10/vector:72,
                 from fuse/fuse.h:23,
                 from fuse/fuse.cpp:32:
/usr/include/c++/10/bits/vector.tcc:109:7: note: candidate: ‘void std::vector<_Tp, _Alloc>::emplace_back(_Args&& ...) [with _Args = {}; _Tp = mount_op; _Alloc = std::allocator<mount_op>]’
  109 |       vector<_Tp, _Alloc>::
      |       ^~~~~~~~~~~~~~~~~~~
/usr/include/c++/10/bits/vector.tcc:109:7: note:   candidate expects 0 arguments, 1 provided
make: *** [Makefile:56: lib/wake/fuse-wake] Error 1
```

PR #583, opened 2021-03-24
Helper functions for 'Job'

mmjconolly:
An invocation that calls all the helpers:
```
$ ./bin/wake -v -x '"(echo hello; env; touch a b) 1>&2" | makePlan2 "do foo" Nil | prependPlanPath "/opt" | setPlanEnvVar "FOO" "baa" | setPlanOutput "a" | runJob | getJobOutputs'
/bin/dash -c '(echo hello; env; touch a b) 1>&2' # launched by: /scratch/matthewc/helpers/wake/lib/wake/fuse-wake .build/6152.1.in.json .build/6152.1.out.json
hello
[ #    ] do foo
PATH=/opt:/usr/bin:/bin
FOO=baa
PWD=/scratch/matthewc/helpers/wake
"(echo hello; env; touch a b) 1>&2" | makePlan2 "do foo" Nil | prependPlanPath "/opt" | setPlanEnvVar "FOO" "baa" | setPlanOutput "a" | runJob | getJobOutputs: List Path = Path "a", Nil
```


share/wake/lib/system/job.wake
```diff
@@ -114,6 +114,31 @@ export def setPlanOnce  v p = editPlanOnce  (\_ v) p
...
+  editPlanEnvironment (addEnvironmentPath value) plan
+
+# Set an environment variable in a Plan
+export def setPlanEnvVar name value plan =
+  def var = "{name}={value}"
+  editPlanEnvironment (var, _) plan
+
+# Set the Job output to a single file
+export def setPlanOutput output plan =
+  setPlanOutputs (output, Nil) plan
```
terpstra, line 136: Hmm. I don't really want to encourage people to set the outputs of a Plan directly. I much prefer to guide people towards having wake to auto-detect them. Using 'obscurity' to guide people towards a different path might not be the best idea either, though.

share/wake/lib/system/job.wake
```diff
@@ -114,6 +114,31 @@ export def setPlanOnce  v p = editPlanOnce  (\_ v) p
 export def setPlanKeep  v p = editPlanKeep  (\_ v) p
 export def setPlanShare v p = editPlanShare (\_ v) p
 
+# Create a labeled plan.
+# 'command' is a String, not List String.
+export def makeLabeledShellPlan label visible command =
+  makeShellPlan command visible
+  | setPlanLabel label
+
+export def makePlan2 = makeLabeledShellPlan
```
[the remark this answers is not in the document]
terpstra: Actually, I think the above methods should replace `makeShellPlan`.
We should have an analogous change for `makePlan` that still takes a list of arguments instead of using a shell script.

Alternatively, we could flip it and have shell plans be the default and non-shell plans be the exception. `makePlan` and `makeNativePlan` or something.

share/wake/lib/system/job.wake
terpstra: Perhaps a better helper function would be:
```
def filterPlanOutputs regExp plan = editPlanOutputFn (filter (matches regExp)) plan
```

Then your example would be: `| filterPlanOutputs `a``

mmjconolly: It's not exacally encouragement, it's a helper function in a source rather rather than something in a beginners how-to guide.

> | filterPlanOutputs `a` 

Would that work with setPlanLocalOnly?
I'm not actually looking to "filter" anything so I wouldn't have grep-ed for that

terpstra: It has backtics because it's a regular expression.

This proposal would NOT work with LocalOnly, which was kind of the point. ;-) I don't want to encourage LocalOnly.

Actually, what I think we might want is to change `setPlanLocalOnly` to accept the "output list" as its argument, rather than taking True/False.

terpstra: So, LocalOnly is its own thing. If you're using that, you are a power-user and you're using dangerous APIs. We could/should still make it harder to mess up, so I think the `setPlanLocalOnly` accepting the output list helps there.

The risk with setPlanOutput vs. filterPlanOutput is you can specify a file that does not actually get created by the Job. That introduces a new failure mode, that is not very easy to handle gracefully. For LocalOnly, we have to take that risk, but for non-LocalOnly, I would rather steer people away from the hole.

mmjconolly:
Changes since review:
* functions that set the plan outputs are removed
* makePlan is now a labeled shell plan, previous makePlan is called makeCmdArgsPlan

terpstra (approved):
Maybe rename `makeCmdArgsPlan` to `makeExecPlan`?

PR #587, opened 2021-04-02
wakebox: Metadata from within squashfs

mmjconolly:
This allows the squashfs to self-describe its expected mountpoint, environment modifications, and any required helper mounts.

A rootfs must specify it's destination in the params json, but a toolchain squashfs can place the expected mountpoint with itself at `.wakebox/mountpoint`

To modify environment variables, a sh-compatible file `.wakebox/environment` will be sourced.
If a rootfs requires additional helper mounts, `.wakebox/mounts` will contain json-formatted mount-ops (tmpfs & bind only for now).

Example wakebox input params json:
```
{
  "command": ["bash"],
  "user-id": 0,
  "group-id": 0,
  "mount-ops": [
    { "type": "squashfs", "source": "centos8.squashfs", "destination": "/"},
    { "type": "squashfs", "source": "gcc.squashfs"},
    { "type": "workspace", "destination": "/workspace"}
  ]
}
```
The files that provided by the mounts (viewed from within wakebox):
```
root@computer:/workspace$ cat /.wakebox/environment 
export PS1="\\u@\\H:\\w\$ "
export PATH=/bin:/usr/bin:/sbin:$PATH
[ -z "$USER" ] && export USER=root
[ -z "$HOME" ] && export HOME=/root
root@computer:/workspace$ cat /.wakebox/mounts      
{
  "mount-ops": [
    { "type": "bind", "source":"/proc", "destination": "/proc"},
    { "type": "bind", "source":"/sys", "destination": "/sys"},
    { "type": "bind", "source":"/dev", "destination": "/dev"},
    { "type": "tmpfs", "destination": "/tmp"},
    { "type": "tmpfs", "destination": "/run"},
    { "type": "tmpfs", "destination": "/var"},
    { "type": "tmpfs", "destination": "/home"},
    { "type": "tmpfs", "destination": "/tools"}
  ]
}
root@computer:/workspace$ cat /tools/gcc/.wakebox/mountpoint 
/tools/gcc
root@computer:/workspace$ cat /tools/gcc/.wakebox/environment
export PATH=/tools/gcc/bin:$PATH
root@computer:/workspace$ echo $PATH
/tools/gcc/bin:/bin:/usr/bin:/sbin:
```

### DIFF
--- a/common/execpath.cpp
+++ b/common/execpath.cpp
@@ -71,6 +71,15 @@ std::string find_path(const char *const * env) {
   return ".:/bin:/usr/bin";
 }
 
+std::string find_path(const std::vector<std::string> &env) {
+  for (auto &s : env) {
+    if (!s.compare(0, 5, "PATH=")) {
+      return s.substr(5);
+    }
+  }
+  return ".:/bin:/usr/bin";
+}
+
 std::string get_cwd() {
   std::vector<char> buf;
   buf.resize(1024, '\0');
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -34,6 +34,7 @@
 #include "execpath.h"
 #include "membytes.h"
 #include "namespace.h"
+#include "shell.h"
 
 bool json_as_struct(const std::string& json, json_args& result) {
 	JAST jast;
@@ -85,8 +86,7 @@ static int execve_wrapper(const std::vector<std::string>& command, const std::ve
 		env.push_back(e.c_str());
 	env.push_back(0);
 
-	std::string command_path = find_in_path(cmd_args[0], find_path(env.data()));
-	execve(command_path.c_str(),
+	execve(command[0].c_str(),
 		const_cast<char * const *>(cmd_args.data()),
 		const_cast<char * const *>(env.data()));
 	return errno;
@@ -165,6 +165,8 @@ int run_in_fuse(const fuse_args& args, std::string& result_json) {
 
 	pid_t pid = fork();
 	if (pid == 0) {
+
+		std::vector<std::string> command = args.command;
 #ifdef __linux__
 		if (!setup_user_namespaces(
 			args.userid,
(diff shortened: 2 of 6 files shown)


common/execpath.cpp
```diff
@@ -71,6 +71,15 @@ std::string find_path(const char *const * env) {
   return ".:/bin:/usr/bin";
 }
 
+std::string find_path(const std::vector<std::string> env) {
```
terpstra, line 74: Should be a reference `&env`

terpstra: It would be nice if down the road the old method died.

common/execpath.cpp
```diff
@@ -71,6 +71,15 @@ std::string find_path(const char *const * env) {
...
+  for (auto& s : env) {
+    if (!strncmp(s.c_str(), "PATH=", 5)) {
```
terpstra, line 76: Check out: http://www.cplusplus.com/reference/string/string/compare/
You should use s.compare(0, 5, "PATH=")

fuse/fuse.cpp
```diff
@@ -183,10 +184,26 @@ int run_in_fuse(const fuse_args& args, std::string& result_json) {
...
+		// 'Source' the environments provided by any mounts before running command.
+		// The shell will search the PATH for the executable location.
+		std::vector<std::string> command = {"/bin/sh", "-c"};
+		std::stringstream cmd_ss;
+		for (auto& e : envs_from_mounts)
+			cmd_ss << ". " << shell_escape(e) << "; ";
+
+		cmd_ss << "exec ";
+		for (auto& s : args.command)
+			cmd_ss << shell_escape(s) << " ";
```
terpstra, line 197: A nice trick is to go: << " " << shell_escape and skip the " " on exec.

fuse/fuse.cpp
```diff
@@ -183,10 +184,26 @@ int run_in_fuse(const fuse_args& args, std::string& result_json) {
 			exit(1);
 		}
 		dir = dir + "/" + args.directory;
+
+		// 'Source' the environments provided by any mounts before running command.
+		// The shell will search the PATH for the executable location.
+		std::vector<std::string> command = {"/bin/sh", "-c"};
+		std::stringstream cmd_ss;
+		for (auto& e : envs_from_mounts)
```
terpstra, line 192: If `envs_from_mounts` is empty, this exec is unnecessary.

terpstra (changes requested):
See if the destination can be more 'straight-line'.

terpstra (approved):
LGTM

PR #589, opened 2021-04-09
fuse-wake: refactor daemon communication

mmjconolly:
This starts the separating-out of the fuse daemon logic. 
This PR shouldn't result in a behavior change

-------------------------------------
Longer term:

In pseudocode, our current flow looks like
```
contact_fuse_daemon()
for mount_ops:
    if op == workspace
        bind_workspace_mount()
    else
        do_other_mount_op()
collect_results()
```
I'd like to make the 'contact the fuse daemon' logic become part of the 'workspace' mount op
```
for mount_ops:
    if op == workspace:
       contact_fuse_daemon()
       bind_workspace_mount()
    else
       do_other_mount_op()
if want_results:
   collect_results()
```

In json our current flow the input json looks like:
```
{
  "command": ["bash", "-c", "findmnt"],
  "visible": ["a-file", "b-file"],
  "mount-ops": [
    { "type": "workspace", "destination": "."}
  ]
}
```
but it could instead be:
```
{
  "command": ["bash", "-c", "findmnt"],
  "mount-ops": [
    { "type": "workspace", "destination": ".", "visible": ["a-file", "b-file"] }
  ]
}
```

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -36,7 +36,36 @@
 #include "namespace.h"
 #include "shell.h"
 
-bool json_as_struct(const std::string& json, json_args& result) {
+struct daemon_locations {
+	// Path to the fuse-waked daemon executable.
+	std::string executable;
+	// Location that the fuse filesystem is mounted.
+	std::string mount_path;
+	// Subdir in the fuse filesystem mount that will be used by this fuse-wake's job.
+	std::string mount_subdir;
+	// Path that the fuse daemon will write result metadata to.
+	std::string output_path;
+	// File that exists when the daemon is running/active.
+	std::string is_running_path;
+	// File held open by each child of fuse-wake. When all children close it,
+	// the daemon releases the resources for that job.
+	std::string subdir_live_file;
+	// JSON input file to the fuse daemon, listing which files should be visible.
+	std::string visibles_path;
+
+	daemon_locations(const std::string &exec_path, const std::string &working_dir) {
+		const std::string pid = std::to_string(getpid());
+		executable       = exec_path;
+		mount_path       = working_dir + "/.fuse";
+		mount_subdir     = mount_path + "/" + pid;
+		output_path      = mount_path + "/.o." + pid ;
+		is_running_path  = mount_path + "/.f.fuse-waked";
+		subdir_live_file = mount_path + "/.l." + pid ;
+		visibles_path    = mount_path + "/.i." + pid ;
+	}
+};
+
+bool json_as_struct(const std::string &json, json_args &result) {
 	JAST jast;
 	if (!JAST::parse(json, std::cerr, jast))
 		return false;
@@ -75,7 +104,7 @@ bool json_as_struct(const std::string& json, json_args& result) {
 	return true;
 }
 
-static int execve_wrapper(const std::vector<std::string>& command, const std::vector<std::string>& environment) {
+static int execve_wrapper(const std::vector<std::string> &command, const std::vector<std::string> &environment) {
 	std::vector<const char *> cmd_args;
 	for (auto &s : command)
 		cmd_args.push_back(s.c_str());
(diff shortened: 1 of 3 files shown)


fuse/fuse.cpp
```diff
@@ -36,7 +36,40 @@
...
+	// File that exists when the daemon is running/active.
+	std::string is_running_path;
+	// Lock file held by each child of fuse-wake. When all children close it,
+	// the daemon releases the resources for that job.
+	std::string daemon_lock;
+	// JSON input file to the fuse daemon, listing which files should be visible.
+	std::string visibles_path;
+};
+
+void get_daemon_locations(
```
terpstra, line 57: This should maybe be the constructor?

fuse/fuse.cpp
```diff
@@ -92,34 +125,15 @@ static int execve_wrapper(const std::vector<std::string>& command, const std::ve
...
-	std::string rpath = mpath + "/" + name;
-	// Lock file held by each child of fuse-wake. When all children close it,
-	// the daemon releases the resources for that job
-	std::string lpath = mpath + "/.l." + name;
-	// Input (as json) to the fuse daemon to setup visible files.
-	std::string ipath = mpath + "/.i." + name;
-	// Result metadata from the daemon
-	std::string opath = mpath + "/.o." + name;
-
+bool contact_daemon(struct daemon_locations &dls, int &livefd, std::vector<std::string> &visible){
```
terpstra, line 128: please rename dls to daemon, then `daemon.is_running` and `daemon.executable` all make sense to read

fuse/fuse.cpp
```diff
@@ -36,7 +36,40 @@
...
+	const std::string &working_dir,
+	struct daemon_locations &dls
+) {
+	const std::string pid = std::to_string(getpid());
+	dls.executable      = exec_path;
+	dls.mount_path      = working_dir + "/.fuse";
+	dls.mount_subdir    = dls.mount_path + "/" + pid;
+	dls.output_path     = dls.mount_path + "/.o." + pid ;
+	dls.is_running_path = dls.mount_path + "/.f.fuse-waked";
+	dls.daemon_lock     = dls.mount_path + "/.l." + pid ;
```
terpstra, line 68: subdir_live_file

fuse/fuse.cpp
terpstra, line 128: If you are taking arguments by reference for performance, and then destroying them, a comment might be nice.

fuse/fuse.cpp
```diff
@@ -132,33 +146,91 @@ int run_in_fuse(const fuse_args& args, std::string& result_json) {
...
-		return 1;
+		std::cerr << "write " << dls.visibles_path << ": " << strerror(errno) << std::endl;
+		return false;
 	}
 	ijson.close();
+	return true;
+}
+
+bool collect_result_metadata(
+	struct timeval &start,
```
terpstra, line 180: should be const

fuse/fuse.cpp
```diff
@@ -132,33 +146,91 @@ int run_in_fuse(const fuse_args& args, std::string& result_json) {
...
+	struct timeval &stop,
+	std::string daemon_output_path,
```
terpstra, line 182: const std::string &foo

fuse/fuse.cpp
```diff
@@ -132,33 +146,91 @@ int run_in_fuse(const fuse_args& args, std::string& result_json) {
...
+	pid_t pid,
+	int livefd,
+	int status,
+	struct rusage &rusage,
+	std::string &result_json
+){
+	// Cause the daemon_output_path to be generated (this write will fail)
+	(void)!write(livefd, &stop, 1); // the ! convinces older gcc that it's ok to ignore the write
```
terpstra, line 190: "x" might have been better than &stop

terpstra (approved):
LGTM minus a few minor nits

terpstra:
The idea of two different views to hide not just the workspace but other things too, is interesting. Needs more concrete use-case and examples to reason about.

PR #590, opened 2021-04-09
Change busy database print to only use one line

jackkoenig:
Instead of printing the same message every second, it now prints the
message once and just adds an additional period per second.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -191,10 +191,14 @@ std::string Database::open(bool wait, bool memory) {
     "  content text,"
     "  unique(job_id, uri) on conflict replace);";
 
+  bool waiting = false;
   while (true) {
     char *fail;
     ret = sqlite3_exec(imp->db, schema_sql, 0, 0, &fail);
     if (ret == SQLITE_OK) {
+      if (waiting) {
+        std::cerr << std::endl;
+      }
       // Use an empty entropy table as a proxy for a new database (it gets filled automatically)
       const char *get_version = "select (select count(row_id) from entropy), (select max(version) from schema);";
       const char *set_version = "insert or ignore into schema(version) values(" SCHEMA_VERSION ");";
@@ -211,10 +215,18 @@ std::string Database::open(bool wait, bool memory) {
     sqlite3_free(fail);
 
     if (!wait || ret != SQLITE_BUSY) {
+      if (waiting) {
+        std::cerr << std::endl;
+      }
       close();
       return out;
     } else {
-      std::cerr << "Database wake.db is busy; waiting 1 second ..." << std::endl;
+      if (waiting) {
+        std::cerr << ".";
+      } else {
+        waiting = true;
+        std::cerr << "Database wake.db is busy; waiting .";
+      }
       sleep(1);
     }
   }


src/database.cpp
```diff
@@ -211,10 +215,18 @@ std::string Database::open(bool wait, bool memory) {
     sqlite3_free(fail);
 
     if (!wait || ret != SQLITE_BUSY) {
+      if (waiting) {
+        std::cerr << std::endl;
+      }
       close();
       return out;
```
jackkoenig, line 222: I'm not sure when this branch is executed but I added it anyway 🤷

src/database.cpp
```diff
@@ -211,10 +215,18 @@ std::string Database::open(bool wait, bool memory) {
...
     } else {
-      std::cerr << "Database wake.db is busy; waiting 1 second ..." << std::endl;
+      if (waiting) {
+        std::cerr << ".";
+      } else {
+        waiting = true;
+        std::cerr << "Database wake.db is busy; waiting 1 second ...";
```
mmjconolly, line 228: Maybe something like `s/waiting 1 second .../waiting ./` as we might wait longer than 1 second

src/database.cpp
```diff
@@ -200,6 +201,9 @@ std::string Database::open(bool wait, bool memory) {
       const char *set_version = "insert or ignore into schema(version) values(" SCHEMA_VERSION ");";
       ret = sqlite3_exec(imp->db, get_version, &schema_cb, 0, 0);
       if (ret == SQLITE_OK) {
+        if (waiting) {
+          std::cerr << std::endl;
+        }
```
mmjconolly, line 206: What motivates this print statement?

src/database.cpp
jackkoenig: Sounds good to me, mind phrasing it as a `suggestion`?

src/database.cpp
terpstra: I think this needs to be move to the top of the enclosing 'of'; ie: right before the "use an empty entropy table" comment.

src/database.cpp
mmjconolly, line 228:
```suggestion
        std::cerr << "Database wake.db is busy; waiting.";
```

jackkoenig: Done.

src/database.cpp
jackkoenig: Done.

src/database.cpp
jackkoenig: Oh sorry, I missed that you had done the suggestion, I made the change though.

src/database.cpp
jackkoenig: @terpstra does this one need to be here? I'm not sure when we're returning vs. breaking from the loop.

mmjconolly:
Just tested it, seems to work as expected
```
$ ../bin/wake --no-tty -x '"sleep 5m" | makePlan "" Nil | runJob' &
$ ../bin/wake -x '"sleep 5m" | makePlan "" Nil | runJob' 
Database wake.db is busy; waiting .......^C
```

mmjconolly (approved):

terpstra:
Didn't I already review this? I don't see my comments anywhere :-/

terpstra (approved):
LGTM. I could swear I asked for a change to be made, but when I look over the PR now, everything seems to be correct.

PR #595, opened 2021-04-20
gopt: allow repeated options that have a value

mmjconolly:
Our gopt implementation allows you to repeat an option, but it will only store the value associated with the option _once_.
This PR changes that to store the repeated option's values in an array.
The array is allocated outside of gopt and passed in to the option struct.

Code example:
```
int main(int argc, char *argv[])
{
        unsigned int max_vals = argc / 2;
        char* vals[max_vals];
        struct option options[] {
                {'v', "value", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, vals, max_vals},
                {0,   0,       GOPT_LAST }
        };
        argc = gopt(argv, options);
        gopt_errors("test-cli", options);
        for (unsigned int i = 0; i < arg(options, "value")->count; i++)
                std::cout << "repeated value: " << vals[i] << std::endl;
```

It handles short, short without space, and long version of the argument
```
./cli-tester -v hello --value my -vname -v is positional-arg --value unknown                           
repeated value: hello
repeated value: my
repeated value: name
repeated value: is
```
`gopt_errors` will let you know if you failed to allocate enough space for the array of char*

### DIFF
--- a/gopt/gopt.h
+++ b/gopt/gopt.h
@@ -34,6 +34,14 @@ struct option
   /* input/output: */
   unsigned int  flags;
 
+  /* input/output: array of char* allocated by caller, each char* filled by gopt.
+     Required for GOPT_REPEATABLE_VALUE, unused otherwise. */
+  char        **arguments;
+
+  /* input,
+     Required for GOPT_REPEATABLE_VALUE, unused otherwise */
+  unsigned int  max_args;
+
   /* output: */
   unsigned int  count;
   char         *argument;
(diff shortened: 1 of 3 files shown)


gopt/gopt.h
```diff
@@ -34,6 +34,12 @@ struct option
   /* input/output: */
   unsigned int  flags;
 
+  /* char* array allocated by caller, each char* filled by gopt. GOPT_REPEATABLE_VALUE only. */
+  char        **arguments;
+
+  /* input, GOPT_REPEATABLE_VALUE only. */
```
terpstra, line 40: Not just only used when GOPT_REPEATABLE_VALUE is set, but required if GOPT_REPEATABLE_VALUE is set.

terpstra (approved):
LGTM

terpstra:
Maybe contact the author at https://www.purposeful.co.uk/gopt/ and see if we can get this upstreamed?

PR #596, opened 2021-04-21
fuse.cpp: change dir before finding executable in PATH

mmjconolly:
A recent change caused the full path to the executable to be not found when these are true:
* the job asks to be run within a subdir of the workspace
* the PATH contains relative locations
* the executable is contained in one of the relative PATH locations

This PR restores the previous behaviour where we would `chdir` into the subdirectory and only then search the PATH for the executable.

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -255,6 +255,11 @@ int run_in_fuse(fuse_args &args, std::string &result_json) {
 		}
 		dir = dir + "/" + args.directory;
 
+		if (chdir(dir.c_str()) != 0) {
+			std::cerr << "chdir " << dir << ": " << strerror(errno) << std::endl;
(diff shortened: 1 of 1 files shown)


terpstra:
Why is there no `chdir` removed in this PR? I can see you added one in the place I'd expect, but I don't see you removing the other one?

mmjconolly:
> Why is there no `chdir` removed in this PR? I can see you added one in the place I'd expect, but I don't see you removing the other one?

Previously, the `chdir` was outside the #ifdef blocks, now there is two copies one each within the two branches of the #ifdef

terpstra (approved):
Got it. The diff makes it look like one chdir was added and one left alone. Actually what happened is that one chdir was removed and two new chdirs were added (one for each limb of the if/else).

PR #600, opened 2021-04-28
wakebox: cli interactive mode options

mmjconolly:
This replaces most of `wakebox.cpp` to provide a second mode of option from the CLI.
There is scope for more CLI work, but this is likely sufficient for use in the medium term

There is now
* batch mode, for use with json params files
* interactive mode, for adhoc cli usage

There are also some refactors
* wakebox only builds on linux
* the working directory of the eventual command to execute is now set in main.cpp/client.cpp, not in namespace.cpp
* the #ifdef linux section in fuse.cpp is much smaller

### DIFF
--- a/wakebox/main.cpp
+++ b/wakebox/main.cpp
@@ -23,6 +23,7 @@
 #include <string>
 #include <fcntl.h>
 #include <unistd.h>
+#include <stdio.h>
 
 #include "execpath.h"
 #include "fuse.h"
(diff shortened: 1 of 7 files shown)


wakebox/main.cpp
```diff
@@ -32,57 +33,120 @@
...
-		{'o', "output-stats", GOPT_ARGUMENT_REQUIRED},
-		{'i', "interactive",  GOPT_ARGUMENT_FORBIDDEN},
-		{'s', "force-shell",  GOPT_ARGUMENT_FORBIDDEN},
-		{'h', "help",         GOPT_ARGUMENT_FORBIDDEN},
-		{0,   0,              GOPT_LAST }
-	};
+	// If we have a workspace mount, we want to default to that location.
+	for (auto &x : mount_ops) {
+		if (x.type == "workspace") {
+			if (x.destination.at(0) == '/')
```
terpstra, line 71: destination[0]

wakebox/main.cpp
```diff
@@ -32,57 +33,120 @@
...
+	return host_workspace_dir + "/" + subdir;
+}
 
-	const char* params  = arg(options, "params" )->argument;
-	const char* result_path = arg(options, "output-stats")->argument;
+int run_interactive(
+	const std::string &rootfs,
+	const std::vector<std::string> toolchains,
+	const std::vector<mount_op> binds,
+	const std::vector<std::string> command
```
terpstra, line 96: Missing &

wakebox/main.cpp
```diff
@@ -32,57 +33,120 @@
...
+
+	fa.command_running_dir = command_running_dir(fa.mount_ops, fa.working_dir, "");
+
+	std::string result;
+	return run_in_fuse(fa, result);
+}
+
+
+int run_batch(
+	const char* params_path,
```
terpstra, line 129: You could have used `const std::string &` and it would auto-convert the type at call sites.

wakebox/main.cpp
```diff
@@ -32,57 +33,120 @@
...
+		fa.environment.push_back(std::string("USER=") + std::getenv("USER"));
 	}
 
+	const char *term = std::getenv("TERM");
+	fa.environment.push_back(std::string("TERM=") + term);
+
+	fa.command_running_dir = command_running_dir(fa.mount_ops, fa.working_dir, "");
+
+	std::string result;
+	return run_in_fuse(fa, result);
```
terpstra, line 124: Feels like maybe ignoring a potential error we could report?

wakebox/main.cpp
```diff
@@ -124,3 +188,90 @@ int main(int argc, char *argv[])
 
 	return res;
 }
+
+int main(int argc, char *argv[])
+{
+	unsigned int max_pairs = argc / 2;
+	char *tools[max_pairs];
+	char *binds[max_pairs];
```
terpstra, line 196: `std::vector<char*> tools(max_pairs, nullptr);`
`tools.data()` -> get the underlying C type

wakebox/main.cpp
```diff
@@ -124,3 +188,90 @@ int main(int argc, char *argv[])
...
+
+		std::vector<std::string> toolchains;
+		for (unsigned int i = 0; i < arg(options, "toolchain")->count; i++)
+			toolchains.push_back(tools[i]);
+
+		std::vector<mount_op> bind_ops;
+		for (unsigned int i = 0; i < arg(options, "bind")->count; i++) {
+			const std::string s = binds[i];
+			std::string source = s.substr(0, s.find(":"));
+			std::string destination = s.substr(s.find(":")+1);
```
terpstra, line 246: I assume you tested the 'no :' case? I usually explicitly test the result of s.find.

wakebox/main.cpp
```diff
@@ -124,3 +188,90 @@ int main(int argc, char *argv[])
...
+	bool has_help = arg(options, "help")->count > 0;
+	bool has_params_file = arg(options, "params")->count > 0;
+	bool has_positional_cmd = argc > 1;
+
+	if (has_help) {
+		print_help();
+		return 1;
+	}
+
+	if (has_positional_cmd && has_params_file) {
```
terpstra, line 225: In a future PR, I'd like to see the capability of combining `-r` `-t` and `-p`.

terpstra (commented):
Looks like good forward progress!

terpstra (approved):
Meant to pre-approve

wakebox/main.cpp
terpstra: (to avoid stack overflow on very large command lines)

PR #601, opened 2021-04-29
fuse-wake: add label to json params file and in wake --failed

mmjconolly:
Helps with identifying Jobs when looking at params files, or in `wake --failed -v`


terpstra (approved):
Yes to JSON no to duplicated --failed output.

PR #605, opened 2021-05-06
wakebox: Add squashfs files to the tracked inputs list

mmjconolly:
This PR adds the squashfs files used as mountpoint to the list of tracked inputs that appear in the result json.
Also removed an unneeded 'pid' argument.

An input json like the following:
```
{
  "command": ["bash", "-c", "ls -l; cat foo"],
  "user-id": 0,
  "group-id": 0,
  "visible": ["foo"],
  "mount-ops": [
    { "type": "squashfs", "source": "centos8.squashfs", "destination": "/"},
    { "type": "squashfs", "source": "/home/user/dummy.squashfs"},
    { "type": "workspace", "destination": "/workspace"}
  ]
}
```
results in an output json metadata file:
```
{
  "usage": {
    "status": 0,
    "membytes": 2912256,
    "inbytes": 6,
    "outbytes": 0,
    "runtime": 0.114621,
    "cputime": 0.008
  },
  "outputs": [],
  "inputs": [
    "foo",
    "centos8.squashfs",
    "/home/user/dummy.squashfs"
  ]
} 
```
For the tracked input file outside the workspace, like `/home/user/dummy.squashfs`, would wake hash this file correctly in the event of a job re-run?

### DIFF
--- a/fuse/fuse.cpp
+++ b/fuse/fuse.cpp
@@ -95,9 +95,9 @@ static int execve_wrapper(const std::vector<std::string> &command, const std::ve
 
 bool collect_result_metadata(
 	const std::string daemon_output,
+	std::vector<std::string> inputs_from_mounts,
 	const struct timeval &start,
 	const struct timeval &stop,
-	const pid_t pid,
 	const int status,
 	const struct rusage &rusage,
 	std::string &result_json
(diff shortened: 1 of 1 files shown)


terpstra (approved):

terpstra:
"For the tracked input file outside the workspace, like /home/user/dummy.squashfs, would wake hash this file correctly in the event of a job re-run?"

Yes

terpstra:
> How do we want to generate visible files from outside the workspace? Maybe we should not be doing this. 

We agreed in zoom this a good guiding principle (though we can't enforce it).

terpstra:
We are concerned that this PR might lead to inputs that are not visible.
Perhaps this addition should happen in the wakelang wakebox bindings (which would also make it possible to add these to the visible list).

PR #607, opened 2021-05-10
Export tuples: RunnerInput, RunnerOutput, Usage

mmjconolly:
Allows a user to implement a base runner in their own codebase outside of the `wake` package

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -15,7 +15,7 @@
 
 package wake
 
-tuple Usage = # Unknown quantities are 0
+export tuple Usage = # Unknown quantities are 0
   export Status:   Integer
   export Runtime:  Double
   export CPUtime:  Double
@@ -27,7 +27,7 @@ export def getUsageThreads (Usage _ run cpu _ _ _) =
   if run ==. 0.0 then cpu else cpu /. run
 
 # RunnerInput is a subset of the fields supplied in the execution Plan
-tuple RunnerInput =
+export tuple RunnerInput =
   export Label:       String
   export Command:     List String
   export Visible:     List Path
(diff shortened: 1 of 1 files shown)


terpstra:
I would like to change the runner interface in a more substantial way than just this.

terpstra (approved):
I thought you were trying to expose the `tuple Runner`. These are harmless enough, I suppose.

terpstra:
Well, it will make it impossible to extend RunnerInput in the future, though.

terpstra:
So, I've had a thought I want to write down: the constructor/deconstructor was made private in order to support extending the tuple in a backwards compatible way (without needing proxy types in v0_19_wake). The reason you made this PR is to access to deconstructor so you can write a multi-member extractor instead of a thousand getXYZs. I think you'd have been happy to continue with a makeXYZ pattern, right? If so, maybe we should invest in a view pattern:

```
tuple Foo =
  export A: Integer
  export B: String
  export C: Double # added later

export tuple FooView =
  export A: Integer
  export B: String

def fooView (Foo a b c) = FooView a b
```

Nothing stops us from having multiple views. Views can even return the contents in different ways. For example, a complex variable could have a cartesian-coordinates view and/or a polar-coordinates view. eg:

```
def myFunc x =
  def ComplexPolar length angle = complexPolar x
  def ComplexCartesian x y = complexCartesian x
  def delta = length - sqrt (x*x + y*y)
  ...
```

Even for less esoteric types, having different Views would not only let us keep the extensibility of a private constructor on the 'core' type, but also allow more efficient use. For example, a really big tuple (Plan) might have some fields that are more interesting than others. A view of the interesting fields would let you write an extractor pattern without having to write a ton of `_`s for the irrelevant fields.

Issue #551, opened 2021-01-28
claimFileAsPathIn needs Persistence Once (not ReRun)


terpstra:
While uses of runIncrementalJob should probably be wrapped in a target, if they are not, you can accidentally end up in a situation where claimFileAsPathIn can run on the same file. This then fails with a multiple output error.

PR #614, opened 2021-05-19
claimFileInPath: improve for use as an 'input file' processing API

terpstra:
This makes claimFileAsPathIn work when:
- invoked multiple times
- used to claim a file from another file system


terpstra:
Fixes #551

mmjconolly (commented):
Say if someone wants to claim a gigantic file that would be awkward to copy, what function should they use instead?

terpstra:
> Say if someone wants to claim a gigantic file that would be awkward to copy, what function should they use instead?

If you had a multigigabyte file as an input, this was probably the wrong API already anyway.

What is that file? A toolchain? Maybe it should have been part of the environment and/or wakebox?
If it's a REAL input file, like it's a netlist or something from a 3rd party, then we probably DO want to copy it from /nfs/folder, no matter how large into the /local/build/dir.

mmjconolly (approved):

PR #616, opened 2021-05-22
Unit testing

terpstra:
wake has been lacking a unit testing framework to detect regressions for a very long time.
This adds an infrastructure to support adding tests and adds the first type-system test.


share/wake/lib/core/list.wake
```diff
@@ -226,3 +226,16 @@ export def unzip = match _
   Pair a b, t =
     def Pair u v = unzip t
     Pair (a, u) (b, v)
+
+# Group a list by some comparison function.
+# Elements which compare equal are placed into the same bucket.
+# Buckets preserve the order of the original list.
+# The buckets appear in the sort-order specified by cmp.
+# Example: seq 5 | groupBy (_/3 < /_3)
+#          (0,1,2,Nil), (3,4,Nil), Nil
```
mmjconolly, line 235: What does `(_/3 < /_3)` actually mean? 

`(something divided by 3) is less than (something divided by 3)` is how it reads to me, but that doesn't connect to what the function outputs, which is 2 lists

terpstra: Your reading is correct.
So it says group everything together that is equal after being divided by 3.

In other words, group (0,1,2), (3,4,5), (6,7,8), ...

terpstra: `/_3` is a typo.

terpstra: I've made a separate PR for this function; please review it independently.
This PR is just based on top of that one so I can test CI.

mmjconolly: my attempt at reading doesn't lead to understanding though

What does each symbol do?
What is `/_` or is it that `/_` is an operator, or that `_3` is the name of a function

build.wake
```diff
@@ -81,13 +81,21 @@ def tarball Unit =
...
   # Identify those sources files to include in the tarball
   def srcs =
-    sources "." `([^.][^/]*/)*[^.][^/]*` # ignore .foo/ files srcs
+    sources "." `.*`
     | filter (! matches `(debian/|wake.spec).*` _.getPathName) # omit packaging files
-  def pub =
-    srcs
+    | filter (! matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
+    | filter (! matches `(.*/)*\.gitignore` _.getPathName) # omit git files
+  def Pair tests core = splitBy (matches `tests/.*` _.getPathName) srcs
```
mmjconolly, line 88: What is `core`?

tests/README.md
```diff
@@ -0,0 +1,19 @@
+In order to create a new test, you need three things:
+  1. a directory wake/tests/category/testname
+  2. an empty `.wakeroot` file in that directory
```
mmjconolly, line 3: If we end up writing a lot of tests, the boiler plate of creating .wakeroot files could be part of the test run

tests/tests.wake
```diff
@@ -0,0 +1,143 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package test_wake
+from wake import _
+
+export topic wakeTestBinary: Unit => Result (Pair String (List Path)) Error
```
mmjconolly, line 19: Comment describing what the topic is for

tests/tests.wake
```diff
@@ -0,0 +1,143 @@
...
+            def failed = len failures
+            def total = len list
+            def pass = total - failed
+            "  - {category}: {str pass}/{str total} tests pass",
+            failures
+        catWith "\n" text
+
+    def report =
+        results
+        | groupBy (_.getTripleFirst <* _.getTripleFirst)
```
mmjconolly, line 64: `isLT` is much easier to read than `<*` (had to look it up)

mmjconolly (approved):
just minor comments

tests/README.md
terpstra: That's an option, but it also means you won't be able to run the pass.sh script by hand any more. One could alternatively include a wake --init call in the script. What's your preference? Mine was this PR in its current form.

tests/tests.wake
terpstra: isLT would need an order, so you'd have to combine it with `<=>*`; ie: `(_.getTripleFirst <=>* _.getTripleFirst | isLT)`. I'm not sure that's an improvement.

build.wake
terpstra: renamed to wakeSources and testSources.

tests/tests.wake
terpstra: Added

PR #617, opened 2021-05-22
list: add a groupBy method

terpstra:
It's very useful to take a list and break it into user-defined categories.
This function makes that possible.

### DIFF
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -226,3 +226,28 @@ export def unzip = match _
   Pair a b, t =
     def Pair u v = unzip t
     Pair (a, u) (b, v)
+
+# Group a list by some comparison function.
+# Elements which compare equal are placed into the same bucket.
+# Buckets preserve the order of the original list.
+# The buckets appear in the sort-order specified by cmp.
+# Examples:
+#
+#   def cmp a b = a%3 < b%3
+#   def list = seq 10
+#   groupBy cmp list
+#   ---
+#   (0,3,6,9,Nil), (1,4,7,Nil), (2,5,8,Nil), Nil
+#
+#   def cmp a b = a/3 > b/3
+#   def list = seq 10
+#   groupBy cmp list
+#   ---
+#   (9,Nil), (6,7,8,Nil), (3,4,5,Nil), (0,1,2,Nil), Nil
+#
+export def groupBy cmp list =
+  def sorted = sortBy cmp list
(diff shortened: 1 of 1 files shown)


share/wake/lib/core/list.wake
```diff
@@ -226,3 +226,16 @@ export def unzip = match _
   Pair a b, t =
     def Pair u v = unzip t
     Pair (a, u) (b, v)
+
+# Group a list by some comparison function.
+# Elements which compare equal are placed into the same bucket.
+# Buckets preserve the order of the original list.
+# The buckets appear in the sort-order specified by cmp.
+# Example: seq 5 | groupBy (_/3 < _/3)
+#          (0,1,2,Nil), (3,4,Nil), Nil
```
mmjconolly, line 235: Can we make examples low-density, and to have explicit arguments (ie, no `_`)
```
def cmp a b = (a/2 < b/3)
def list = seq 10
groupBy cmp list
   <output>
```

terpstra: Sure. How about the new version?

terpstra: I've included two examples now. The modulo one (which I think you were expecting) and the mind bending one.

share/wake/lib/core/list.wake
```diff
@@ -226,3 +226,28 @@ export def unzip = match _
...
     def Pair u v = unzip t
     Pair (a, u) (b, v)
+
+# Group a list by some comparison function.
+# Elements which compare equal are placed into the same bucket.
+# Buckets preserve the order of the original list.
+# The buckets appear in the sort-order specified by cmp.
+# Examples:
+#
+#   def cmp a b = a%3 < b%3
```
mmjconolly, line 236: Is it happenstance that we get `3` on both sides? What happens if they're different?

terpstra: If they are different, it will probably do something odd, since it's not a legal comparator.
(ie: a < b should imply !(b < a))

share/wake/lib/core/list.wake
mmjconolly: Yeah it looks better

share/wake/lib/core/list.wake
terpstra: I did consider taking two arguments; a project function and a compare function, so you would do this:
`groupBy (_/3) icmp (seq 10)`

mmjconolly (approved):

PR #619, opened 2021-05-24
wakebox: enable network by default on interactive mode

mmjconolly:
Value wasn't initialized, matches non-interactive default for now until we provide explicit knobs

### DIFF
--- a/wakebox/main.cpp
+++ b/wakebox/main.cpp
@@ -91,6 +91,8 @@ const std::string command_running_dir(
 	return host_workspace_dir + "/" + subdir;
 }
 
+// Interactive mode does not provide userid control at this time.
+// Allows networking by default, no user control yet.
 int run_interactive(
 	const std::string &rootfs,
 	const std::vector<std::string> &toolchains,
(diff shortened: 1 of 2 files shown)


wakebox/main.cpp
```diff
@@ -104,6 +104,9 @@ int run_interactive(
 	fa.userid = 0;
 	fa.groupid = 0;
 
+	// Allow networking by default
+	fa.isolate_network = false;
+
```
terpstra, line 109: Shouldn't all of these assignments have been in the fuse_args constructor?
And, shouldn't the fuse_args constructor be split into a json_args constructor as well?
And also, should fuse_args have a json_args member rather than using inheritance?

terpstra (changes requested):
Move assignments to constructor.

wakebox/main.cpp
mmjconolly: we did discuss it at the review time

I didn't want `fuse_args` to have a `json_args` field as the fuse_client implementation shouldn't care/know about json inputs at all

I guess we could copy all the fields from json_args into fuse_args

terpstra: If you want to leave the inheritance, that's fine. However, I do think your json_args should have a constructor that initialized all the those member fields.

mmjconolly: I've added a 
```
json_args() : isolate_network(false), userid(0), groupid(0) {} 
```
or did you also mean the vectors and strings too?

terpstra: Nope. You did what I wanted.
Only old-school C types (ints/bools/chars/etc) need to be explicitly initialized as they don't have a default constructor of their own. std::string/vector/etc auto-initialize to empty.

terpstra (approved):
thanks

PR #620, opened 2021-05-24
wakebox: Allow create-dir in helper mounts

mmjconolly:
Provides a little more freedom creating directory structure


terpstra (approved):
Seems useful.

PR #621, opened 2021-05-25
wake: move fuse-wake and preload-wake into /usr/bin

terpstra:
These tools could potentially be rerun by a user from the command-line.
It makes sense with them to be in /usr/bin like wakebox.

This contrasts with:
  - fuse-waked (only invoked by fuse-wake/wakebox)
  - bsp-wake   (only invoked by an editor plugin via pipes)
  - shim-wake  (only invoked by wake to exploit vfork)


terpstra:
This fixes a bug introduced where the location of fuse-wake was expected to be lib/wake, but was actually bin/ for rpm/deb builds.

mmjconolly (approved):

PR #622, opened 2021-05-26
Typed patterns

### LINKED ISSUE #609: Support Type Annotations
The hardest thing in reading Wake code is understanding the types of functions. Being able to specify types via `: <type>` (or some other style) would be so helpful.

While it would be ideal if this feature were supported across all features, I think just getting it on defined `defs` is useful enough to be worth doing even if it's harder to support anonymous functions or match statements. 

This issue https://github.com/sifive/wake/issues/369 suggests that this feature is already supported to some extent, but trying it out didn't seem to work.

terpstra:
This PR introduces support for type annotations in patterns. This makes it possible to specify a function's type in a natural way.
```
def f (x: Integer): Integer = x+1
```

Naturally, type signatures need to support polymorphism. To that end, free type variables (lower-case types) can be introduced in the type signature of a `def`. These may be further referenced in non-`def` type signatures. For example,
```
def map2 (fn: a => b): (List a => List b) = match _
  Nil = Nil
  (x: a), (r: List a) = (fn x): b, (map2 fn r): List b
```

Concretely, type signatures are now supported in:
- match patterns
- require patterns
- lambda patterns
- def patterns (may introduce type variables)

Fixes #609.


terpstra:
I obviously don't expect a review of the C++ implementation itself, but rather a review of "does it work as expected".
I am going to add a bunch of tests to this PR before we merge it.

mmjconolly:
Is there a correct way to indent long lines that have types, for example:
```
def makeWakeboxRunner 
    (score : Plan => Result Double String) 
    (toolchainProvider:  List String => Result (List String) Error)
    : Runner = 

    <some impl>
```

terpstra:
> Is there a correct way to indent long lines that have types, for example:
> 
> ```
> def makeWakeboxRunner 
>     (score : Plan => Result Double String) 
>     (toolchainProvider:  List String => Result (List String) Error)
>     : Runner = 
> 
>     <some impl>
> ```

Not currently. The parser requires patterns fit entirely in one line.

terpstra:
If we wanted to support multi-line patterns, one fairly parsimonious with the extant language might be:
```
def (
  myFun
  (x: Integer)
  (y: String)
  : Whatever
) =
  body
```

And:
```
def
  myFun
  (x: Integer)
  (y: String)
  : Whatever
=
  body
```
... where the first form would also work for `match` and `require` patterns, but the later would be allowed for `def`s.

terpstra:
When/if I make a change like that, though, I'd like it to be done at the same time as a rewrite of the parser using the lemon grammar generator. There are already too many constructs in wake where I am concerned about context freedom, and the multiline expression syntax in wake is one of them. I don't want to double down on that until I have a tool generated parser, so I can be sure I'm not adding more tech debt.

PR #623, opened 2021-05-26
wakebox: mkdir with parent dirs

mmjconolly:
This allows the 'create-dir' mount option to create parent directories without having to specify each parent

Also allows toolchains mountpoints the same leeway when specifying a mountpoint


common/mkdir_parents.cpp
```diff
@@ -0,0 +1,42 @@
...
+// Similar to the shell command: 'mkdir -p'.
+int mkdir_with_parents(const std::string &path, mode_t mode) {
+	// Call 'mkdir' on each path component starting at the base.
+	// A '/' at the start is not considered, as it always exists.
+	// If there is no '/' component at all, the 'substr' will consider whole path.
+	size_t slash_pos = 0;
+	size_t prev_slash_pos = 0;
+	while (slash_pos != std::string::npos) {
+		slash_pos = path.find("/", prev_slash_pos+1);
+		prev_slash_pos = slash_pos;
```
terpstra, line 35: prev_slash_pos is never actually used. You could just replace it on line 34 with slash_pos.

wakebox/main.cpp
```diff
@@ -271,8 +271,9 @@ int main(int argc, char *argv[])
 			bind_ops.push_back({"bind", source, destination});
 		}
 		if (arg(options, "bind-cwd")->count > 0) {
-			bind_ops.push_back({"create-dir", "", "/tmp/workspace"});
-			bind_ops.push_back({"bind", get_cwd(), "/tmp/workspace"});
+			std::string cwd = get_cwd();
+			bind_ops.push_back({"create-dir", "", cwd});
```
terpstra, line 275: This seems an odd choice. Why would you want to keep the same path as outside wakebox?
I'd have thought the prior behavior was better default.

terpstra (approved):
Minor nits

wakebox/main.cpp
mmjconolly: I think it make it less jarring for people if the working directory is the same when they drop in to wakebox.
It won't affect any reproduction of jobs from a json file or anything like that

Issue #606, opened 2021-05-07
Detect Path type in within string { } syntax


mmjconolly:
Say we have a function like:
```
def untar tarball dest =
  "tar xf {tarball.getPathName} --directory {dest.getPathName}"
  | makePlan "untar: {tarball.getPathName}" (tarball, dest_dir, Nil)
  | runJob 
```
We find that we need to use getPathName a lot, because we want _both_ the String and Path versions of the variables for most jobs we run, once for the command and once for the visibles list.

It would be nice if the `{ }` syntax were capable of noticing that it had a Path type and magically called getPathName for us.
For example:
```
def untar tarball dest =
  "tar xf {tarball} --directory {dest}"
  | makePlan "untar: {tarball}" (tarball, dest_dir, Nil)
  | runJob 
```

terpstra:
Concretely, it used to be the case that `"hello {foo}"` would implicitly call `format` on foo.

terpstra:
... and there were some type-based explicit exceptions to that rule. For example, `String` was dumped as-is. We decided to require `String` type so that you would get a nice error message in case you used it improperly, passing something you did not intend.

mmjconolly:
Calling `format` on all types in `{ }` is much more general that what I'm hoping for
Is it possible to make `{ }` work for both `String` and `Path`, but not other types?

Mostly looking to help with the ergonomics of the Path type specifically

terpstra:
I mean, we could also just define a unary operator `$` or something that runs getPathName on the object. That would be at least is somewhat discoverable since you can click or hover over the `$` operator to see what happened. With a completely implicit conversion, it's just magic you need to know about.

terpstra:
eg: `"This is a path {$myFile} but this is a string {str}"`

mmjconolly:
thats seems like a reasonable way to me

V-FEXrt:
Untriaged + wontfix is kind of weird. I'm just marking here that we need to circle around to how `$p` as a shorthand for `p.getPathName` played out

PR #625, opened 2021-05-28
unary $ for path interpolation

mmjconolly:
Fixes https://github.com/sifive/wake/issues/606

### DIFF
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -122,3 +122,9 @@ export def write path content =
   def spath = simplify path
   def parent = simplify "{spath}/.."
   writeImp (mkdir parent, Nil) 0664 spath content
+
+# Unary operator $ turns a Path into a String
+# Useful in string interpolation:
+#   def mypath = source "somefile.cpp"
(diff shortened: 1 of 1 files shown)


terpstra:
What do other people think about this change?

Dedicating a single-character operator for this seems a bit spendy for this convenience. What I like about `getPathName` is it very clearly says what's happening. OTOH, I agree it comes up a lot.

mwachs5:
is it really dedicating it ONLY for paths? Couldn't we define a `$` for things that are not paths (say, integers)?

mmjconolly:
I don't care if it's specifically `$` or something else, just something much shorter / less verbose than getPathName, since it's going to be used in a lot of Plans

terpstra:
> is it really dedicating it ONLY for paths? Couldn't we define a `$` for things that are not paths (say, integers)?

In wake, you cannot have two different functions with the same name, like you can in scala (overloaded by type).

terpstra:
Plan for now is to put something like this into our actual repositories and play around with it before promoting it into the standard library

PR #626, opened 2021-06-02
integerToByte: should return "\x00" when run on 0

terpstra:
Found this bug while writing documentation.

### DIFF
--- a/src/string.cpp
+++ b/src/string.cpp
@@ -550,7 +550,11 @@ static PRIMFN(prim_bin2str) {
   long x = ok ? mpz_get_si(arg0) : 0;
   ok = ok && x >= 0 && x < 256;
   char c[2] = { static_cast<char>(x), 0 };
-  RETURN(String::alloc(runtime.heap, ok ? c : ""));
+  if (ok) {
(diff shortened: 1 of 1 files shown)


src/string.cpp
```diff
@@ -550,7 +550,11 @@ static PRIMFN(prim_bin2str) {
   long x = ok ? mpz_get_si(arg0) : 0;
   ok = ok && x >= 0 && x < 256;
   char c[2] = { static_cast<char>(x), 0 };
-  RETURN(String::alloc(runtime.heap, ok ? c : ""));
+  if (ok) {
+    RETURN(String::alloc(runtime.heap, c, 1));
```
mmjconolly, line 554: I'm not sure I follow what's happening in this allocation, what does c[0] get used for?

terpstra: String::alloc runs the constructor for String.
When passed a single 'const char *', it expects a null-terminated string.
When c[0] = 0, then this creates an empty string.
By passing the additional ", 1", we use the String constructor that takes both a `const char *body` and a `size_t length`.

PR #628, opened 2021-06-02
Document core

terpstra:
This PR adds documentation and type signatures to every function in wake/core, except the JSON functions which I want to replace and the tree functions I will get to later.


share/wake/lib/core/list.wake
```diff
@@ -15,219 +15,455 @@
...
+#   def twice x = x, x, Nil
+#   mapFlat twice (seq 3) = 0, 0, 1, 1, 2, 2, Nil
+#   mapFlat seq (seq 5) = 0, 0, 1, 0, 1, 2, 0, 1, 2, 3, Nil
+export def mapFlat (f: a => List b): (List a => List b) =
+    def helper = match _
+        Nil  = Nil
+        h, t = f h ++ helper t
+    helper
+
+# mapPartial: create a new List by applying a partial funciton to each element of a List.
```
hcook, line 91:
```suggestion
# mapPartial: create a new List by applying a partial function to each element of a List.
```

share/wake/lib/core/option.wake
```diff
@@ -17,100 +17,200 @@ package wake
...
+#   omap (_+1) (Some 4) = Some 5
+#   omap (_+1) None     = None
+export def omap (f: a => b): Option a => Option b = match _
+    Some a = Some (f a)
+    None   = None
+
+# omapPartial: apply partial function 'f' to the optional contents
+# A partial function returns Option; only Some cases result in a value.
+#
+#   def isOdd x = x&1 == 1
```
mmjconolly, line 88: visual nit:
```
def isOdd x = x&1 == 1
```
has the `=` dividing the definition from the implementation but then the implementation has `==` on the same line. 
Just takes an extra second to mentally separate it out compared to
```
def isOdd x =
    x&1 == 1
```

share/wake/lib/core/option.wake
```diff
@@ -17,100 +17,200 @@ package wake
...
+#   def divideEven = match _
+#     x if isOdd x = None
```
mmjconolly, line 90: I'm suprised by the `if` here, I had presumed all `if`s needed a `then/else`
Are there brakets that could be added, for example: `x if (isOdd x)`

share/wake/lib/core/list.wake
```diff
@@ -15,219 +15,455 @@
...
+    _, t = t
 
-export def drop i l =
-  if i <= 0 then l else match l
-    Nil = Nil
-    _, t = drop (i-1) t
+# map: create a new List by applying a function f to each element of a List.
+#
+#   map str    (seq 5) = "0", "1", "2", "3", "4", Nil
+#   map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
```
mmjconolly, line 73: In some examples there is a `---` to separate the wake code from the result of computation
It could be nice to have a consistent way to communicate this

share/wake/lib/core/option.wake
terpstra: This `if` is a so-called "pattern guard". When you have a match expression, you can put an `if` expression to further qualify the pattern. However, in this example it's pointless, so I've removed it.

terpstra:
Since @mmjconolly wants to make a new release, I'm going to stop for now. There are a bunch of non-documentation follow-up fixes that conflict with this PR that I'd like to push after this is merged.

vector.wake was a massive slog.

I did not get to tree.wake, yet. I can do that in a later PR.

share/wake/lib/core/list.wake
terpstra: I agree, though I would like this decision to be guided by whatever the tooling we use expects.

share/wake/lib/core/list.wake
```diff
@@ -15,219 +15,495 @@
...
+            combiningFn element acc, tail
+    loop
+
+# Concatenate two lists into one List.
+# The ++ operator must be provided by the wake package for publish to work.
+#
+#   (1, 2, 3, Nil) ++ (8, 9, Nil) = (1, 2, 3, 8, 9, Nil)
+export def (l: List a) ++ (r: List a): List a = lcat r l
+# curried to expose recursive loop outside of ++
+def lcat = foldr (_,_)
```
mmjconolly, line 174: Missing return type annotation

terpstra: That's a private method.

PR #632, opened 2021-06-03
types: fix multi-arg pattern types and functions with pattern extractors

terpstra:
This fixes two bugs:
- output type signatures on functions with a pattern argument were ignored
- multi-argument patterns (or functions with multiple pattern arguments) ignored their first-level types

### DIFF
--- a/src/bind.cpp
+++ b/src/bind.cpp
@@ -580,7 +580,9 @@ static PatternTree cons_lookup(ResolveBinding *binding, std::unique_ptr<Expr> &e
     expr = std::unique_ptr<Expr>(lambda);
     out.var = 0; // bound
   } else {
-    for (ResolveBinding *iter = binding; iter; iter = iter->parent) {
+    if (ast.name.empty()) {
+      out.sum = multiarg;
+    } else for (ResolveBinding *iter = binding; iter; iter = iter->parent) {
       iter->qualify_def(ast.name);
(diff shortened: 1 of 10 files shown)


terpstra:
@mmjconolly Unit tests exercising the failure added first (see the CI failures) and then followed up by the patch that fixes the relevant bug.

src/bind.cpp
```diff
@@ -655,11 +643,29 @@ static std::unique_ptr<Expr> rebind_match(const std::string &fnname, ResolveBind
...
-    prototype.tree.children = std::move(children);
-    prototype.tree.sum = multiarg;
+    prototype.tree.sum = multiarg = std::make_shared<Sum>(AST(LOCATION));
+    multiarg->members.emplace_back(AST(LOCATION));
+    size_t i = 0;
+    for (auto &a : match->args) {
+      prototype.tree.children.emplace_back(a->location, i);
+      for (auto &p : match->patterns)
+        if (i < p.pattern.args.size())
+          a = ascribe(std::move(a), p.pattern.args[i].region, std::move(p.pattern.args[i].type));
```
terpstra, line 663: The reorganization of this function was entirely to get this line (663) distinct from the other line (652) for the cases of single-argument match and multi-argument match.

mmjconolly (approved):

PR #633, opened 2021-06-04
Allow fuse-wake input/output json files to be kept

mmjconolly:
This is useful for debugging, as `wake --failed -v` doesn't show the full environment


terpstra (approved):
Seems handy.

PR #644, opened 2021-06-10
Fix handling of symbolic links

### LINKED ISSUE #560: Hash the link destination name, not the contents of the file the symlink points at
If we have a job that outputs a broken symlink:
    `workspace/build/foo -> /some/missing/path`

We see messages printed to stderr:
    `hash_open: workspace/build/foo: No such file or directory`

From discussion: 
"It should hash the link destination name, not the contents of the file the symlink points at."

terpstra:
It turns out there were three distinct bugs with symlinks:
- wake's unlink wrapper could not remove them (this led to problems where jobs meant to overwrite a symlink, but instead created the link destination instead)
- the wake shim would hash the contents of the destination instead of the link (Fixes #560)
- the wake 'write' command would not follow a symlink to create the target instead

This PR fixes all of the above issues. It also:
- adds a unit test which fails if any of these three bugs are present
- adds a `getPathHash` method for introspecting hash codes
- fixes a recursive-fuse-bug in the unit testing framework on os/x

### DIFF
--- a/shim/shim.c
+++ b/shim/shim.c
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+#define _XOPEN_SOURCE 700
+
 /* Wake vfork exec shim */
 #include <sys/stat.h>
 #include <fcntl.h>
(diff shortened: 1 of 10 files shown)


mmjconolly:
This is odd, it seems like youre including the correct header
```
shim/shim.c:97:28: error: use of undeclared identifier 'O_NOFOLLOW'
  fd = open(file, O_RDONLY|O_NOFOLLOW);
```

shim/shim.c
```diff
@@ -15,6 +15,15 @@
...
+/* Unfortunately, OS/X so far only implements issue 6.
+ * O_NOFOLLOW was added by issue 7.
+ */
+#if !defined(O_NOFOLLOW)
+#define _DARWIN_C_SOURCE 1
+#endif
+
 /* Wake vfork exec shim */
 #include <sys/stat.h>
 #include <fcntl.h>
```
mmjconolly, line 29: FWIW, I was able to get it to build by just putting `#include <fcntl.h>` before the `#define _XOPEN_SOURCE 700`

terpstra:
Yeah, the problem is that OS/X only implements issue 6 of the open group base specification. O_NOFOLLOW was defined in issue 7 (which my sources depend on). However, OS/X does offer this as feature as an extension for which I need to opt-in.

I did the minimal fix in this PR (which also required removing the non-standard flock call).

shim/shim.c
terpstra: Well, that's not an actual fix. By moving the define later, you've essentially removed its affect.

PR #650, opened 2021-06-24
Add suggestion to error message when source file doesn't exist

jackkoenig:
Pretty self-explanatory. Many a user has had a hell of a time after moving a source file without doing a `git mv`.
I also grepped for the error message and updated it in the tutorial which uses a different error message now.


terpstra (approved):
LGTM

PR #653, opened 2021-07-06
Initial do-nothing wake LSP implementation

elena-pudovkina:
### VSCode LSP extension.
To build:
1) `./bin/wake wasm`
2) `cd extensions/vscode/`
3) `npm install`

To run:
1) Inside "extensions/vscode/": `code .`
2) Pressing Ctrl+F5 will run the extension in a new VSCode window.


extensions/vscode/lsp/.eslintignore
```diff
@@ -0,0 +1,5 @@
+node_modules/**
+client/node_modules/**
+client/out/**
+server/node_modules/**
+server/out/**
```
terpstra, line 5: I believe server/ no longer exists, so this should be removed.

extensions/vscode/lsp/.eslintrc.js
```diff
@@ -0,0 +1,20 @@
...
+		'plugin:@typescript-eslint/recommended',
+	],
+	rules: {
+		'semi': [2, "always"],
+		'@typescript-eslint/no-unused-vars': 0,
+		'@typescript-eslint/no-explicit-any': 0,
+		'@typescript-eslint/explicit-module-boundary-types': 0,
+		'@typescript-eslint/no-non-null-assertion': 0,
+	}
+};
```
terpstra, line 20: Is this file actually used by the build? It seems to be a configuration file for checking code style guidelines for typescript. If we include this file, we should actually run the linter. Do we?

extensions/vscode/lsp/.gitignore
```diff
@@ -0,0 +1,4 @@
+out
+node_modules
+client/server
+.vscode-test
```
terpstra, line 4: client/server?
This seems inaccurate.

We should create a .gitignore ourselves after creating the vsix build rule and seeing what files `git status` reports as cruft after that completes.

extensions/vscode/lsp/client/package.json
```diff
@@ -0,0 +1,23 @@
+{
+	"name": "lsp-wake-client",
+	"description": "VSCode part of a language server for wake",
+	"author": "SiFive",
+	"license": "MIT",
```
terpstra, line 5: wake is licensed under the Apache 2.0 license. The extension should presumably be the same.

extensions/vscode/lsp/client/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+
+
+let client: LanguageClient;
+
+export function activate(context: ExtensionContext) {
+	const serverModule = context.asAbsolutePath('/lsp-wake.wasm-cpp11-release');
+
+	// The debug options for the server
+	// --inspect=6009: runs the server in Node's Inspector mode so VS Code can attach to the server for debugging
+	let debugOptions = { execArgv: ['--nolazy', '--inspect=6009'] };
```
terpstra, line 23: I'm betting this has no effect, since our server is a wasm blob.
Perhaps we should remove the 'debug:' below as well?

extensions/vscode/lsp/client/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+			module: serverModule,
+			transport: TransportKind.stdio,
+			options: debugOptions
+		}
+	};
+
+	// Options to control the language client
+	let clientOptions: LanguageClientOptions = {
+		// Register the server for plain text documents
+		documentSelector: [{ scheme: 'file', language: 'plaintext' }],
```
terpstra, line 42: This is certainly wrong; we should select all the .wake files in a project.

extensions/vscode/lsp/client/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+		synchronize: {
+			// Notify the server about file changes to '.clientrc files contained in the workspace
+			fileEvents: workspace.createFileSystemWatcher('**/.clientrc')
```
terpstra, line 45: Similarly, we probably only want to notify the server about changes to `**/*.wake` files.

extensions/vscode/lsp/client/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+		}
+	};
+
+	// Create the language client and start the client.
+	client = new LanguageClient(
+		'languageServerExample',
+		'Language Server Example',
```
terpstra, line 52: Please put something sensible here.

extensions/vscode/lsp/client/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
```
terpstra, line 1: I think you should probably reorganize the directory structure.
What is currently in `extensions/vscode/lsp` should probably just be: `extensions/vscode`, since it's not just an LSP extension, but the general-purpose vscode extension for wake. Conversely, since there is no `server`, only a `client`, I think the `extensions/vscode/lsp/client` directory should be `extensions/vscode/lsp-client`.

extensions/vscode/lsp/language-configuration.json
```diff
@@ -0,0 +1,30 @@
+{
```
terpstra, line 1: This file was taken from `share/doc/wake/syntax/vscode`. I think you should remove `share/doc/wake/syntax/vscode` as part of this PR (which will then also mark this file as moved instead of freshly created). When you remove `share/doc/wake/syntax/vscode`, please also move the README.md (and perhaps enhance it).

extensions/vscode/lsp/package.json
```diff
@@ -0,0 +1,76 @@
+{
```
terpstra, line 1: Why do we need two package.json files?

lsp/lsp.cpp
```diff
@@ -34,11 +34,8 @@
 #include <sstream>
 #include <fstream>
 
-// Begin log
-std::ofstream logfile;
-
-#include "json5.h"
-#include "execpath.h"
+#include <json5.h>
+#include <execpath.h>
```
terpstra, line 38: In the wake codebase, I use `"foo.h"` to denote a header file in the wake git repository and `<bar.h>` to denote system libraries. Therefore, please do not make this change.

lsp/lsp.cpp
```diff
@@ -124,26 +136,48 @@ int main(int argc, const char **argv)
...
       response.children.emplace_back("id", id);
 
-      JAST &error = response.add("error", JSON_OBJECT);
-      error.add("code", JSON_INTEGER, MethodNotFound);
-      error.add("message", "Method '" + method + "' is not implemented.");
+      if (isShutDown && (method != "exit")) {
+        JAST &error = response.add("error", JSON_OBJECT);
+        error.add("code", JSON_INTEGER, InvalidRequest);
+        error.add("message", "Received a request other than 'exit' after a shutdown request.");
+      }
```
terpstra, line 158: The style in the rest of the repository is to put the `else if` on the same line; eg:
```
if (abc) {
  ...
} else if (def) {
  ...
}
```

lsp/lsp.cpp
```diff
@@ -124,26 +136,48 @@ int main(int argc, const char **argv)
...
+        error.add("code", JSON_INTEGER, ServerNotInitialized);
+        error.add("message", "Must request initialize first");
+      }
+      else if (method == "shutdown") {
+        JAST empty(JSON_OBJECT);
+        response.children.emplace_back("result", empty);
+        isShutDown = true;
+      }
+      else if (method == "exit") {
+        if (isShutDown) return 0;
```
terpstra, line 173: Since `return` is such an important directive, it should be highly visible, not hidden on the end of a line.
There are two ways you can achieve that:

```
if (isShutDown) {
  return 0;
} else {
  return 1;
}
```

Or, my preference: `return isShutdown?0:1;`
I'd say you should either p

terpstra (commented):
This is some great progress.

Before we merge this PR, we need to include rules to actually build the vsix and run those rules as part of the pull-request-triggered automated testing.

terpstra:
Also, please update the top-level PR comment with at least some bare-bones instructions on how to build and test the extension.

extensions/vscode/lsp/client/src/extension.ts
terpstra: According to https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#documentFilter this should probably be:
```
documentSelector: [{ language: 'wake', pattern: '**/*.wake' }]
```

extensions/vscode/lsp/.eslintignore
elena-pudovkina: I agree.

extensions/vscode/lsp/.gitignore
elena-pudovkina: Okay.

extensions/vscode/lsp/client/package.json
elena-pudovkina: I changed it to Apache 2.0.

extensions/vscode/lsp/client/src/extension.ts
elena-pudovkina: Changed it.

extensions/vscode/lsp/client/src/extension.ts
elena-pudovkina: Done.

extensions/vscode/lsp/client/tsconfig.json
elena-pudovkina: Done.

extensions/vscode/lsp/language-configuration.json
elena-pudovkina: Done.

lsp/lsp.cpp
elena-pudovkina: Done.

lsp/lsp.cpp
elena-pudovkina: Done.

lsp/lsp.cpp
elena-pudovkina: Done.

extensions/vscode/lsp/client/src/extension.ts
elena-pudovkina: Done.

extensions/vscode/lsp/.eslintrc.js
elena-pudovkina: It seems this file is unnecessary.

extensions/vscode/lsp/client/src/extension.ts
elena-pudovkina: Done.

extensions/vscode/lsp/package.json
elena-pudovkina: Merged it into one.

terpstra:
Closing in favor of #656

PR #659, opened 2021-07-13
FUSE daemon resilience

terpstra:
This PR makes it far harder for the client to miss a connection to the daemon and also enables a debug facility which can be enabled to diagnose the cause.

### DIFF
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -1532,6 +1532,10 @@ int main(int argc, char *argv[])
 	fl.l_len = 0; // 0=largest possible
 	if (fcntl(log, F_SETLK, &fl) != 0) {
 		if (errno == EAGAIN || errno == EACCES) {
+			if (enable_trace) {
+				fprintf(stderr, "fcntl(%s.log): %s -- assuming another daemon exists\n",
+					path.c_str(), strerror(errno));
+			}
 			status = 0; // another daemon is already running
 		} else {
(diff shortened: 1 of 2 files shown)


fuse/daemon.cpp
```diff
@@ -1525,6 +1532,10 @@ int main(int argc, char *argv[])
 	fl.l_len = 0; // 0=largest possible
 	if (fcntl(log, F_SETLK, &fl) != 0) {
 		if (errno == EAGAIN || errno == EACCES) {
+			if (enable_trace) {
+				fprintf(stderr, "fcntl(%s.log): %s -- assuming another daemon exists\n",
+					path.c_str(), strerror(errno));
+			}
 			status = 0; // another daemon is already running
```
mmjconolly, line 1539: Is status = 0 the same as when it starts up a new daemon, it could be nice to have a separate return code so that the client can know what happened

mmjconolly: Could check it in the `do waitpid(pid, &status, 0);`

mmjconolly (approved):

fuse/daemon.cpp
terpstra: Unfortunately, by that point the daemon already ran fork twice to cut its parent-client relationship with the client, so the client cannot recover the exit status using waitpid.

terpstra: The 'waitpid' call in the client is just there to reap the exit status of the pre-fork daemon child so it doesn't pollute the process tree.

PR #661, opened 2021-07-14
fuse: daemon exit policy should differentiate between linger + retry

terpstra:
We want to wait a while before the daemon tries to quit the first time.
Once we start trying to exit, though, we need to get the job done!

### DIFF
--- a/fuse/daemon.cpp
+++ b/fuse/daemon.cpp
@@ -54,7 +54,12 @@
 #define BAD_FD STDIN_FILENO
 
 // How long to wait for a new client to connect before the daemon exits
-static int min_timeout;
+static int linger_timeout;
+
+// How to retry umount while quitting
+// (2^8-1)*100ms = 25.5s worst-case quit time
+#define QUIT_RETRY_MS		100
+#define QUIT_RETRY_ATTEMPTS	8
 
 struct Job {
 	std::set<std::string> files_visible;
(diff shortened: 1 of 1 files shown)


terpstra:
Before #659, the daemon would worst-case quit (4+8+12)=24s after starting to exit the first time (2 << exit_attempts). Now, it will quit (1+2+4+8+...=255)*100ms=25.5s after starting to exit, keeping the retry policy more-or-less the same as it has been before.

terpstra:
For posterity, here is the failure scenario that #659 and #661 are attempting to resolve:

terpstra:
#659 fixes the issue by ensuring that the daemon will linger for at least as long as the client is waiting for it to start. Thus, the failing client in the example does not get starved by competing jobs. However, that PR messed up the retry daemon umount logic, which led to the possibility of the daemon taking a VERY long time to exit on retry while still holding the lock which prevents the next daemon from starting.

fuse/daemon.cpp
```diff
@@ -1405,10 +1420,12 @@ static void handle_exit(int sig)
...
 		do waitpid(pid, &status, 0);
 		while (WIFSTOPPED(status));
+		// Attempts numbered counting from 1:
+		fprintf(stderr, "Unable to umount on attempt %d\n", exit_attempts);
 	}
 
-	if (exit_attempts == 3) {
-		fprintf(stderr, "Unable to cleanly exit after 4 unmount attempts\n");
+	if (exit_attempts == QUIT_RETRY_ATTEMPTS) {
+		fprintf(stderr, "Too many umount attempts; unable to exit cleanly. Leaving a broken mount point behind.\n");
```
mmjconolly, line 1428: What causes unmount failures?

mmjconolly (approved):

fuse/daemon.cpp
terpstra: Anything lingering in the .fuse/ folder will cause the umount call too fail. Unfortunately, there is no way to do the umount synchronously within the FUSE loop, so it's inherently failure-prone and racey. If the daemon calls `exit(1)` before the umount succeeds, you get one of those weird dangling mount points that report '????' for their attributes and you have to manually `fusermount -u`. That's why the daemon retries pretty hard not to do an exit.

PR #665, opened 2021-07-20
src: split into directories that reflect the compiler stages

elena-pudovkina:


terpstra (approved):
Approved!

In the future, please:
- write a comment on the pull request describing the change
- avoid changing files you don't need to (I used `git checkout master lsp extensions` to restore these files)

terpstra:
I'm going to squash merge this PR, because the individual commits are not factored into individual changes, but rather organized by implementation order.

elena-pudovkina:
> I'm going to squash merge this PR, because the individual commits are not factored into individual changes, but rather organized by implementation order.

Okay, thank you!

PR #669, opened 2021-07-26
getJobResult

mmjconolly:
Adding this as I don't think we have a one liner to do `Job => Result (List Path) Error`

Regardless of future api changes to Plan/Job, we should have something like this for now so we're not asking people to do something like this on each job:
```
job
| getJobOutputs
| map getPathResult
| findFail
```

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -323,9 +323,7 @@ export def runJob (p: Plan): Job = match p
 
 # Get result for job with the list of output files/dirs.
 export def getJobResultOutputs (job: Job): Result (List Path) Error =
-    map getPathResult job.getJobOutputs
-    | findFail
-
+    findFailFn getPathResult job.getJobOutputs
 
 # Get result for job with exactly one output Path.
 export def getJobResultOutput (job: Job): Result Path Error =
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -320,6 +320,21 @@ export def runJob (p: Plan): Job = match p
         def _ = wait (\_ badfinish job error) (badlaunch job error)
         job
 
+
+# Get result for job with the list of output files/dirs.
+export def getJobResultOutputs (job: Job) : Result (List Path) Error =
+  map getPathResult job.getJobOutputs
+  | findFail
+
+
+# Get result for job with exacally one output Path.
+export def getJobResultOutput (job: Job) : Result Path Error =
+  match job.getJobOutputs
+    path, Nil = path.getPathResult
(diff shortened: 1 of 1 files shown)


hcook:
Yeah this is everywhere. `getJobOutput` as opposed to `getJobOutputs` also seems quite popular, so might be worth folding in a helper to expect / extract a single file

mmjconolly:
I've added a single Path version

share/wake/lib/system/job.wake
```diff
@@ -320,6 +320,20 @@ export def runJob p = match p
...
         job
 
+
+# Get result for job with the list of output files/dirs.
+export def getJobResultOutputs (job: Job) : Result (List Path) Error =
+  map getPathResult job.getJobOutputs
+  | findFail
+
+
+# Get result for job with exacally one output Path.
```
hcook, line 330:
```suggestion
# Get result for job with exactly one output Path.
```

share/wake/lib/system/job.wake
```diff
@@ -320,6 +320,20 @@ export def runJob p = match p
...
+
+# Get result for job with the list of output files/dirs.
+export def getJobResultOutputs (job: Job) : Result (List Path) Error =
+  map getPathResult job.getJobOutputs
+  | findFail
+
+
+# Get result for job with exactly one output Path.
+export def getJobResultOutput (job: Job) : Result Path Error =
+  match job.getJobOutputs
```
terpstra, line 332: Somehow I overlooked this PR. Unfortunately, this line is no good, because a failing job has a single BadPath output.

ag-eitilt: Is the edit I just pushed what you're getting at, or did I misunderstand what you're wanting?

terpstra: Yes

terpstra (approved):
Since we're not fixing the Job API next release, let's merge this.

ag-eitilt:
Good catch on the `findFailFn`, thanks!  How long after this is merged should I wait to start using them in federation?  My feel would be the next Wake release, especially if we add a versioned package, but I don't know what the common practice is here.

terpstra:
Hmm. There are a bunch of helper methods currently only available in the unversioned `wake`. We've just been importing them like `from wake import xyz`, though I admit this is not a very good solution.

ag-eitilt:
My main concern is that, even with the `from wake import xyz`, if we start adding these to federation before people's Wake gets updated, things break.  Even worse if people get the new functions before CI does.

mmjconolly:
> if we start adding these to federation before people's Wake gets updated

We just need to do a wake release, then the users can all migrate at once

PR #670, opened 2021-07-27
wakebox executable at bin/wakebox

mmjconolly:
Copies wakebox executable into `bin/` when building without any suffix, it used to do this, regressed during a refactor.
Also added types to the file while doing so.

The real changeset is adding this to the end of the file:
```
 54     require Pass paths =
 55         linkO cpplib libs.getSysLibLFlags objFiles "bin/wakebox"
 56 
 57     require Pass _ =
 58         match paths
 59             bin, Nil = installAs "bin/wakebox" bin | getPathResult
 60             _        = Fail "missing wakebox executable".makeError
 61 
 62     Pass paths
```

### DIFF
--- a/wakebox/build.wake
+++ b/wakebox/build.wake
@@ -17,42 +17,46 @@ package build_wake
 from wake import _
 from gcc_wake import _
 
-def buildWakeBox (Pair cpplib clib) =
-    def commonLibResult =
+def buildWakeBox (Pair (cpplib: String) (clib: String)) : Result (List Path) Error =
+    require Pass (commonLib: SysLib) =
         common cpplib
 
-    def fuseLibResult =
-        buildFuseLibObjs cpplib
-
-    def goptLibResult =
+    require Pass (goptLib: SysLib) =
         gopt clib
 
-    require Pass commonLib = commonLibResult
-    require Pass goptLib = goptLibResult
-    require Pass fuseLib = fuseLibResult
+    require Pass (fuseLib: List Path) =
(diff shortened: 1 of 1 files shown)


terpstra:
So, when installed, it has the correct name already, right?

Is this just to support running wake inside the build folder, because the suffix is there on purpose in the build folder.

mmjconolly:
> So, when installed, it has the correct name already, right?

no it only has the one with the suffix

Concretely: 
```
$ git checkout origin/master
$ make
$ ls bin | grep wakebox
wakebox.native-cpp11-release
$ git checkout wakebox-exec # this PR branch
$ make
$ ls bin | grep wakebox
wakebox
wakebox.native-cpp11-release
```

terpstra:
`make` does not install wake. The behavior you see is working as intended.

Try: `make DESTDIR=/xyz install`

terpstra:
The unsuffixed names are produced by the bootstrap `make`-phase of the build.

mmjconolly:
> `make` does not install wake. The behavior you see is working as intended.

The intended behaviour doesn't seem correct, `install` phases should be optional for functionality

For any runner that uses wakebox it would expect a particular path name which wont exist currently without the install phase

terpstra:
Most complex software projects do not run out of the project's build directory. I'll grant that wake comes pretty close. In any case, if you want to fix this, you need to change the Makefile to produce 'wakebox', not the second stage build files like this PR attempts.

PR #672, opened 2021-08-02
frontend: add reporter facility

elena-pudovkina:
Create a reporter facility for reporting errors and warnings in .wake files.
Change parser.cpp to use the facility for errors with one location.
Change main.cpp to define the reporter object.

PR #674, opened 2021-08-02
lsp: enable squiggly lines

elena-pudovkina:
Restructure lsp.cpp. Rewrite it to define the reporter object and use it for sending diagnostics to the client.
Revert extension.ts back to using the default build of the server.

### DIFF
--- a/lsp/lsp.cpp
+++ b/lsp/lsp.cpp
@@ -1,4 +1,4 @@
-/* Wake Build Server Protocol implementation
+/* Wake Language Server Protocol implementation
  *
  * Copyright 2020 SiFive, Inc.
  *
@@ -33,9 +33,20 @@
 #include <map>
 #include <sstream>
 #include <fstream>
+#include <chrono>
+#include <ctime>  
+#include <functional>
 
 #include "json5.h"
 #include "execpath.h"
+#include "location.h"
+#include "frontend/parser.h"
+#include "frontend/symbol.h"
+#include "runtime/runtime.h"
+#include "frontend/expr.h"
+#include "runtime/sources.h"
+#include "frontend/diagnostic.h"
+
 
 #ifndef VERSION
 #include "../src/version.h"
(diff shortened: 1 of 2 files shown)


terpstra:
Rebase this on top of #672 once it's merged

lsp/lsp.cpp
```diff
@@ -62,8 +73,12 @@ static const char *MethodNotFound       = "-32601";
 static const char *ServerNotInitialized = "-32002";
 //static const char *UnknownErrorCode     = "-32001";
 
+std::vector<Diagnostic> diagnostics;
+Runtime runtime(nullptr, 0, 4.0, 0);
 bool isInitialized = false;
+std::string rootUri = "";
 bool isShutDown = false;
+std::map<std::string, std::string> changedFiles;
```
terpstra, line 81: I'm not super thrilled that these are global variables. Could they perhaps be member variables of a class and then the methods which use them are member methods on that class?

lsp/lsp.cpp
```diff
@@ -74,25 +89,192 @@ static void sendMessage(const JAST &message) {
...
+  {"initialize", initialize},
+  {"initialized", initialized},
+  {"textDocument/didOpen", didOpen},
+  {"textDocument/didChange", didChange},
+  {"textDocument/didSave", didSave},
+  {"textDocument/didClose", didClose},
+  {"workspace/didChangeWatchedFiles", didChangeWatchedFiles},
+  {"shutdown", shutdown},
+  {"exit", serverExit}
+};
```
terpstra, line 260: Something about putting function pointers into a table just rubs me the wrong way. I guess it's fine.

terpstra (approved):

terpstra:
I'm going to merge this as-is. However, I would like to see this refactored to have an LSP object that contains all these global variables (including the reporter). Then, you can instantiate that class inside the main method. Transform all the methods into member methods, and you'll be able to access the members just as you do the globals now.

PR #675, opened 2021-08-03
build: refactor build a bit to work on OS/X

terpstra:
Moved profile+status to runtime for now.
Moved library dependencies from wake to runtime.


elena-pudovkina (commented):
Approved!

PR #676, opened 2021-08-04
lsp.cpp: move global variables into the LSP class

elena-pudovkina:
Move all global variables (except for reporter) into the LSP class.
Turn processing requests into an LSP method.


lsp/lsp.cpp
```diff
@@ -73,265 +73,281 @@ static const char *MethodNotFound       = "-32601";
...
-  JAST message = createResponseMessage(receivedMessage);
-  message.add("result", JSON_NULLVAL);
-  isShutDown = true;
-  sendMessage(message);
-}
+    void diagnoseFile(std::string fileUri) {
+      std::string filePath = fileUri.substr(rootUri.length() + 1, std::string::npos);
+      std::unique_ptr<Top> top(new Top);
+
+      auto fileChangesPointer = changedFiles.find(fileUri);
```
terpstra, line 281: I'd suggest you create a new/local LSPReporter here, and then assign the global pointer to it. In the future, we'd like to pass the reporter TO the lexer, so there is no need for it to exist for longer than this call.

lsp/lsp.cpp
```diff
@@ -73,265 +73,281 @@ static const char *MethodNotFound       = "-32601";
...
+            callMethod(method, request);
+          }
+        }
+      }
+    }
 
+    void callMethod(std::string method, JAST request) {
+      auto functionPointer = methodToFunction.find(method);
+      if (functionPointer != methodToFunction.end()) {
+          ((*this).*(functionPointer->second))(request);
```
terpstra, line 171: Does `this->*(functionPointer->second)(request)` work?

terpstra (approved):
This looks great!

lsp/lsp.cpp
elena-pudovkina: Done.

lsp/lsp.cpp
elena-pudovkina: Yes, thank you

PR #677, opened 2021-08-16
lsp: enable type errors and goto definition

elena-pudovkina:
The type checker now runs on the whole project and updates the errors whenever a change is made.
You can also go to a symbol definition by pressing F12.

### DIFF
--- a/common/location.h
+++ b/common/location.h
@@ -59,7 +59,7 @@ struct FileLocation {
 std::ostream & operator << (std::ostream &os, FileLocation location);
 
 struct Location {
-  std::string filename;
+  const char *filename;
   Coordinates start, end;
 
   Location(const char *filename_, Coordinates start_, Coordinates end_)
--- a/common/location.h
+++ b/common/location.h
@@ -59,7 +59,7 @@ struct FileLocation {
 std::ostream & operator << (std::ostream &os, FileLocation location);
 
 struct Location {
-  const char *filename;
+  std::string filename;
   Coordinates start, end;
 
   Location(const char *filename_, Coordinates start_, Coordinates end_)
--- a/lsp/lsp.cpp
+++ b/lsp/lsp.cpp
@@ -143,10 +143,11 @@ private:
     bool isInitialized = false;
     bool isShutDown = false;
     Runtime runtime;
+    std::vector<std::string> allFiles;
     std::map<std::string, std::string> changedFiles;
-    std::map<std::string, std::vector<Diagnostic>> diagnostics;
     std::vector<std::pair<Location, Location>> uses;
     std::vector<std::pair<std::string, Location>> definitions;
+
     std::map<std::string, LspMethod> methodToFunction = {
       {"initialize",                      &LSP::initialize},
       {"initialized",                     &LSP::initialized},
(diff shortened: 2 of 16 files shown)


src/types/type.h
```diff
@@ -20,20 +20,23 @@
 
 #include "dsu.h"
 #include <ostream>
-struct Location;
+#include "location.h"
+//struct Location;
```
terpstra, line 24: Please remove comment.

src/types/type.cpp
```diff
@@ -189,10 +191,12 @@ bool TypeVar::tryUnify(TypeVar &other) {
   return ok;
 }
 
+extern DiagnosticReporter *reporter;
```
terpstra, line 194: Can we put this into the DiagnosticReporter header for now instead?

src/types/bind.cpp
```diff
@@ -20,6 +20,7 @@
 #define _POSIX_C_SOURCE 200809L
 
 #include <iostream>
```
terpstra, line 22: Do we still need iostream?

lsp/lsp.wake
```diff
@@ -51,7 +51,7 @@ def buildLSP (Pair variant _): Result (List Path) Error =
         objFiles.flatten ++
         deps.getSysLibObjects
 
-    linkO variant (deps.getSysLibLFlags) allObjs "lib/wake/lsp-wake"
+    linkO variant (deps.getSysLibLFlags) allObjs "bin/lsp-wake"
```
terpstra, line 54: I see. This solves the 'find the standard library' problem.

common/location.h
```diff
@@ -59,7 +59,7 @@ struct FileLocation {
 std::ostream & operator << (std::ostream &os, FileLocation location);
 
 struct Location {
-  const char *filename;
+  std::string filename;
```
terpstra, line 62: This can't stay. I'll figure it out.

terpstra (approved):
Looks good! Just a few minor nits and the Location:filename issue.

terpstra:
This PR is using the same branch as your last one. Generally, you should make a new branch for each feature change.

src/types/type.cpp
terpstra: Done

src/types/type.h
terpstra: Done

src/types/bind.cpp
terpstra: Done

common/location.h
terpstra: Done

lsp/lsp.cpp
```diff
@@ -460,8 +463,16 @@ class LSP {
...
+  if (argc == 2) {
+    stdLib = argv[1];
+  } else {
+    stdLib = find_execpath() + "/../../share/wake/lib";
+  }
+  if (access((stdLib + "/core/boolean.wake").c_str(), F_OK) != -1) {
+    LSP lsp(stdLib);
+    // Process requests until something goes wrong
+    lsp.processRequests();
+  }
```
terpstra, line 477: You probably should print an error message and return non-zero if the access fails.

terpstra: Whatever. Fix it in a follow-up. This PR is green and already very large.

PR #678, opened 2021-08-18
lsp: add "find references" and "highlight occurrences"

elena-pudovkina:
It is now possible to see all references to a symbol in a project by pressing Shift+F12.
If you click on a symbol once, all its occurrences in the document will be highlighted.
Additionally, find_all_wakefiles() was moved to frontend to lessen its dependency on runtime.


src/frontend/wakefiles.cpp
```diff
@@ -0,0 +1,361 @@
...
+#include <sstream>
+#include <fstream>
+#include <iostream>
+#include <algorithm>
+
+#include "primfn.h"
+#include "execpath.h"
+#include "frontend/wakefiles.h"
+
+bool push_files(std::vector<std::string> &out, const std::string &path, int dirfd, const RE2 &re, size_t skip) {
```
terpstra, line 36: Damn. RE2 is the library we don't want to depend on.

terpstra (approved):
LGTM!

PR #684, opened 2021-08-28
LSP: re-enable build using wasm in CI

terpstra:
This restores the ability to compile the LSP using emscripten.
The build system pulls in re2 and builds it along with the rest of the project.
The wake standard library is also now included into the vsix.


terpstra:
It even worked in federation! ... crashes are much worse under wasm than native, though. :-(

elena-pudovkina:
Looks good!

PR #686, opened 2021-08-29
LSP: remove the log file for every request

terpstra:
This file gets enormous.
We're also past the point it's useful.

@elena-pudovkina Or are you still using it?


elena-pudovkina:
> This file gets enormous.
> We're also past the point it's useful.
> 
> @elena-pudovkina Or are you still using it?

I use it sometimes, but it's not a big issue for me to add it in again. Better to remove it from master.

elena-pudovkina (approved):

PR #697, opened 2021-09-02
LSP: defer reparsing of the project until 1s of no typing

terpstra:
Whenever a file is potentially changed, we record we're out-of-date.
Whenever a query arrives, if we are out-of-date, we update.
If no new requests arrive for 1 second, we update.

This should improve the perceived performance.


elena-pudovkina (approved):
These look great!

PR #704, opened 2021-09-08
lsp restructure

elena-pudovkina:
Split lsp.cpp into server part, ast part, and json converter part.


terpstra (approved):
There are a bunch of minor things I'd like to see improved, but I think these can all just be done in follow-up PRs down the road.

Issue #705, opened 2021-09-09
Type checker crash on very particular type annotation


ag-eitilt:
The VS Code extension is in general very helpful and surprisingly stable, but I have found one situation where, for some reason, the entire thing panics.  I've been adding type annotations to the standard library, and when I tried editing the following line to the minimal example below (or the line below that), one of the VS Code notifications popped up with "request failed" and I lose every feature provided by the LSP, until I revert the change.

https://github.com/sifive/wake/blob/1001abb13f8cdd96e33709417eebe57f855dcba2/share/wake/lib/versions/v0_19.wake#L34
```
export def vfoldmap f a g (v: Vector a): b = vmapReduce g f a v
```

Interestingly, I've had it happen before I could finish typing the full "Vector", so it might be an issue with any unknown input type (note that Vector doesn't seem to be imported in that file) when there's also an annotation on the output type.  I've not experimented to try to reduce the minimal example further, so I don't know if it needs anything else from that file's environment, or if it doesn't need anything beyond a single (similar) line.

The message thrown when that happens is copied below, in case reading it's easier than triggering the error manually.

```
Assertion failed: var_dob, at: src/types/type.cpp,127,do_unify
RuntimeError: abort(Assertion failed: var_dob, at: src/types/type.cpp,127,do_unify). Build with -s ASSERTIONS=1 for more info.
/home/sammay/.vscode-server/extensions/sifive.wake-0.1.5/lsp-server/lsp-wake.js:1
[...a very long line cut...]
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            

RuntimeError: abort(RuntimeError: abort(Assertion failed: var_dob, at: src/types/type.cpp,127,do_unify). Build with -s ASSERTIONS=1 for more info.). Build with -s ASSERTIONS=1 for more info.
    at process.abort (/home/sammay/.vscode-server/extensions/sifive.wake-0.1.5/lsp-server/lsp-wake.js:1:8635)
    at process.emit (events.js:315:20)
    at processPromiseRejections (internal/process/promises.js:245:33)
    at processTicksAndRejections (internal/process/task_queues.js:94:32)
[Info  - 2:19:27 PM] Connection to server got closed. Server will restart.
[Error - 2:19:27 PM] Request textDocument/documentSymbol failed.
Error: Connection got disposed.
[...]

PR #707, opened 2021-09-09
Export type Vector

ag-eitilt:
Mostly just for type annotations in non-Wake code; it originally came up in documenting the `v0_19_wake` module, but in doing so did reveal #705.


ag-eitilt:
Oops, the original PR has my entire, monolithic annotation commit.  One second, that's not meant to be on the branch.

terpstra (approved):
I'm surprised that was not exported already!

ag-eitilt:
Every other (user-facing) datatype is indeed released; this one just must have gotten missed somehow.

PR #710, opened 2021-09-10
core: add basic type annotations to exported symbols

ag-eitilt:
I've been reading through the Wake code and making sure I stay attentive by adding type annotations.  This is unfortunately a rather large PR, but it should hopefully be relatively straight-forward to review (I extracted the harder-to-glance-at patches to be sent later).  Let me know if you'd prefer smaller chunks, in which case I'll probably just break this up by file.

### DIFF
--- a/share/wake/lib/core/boolean.wake
+++ b/share/wake/lib/core/boolean.wake
@@ -24,7 +24,11 @@ export data Boolean =
 # Unary operator for Boolean NOT.
 # !True  = False
 # !False = True
-export def !x = if x then False else True
+export def !(x: Boolean): Boolean =
+    if x then
+        False
+    else
+        True
 
 # Binary operator for Boolean AND.
 # BEWARE: unlike other languages, in wake, expression 'y' is evaluated even if 'x' is False
@@ -32,7 +36,11 @@ export def !x = if x then False else True
 # False && True  = False
 # True  && False = False
 # False && False = False
-export def x && y = if x then y else False
+export def (x: Boolean) && (y: Boolean): Boolean =
+    if x then
+        y
+    else
+        False
 
 # Binary operator for Boolean OR.
 # BEWARE: unlike other languages, in wake, expression 'y' is evaluated even if 'x' is True
@@ -40,4 +48,8 @@ export def x && y = if x then y else False
 # False || True  = True
 # True  || False = True
 # False || False = False
-export def x || y = if x then True else y
+export def (x: Boolean) || (y: Boolean): Boolean =
+    if x then
+        True
+    else
+        y
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -42,7 +42,7 @@ export data List a =
 #   empty (seq 0)  = True
 #   empty (1, Nil) = False
 #   empty (seq 9)  = False
-export def empty = match _
+export def empty: List a => Boolean = match _
     Nil = True
     _   = False
 
@@ -71,7 +71,7 @@ export def tail: List a => List a = match _
 #
 #   map str    (seq 5) = "0", "1", "2", "3", "4", Nil
 #   map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
-export def map (f: a => b): (List a => List b) =
+export def map (f: a => b): List a => List b =
     def loop = match _
         Nil  = Nil
         h, t = f h, loop t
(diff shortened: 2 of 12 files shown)


share/wake/lib/core/list.wake
```diff
@@ -71,7 +71,7 @@ export def tail: List a => List a = match _
 #
 #   map str    (seq 5) = "0", "1", "2", "3", "4", Nil
 #   map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
-export def map (f: a => b): (List a => List b) =
+export def map (f: a => b): List a => List b =
```
terpstra, line 74: Any reason for removing the `()`s?
They were there to indicate the idea that map turns a function on elements into a function on lists.

share/wake/lib/core/boolean.wake
```diff
@@ -24,20 +24,32 @@ export data Boolean =
 # Unary operator for Boolean NOT.
 # !True  = False
 # !False = True
-export def !x = if x then False else True
+export def !(x: Boolean): Boolean =
+    if x then
+        False
+    else
+        True
```
terpstra, line 31: (lol)

share/wake/lib/core/vector.wake
```diff
@@ -61,15 +61,15 @@ export def vectorToList: Vector a => List a =
 #
 #   vempty (vseq 4) = False
 #   vempty (vseq 0) = True
-export def vempty (Vector _ s e): Boolean =
+export def vempty (Vector _ s e: Vector a): Boolean =
```
terpstra, line 64: I'm uncertain that a type signature of the same name as the destructor actually improves readability.

terpstra (commented):
I'm not sure what I think about this PR.

The type signatures outside of the standard library are all certainly welcome.
The few overlooked ones in Integer, Double, and empty are fine, too.
I think the `()` removal in List is actually a regression.

The reason json/tree/job did not have them is because I wanted to see those files completely refactored to follow the style guidelines and have comment blocks describing them. I'd be a lot happier to see those changes made to these files, rather than just the superficial type signatures added.

share/wake/lib/core/list.wake
ag-eitilt: No specific reason beyond the fact that this file is the only one which includes the parentheses; I assumed it was an early implementation of the type annotation syntax which was never updated.  I'm perfectly happy to amend the commit to leave them in, but I will say that (A), the extension hover info boxes don't include the parentheses even if the definition does, and (B), with my familiarity with functional programming, I assume the connotations of that currying even without it being called out.

share/wake/lib/core/boolean.wake
ag-eitilt: Yeah, something about expanding the oneliner to match the style guide makes the humor of the definition stand out even more.  Or am I missing something?

share/wake/lib/core/vector.wake
ag-eitilt: Maybe not, and this might indeed be one place where we explicitly go against the style guide.  The one thing it does emphasize is that the function can take a `Vector` *of any type*.

share/wake/lib/core/list.wake
terpstra: I guess it makes less sense now than when it was typed as `(a => b) => (List a => List b)`.

The hover extension does not show the `()`s, because it reformats (with minimal `()`s) the unified final type.

share/wake/lib/core/vector.wake
terpstra: Actually, probably what would be more readable would be:
```
def vempty (v: Vector a): Boolean =
  def Vector _ s e = v
  ...
```

That function was written before we had the def extractor syntax available.

share/wake/lib/core/boolean.wake
terpstra: There's nothing wrong with the change. It's just a funny function, made more funny by being 4 lines long.

terpstra:
> Makes sense, and I can certainly update the style throughout the files; I actually already did a good chunk of it, and just cherry-picked the one-line fixes to give you a bit smaller of a PR to review initially.

If you are doing it as a follow-up review, then I approve!
The missing type signatures were just there to shame me into the larger action.

> I think comments should be done as part of a separate pass through the standard library, though, after we've formalized @elena-pudovkina's Markdown format.

Well, I think there's writing the comments and there's formatting the comments. When that comment reformatting PR comes, it's going to be mostly mechanical brain-powered-off work. Having the hard part (correct docs) already in place does not seem a waste of time to me.

> Should I change the indentation to match the four-space guideline in the same PR? I don't know how GitHub presents that, but in my tools that would result in a rather hard-to-read diff.

You can render with `?w=1` on the end of a github PR URL to hide the whitespace changes. Also, you can review PRs in vscode ... with the wake extension!

terpstra (approved):
Ok!

PR #716, opened 2021-09-14
extensions/vscode: make the standard library install read-only

terpstra:
This makes it less likely someone will accidentally modify the standard
library included with the extension in ~/.vscode/extensions/


terpstra:
Screen Shot 2021-09-13 at 5 34 30 PM

elena-pudovkina:
Ooh, very nice!

PR #717, opened 2021-09-14
extension: make the vsix file smaller

elena-pudovkina:
All images needed by the extension's README.md are now compressed and moved to online_images/
They are excluded from the vsix file and are downloaded from the github page instead.


terpstra (approved):
Thanks for this!

PR #718, opened 2021-09-14
lsp: support creating/deleting/renaming files

elena-pudovkina:
When a file is deleted, its diagnostics are cleaned.
When a file is created/deleted, needsUpdate is set to true to account for created/deleted/renamed files.

This allows to create/delete/rename not only files, but folders too.

### DIFF
--- a/tools/lsp-wake/lsp.cpp
+++ b/tools/lsp-wake/lsp.cpp
@@ -530,15 +530,29 @@ private:
 
     void didChangeWatchedFiles(const JAST &receivedMessage) {
       JAST jfiles = receivedMessage.get("params").get("changes");
-      size_t changed = 0;
       for (auto child: jfiles.children) {
         std::string fileUri = child.second.get("uri").value;
-        changed += astree.changedFiles.erase(JSONConverter::decodePath(fileUri));
+        std::string filePath = JSONConverter::decodePath(fileUri);
+        astree.changedFiles.erase(filePath);
+
+        int eventType = stoi(child.second.get("type").value);
+        switch (eventType) {
+          case 1: // The file got created
+            needsUpdate = true;
+            break;
+          case 2: // The file got changed
+            break;
+          case 3: // The file got deleted
+            // Clean its diagnostics
+            std::vector<Diagnostic> emptyDiagnostics;
+            JAST fileDiagnosticsJSON = JSONConverter::fileDiagnosticsToJSON(filePath, emptyDiagnostics);
+            sendMessage(fileDiagnosticsJSON);
+
+            needsUpdate = true;
+        }
       }
-      if (changed) {
-        needsUpdate = true;
-        refresh("watch-list-updated");
-      }
+
+      refresh("files-created-or-deleted");
     }
 
     void shutdown(const JAST &receivedMessage) {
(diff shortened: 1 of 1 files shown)


tools/lsp-wake/lsp.cpp
```diff
@@ -530,15 +530,29 @@ class LSPServer {
...
-        refresh("watch-list-updated");
+        std::string filePath = JSONConverter::decodePath(fileUri);
+        astree.changedFiles.erase(filePath);
+
+        int eventType = stoi(child.second.get("type").value);
+        switch (eventType) {
+          case 1: // The file got created
+            needsUpdate = true;
+            break;
+          case 2: // The file got changed
```
terpstra, line 543: I think we need to update in this case, too. This indicates someone modified a file external to vscode. We need to refresh to see if that change affected our diagnosis.

tools/lsp-wake/lsp.cpp
```diff
@@ -530,15 +530,29 @@ class LSPServer {
...
+            // Clean its diagnostics
+            std::vector<Diagnostic> emptyDiagnostics;
+            JAST fileDiagnosticsJSON = JSONConverter::fileDiagnosticsToJSON(filePath, emptyDiagnostics);
+            sendMessage(fileDiagnosticsJSON);
+
+            needsUpdate = true;
+        }
       }
+
+      refresh("files-created-or-deleted");
```
terpstra, line 555: Probably we should just put a `needsUpdate |= !jfiles.children.empty();` before this refresh. Then, we can replace the `switch` with an `if (deleted) ...`

terpstra (changes requested):
What do you think?

terpstra:
While testing this, I've noticed that if you rename a MODIFIED file, we lose the updated version until they type again. I'm not sure how to fix this, as the server doesn't seem to give us a notification of rename...

terpstra:
This 'externally modified' notification is amazing, though! I can edit files with joe while vscode is running and the problems window auto updates!

terpstra:
I'm going to merge this as-is, because I think dealing with rename of a dirty (modified) vscode file is a separate issue and this PR already makes things much better!

elena-pudovkina:
> While testing this, I've noticed that if you rename a MODIFIED file, we lose the updated version until they type again. I'm not sure how to fix this, as the server doesn't seem to give us a notification of rename...

Yeah that makes sense, there's no way to distinguish between a renamed and a deleted file..

tools/lsp-wake/lsp.cpp
```diff
@@ -530,14 +530,22 @@ class LSPServer {
...
+          JAST fileDiagnosticsJSON = JSONConverter::fileDiagnosticsToJSON(filePath, emptyDiagnostics);
+          sendMessage(fileDiagnosticsJSON);
+        }
       }
-      if (changed) {
+
+      if (!jfiles.children.empty()) {
         needsUpdate = true;
-        refresh("watch-list-updated");
+        refresh("files-created-or-deleted");
```
elena-pudovkina, line 548: Maybe it's something like "watched-files-changed"? Since we refresh for modified files also.

PR #720, opened 2021-09-14
wake: add an '--lsp' option to start the LSP service

terpstra:
This is useful, because wake is in the PATH while lsp-wake is not.


mmjconolly:
thanks!

PR #723, opened 2021-09-15
map: include an example of how to format comments

terpstra:
Notice in the PR how I also added a name for the inner function argument and it then made it into the type signature shown in the tool-tip, matching the documentation.

Screen Shot 2021-09-14 at 5 07 41 PM

### DIFF
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -67,16 +67,16 @@ export def tail: List a => List a = match _
     Nil  = Nil
     _, t = t
 
-# Create a new List by applying the function *mapFn* to each element of *list*.
-# The *map* function (along with *foldl*) is generally how one implements loops in wake.
-# This function (like most in wake) runs *mapFn* in parallel.
+# Create a new List by applying the function `mapFn` to each element of `list`.
+# The `map` function (along with `foldl`) is generally how one implements loops in wake.
+# This function (like most in wake) runs `mapFn` in parallel.
 #
 # Parameters:
-#  - *mapFn*: The function to apply to each element
-#  - *list*: The List of elements to feed to *mapFn*
+#  - `mapFn`: The function to apply to each element
+#  - `list`: The List of elements to feed to `mapFn`
 #
 # Guarantees:
-#  - The resultant List has the same length as *list*
+#  - The resultant List has the same length as `list`
 #
 # ```
 # map str     (3, 9, Nil) = "3", "9", Nil
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -360,17 +360,29 @@ export def forall (f: a => Boolean): List a => Boolean =
     def not = ! f _
     ! exists not _
 
-# splitBy: partition a List into those elements with `f` True and False
+# Partition one `list` into two Lists based on the output of `acceptFn`.
+# Every element of `list` appears in exactly one of the output Lists.
+# Two elements in an output List retain the order they had in `list`.
 #
-#   def isEven x = x%2 == 0
-#   splitBy isEven (seq 6) = Pair (0, 2, 4, Nil) (1, 3, 5, Nil)
-export def splitBy (f: a => Boolean): List a => Pair (List a) (List a) =
-    def loop = match _
+# Parameters:
+#  - `acceptFn`: The Boolean function which categorizes each element
+#  - `list`: The List of elements to be categorized by `True` / `False`
+#
+# Returns `Pair true false`, where:
+#  - `true`:  List of elements from `list` for which `acceptFn` returned `True`
+#  - `false`: List of elements from `list` for which `acceptFn` returned `False`
+#
+# ```
+# def isEven x = x%2 == 0
+# splitBy isEven (0, 1, 3, 5, 6, Nil) = Pair (0, 6, Nil) (1, 3, 5, Nil)
+# ```
+export def splitBy (acceptFn: a => Boolean): List a => Pair (true: List a) (false: List a) =
+    def loop list = match list
         Nil  = Pair Nil Nil
         h, t =
             # don't wait on f to process tail:
             def Pair u v = loop t
-            if f h then
+            if acceptFn h then
                 Pair (h, u) v
             else
                 Pair u (h, v)
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -74,14 +74,13 @@ export def tail: List a => List a = match _
 # Parameters:
 #  - *mapFn*: The function to apply to each element
 #  - *list*: The List of elements to feed to *mapFn*
-#  - returns: A new List with the outputs of *mapFn*
 #
 # Guarantees:
 #  - The resultant List has the same length as *list*
 #
 # ```
-# map str    (seq 5) = "0", "1", "2", "3", "4", Nil
-# map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
+# map str     (3, 9, Nil) = "3", "9", Nil
+# map (_+100) (3, 9, Nil) = 103, 109, Nil
 # ```
 export def map (mapFn: a => b): List a => List b =
     def loop list = match list
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -67,14 +67,26 @@ export def tail: List a => List a = match _
     Nil  = Nil
     _, t = t
 
-# map: create a new List by applying a function f to each element of a List.
+# Create a new List by applying the function *mapFn* to each element of *list*.
+# The *map* function (along with *foldl*) is generally how one implements loops in wake.
+# This function (like most in wake) runs *mapFn* in parallel.
 #
-#   map str    (seq 5) = "0", "1", "2", "3", "4", Nil
-#   map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
-export def map (f: a => b): List a => List b =
-    def loop = match _
+# Parameters:
+#  - *mapFn*: The function to apply to each element
+#  - *list*: The List of elements to feed to *mapFn*
+#  - returns: A new List with the outputs of *mapFn*
+#
+# Guarantees:
+#  - The resultant List has the same length as *list*
+#
+# ```
+# map str    (seq 5) = "0", "1", "2", "3", "4", Nil
+# map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
+# ```
+export def map (mapFn: a => b): List a => List b =
+    def loop list = match list
         Nil  = Nil
-        h, t = f h, loop t
+        h, t = mapFn h, loop t
     loop
 
 # mapFlat: create a new List by applying a function f to each element and concatenating the output.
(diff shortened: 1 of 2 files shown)


terpstra:
I've also noticed there is a sweet spot in comment length. If you add too much documentation, the tooltip has a scroll bar in it and does not display more. This about as big as you can make it while keeping the hover text useful.

ag-eitilt (approved):
Looks good to me!  The two questions I have aren't anything to block this: first, whether it might look better to have symbol names formatted as `` `code` `` rather than in *\*italics\**.  I do see it as more likely to have issues if the default font is changed to a monospaced one (is that possible in VS Code?), though, so it may not be the best idea even if it does stand out more with the default settings.  Second, how does this comment format look in NeoVim?

Also, I just want to say how nice of a feature it is that the LSP picks up the variable names even of eta-reduced functions!

share/wake/lib/core/list.wake
```diff
@@ -67,14 +67,26 @@ export def tail: List a => List a = match _
     Nil  = Nil
     _, t = t
 
-# map: create a new List by applying a function f to each element of a List.
-#
-#   map str    (seq 5) = "0", "1", "2", "3", "4", Nil
-#   map (_+10) (seq 5) = 10, 11, 12, 13, 14, Nil
-export def map (f: a => b): List a => List b =
-    def loop = match _
+# Create a new List by applying the function *mapFn* to each element of *list*.
```
richardxia, line 70: Any reason you chose to use italics (`*`) instead of monospace (<code>`</code>) for variable/parameter names? I think monospace formatting is more common in documentation comments across various languages.

share/wake/lib/core/list.wake
```diff
@@ -67,14 +67,26 @@ export def tail: List a => List a = match _
...
+# The *map* function (along with *foldl*) is generally how one implements loops in wake.
```
richardxia, line 71: One style convention that I see often in other languages is to put a blank line between the first line/sentence and subsequent sentence. The style also involves making sure that first sentence is a good, single-line summary sentence.

I don't know what it'll look like in VSCode, since I don't use it, but many documentation site generators will often display only that first line in scenarios where real estate is limited. For example, in Scala, [only that first line is displayed](https://www.scala-lang.org/api/2.13.3/scala/collection/immutable/List.html) in the list of all methods, and you have to click on the method to see the longer description.

Here's examples of language style guides for the separate summary line:

- Scala: https://docs.scala-lang.org/style/scaladoc.html
- Python: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings
- Doxygen (many languages, but probably mostly C++): https://www.doxygen.nl/manual/docblocks.html#cppblock

share/wake/lib/core/list.wake
```diff
@@ -67,14 +67,26 @@ export def tail: List a => List a = match _
...
+# This function (like most in wake) runs *mapFn* in parallel.
+#
+# Parameters:
+#  - *mapFn*: The function to apply to each element
+#  - *list*: The List of elements to feed to *mapFn*
+#  - returns: A new List with the outputs of *mapFn*
```
richardxia, line 77: I normally see return values documented separately from the parameters, and oftentimes not really documented in a way that is separate from the main description.

I couldn't find many examples of this in other languages, since most go with the directive-style thing where you add something like `@param`, `:param`, or `\param` for each parameter, and have a separate directive for `@returns`. Swift has a Markdown-style format where they have [separate Parameters and Returns sections](https://github.com/apple/swift/blob/3680f9c2602e7a8c6504f512c017922eaa20ffc2/docs/DocumentationComments.md#returns-field).

share/wake/lib/core/list.wake
```diff
@@ -67,14 +67,26 @@ export def tail: List a => List a = match _
...
+#
+# Guarantees:
+#  - The resultant List has the same length as *list*
+#
+# ```
+# map str    (seq 5) = "0", "1", "2", "3", "4", Nil
```
richardxia, line 83: Maybe a bit of a nit that is specific to this function's description, but I'm wondering if the examples here should avoid using the `seq` function, since you'd have to understand that it returns a list counting up to the argument. If you don't know `map`, I feel like it's even less likely that you know `seq`.

terpstra:
Here's what it looks like in neovim using a `:
Screen Shot 2021-09-14 at 6 04 22 PM

Here's what it looks like in neovim using a *:
Screen Shot 2021-09-14 at 6 06 02 PM

share/wake/lib/core/list.wake
terpstra: So I tried using `.

It looks ok in vscode:
Screen Shot 2021-09-14 at 6 09 32 PM

But see below for neovim, where it does not work at all.

share/wake/lib/core/list.wake
terpstra: Sure.

share/wake/lib/core/list.wake
terpstra: We could just not have a returns. It is usually pretty redundant with the function description.

terpstra: Giving it a section entirely for `return` wastes too much of the precious tooltip area IMO.

share/wake/lib/core/list.wake
terpstra: Here's what the special first-line style produces:

Screen Shot 2021-09-14 at 6 14 15 PM

share/wake/lib/core/list.wake
terpstra: I think stylistically and for neovim support, I prefer using *. However, I agree that most people would naturally write a `. It's also probably fine if someone does that, because they have not read the style guide. However, since the style guide defines what people SHOULD do, I think we're not really bound, either way.

What do you guys think?

share/wake/lib/core/list.wake
terpstra: Having the first sentence be a succinct summary is a good idea.

However, I'm not sure we need to put a blank line in the comment between it and the next line. You can have a good first sentence that can be pulled out by hypothetical future tooling and still have more sentences that end up in the same paragraph if there's space.

I think I lean towards the formatting of the original proposal, but adding the requirement that the first sentence should be a self-contained description, suitable for extract.

terpstra: Maybe the rule is: the first line in the wake comment should be a complete self-contained sentence. Subsequent text that elaborates can follow on the next line to continue the paragraph.

richardxia:
I think you should still go with backticks, and I think your neovim just doesn't have a good theme installed. Here's what it looks like for me (with vim + coc.nvim, not neovim).

With backticks (they're invisible):

Screen Shot 2021-09-14 at 8 36 47 PM

With asterisks:

Screen Shot 2021-09-14 at 8 37 19 PM

---

Something interesting that I'm noticing is that the Markdown is actually getting rendered for me, since you can see how my editor rendered the `-`-style bulleted list as a `*`-style bulleted list.

terpstra:
> I think you should still go with backticks, and I think your neovim just doesn't have a good theme installed.

@mmjconolly Do you think we can fix neovim to render the backticks properly with a new theme?

> Here's what it looks like for me (with vim + coc.nvim, not neovim).

I agree backticks look better i

> Something interesting that I'm noticing is that the Markdown is actually getting rendered for me

neovim is rendering it too (note the italics and example code). The LSP specification specifically calls out markdown as an option for this output, so clients are expected to render it, somehow.

share/wake/lib/core/list.wake
ag-eitilt: A description of the returned values can occasionally be helpful, particularly in cases where we're returning a `Pair` or `Triple`; consider `splitBy` where it would probably be best to explicitly say that the `True` values are the first element.  In most cases, though, I agree that it's not required, and maybe not even desirable.

share/wake/lib/core/list.wake
ag-eitilt: I like that formulation of the rule.  It also leads naturally into keeping that summary short, as too long of a summary will result in the first sentence taking up multiple lines.  The only thing I can see tripping it up is that, if we ever get code formatters with knowledge of comment blocks, they may want to update the line breaks in that paragraph to "better" justify the right edge; I don't think Vim's paragraph detection, at least, would be sophisticated enough to understand "ignore the first line of a comment paragraph if and only if it is the first paragraph in a documentation block".  That's probably not a big enough issue to block the guideline, though it might suggest leaning toward inserting a blank line.

share/wake/lib/core/list.wake
terpstra: Ah. In that case, I think a full section like for Parameters might be justified, along the lines of:

```
# Returns `Pair true false`, where:
#  - `true`: List of the elements for which `acceptFn` returned `True`
#  - `false`: List of the elements for which `acceptFn` returned `False
```

share/wake/lib/core/list.wake
terpstra: Ok. Backticks.

share/wake/lib/core/list.wake
ag-eitilt: I do agree that's probably the best option, though it's heavyweight enough that we will want to be a bit careful about where we use it.

mmjconolly:
> > I think you should still go with backticks, and I think your neovim just doesn't have a good theme installed.
> 
> @mmjconolly Do you think we can fix neovim to render the backticks properly with a new theme?

No idea, but I think the backticks look ok-ish anyway, as I'm used to that meaning 'code snippet` from things like slack and github and markdown

share/wake/lib/core/list.wake
terpstra: Sounds like we have a consensus, then. Add a "Returns:" section with multiple bullet points only when it's warranted.

terpstra:
Ok, it sounds like we have a rough consensus now on the comment block for `map`.

terpstra:
I've just added two more code comment blocks to this PR to cover some of the other corner cases we discussed.

In `splitBy`, notice I used tags on the return type to control names in the output tuple. Here is how `splitBy` renders:

Screen Shot 2021-09-15 at 10 52 46 AM

Here is how the `unreachable` version renders:

Screen Shot 2021-09-15 at 10 48 57 AM

Because it is so big, you can scroll down in the tooltip:
Screen Shot 2021-09-15 at 10 49 23 AM

elena-pudovkina:
Also, about the fenced code block syntax highlighting.
The language can be specified as such:
```
```ruby
require 'redcarpet'
markdown = Redcarpet.new("Hello World!")
puts markdown.to_html
```
```
If no language is set, the default is used. I'm not sure which one the vscode client uses and how to make it use a custom language highlighting rule...

elena-pudovkina:
> I think you should still go with backticks, and I think your neovim just doesn't have a good theme installed. Here's what it looks like for me (with vim + coc.nvim, not neovim).
> 
> With backticks (they're invisible):
> 
> Screen Shot 2021-09-14 at 8 36 47 PM
> 
> With asterisks:
> 
> Screen Shot 2021-09-14 at 8 37 19 PM
> 
> Something interesting that I'm noticing is that the Markdown is actually getting rendered for me, since you can see how my editor rendered the `-`-style bulleted list as a `*`-style bulleted list.

Richard, do you use something except the lsp to get syntax highlighting? Yours looks more informative.

richardxia:
> @richardxia Richard, do you use something except the lsp to get syntax highlighting? Yours looks more informative.

For syntax highlighting, I use the [vim syntax highlighting file in this repo](https://github.com/sifive/wake/blob/cb041ef9ae1711bb1b81f34685d179fe93d6186a/share/doc/wake/syntax/vim/syntax/wake.vim), which I helped write.

PR #724, opened 2021-09-15
stdlib: include a .wakeroot

terpstra:
This helps the neovim LSP operate when you follow a definition into the
standard library.  Unlike vscode, neovim can have multiple open workspaces
in the same editor.  This means that it cannot associate the standard
library with any particular project, so the standard library needs to be its
own project.  neovim defines a project by the presence of a .wakeroot
boundary.  Thus, by adding this file, when you go-to-definition into the
standard library, the LSP will work there and you can jump between uses/defs
within the standard library, but not back from the standard library to the
other workspaces.  This seems acceptable, for now.


mmjconolly (approved):
assuming it's simply copied during the install step

terpstra:
> assuming it's simply copied during the install step

Yeah, I confirmed that before I made the PR.

PR #731, opened 2021-09-16
Type annotation and style formatting

ag-eitilt:
Followup to #710, focusing on bringing `json.wake`, `tree.wake`, and `job.wake` up to parity with the style used elsewhere in the standard library.  If I revisit this again, I'll probably look at following the suggestion in https://github.com/sifive/wake/pull/710#discussion_r707616195 to deconstruct parameters on `def` lines rather than at function arguments, though doing so might align well with improving the comments and giving parameters more descriptive names.

EDIT: Sorry, I didn't realize how big this had gotten.  It probably would have been better to do a file-per-PR style (or otherwise smaller chunks) instead.  Sorry!

EDIT: The Mac failure is weird -- I don't think any of this should trigger duplicate symbol errors in C++ code, or have I done something strange?

### DIFF
--- a/build.wake
+++ b/build.wake
@@ -18,31 +18,50 @@ from wake import _
 
 # Useful build variants (arguments to all)
 def toVariant = match _
-    "default" = Pass (Pair "native-cpp11-release" "native-c11-release")
-    "static"  = Pass (Pair "native-cpp11-static"  "native-c11-static")
-    "debug"   = Pass (Pair "native-cpp11-debug"   "native-c11-debug")
-    "wasm"    = Pass (Pair "wasm-cpp11-release"   "wasm-c11-release")
-    s = Fail "Unknown build target ({s})".makeError
+    "default" =
+        Pass (Pair "native-cpp11-release" "native-c11-release")
+    "static"  =
+        Pass (Pair "native-cpp11-static"  "native-c11-static")
+    "debug"   =
+        Pass (Pair "native-cpp11-debug"   "native-c11-debug")
+    "wasm"    =
+        Pass (Pair "wasm-cpp11-release"   "wasm-c11-release")
+    s =
+        Fail "Unknown build target ({s})".makeError
 
 export def build: List String => Result String Error = match _
-    "tarball", Nil = tarball Unit | rmap (\_ "TARBALL")
+    "tarball", _ =
+        tarball Unit
+        | rmap (\_ "TARBALL")
     kind, Nil =
-        require Pass variant = toVariant kind
-        all variant | rmap (\_ "BUILD")
-    _ = Fail "no target specified (try: build default/debug/tarball)".makeError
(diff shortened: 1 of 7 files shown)


share/wake/lib/core/json.wake
```diff
@@ -17,155 +17,284 @@ package wake
...
+                    , '"'
+                    , jsonEscape key
+                    , "\":{space}"
+                    , rec acc deeper value
+                if list.empty then
+                    '{}'
+                    , rhs
+                else
+                    "\{{tabbeder}"
+                    , foldr helper ("{tabbed}\}", rhs) list | tail
```
ag-eitilt, line 168: Splitting lists like this was almost entirely because my mind wants to think that `,` binds more tightly than `|`; putting each item on separate lines emphasizes that the pipe drops the first value of the fold rather than the `"\{{tabbeder}"`.  The comas at the beginning is a habit I've picked up from Haskell, where for some reason the common style is to put commas at the beginning of the line instead of the end.

ag-eitilt: For reference, the Haskell style would be:

```haskell
else
    [ "{"
    , tabbeder
    , tail $ foldr helper [tabbed, "}", rhs] list
    ]
```

build.wake
```diff
@@ -132,22 +173,29 @@ def tarball Unit =
...
-        "wake-static_{release}.tar.xz", map getPathName files | sortBy (_<*_)
+        which "tar",
+        "--mtime={time}",
+        "--transform=s@^tmp/@wake-{release}/@",
+        "--owner=0",
+        "--group=0",
+        "--numeric-owner",
+        "-cJf",
+        "wake-static_{release}.tar.xz",
+        map getPathName files | sortBy (_<*_)
```
ag-eitilt, line 191: I don't know what I was thinking with this change. I'll still probably put the map call on a separate line, and maybe have the which "tar" separate as well, but the rest should probably stay together.

share/wake/lib/core/tree.wake
```diff
@@ -31,431 +32,622 @@ def ratioQ = 3
...
-      EQ = True
-      _  = False
+    match t (tupperLE y t)
+        _ None =
+            False
+        (Tree cmp _) (Some (Pair x _)) = match (cmp x y)
+            EQ =
+                True
+            _ =
+                False
```
ag-eitilt, line 552: To describe some of where I'm coming from, I've grown to quite dislike vertical alignment by spaces almost purely due to the fact that it ties mostly-unrelated lines together.  As soon as we decide to rename `x` to be more descriptive, or if the return stops being a `Some (Pair _ _)`, we have to change the otherwise-fine `None` line to match the new length.  Git isn't quite as sensitive to that as some other VCS, but it still adds a lot of unnecessary noise to diffs and blames.

PR #734, opened 2021-09-24
system: add additional future-compatible 'Result' shims

ag-eitilt:
Follow-up to #669 which added two shims in preparation of `BadPath` getting promoted to `Fail`, this adds similar shims to the other functions which might fail to retrieve paths.  Internal uses of those functions have been updated to use the `Result` varieties.  A few notes for discussion:

* No shim was created for `sources` in favor of only populating the returned list with successful paths; if any file is matched by the regex but would not be returned by `sourceResult`, `sources` silently discards it.
* `runIncrementalJob` doesn't verify that the post-job tarball is created successfully.  It doesn't seem to be relied upon anywhere, but if it is indeed a critical artifact, a shim can be created for that function as well.
* The shim names are a bit unwieldly.  My suggestion for migration would be to:
  - Release these as part of the next Wake version (with a versioned package)
  - Change all uses of the non-`Result` functions to either use these and/or to explicitly import that version
  - Update those functions to have the same return type as the shims in a future version, and remove `BadPath`
  - Potentially change all uses back and deprecate/eventually remove the shims?


ag-eitilt:
Ah, one second.  I'll fix that CI failure.

share/wake/lib/system/sources.wake
```diff
@@ -38,27 +38,38 @@ def raw_source file =
...
-        Nil = makeBadPath "{file}: not a source file".makeError
-        x, Nil = x
+        Nil = failWithError "{file}: not a source file"
+        x, Nil = Pass x
         _ = unreachable "a quoted RegExp can never match more than one distinct String"
 
 # Find sources files
 export def sources (dir: String) (filterRegexp: RegExp): List Path =
+    sourcesResult dir filterRegexp
+    | pathListResultShim
```
terpstra, line 66: This still removes Failure instead of BadPath..

terpstra (changes requested):
I don't think we can merge this as-is. It changes the behaviour of existing APIs that need to continue to return BadPath to elide those elements.

share/wake/lib/system/sources.wake
ag-eitilt: I thought I checked that, but apparently I was just making bad assumptions.  Sorry!

ag-eitilt:
If we can still slide it in (Wake has a shorter CI, after all), I think I fixed the `List Path` issues, but do double-check.  Must have made my final communications check last night right before you started looking at this.

PR #735, opened 2021-09-30
Deliberately fail in such a way to break the LSP

ag-eitilt:
This changes the result type enough to throw an error which is given the wrong source location (`bind.cpp:444`).

This PR is *not* meant to be merged, but simply as an easy minimal bug reproduction.


ag-eitilt:
Ah, that's the label.  Thanks!

mmjconolly:
> Ah, that's the label. Thanks!

just created it now for this PR

terpstra:
Generally, this sort of thing should be submitted as an issue, not a PR. ;)

However, thanks for the report.

PR #748, opened 2021-10-05
wake: --no-tty no longer implies --stdout=info,...

### LINKED ISSUE #747: Don't add 'info' to stdout for --no-tty
While it makes sense to add 'echo' so that the disentangled outputs match-up with the echo commands, we probably don't want to change the log detail level without an explicit opt-in like '-v'.

terpstra:
Fixes #747.

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -538,7 +538,7 @@ int main(int argc, char **argv) {
   if (debug   && !fd1) fd1 = "debug,info,echo,report,warning,error";
   if (verbose && !fd1) fd1 = "info,echo,report,warning,error";
   if (quiet   && !fd1) fd1 = "error";
-  if (!tty    && !fd1) fd1 = "info,echo,report,warning,error";
+  if (!tty    && !fd1) fd1 = "echo,report,warning,error";
   if (!fd1) fd1 = "report,warning,error";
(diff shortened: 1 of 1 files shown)


tools/wake/main.cpp
```diff
@@ -538,7 +538,7 @@ int main(int argc, char **argv) {
   if (debug   && !fd1) fd1 = "debug,info,echo,report,warning,error";
   if (verbose && !fd1) fd1 = "info,echo,report,warning,error";
   if (quiet   && !fd1) fd1 = "error";
-  if (!tty    && !fd1) fd1 = "info,echo,report,warning,error";
+  if (!tty    && !fd1) fd1 = "echo,report,warning,error";
```
mmjconolly, line 541: do the cli flags debug/verbose have effect when running --no-tty?

ag-eitilt: Yeah, they'll be matched first and override this line in specifying output levels.  I think the other things `--no-tty` does (remove spinners, etc.) will still be in effect.

terpstra: Note also that 'tty=false' not just for '--no-tty', but also for just running wake without a tty.

ag-eitilt: Ah, I missed the later `term_init`, thanks!

PR #749, opened 2021-10-05
wake: allow independent control of memory and CPU usage

### LINKED ISSUE #591: Better resource management
Make it possible to control max heap and cpu usage separately in both percentage and absolute values.

terpstra:
This will hopefully help fitting wake jobs into container slices.

Fixes #591.

### DIFF
--- a/src/runtime/job.h
+++ b/src/runtime/job.h
@@ -19,15 +19,34 @@
 #define JOB_H
 
 #include <memory>
+#include <string>
 
 struct Database;
 struct Runtime;
+
+struct ResourceBudget {
+  double percentage;
+  uint64_t fixed;
+
+  static const char *parse(const char *str, ResourceBudget &output);
+
+  ResourceBudget() : percentage(0), fixed(0) { }
+  ResourceBudget(double percentage_) : percentage(percentage_), fixed(0) { }
+
+  uint64_t get(uint64_t max_available) const {
+    if (fixed) {
+      return fixed;
+    } else {
+      return max_available * percentage;
+    }
+  }
+};
 
 struct JobTable {
   struct detail;
   std::unique_ptr<detail> imp;
 
-  JobTable(Database *db, double percent, bool debug, bool verbose, bool quiet, bool check, bool batch);
+  JobTable(Database *db, ResourceBudget memory, ResourceBudget cpu, bool debug, bool verbose, bool quiet, bool check, bool batch);
   ~JobTable();
 
   // Wait for a job to complete; false -> no more active jobs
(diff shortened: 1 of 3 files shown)


src/runtime/job.h
```diff
@@ -19,15 +19,34 @@
...
 
 #include <memory>
+#include <string>
 
 struct Database;
 struct Runtime;
 
+struct ResourceBudget {
+  double percentage;
+  uint64_t fixed;
```
richardxia, line 29: It seems like `percentage` and `fixed` have a specific rule that decides which one is actually valid (which is that `fixed` has priority if it's non-zero). Could you either document that, or change this from a `struct` to a `class` so that the members are private?

src/runtime/job.h
```diff
@@ -19,15 +19,34 @@
...
+
+  static const char *parse(const char *str, ResourceBudget &output);
```
richardxia, line 31: It's a bit unintuitive, at least to me, that this returns a string error message on failure or a nullptr otherwise just based on its type signature, so could that be documented in a short comment here?

richardxia (approved):
Minor PR comments and nits, but this LGTM.

Also, if this is for playing better with containers, are you checking the total available CPU and memory based on cgroups? It's non-blocking, but it'd be much better to check based on cgroups, which is what you'll have in a container environment like Kubernetes.

src/runtime/job.h
terpstra: Done

src/runtime/job.h
terpstra: Done.

tools/wake/main.cpp
```diff
@@ -211,6 +214,8 @@ int main(int argc, char **argv) {
   bool exports = arg(options, "exports")->count;
 
   const char *percents= arg(options, "percent")->argument;
+  const char *jobs    = arg(options, "jobs")->argument;
+  const char *memorys = arg(options, "memory")->argument;
```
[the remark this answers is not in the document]
terpstra: Done.

terpstra:
> ... check the total available CPU and memory based on cgroups

We are using cgroups also with slurm, atm, so this is a good point. It would be much nicer to just say '-m90%' instead of having to manually match the cgroup configuration using '-m50G'.

I'm not sure how to introspect this information, though. Looking now. (It's also possible though unlikely that the sys calls I use already do the right thing).

terpstra:
`/sys/fs/cgroup/memory/memory.limit_in_bytes` may be a lower limit than _PHYS_PAGES, so I should check both on linux, I guess.

It seems on linux I need to use sched_getaffinity to detect when less cores are available.

PR #750, opened 2021-10-06
fuse-waked: fix unmounting on macfuse-4

terpstra:
For whatever reason, fuse_unmount after a fork doesn't work anymore.
However, on OSX we can just call good old unmount().


richardxia (approved):
LGTM, I can confirm this fixes my mounting issues. Thanks!

PR #762, opened 2021-10-13
job: report how many resources wake will use in the build

terpstra:
Do you guys think this would be useful to see in CI logs?

```
terpstra@peach:~/wake$ ./bin/wake --no-tty build default
wake: targeting utilization for 7 threads and 15461882265 bytes of memory.
'<mkdir>' -m 0775 lib
'<mkdir>' -m 0775 lib/wake
...
```


mmjconolly:
For interactive use, it probably shouldn't display without `-v`
For non-interactive use, I don't think it matters too much assuming it's only printed once

PR #772, opened 2021-10-20
core: add transformation functions for simplifying/verifying JSON

ag-eitilt:
This is the normalization part of #709 extracted to a separate issue.  Essentially, it provides a framework for transforming the values of a JSON structure without (too many) changes to its overall shape -- a `JString` will remain a `JString`.  This will probably be most helpful in ensuring that JSON written by Wake is able to be parsed by some other tool used by any given project, but I'm sure people will find other uses as well.

The next several comments are copied from the latest review on that PR for followup here.


ag-eitilt:
```patch
+    def simplifyKeyGroups = match _
+        Nil =
+            failWithError "groupBy invariant reached in deduplicateJObjectKeys"
```
> This one should probably be `unreachable`.

I'm not a fan of `unreachable` on principle.  Since the return type already comes with a built-in `Error`, this can easily slot in to that infrastructure and never be run either way.

ag-eitilt:
```patch
+        _ _ =
+            failWithError "For compatibility, JSON objects should not have duplicate keys. While we try to recover from this where possible, not every type may be combined."
```
> I'd include in the error message the snippets that can't be aligned. Probably formatted as a JSON serialization without whitespace pattern. I'd also include the path to the offending key in the error.

Data from within the current object has been added.  Trying to print a full path from the file root would require more calculation than the current function can support.

ag-eitilt:
```patch
+        failWithError "For compatibility, JSON doubles should not be infinite."
```
> It would be pretty useful to indicate the 'path' to the object where this error occurred in the message. (eg: `Fail "document/jingle-bells/pages: infinite Double forbidden in json <= json5"`)

It would be, but the only input any of these functions have is a `JValue`.  Adding the file path will have to be up to the calling function to include an `addErrorContext`.

ag-eitilt:
```patch
+# Target the minimum JSON language described by the specifications.
+export def normalizeJSONBasic: JSONNormalize =
```
> Is normal json json4? If so, I'd include a '4' in the name and also add a 'json5' version, which does not mess around with Doubles.

"Basic" (now renamed "Compat") JSON is JSON according to the standards published by IETF/ECMA/ISO.  JSON5 is a (presumably widespread but unofficial) effort to bring it in line with ECMAScript 5 (it's not an incremental version).  I'm not convinced we need a `normalizeJSON5` object predefined; if it's critical for anyone's workflow, they can simply bypass the `JDouble` normalization:

```
def normalizeJSON5: JSONNormalize =
    normalizeJSONCompat
    | setJSONNormalizeDouble Pass
```

ag-eitilt:
```patch
+    JObject a =
```
> Both this and the JArray cases fail to run normalizeJSON recursively after they apply the local fix-up.

I was originally wanting the functions to be able to control their own normalization, but I originally implemented it incorrectly, and working through the logic, even what I thought was "correct" still could never work.  Fixed.

ag-eitilt:
```patch
+export def normalizeJSON (fmt: JSONNormalize): JValue => Result JValue Error = match _
```
> Perhaps call this `normalizeJSONWith`, and then we can have a `normalizeJSON` which defaults to the json4 variant.

`With`, sure, and we can *maybe* add a plain alias later, but I think this is probably something that every project using wake would have different opinions on depending on their toolchain: some might want the most compatibility, some might only need canonical strings, some might require objects conform a stricter schema...

ag-eitilt:
```patch
+                Pass (JArray (val ++ acc))
+        (JObject val) (JObject acc) =
+            val ++ acc
+            | deduplicateJObjectKeys
```
> The recursive call here is inappropriate. You should leave that to `normalizeJSON`, like the `JArray` case above does. ie: `Pass (JObject (val ++ acc))`

The restrictions are different.  The `JArray` case just takes the elements of one, sticks them on the elements of the other, and calls it done.  For `JObject`, both `val` and `acc` are guaranteed to not have any duplicate keys internally by this point, but the same key might occur in both and thus still needs to be filtered.  The alternate implementation would be to have `normalizeJSON` recurse *after* processing each level rather than before, but then you lose the ability to have it perform the heavy lifting for the above equality tests.

ag-eitilt:
> There are two other things missing in this PR:
>
> * The JSON pretty printing code should now format Infinitity according to JSON5, since we expect normalization to eliminate inifinity for us.
> * There needs to be a collection of unit tests for this code (see: `tests/` -- I'd recommend having a bunch of `1.in.json` and `1.out.json` files that you compare)

I'll update those in further commits before this PR should be merged.

ag-eitilt:
> @colinschmidt @ag-eitilt Could this function also be used to combine/merge multiple JValues together? This seems it may be useful for 'overrides'.

If the `JValue` are added to a `JObject` as keys with the same name, that might currently be possible, though nested `JArray` or `JObject` values would be merged rather than overwritten.  However, the code could indeed be relatively easily adapted to a more user-friendly merge/override function; I'll see what I can do towards that.

terpstra:
I don't know how to reply to individual comments. This is going be a giant mess if I just starting replying out-of-order.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+# with multiple values of different types or with different values of a type
+# which cannot be combined.
+def deduplicateJObjectKeys (dict: List (Pair String JValue)): Result (List (Pair String JValue)) Error =
+    def cmpKeysOnly (Pair k1 _) (Pair k2 _) =
+        k1 <* k2
+    def simplifyKeyGroups = match _
+        Nil =
+            # `simplifyKeyGroups` is only ever mapped over the output of
+            # `groupBy`, which is always a list of *non-empty* lists.
+            failWithError "groupBy invariant reached in deduplicateJObjectKeys"
```
ag-eitilt, line 284: > This one should probably be `unreachable`.

I'm not a fan of `unreachable` on principle. Since the return type already comes with a built-in `Error`, this can easily slot in to that infrastructure and never be run either way.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+        (JString val) (JString acc) if val ==* acc =
+            Pass (JString acc)
+        (JInteger val) (JInteger acc) if val == acc =
+            Pass (JInteger acc)
+        (JDouble val) (JDouble acc) if val ==. acc =
+            Pass (JDouble acc)
+        (JBoolean val) (JBoolean acc) if enor val acc =
+            Pass (JBoolean acc)
+        (JArray val) (JArray acc) =
+            Pass (JArray (val ++ acc))
```
ag-eitilt, line 303: > I think you're overthinking this one. Your rule is too magical. It can switch behaviours for a user-inexplicable reason. Pick one behaviour, only. I'd just go for `JArray (val ++ acc)` unconditionally, since that's the more helpful of the two options.

I went with that because then `deduplicateJObjectKeys (x ++ x) == deduplicateJObjectKeys x` (i.e. merging an object into itself wouldn't change anything), but it is admittedly rather magical. The only other solution I have would be to filter out duplicates in the second array (union of the two), but that's probably too prone to errors such as deliberately-duplicated values. Ultimately, I've just wound up going with your suggestion for an unconditional append.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+    Boolean: Boolean => Result Boolean Error
+    Object: List (Pair String JValue) => Result (List (Pair String JValue)) Error
+    Array: List JValue => Result (List JValue) Error
+
+# A basic normalization ruleset which returns inputs unchanged.
+export def normalizeJSONIdentity: JSONNormalize =
+    JSONNormalize Pass Pass Pass Pass Pass Pass
+
+# Target the minimum JSON language described by the specifications.
+export def normalizeJSONCompat: JSONNormalize =
```
ag-eitilt, line 246: > Is normal json json4? If so, I'd include a '4' in the name and also add a 'json5' version, which does not mess around with Doubles.

"Compat" (renamed from "Basic") JSON is JSON according to the standards published by IETF/ECMA/ISO. JSON5 is a (presumably widespread but unofficial) effort to bring it in line with ECMAScript 5 (it's not an incremental version). I'm not convinced we need a `normalizeJSON5` object predefined; if it's critical for anyone's workflow, they can simply bypass the `JDouble` normalization:

```
def normalizeJSON5: JSONNormalize =
    normalizeJSONCompat
    | setJSONNormalizeDouble Pass
```

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+            getJSONNormalizeDouble fmt a
+            | rmap JDouble
+            | addErrorContext path
+        JBoolean a =
+            getJSONNormalizeBoolean fmt a
+            | rmap JBoolean
+            | addErrorContext path
+        JNull =
+            Pass JNull
+        JObject a =
```
ag-eitilt, line 203: > Both this and the JArray cases fail to run normalizeJSON recursively after they apply the local fix-up.

I was originally wanting the functions to be able to control their own iteration, but I originally implemented it incorrectly, and working through the logic, even what I thought was "correct" still could never work. Fixed.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+export tuple JSONNormalize =
+    String: String => Result String Error
+    Integer: Integer => Result Integer Error
+    Double: Double => Result Double Error
+    Boolean: Boolean => Result Boolean Error
+    Object: List (Pair String JValue) => Result (List (Pair String JValue)) Error
+    Array: List JValue => Result (List JValue) Error
+
+# A basic normalization ruleset which returns inputs unchanged.
+export def normalizeJSONIdentity: JSONNormalize =
```
ag-eitilt, line 242: > Pretty sure we don't want this one to be the default!
> Maybe call this 'Identity?

Renamed, but I think a noop being the default value of a tuple is legitimate.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+# For example, Wake's implementation of JSON is more lenient than the base
+# standards, so if maximum compatibility is desired in situations where some
+# other implementation may be parsing generated output, `normalizeJSONCompat`
+# may be used to target that lowest common denominator.
+#
+# Note that the relevant normalization is applied to each member of a `JArray`
+# or `JObject` *before* the enclosing container. This means that if some
+# normalization fails anywhere in the JSON tree the entire call will fail, even
+# if the specific branch causing the failure would later have been pruned.
+export def normalizeJSONWith (fmt: JSONNormalize) (json: JValue): Result JValue Error =
```
ag-eitilt, line 183: > Perhaps call this `normalizeJSONWith`, and then we can have a `normalizeJSON` which defaults to the json4 variant.

`With`, sure, and we can *maybe* add a plain alias later, but I think this is probably something that every project using wake would have different opinions on depending on their toolchain: some might want the most compatibility, some might only need canonical strings, some might require objects conform a stricter schema...

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+        (Pair key value), ps =
+            (value, map getPairSecond ps)
+            | rfoldr (simplifyValues key) JNull
+            | rmap (Pair key)
+    def simplifyValues key: JValue => JValue => Result JValue Error = match _ _
+        val JNull =
+            Pass val
+        JNull acc =
+            Pass acc
+        (JString val) (JString acc) if val ==* acc =
```
ag-eitilt, line 294: > This is tricky. You've picked a normalization here that might differ from `getJSONNormalizeString`. Indeed, you are using `==~` which is normalized with `unicodeIdentifier`, which does not match the policy of `unicodeCanonical` applied in `normalizeJSONBasic`. Perhaps `simplifyKeyGroups` should take a `String => String` normalization method argument.

I've rewritten `normalizeJSON` to recurse depth-first, this to use `==*`, and key comparison to use `<*`; that should bypass any issues where the comparison functions in different parts of the code mismatch.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+        (JDouble val) (JDouble acc) if val ==. acc =
+            Pass (JDouble acc)
+        (JBoolean val) (JBoolean acc) if enor val acc =
+            Pass (JBoolean acc)
+        (JArray val) (JArray acc) =
+            Pass (JArray (val ++ acc))
+        (JObject val) (JObject acc) =
+            val ++ acc
+            | deduplicateJObjectKeys
+            | rmap JObject
```
ag-eitilt, line 307: > The recursive call here is inappropriate. You should leave that to `normalizeJSON`, like the `JArray` case above does. ie: `Pass (JObject (val ++ acc))`

The restrictions are different.  The `JArray` case just takes the elements of one, sticks them on the elements of the other, and calls it done.  For `JObject`, both `val` and `acc` are guaranteed to not have any duplicate keys internally by this point, but the same key might occur in both and thus still needs to be filtered.  The alternate implementation would be to have `normalizeJSON` recurse *after* processing each level rather than before, but then you lose the ability to have it perform the heavy lifting for the above equality tests.

share/wake/lib/core/json.wake
terpstra: I'm not sure we want this property (`x ++ x == x`).

share/wake/lib/core/json.wake
terpstra: So, I was misled into believing JSON5 was an actual standard. I agree we should default to the 'real' JSON and normalize to that by default. Don't bother with a JSON5 version.

share/wake/lib/core/json.wake
terpstra: Sure. I was hoping to have a default already included in `normalizeJSON`, so there was no need to name one of the tuples as `Default`, since the default is whatever `normalizeJSON` passed to `normalizeJSONWith`. I think the right default is what you've called `Compat`.

share/wake/lib/core/json.wake
terpstra: Well, I totally agree we need to have an escape hatch to give people alternative normalizations. However, since we've decided JSON5 is non-standard, and IMO your 'compat' version makes quite sensible and non-controversial choices, I think it makes sense to stream-line the common case. People probably want to just do `normalizeJSON myJSON` and not worry about the tuple.

share/wake/lib/core/json.wake
terpstra: Yes. I want `normalizeJSONWith` to be the one who does the recursion. It should work that way to ensure only a single pass processes any given JSON node. Furthermore, `normalizeJSONWith` has to be the one doing the recursion, because you want to apply the appropriate leaf normalizations in the descent.

share/wake/lib/core/json.wake
ag-eitilt: With `unreachable` actually being a compiler hint, I agree that makes more sense here.  I'll change it over.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+            require Pass children =
+                findFail normalized
+            children
+            | getJSONNormalizeObject fmt
+            | rmap JObject
+            | addErrorContext path
+        JArray a =
+            def normalized =
+                zip (seq a.len) a
+                | map (\(Pair i x) normalizeJSON "{path}/[{format i}]" x)
```
terpstra, line 220: A thought: constructing the path is a lot of work, and is discarded on normal invocation. One could also have a `addErrorContextFn` that prepends the path on the way out of the call stack on failure. That would also mean you don't need to pass a path into the helper function at all.

share/wake/lib/core/json.wake
ag-eitilt: It's definitely not something we'd want to pursue, but just due to how everything else works, everything but arrays coincidentally happens to have that property.  It's basically weighing which rule we want the exception on (i.e. "arrays append *except when*" vs "it's idempotent *except for*"), and I do think you're right that the latter's the better way to go -- especially since if we never point it out in the docs, there's probably not going to be as many people noticing that one on their own.

share/wake/lib/core/json.wake
```diff
@@ -169,3 +169,143 @@ export def x ==/ y = match x y
...
+            getJSONNormalizeBoolean fmt a
+            | rmap JBoolean
+            | addErrorContext path
+        JNull =
+            Pass JNull
+        JObject a =
+            def normalizeJObjectValue entry =
+                def Pair key value =
+                    entry
+                normalizeJSON "{path}/{key}" value
```
terpstra, line 207: So, what I was proposing is that you first combine the keys shallowly, using 'JSONNormalizeObject'. Then, you recurse on the combine list.

share/wake/lib/core/json.wake
terpstra: Yeah. Also for consistency, I think since we "append" objects, we should "append" arrays.

share/wake/lib/core/json.wake
terpstra: I'm pretty concerned that your current algorithm takes `O(n^2)` time a JSON with nesting `n` levels deep.

ag-eitilt: That is definitely bad, and I will then need to rework the `deduplicateJObjectKeys`.  Applying the `normalizeJSONWith` recursion *after* joining the keys might indeed wind up being the only way to avoid it, but the reason I'm balking at that is because doing so means that if a user writes a so-called "normalize" function to make every int 0, I think they'd expect that every int in an object should be able to be collapsed rather than an error being thrown on what will become identical.  I'm pretty sure I can make `deduplicateJObjectKeys` recurse only to `JObjects` that are currently being joined due to duplication; that would still be `O(n^2)` at worst, but the most common case (without duplicate keys) would be `O(n)` or thereabouts, and even a duplicate key every other level wouldn't be too much more complex.  Would that be acceptable?

share/wake/lib/core/json.wake
ag-eitilt: `normalizeJSONWith` definitely has to recurse for that exact reason, the question is just whether it processes top-down or bottom-up.  My argument for doing children before the current node (which does admittedly mean we can't always do just a single pass) is below.

share/wake/lib/core/json.wake
ag-eitilt: The more realistic use case would be string canonicalization: single-pass recursion purely through `normalizeJSONWith` -- i.e. children after the current node -- couldn't catch `{"eacute":"é", "eacute":"é"}` and would fail to deduplicate those.  Normalizing children first would collapse the second (with a combining acute) into a string identical to the first (precomposed), and when it then looks at the enclosing object, would see two duplicate keys with identical values and happily return `{"eacute":"é"}`.

ag-eitilt:
I wasn't really expecting it to be, but the error is *not* in the deliberately-bad JSON files I created for testing.

PR #773, opened 2021-10-21
Add non-unicode inline pair constructor

colinschmidt:
This is a nice to have since I don't have a `→` key on my keyboard.
I'm not sure if there is some other place to put functions that are just aliases, or if my syntax is correct here.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -492,29 +492,29 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
         def pmkdir m p = prim "mkdir"
         def pwrite m p d = prim "write"
         def json = JObject (
-          "label"       → JString label,
-          "command"     → command     | map JString | JArray,
-          "environment" → environment | map JString | JArray,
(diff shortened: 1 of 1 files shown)


terpstra:
So, the problem with proposal is that the precedence of `-` is far too high:

Try this code:
```
def foo = (
  "hello" -> 23 | str,
  "bye" -> "asd",
  Nil
)
```

terpstra:
I suppose you could try `$->`, but that's right-associative vs `→` which is left-associative.

terpstra:
I thought `:->` would be legal, but it seems it is not, currently. Normal type ascription is also right-associative, though, so even if we made it legal to define operators starting with `:`, we'd probably want `:->` to be right-associative too, for consistency.

mmjconolly:
How far down the table of operators does it need to be, Below `OrOp` in https://github.com/sifive/wake/blob/master/share/doc/wake/quickref.md ?

Ill throw one out there: `$~~>`

terpstra:
So, that table is out-of-date since the new parser. There are substantially less precedence levels now. I should update it.

FYI, only the first character of the operator matters. `$->` is the same as `$$~^>`.

PR #792, opened 2021-10-29
json: add getters for the common pattern of 'getJ* | getOrElse'

ag-eitilt:
It has been noted that `getJ*` functions for accessing `JValue`s are very often followed by `getOrElse` calls to provide a default value if the JSON is of the wrong type (@jackkoenig, was that you?).  This PR adds helpers for those common situations.  It doesn't save writing much code and winds up adding new places which will need maintenance going forward, so I'm not actually sure adding these is desirable, but it was easy enough to code and I figured this would be a good way to generate discussion.

### DIFF
--- a/share/wake/lib/core/json.wake
+++ b/share/wake/lib/core/json.wake
@@ -25,31 +25,90 @@ export data JValue =
   JObject  (List (Pair String JValue))
   JArray   (List JValue)
 
-export def getJString: JValue => Option String = match _
-  JString x               = Some x
-  JArray (JString x, Nil) = Some x
-  _                       = None
-export def getJInteger: JValue => Option Integer = match _
-  JInteger x               = Some x
-  JArray (JInteger x, Nil) = Some x
-  _                        = None
-export def getJDouble: JValue => Option Double = match _
-  JDouble x               = Some x
-  JInteger x              = dint x
-  JArray (JDouble x, Nil) = Some x
-  JArray (JInteger x, Nil)= dint x
-  _                       = None
-export def getJBoolean: JValue => Option Boolean = match _
-  JBoolean x               = Some x
-  JArray (JBoolean x, Nil) = Some x
-  _                        = None
-export def getJObject: JValue => Option (List (Pair String JValue)) = match _
-  JObject x               = Some x
-  JArray (JObject x, Nil) = Some x
-  _                       = None
-export def getJArray: JValue => Option (List JValue) = match _
-  JArray x = Some x
-  _        = None
+# Attempt to retrieve the value of a textual JSON structure.
+# This succeeds if passed a `JString` or a `JArray` containing only a single
+# `JString`, and returns `None` otherwise.
+export def getJString: (json: JValue) => Option String = match _
+    JString x               = Some x
+    JArray (JString x, Nil) = Some x
+    _                       = None
+# Retrieve the value of a JSON text, or a fallback if the type is incompatible.
+# This returns the represented value if passed a `JString` or a `JArray`
+# containing only a single `JString`, and returns `default` otherwise.
+export def getJStringOrElse (default: String) (json: JValue): String =
+    getJString json
+    | getOrElse default
+
+# Attempt to retrieve the value of an integer JSON structure.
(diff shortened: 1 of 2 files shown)


jackkoenig:
> @jackkoenig, was that you?

It was indeed, that change dates back to when wake had exceptions which were unwieldy and later removed.

I'm not sure how I feel about these extra utilities since it's fairly standard practice in ML-style languages to return `Option` for functions that may not return a value (eg. `head` on a `List`, indexing any sequence) so having these special cases would be inconsistent with every other `Option` returning function. I don't have a strong option though, optimizing for common cases can be useful.

terpstra:
So I know I'm in the minority, here, but I generally prefer to see two independent methods used unless there is a substantial savings in code complexity. My rationale is that it means people need to remember less methods in their heads. Maybe IDE integration changes this equation somewhat.

terpstra:
I *do* think we should refactor the JSON query methods to support a different scheme than they currently, use, however.

In particular, I do not like how `//` and friends use a `JArray`. I think we should follow `jq`-style and have the query language operate on explicit `List JObject`s. The final extractor methods like `getJString` would then no longer accept a `JArray` that contains a `JString`, but rather perform a mapping `List JObject => List String`. (Or potentially, `JObject => Option String` and we expect you to compose it with `mapPartial`)

ag-eitilt:
Sounds like you're actually in the majority for this group.  I'll close this in the near future unless anyone else wants to argue on their behalf.

I'll put my vote in for `mapPartial`, but will need to dig into JQuery a bit more before being able to weigh in on `//`.  On the surface, I do like the sound of `List JValue`.

ag-eitilt:
@ngraybeal, your comment in federation about your discussion with @davidbiancolin and @colinschmidt was what prompted this PR.  Do any of you want to weigh in on the `getJ* _ | getOrElse` vs. `getJ*OrElse` pattern?  Personally, I do agree that there's a few common helpers that could be pulled, but I'm not sure these ones to save a single line are necessary.

colinschmidt:
My code is littered with two versions of this pattern:
```
def valueWithADefault =
    jvalue // `key`
    | getJString
    | getOrElse "default"
```
and
```
require Pass requiredValue =
    jvalue // `key`
    | getJString
    | getOrFail "This value is required".makeError
```
If I could do these in a single line I would be pleased.
I have 28 instances of the above patterns in a 500 line wake file(~20% of the file is this pattern), and only a single use of `getJString` in that file that isn't in on of these patterns.

davidbiancolin:
Ditto to colin's comment. I wrote my [own helpers](https://github.com/sifive/federation/pull/17528/files#diff-f62d649594cf604f040e5ef8cfa2cf5c1cd4677a3832d1caae58fff0d2555235R18) to do this one line and i figure others will too.

ag-eitilt:
Those particular helpers do make sense to me, as they additionally wrap the JSON lookup and can provide a standardized error message.  I also agree with the fact that virtually every `getJ*` function is followed by *one of* a `getOrElse`, `getOrFail`, `mapPartial` or `isSome`.  What I'm not convinced about is making an extra set of functions for the sole purpose of being able to drop a single line for a single one of those processors; @davidbiancolin's helper would still be required even if this is merged, and replacing @colinschmidt's 28 instances would only trim the file by about 5%.

terpstra:
> My code is littered with two versions of this pattern:
> 
> ```
> def valueWithADefault =
>     jvalue // `key`
>     | getJString
>     | getOrElse "default"
> ```

I agree that a helper method may be called for, but why stop at combining just the getJString and getOrElse? I'd rather have a helper that combines all three steps and/or processes an entire tuple at a time. I'm also not convinced that the above code pattern is even correct, since it will silently take the default value if someone has `"key":4` since `4` is not a string.

colinschmidt:
Yeah I was assume we would create a helper method that allows me to do:
`def stringValueWithADefault = getJStringOrDefault jvalue `key` "default"`
And you are right @terpstra this code doesn't include any validation of the type which would have bloated the code even further.

ag-eitilt:
Sounds like there might be room to look at what helper functions are desirable, but I closed this as it's not these.

PR #806, opened 2021-11-12
bind: fix publish type error locations

### LINKED ISSUE #805: Type errors with 'topic' are reported poorly
When an object is `publish`ed to a topic with the wrong type, the type mismatch error points to the topic definition rather than the `publish` call; this is a particular issue when you want to change the type of the topic itself; you can get many errors all pointing to the same line and have to resort to a project-wide search to find everywhere that needs fixing ("Go to References" also doesn't work for topics, though I imagine that's due to them being a bit type-like).  A smaller issue is that the underlying `List` representation is printed in the error message, so you have "does not match explicit type ascription of: `List a`" when you've only written `topic accum: a`.

terpstra:
A `publish` with the wrong type should put the error message at the location
of the publish, not the topic.

Fixes #805.

### DIFF
--- a/src/dst/bind.cpp
+++ b/src/dst/bind.cpp
@@ -1045,23 +1045,25 @@ static std::unique_ptr<Expr> fracture(std::unique_ptr<Top> top) {
           def.expr = fracture(*top, false, trim(def.name), std::move(def.expr), &gbinding);
           ++gbinding.current_index;
 
+          FileFragment topic = def.expr->fragment;
+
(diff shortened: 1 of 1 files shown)


ag-eitilt (approved):
Seems to work well!  The "explicit type ascription" language is still slightly weird, but it makes a lot more sense to have the `List` visible on the `publish` than it did on the definition, so I'm perfectly happy with not needing to rewrite the entire error message for this specific situation.

PR #808, opened 2021-11-12
core: use 'Fail' for bad paths in standard functions

ag-eitilt:
This should provide safe `Path` use throughout the stdlib.  I've not yet removed `makeBadPath` or the `BadPath` constructor since I'm not quite sure what the best way to do so would be while (easily) providing older-versioned libraries, but we can revisit that later; I know you were thinking about adding a hash to the `Path` constructor as well, so we can probably fold both the type reworks into a single refactor.

As far as I can tell, the `v0_22_wake` implementations should function just like they used to, but I missed things the last time I thought that, so let me know.  The one thing I'm not quite happy with is that `runIncrementalJob` doesn't provide any feedback if its `_postTar` hook fails -- it was ignored previously as well, but now we have a way of checking for that -- but I couldn't figure out any way to backport it without copying a *lot* of code.

The one thing I haven't done that was talked about towards the beginning of this was making `runJob` return a `Result Job Error` if the invocation fails.  Is that something we want to add now, or should that be part of the overall `Job` rework?  (If now, `runIncrementalJob` should probably be figured out to match.)


share/wake/lib/core/json.wake
```diff
@@ -58,12 +58,11 @@ export def parseJSONBody (body: String): Result JValue Error =
...
-  match (getPathResult path)
-    Pass f = match (imp f.getPathName)
-      Pass body = Pass body
-      Fail f = Fail (makeError f)
-    Fail f = Fail f
+    def imp f =
+        prim "json_file"
+    match (imp path.getPathName)
+        Pass body = Pass body
+        Fail f = Fail (makeError f)
```
terpstra, line 65: I think the standard library still needs to defensively use `getPathResult`.

share/wake/lib/system/path.wake
```diff
@@ -18,8 +18,7 @@ package wake
 # Read the file contents of a Path
 export def read (path: Path): Result String Error =
     def imp p = prim "read"
-    require Pass f = getPathResult path
-    match (imp f.getPathName)
+    match (imp path.getPathName)
```
terpstra, line 21: Again, I think this one needs to be defensive until BadPath is really gone.

share/wake/lib/versions/v0_19.wake
```diff
@@ -1,15 +1,15 @@
...
...setRunnerOutputOutputs setRunnerOutputUsage setRunnerScore setTripleFirst setTripleSecond setTripleThird setUsageCPUtime setUsageInBytes setUsageMemBytes setUsageOutBytes setUsageRuntime setUsageStatus shellJob simplify sortBy sortPaths sortStrings source sources splitAt splitBy splitUntil sqrt stack str strlen sum sysname tab tail take takeUntil tap tappi tat tcontains tdelete tdistinctBy tdistinctRunBy tdrop tdropUntil tempty tequal texists tfilter tfind tfold tfoldl tfoldmap tfoldr tforall tinsert tinsertMulti tinsertReplace tintersect tlen tlowerGE tlowerGT tmax tmin tnew tokenize transpose treeToList treeToVector tsplit tsplitAt tsplitBy tsplitUntil tsubset tsubtract ttake ttakeUntil tunion tunionMulti tupperLE tupperLT unicodeCanonical unicodeIdentifier unicodeLowercase unicodeToInteger unsetEnvironment unzip vat vcmp vdistinctBy vdistinctRunBy vdrop vdropUntil vectorToList vectorToTree vectorToTreeMulti vempty version vexists vfilter vfind vflatten vfoldl vfoldr vforall virtualRunner vlen vmap vreverse vscan vscanl vscanr vseq vsortBy vsplitAt vsplitBy vsplitUntil vtab vtake vtakeUntil vtranspose vunfoldl vunzip vzip wOK wait which whichIn workspace write writeIn xOK xor zip
```
terpstra, line 5: Smart.

share/wake/lib/versions/v0_22.wake
```diff
@@ -1,7 +1,108 @@
...
 from wake export def unary ! + +. - -. ~ ∏ ∏. ∑ ∑. √ ∛ ∜
 from wake export def binary != !=* !=. !=^ !=~ $ % && * *. + ++ +. , - -. . / /. /../ // /| ; < <* <. << <= <=* <=. <=> <=>* <=>. <=>^ <=>~ <=^ <=~ <^ <~ == ==* ==. ==/ ==^ ==~ > >* >. >= >=* >=. >=^ >=~ >> >^ >~ ^ ^. | || → ∈ ∉ ∋ ∌ ∘ ∩ ∪ ⊆ ⊇ ⊈ ⊉ ⊊ ⊋ ⊎
 from wake export topic environment path runner source
+
+from wake export def getJobResultOutput=getJobOutput getJobResultOutputs=getJobOutputs
```
terpstra, line 9: 👍

src/src.wake
```diff
@@ -35,19 +35,28 @@ def src (dir: String) (variant: Pair String String) (internal: List (Pair String
...
+    require Pass versionHeader =
+        version_h Unit
+    require Pass headerPaths =
+        def searchDirs =
+            srcDir,
+            vendorDir,
+            Nil
+        searchDirs
+        | findFailFn (sources _ `.*\.h`)
+        | rmap flatten
```
terpstra, line 54: You used this pattern in a few places. It's correct, but baroque. I wonder if there is a better pithy and straight-forward way.

terpstra: Probably just two `require`s for direct calls to `sources` would be more readable.

terpstra (changes requested):
This PR needs to define a new v0_24_wake package.

terpstra:
Our main repo doesn't typecheck with this PR. Probably we need some `from v0_22_wake import _`s. I'd like to see that work before we merge.

terpstra:
> The one thing I'm not quite happy with is that `runIncrementalJob` doesn't provide any feedback if its `_postTar` hook fails
> The one thing I haven't done that was talked about towards the beginning of this was making `runJob` return a `Result Job Error` if the invocation fails.

I think leaving both of these for later is the right thing to do.

ag-eitilt:
> This PR needs to define a new v0_24_wake package.

I wasn't sure if any of your work would need to be included, but I can definitely add it.

> Our main repo doesn't typecheck with this PR. Probably we need some `from v0_22_wake import _`s. I'd like to see that work before we merge.

I knew I was forgetting something!  PR is in CI now.

share/wake/lib/core/json.wake
ag-eitilt: Makes sense.  I'll restore those.

share/wake/lib/system/path.wake
ag-eitilt: It might even be better to keep it defensive even after removing `BadPath` -- a good `Path` only proves that the `Path` existed when it was created, but things could theoretically be deleted between creation and use if the user/system is doing bad things.  I don't know how much of a performance hit it would be (and almost *any* hit would probably be too much), but once `getPathResult` isn't needed as a simple constructor test, would we want to make it a "is this file *still* good" check?

src/src.wake
ag-eitilt: Yeah, probably.  When there's half a dozen `sources` calls I start looking for ways of removing the boilerplate, but it's definitely baroque.

share/wake/lib/versions/v0_22.wake
```diff
@@ -1,7 +1,108 @@
...
+
+
+# Every other function returning `Result` post-0.24 needs some degree of reimplementation
+
+def treeCompat job typ =
+    def tree j t =
+        prim "job_tree"
+    def treeOk (Pair f h) = match h
+        "BadHash" = makeBadPath (makeError "Could not hash {f}")
+        _ = claimFileAsPath f "build/v0_22_wake/{f}"
```
ag-eitilt, line 78: I found something else that I wanted to ask about: I don't like this `claimFileAsPath`, but is there any better way of doing it?  Since we obviously don't want to export the `Path` constructor, I can't just wrap the valid path names as in job.wake.  `source` is also out since it's likely returning files that were created by the job rather than ones in Git, and I can't just base it off the `Result (List Path) Error` that the stdlib now returns as the `Fail` case for that throws out all the good paths.

share/wake/lib/system/path.wake
terpstra: There is nothing the build can do to protect itself from users who modify the workspace while a build is running. This is why I don't want to use developer builds as a shared cache publisher. ;-) As long as no person is manipulating the workspace during the build, though, you should be fine; wake-fuse protects all extant files during Job execution.

terpstra: So, once BadPath is removed from the `Path` type, I would indeed remove this along the lines of your original PR.

share/wake/lib/versions/v0_22.wake
terpstra: Oh jeeze. Yeah, I overlooked that.

terpstra: So, yeah, we also don't really want to export makeBadPath in v0_24_wake, either.

terpstra: Here's my proposal: move the `data Path` fully exported into a new package `internals_wake`. Then, re-export only the type from `wake`. Then, you can define `makeBadPath` in `v0_22_wake` and also construct a `Path` directly there.

ag-eitilt: That was what I came up with as well.  It seemed ugly, but if you don't have anything better, I guess it's what we'll need to be doing.

terpstra: The other option is to use a different `Path` type in `v0_22` and `v0_24`, and convert between them as `Path@v0_22_wake => Result Path@v0_24_wake Error`. However, I think that will be more painful to support upgrading extant code bases.

ag-eitilt: If they were mostly staying within file boundaries, I can see that *maybe* working, but `Path`s get passed around way too often for me to think that would work.

terpstra: Looks good.

terpstra (approved):
All my concerns have been addressed. Thanks!

ag-eitilt:
Glad to see this get merged, thank *you*!

PR #809, opened 2021-11-12
Subhash as useful warning

### LINKED ISSUE #385: Target subkey mismatch
When looking at an error like:
```
ERROR: Target subkey mismatch for ./some/path/foo.wake:7:[1-100]
To debug, rerun your wake command with these additional options:
  --debug-target=1 to see the unique target arguments (before the '\')
  --debug-target=2 to see the first invocation's extra arguments
  --debug-target=3 to see the second invocation's extra arguments
```
I don't feel there is enough information to understand what is going on without running the `--debug-target` bits

`Target subkey mismatch` describes the problem in terms of the wake internals which the user wont know (what is a subkey? why would it mismatch?). Please describe the issue in vocabulary that the user would have

Additionally, there is probably useful information in the `--debug-target` ids that is worth printing out here without asking the user to copy/paste to discover.

### LINKED ISSUE #570: Difficult to use --debug-target when indirectly invoking wake
A user doesn't always know the command used to invoke wake

If you see a message like
```
ERROR: Target subkey mismatch for share/wake/lib/system/job.wake:259:[8-99]
To debug, rerun your wake command with these additional options:
  --debug-target=4983443480946889521 to see the unique target arguments (before the '\')
  --debug-target=1907088811814402159 to see the first invocation's extra arguments
  --debug-target=17973659819140003278 to see the second invocation's extra arguments
```
when you have invoke wake indirectly (via, for example a Makefile) then you might not be able to easily find out what the initial invocation (via `-x` or whatever) was.

At this point you probably try (in vain) something like this:
```
$ wake --debug-target=4983443480946889521
Available wake targets:
  build
  install
  static
```

terpstra:
Rework how target subkeys are treated in wake. By definition, these keys are not supposed to matter. So, if they are used inconsistently, this should not trigger an error, just a warning. Furthermore, the old error messages were exceedingly unhelpful.

Consider this bad program:
```
export def broken _ =
    require Pass job1 =
        "ls"
        | makePlan "job1" Nil
        | setPlanStdout logError
        | runJob
        | getJobStdout
    require Pass job2 =
        "ls"
        | makePlan "job2" Nil
        | setPlanStdout logWarning
        | runJob
        | getJobStdout
    Pass "{job1}{job2}"
```

The old ERROR message:
```
ERROR: Target subkey mismatch for share/wake/lib/system/job.wake:273:[8-32]
To debug, rerun your wake command with these additional options:
  --debug-target=945013996097089732 to see the unique target arguments (before the '\')
  --debug-target=18216666221716652470 to see the first invocation's extra arguments
  --debug-target=3239709073642377762 to see the second invocation's extra arguments
```

The new WARNING message:
```
Target 'runOnce' <share/wake/lib/system/job.wake:273:[8-32]> was called with inconsistent auxiliary arguments
  It was passed these identical primary key arguments:
    cmd = "/opt/local/bin/dash", "-c", "ls", Nil
    env = "PATH=/opt/local/bin:/usr/local/bin:/usr/bin:/bin", Nil
    dir = "."
    stdin = ""
  It was passed these auxiliary arguments (after the '\') which should also be identical:
    res = Nil
    usage = Usage 0 0.00000000000000000e+00 1e0 0 0 0
    finputs = <share/wake/lib/system/job.wake:472:43>
    foutputs = <share/wake/lib/system/job.wake:472:43>
    vis = Nil
    keep = True
    run = <share/wake/lib/system/job.wake:64:[7-23]>
    echo = "echo"
    (INCONSISTENT) stdout = "warning"
    stderr = "warning"
    (INCONSISTENT) label = "job2"
Pass ""
```

Fixes #385
Fixes #570

### DIFF
--- a/src/dst/todst.cpp
+++ b/src/dst/todst.cpp
@@ -859,12 +859,13 @@ static void dst_def(CSTElement def, DefMap &map, Package *package, Symbols *glob
       FileFragment l = FRAGMENT_CPP_LINE;
 
       Expr *table = new Prim(l, "tnew");
-      for (size_t i = 0; i <= args.size(); ++i)
+      for (size_t i = 0; i < args.size()+2; ++i)
         table = new Lambda(l, "_", table, " ");
 
       std::stringstream s;
       s << defs.front().body->fragment.location();
       table = new App(l, table, new Literal(l, s.str(), &Data::typeString));
+      table = new App(l, table, new Literal(l, std::to_string(tohash), &Data::typeInteger));
 
       for (size_t i = 0; i < args.size(); ++i)
         table = new App(l, table, new Literal(l, std::string(args[i].first), &Data::typeString));
--- a/src/dst/todst.cpp
+++ b/src/dst/todst.cpp
@@ -830,17 +830,20 @@ static void dst_def(CSTElement def, DefMap &map, Package *package, Symbols *glob
     if (target) {
       if (tohash == 0) ERROR(fn.location(), "target definition of '" << name << "' must have at least one hashed argument");
       FileFragment bl = body->fragment;
-      Expr *hash = new Prim(bl, "hash");
-      for (size_t i = 0; i < tohash; ++i) hash = new Lambda(bl, "_", hash, " ");
-      for (size_t i = 0; i < tohash; ++i) hash = new App(bl, hash, new VarRef(bl, args[i].first));
-      Expr *subhash = new Prim(bl, "hash");
-      for (size_t i = tohash; i < args.size(); ++i) subhash = new Lambda(bl, "_", subhash, " ");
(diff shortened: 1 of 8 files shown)


terpstra:
An argument could be made that the sandbox (replacing resources) and visible file list should be in the primary key, but at present they are not. We can consider changing that if it ever comes up.

terpstra:
> An argument could be made that the sandbox (replacing resources) and visible file list should be in the primary key, but at present they are not. We can consider changing that if it ever comes up.

Actually, I remember the rationale behind the current choice: the things in the key are those things likely to lead to distinct output filenames. From that perspective, there is maybe too much in the key (env/stdin).

PR #815, opened 2021-11-17
vscode: add configuration so C++ code works in vscode

terpstra:
Turns out that vscode is pretty handy also for C++.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,20 @@
+{
+    "configurations": [
+        {
+            "name": "Mac",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "/opt/local/include"
+            ],
+            "defines": [],
+            "macFrameworkPath": [
+                "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks"
+            ],
+            "compilerPath": "/usr/bin/clang",
+            "cStandard": "c11",
+            "cppStandard": "c++11",
+            "intelliSenseMode": "macos-clang-x64"
+        }
+    ],
+    "version": 4
+}
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -14,6 +14,17 @@
             "cStandard": "c11",
             "cppStandard": "c++11",
             "intelliSenseMode": "macos-clang-x64"
+        },
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/gcc",
+            "cStandard": "c11",
+            "cppStandard": "c++11",
+            "intelliSenseMode": "linux-gcc-x64"
         }
     ],
     "version": 4
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,9 @@
 manifest.wake
 *.o
-/.vscode
 /.build
 /.fuse
 /.fuse.log
+/.vscode/*.log
 /**/wake.db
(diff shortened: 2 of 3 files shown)


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,20 @@
...
+                "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks"
+            ],
+            "compilerPath": "/usr/bin/clang",
+            "cStandard": "c11",
+            "cppStandard": "c++11",
+            "intelliSenseMode": "macos-clang-x64"
+        }
+    ],
+    "version": 4
+}
```
mwachs5, line 20: do you want newlines here

terpstra: It's machine generated, so if I change anything, it will just change it back.

richardxia:
Any chance that you'd do this in a way that supports multiple different editors instead of doing this just for VSCode? [`clangd`](https://clangd.llvm.org/installation) is a popular language server (which I use with Vim), and it primarily relies on having a `compile_commands.json` file output by the build tool in order to understand the compiler flags and paths. Once that file is created, `clangd` "just works".

terpstra:
> Any chance that you'd do this in a way that supports multiple different editors instead of doing this just for VSCode?

I have no idea how that tool works, but if you want to add an appropriate configuration file to this PR, go for it. I doubt this file would have helped the vscode intellisense plugin, though, so I still need the files in this PR.

richardxia:
> I have no idea how that tool works, but if you want to add an appropriate configuration file to this PR, go for it.

It's less a configuration file and more a build artifact, since it relies on running the actual build and recording which compiler flags (e.g. include flags) are passed to the compiler. I actually tried doing this some time ago, but I was having a lot of difficulty with my weirdo setup of Nix on macOS and with the fact that the Makefile builds another copy of wake using wake itself.

> I doubt this file would have helped the vscode intellisense plugin, though, so I still need the files in this PR.

I don't know what "the" VSCode Intellisense plugin is, but there's a VSCode plugin for clangd: https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd

terpstra:
> It's less a configuration file and more a build artifact, since it relies on running the actual build and recording which compiler flags (e.g. include flags) are passed to the compiler.

I mean, on linux you should not need any compile flags. Just tell your editor that wake uses C++11 / C11 code and it should be fine. You may need to run 'make' one time to create some of the auto-generated header files, though.

> > I doubt this file would have helped the vscode intellisense plugin, though, so I still need the files in this PR.
> 
> I don't know what "the" VSCode Intellisense plugin is

Screen Shot 2021-11-17 at 11 32 14 AM

Seem to be `ms-vscode.cpptools`... which is what you get installed by default when you open your first C++ file and click the "please install the correct plugins" button.

richardxia:
> I mean, on linux you should not need any compile flags. Just tell your editor that wake uses C++11 / C11 code and it should be fine. You may need to run 'make' one time to create some of the auto-generated header files, though.

I haven't tried on Linux.

The issue I'm having is that I can't get symbol definitions to work across different directories (e.g. hover and jump-to-definition don't work), so I think these local compiler flags need to be made known to the language server: https://github.com/sifive/wake/blob/e91928a9edda481651f31d46b370a97fe86c3c40/Makefile#L11

terpstra:
Oh. Interesting. I think it's the `"${workspaceFolder}/**` that makes that "just work" on vscode.

terpstra:
Anyhow, those settings `-Isrc -Ivendor` and the c++ standards version don't change, so I think you can just put a pre-generated JSON file into the repository.

PR #829, opened 2021-11-20
Array: remove from wake public API

terpstra:
This type should never have been exposed.

It is an implementation detail of the actual user-facing `Vector` type.


share/wake/lib/versions/v0_24.wake
```diff
@@ -1,5 +1,5 @@
 package v0_24_wake
-from wake export type Array Boolean Colour Double DoubleClass DoubleFormat Error Integer Intensity JSONFormat JSONNormalize JSONRunnerPlan JValue Job List LogLevel Option Order Pair Path Persistence Plan RegExp Result Runner RunnerInput RunnerOutput Status String Tree Triple Unit Usage Vector
+from wake export type Boolean Colour Double DoubleClass DoubleFormat Error Integer Intensity JSONFormat JSONNormalize JSONRunnerPlan JValue Job List LogLevel Option Order Pair Path Persistence Plan RegExp Result Runner RunnerInput RunnerOutput Status String Tree Triple Unit Usage Vector
```
ag-eitilt, line 2: I definitely agree with removing it from the live `wake` API and future versions.  Removing it from the frozen compatibility packages seems a touch (morally) wrong, but since it *never* appears in any public type signature, doing so at least shouldn't break anything.

ag-eitilt (approved):

PR #832, opened 2021-12-04
system: reduce fragmentation of related functions between files

ag-eitilt:
The .wake files in the `system` directory have somewhat bled into one another (e.g. a number of `Path` evaluation functions were defined in `job.wake`).  This PR doesn't add or remove any actual code, just rearranges it so that closely-related definitions are actually close to what they're related to.  The primary changes are:

* `claimFileAsPath` has been moved alongside the other file->`Path` loading functions in `sources.wake`
* `getPathResult`, `getPathName`, `sortPaths`, and similar are now actually in `path.wake`, which additionally now exports the type.
* `mkdir`, `read`, and `write` join `installAs` as a category of functions which interact with the filesystem; the file `cp.wake` has been renamed `io.wake` to reflect the broader scope, even if the name is a bit more inclusive than even that.


share/wake/lib/system/environment.wake
```diff
@@ -53,6 +53,12 @@ publish path = match (getenv "WAKE_PATH")
 publish path = "/usr/bin", Nil
 publish path = "/bin", Nil
 
+# On MacOS and FreeBSD, many important system binaries are not in /usr/bin
+publish path = match sysname
+    "Darwin" = "/opt/local/bin", "/usr/local/bin", Nil
+    "FreeBSD" = "/usr/local/bin", Nil
```
mmjconolly, line 59: What motivates this change?

terpstra: These were in os.wake previously.

terpstra (approved):
Reorganizing system/ is definitely a good idea.

share/wake/lib/system/environment.wake
ag-eitilt: Admittedly not a lot.  That block was the *only* code in `os.wake`, and the blocks above it are the only other ones that publish to `path`, so it's just a bit of "bring everything related to the `topic` into one place".  If we have more OS-dependent code, keeping the separate file would probably make a bit more sense, but as is this seemed like it had a chance of getting lost.

Issue #821, opened 2021-11-19
LSP: documentation not bound to inner functions


terpstra:
The LSP currently only associates comments with top-level definitions.

For example,
```
# This comment is linked to foo
def foo x =
  # But this comment is not linked to anything
  def bar = x * x
  bar + bar
```

This is not super high priority, since the definition is probably still on the same screen as the use.

PR #835, opened 2021-12-13
LSP: fix documentation hover for `tuple` and `data`

### LINKED ISSUE #820: LSP: documentation hover broken for `tuple` and `data`
The current approach to linking documentation comments to definitions works great for normal top-level `def`s.

However, when a comment precedes a `data` or `tuple` definition, several symbols are introduced. The LSP currently only associates the comments to the first such symbol.

For example,
```
# This comment only shows up for `Nil`, not for `,` (also not `List`)
export data List a =
    Nil
    (head: a), (tail: List a)
```

Another example.
```
# This comment only shows up for Pet, not `getPetName` or `getPetAge`
tuple Pet =
  Name: String
  Age: Integer
```

elena-pudovkina:
Outer comments are now also bound to symbols introduced in a `tuple` or `data` definition. They appear after the inner comment:
image

Definitions which have the same location are now merged into a single definition with the documentation containing:
- Name and type of the symbol
- Inner documentation of the symbol
- List of definitions introduced at this location
- Outer documentation

image

image

Fixes #820.

### DIFF
--- a/src/util/location.h
+++ b/src/util/location.h
@@ -55,8 +55,17 @@ struct Location {
   }
 
   bool operator < (const Location &l) const {
-    if (filename == l.filename) { return start < l.start; }
+    if (filename == l.filename) {
+      if (start == l.start) {
+        return end < l.end;
+      }
+      return start < l.start;
+    }
     return filename < l.filename;
+  }
+
+  bool operator == (const Location &l) const {
+    return this->contains(l) && l.contains(*this);
   }
 };
 
--- a/src/util/location.h
+++ b/src/util/location.h
@@ -55,17 +55,17 @@ struct Location {
   }
 
   bool operator < (const Location &l) const {
-    if (filename == l.filename) {
-      if (start == l.start) {
-        return end < l.end;
-      }
-      return start < l.start;
-    }
-    return filename < l.filename;
+    if (filename != l.filename) return filename < l.filename;
+    if (start != l.start) return start < l.start;
+    return end < l.end;
   }
 
   bool operator == (const Location &l) const {
     return this->contains(l) && l.contains(*this);
+  }
+
+  bool operator != (const Location &l) const {
+    return !(*this == l);
   }
 };
 
--- a/tools/lsp-wake/astree.cpp
+++ b/tools/lsp-wake/astree.cpp
@@ -43,6 +43,15 @@ ASTree::ASTree() {}
 
 ASTree::ASTree(std::string _absLibDir) : absLibDir(std::move(_absLibDir)) {}
 
+void ASTree::recordComments(CSTElement def) {
+  if (def.id() == TOKEN_COMMENT || def.id() == TOKEN_NL) {
+    comments.emplace_back(def.segment().str(), def.location());
+  }
+  for (CSTElement innerdef = def.firstChildElement(); !innerdef.empty();  innerdef.nextSiblingElement()) {
+    recordComments(innerdef);
+  }
+}
+
 void ASTree::diagnoseProject(const std::function<void(FileDiagnostics &)> &processFileDiagnostics) {
   usages.clear();
   definitions.clear();
@@ -77,11 +86,7 @@ void ASTree::diagnoseProject(const std::function<void(FileDiagnostics &)> &proce
 
     CST cst(*fcontent, lspReporter);
     dst_top(cst.root(), *top);
-    for (CSTElement topdef = cst.root().firstChildElement(); !topdef.empty(); topdef.nextSiblingElement()) {
-      if (topdef.id() == TOKEN_COMMENT || topdef.id() == TOKEN_NL) {
-        comments.emplace_back(topdef.segment().str(), topdef.location());
-      }
-    }
+    recordComments(cst.root());
   }
   flatten_exports(*top);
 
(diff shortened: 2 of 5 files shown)


src/util/location.h
```diff
@@ -55,9 +55,18 @@ struct Location {
...
 
   bool operator < (const Location &l) const {
-    if (filename == l.filename) { return start < l.start; }
+    if (filename == l.filename) {
+      if (start == l.start) {
+        return end < l.end;
+      }
+      return start < l.start;
+    }
     return filename < l.filename;
```
terpstra, line 64: So, to prevent nesting when comparing multiple tuple elements, people usually write this as:
```
if (a != b) return a < b;
if (c != d) return c < d;
if (e != f) return e < f;
return g < h;
```

tools/lsp-wake/symbol_definition.h
```diff
@@ -43,12 +44,20 @@ struct SymbolDefinition {
...
     std::string documentation;
+    std::string outerDocumentation;
+    std::vector<std::pair<std::string, std::string>> introduces;
 
     SymbolDefinition(std::string _name, Location _location, std::string _type, SymbolKind _symbolKind, bool _isGlobal) :
     name(std::move(_name)), location(std::move(_location)), type(std::move(_type)), symbolKind(_symbolKind),
     isGlobal(_isGlobal) {}
 
     bool operator < (const SymbolDefinition &def) const {
+      if (!(this->location < def.location) && !(def.location < this->location)) {
```
terpstra, line 55: Isn't this just `==`?

tools/lsp-wake/symbol_definition.h
```diff
@@ -43,12 +44,20 @@ struct SymbolDefinition {
...
+        if (this->name == def.name) {
+          return this->type < def.type;
+        }
+        return this->name < def.name;
+      }
```
terpstra, line 60: Also, I'd recommend the same pattern for comparing tuples from earlier.

tools/lsp-wake/astree.cpp
```diff
@@ -437,10 +482,28 @@ std::string ASTree::sanitizeComment(std::string comment) {
...
+  comment.back().first += text;
+}
+
 ASTree::SymbolUsage::SymbolUsage(Location _usage, Location _definition) : usage(std::move(_usage)), definition(std::move(_definition)) {}
 
-ASTree::Comment::Comment(std::string _comment_text, Location _location)
-: comment_text(std::move(_comment_text)), location(std::move(_location)) {}
+ASTree::Comment::Comment(std::string _comment_text, Location _location, int _level)
+: comment_text(std::move(_comment_text)), location(std::move(_location)) {
+  level = _level;
```
terpstra, line 505: You can (and probably should) just add a `level(_level)`

tools/lsp-wake/astree.cpp
```diff
@@ -437,10 +482,28 @@ std::string ASTree::sanitizeComment(std::string comment) {
   return comment;
 }
 
+std::string ASTree::composeOuterComment(std::vector<std::pair<std::string, int>> comment) {
+  std::string composed;
+  for (int i = (int) comment.size() - 2; i >= 0; i--) {
```
terpstra, line 487: So, the reason you get a warning here is that `size()` might not fit in an `int`. You're probably correct, and it will fit, so the override is ok. I generally do this sometimes, too. However, it is usually better to use a `size_t` to be safe (also the compiler likes these better when optimizing). Unfortunately, `size_t` cannot accommodate negative numbers.

I would probably leave the code how you wrote it, but wanted to point out the pitfalls of the cast.

terpstra (approved):
Looks good!

terpstra:
I saw you closed a bunch of issues and PRs by hand. It's better to write:

Closes #833.
Fixes #820.
Fixes #821.

These now get linked to this PR and closed when this PR merges.

src/util/location.h
elena-pudovkina: Done

tools/lsp-wake/astree.cpp
elena-pudovkina: Done

tools/lsp-wake/symbol_definition.h
elena-pudovkina: It is, fixed

tools/lsp-wake/symbol_definition.h
elena-pudovkina: Done

tools/lsp-wake/astree.cpp
elena-pudovkina: Yes, I understand. Left as is

elena-pudovkina:
> I saw you closed a bunch of issues and PRs by hand. It's better to write:
> 
> Closes #833. Fixes #820. Fixes #821.
> 
> These now get linked to this PR and closed when this PR merges.

I closed PR #833 because it doesn't fix issue #821. It is still not fixed. Should I write about it somewhere?

terpstra:
> > I saw you closed a bunch of issues and PRs by hand. It's better to write:
> > Closes #833. Fixes #820. Fixes #821.
> > These now get linked to this PR and closed when this PR merges.
> 
> I closed PR #833 because it doesn't fix issue #821. It is still not fixed. Should I write about it somewhere?

Not necessary.

PR #837, opened 2021-12-29
Update doc comments for list.wake

JakeSiFive:
This change performs three updates:
1) Updates all functions to use markdown
2) Updates a few functions with nicer markdown
3) Updates many (but not all) functions with extra info

Please check the following:
* Spelling
* All added examples
* All added properties
* Grammar and general clarity of added content
* For needlessly verbose or unwanted examples/conditions

### DIFF
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -23,14 +23,20 @@ package wake
 #   - the comma operator to prepend elements to a List
 #
 # Because the comma operator is right associative, a List parses like this:
+#   ```
 #   0, 1, 2, Nil = (0, (1, (2, Nil)))
+#   ```
 #
 # Since List has a type parameter, any type can be legally stored in a list
+#   ```
 #   0, 1, 2, Nil         # ok! This is a `List Integer`
 #   "0", "1", "2", Nil   # ok! This is a `List String`
+#   ```
 #
 # However, mixing types it NOT legal:
+#   ```
 #   0, "1", Nil          # bad! This will trigger a type-check failure.
+#   ```
 export data List a =
     Nil
     (head: a), (tail: List a)
@@ -38,10 +44,13 @@ export data List a =
 # empty: report if the list contains no elements.
 # If you find yourself using the function, consider using match instead.
 #
+# Examples:
+#   ```
 #   empty Nil      = True
 #   empty (seq 0)  = True
 #   empty (1, Nil) = False
 #   empty (seq 9)  = False
+#   ```
 export def empty: List a => Boolean = match _
     Nil = True
     _   = False
@@ -49,9 +58,12 @@ export def empty: List a => Boolean = match _
 # head: retrieve the first element of the list, else None.
 # If you find yourself using the function, consider using match instead.
 #
+# Examples:
+#   ```
 #   head Nil = None
 #   head ("a", "b", Nil) = Some "a"
 #   head (seq 10) = Some 0
+#   ```
 export def head: List a => Option a = match _
     Nil  = None
     h, _ = Some h
@@ -59,10 +71,13 @@ export def head: List a => Option a = match _
 # tail: remove the first element from the List
 # If you find yourself using the function, consider using match instead.
 #
+# Examples:
+#   ```
 #   tail (seq 5) = 1, 2, 3, 4, Nil
 #   tail ("a", Nil) = Nil
 #   tail Nil = Nil
 #   tail (pi, 1.0, Nil) = 1.0
+#   ```
 export def tail: List a => List a = match _
     Nil  = Nil
     _, t = t
@@ -78,10 +93,11 @@ export def tail: List a => List a = match _
 # Guarantees:
 #  - The resultant List has the same length as `list`
 #
-# ```
-# map str     (3, 9, Nil) = "3", "9", Nil
-# map (_+100) (3, 9, Nil) = 103, 109, Nil
-# ```
+# Examples:
+#   ```
+#   map str     (3, 9, Nil) = "3", "9", Nil
+#   map (_+100) (3, 9, Nil) = 103, 109, Nil
+#   ```
 export def map (mapFn: a => b): List a => List b =
     def loop list = match list
         Nil  = Nil
@@ -90,19 +106,36 @@ export def map (mapFn: a => b): List a => List b =
 
 # mapFlat: create a new List by applying a function f to each element and concatenating the output.
 #
+# Parameters:
+#  - `mapFn`: The function to apply to each element
+#  - `list`: The list of elements to feed to `mapFn`
+#
+# Examples:
+#   ```
 #   def twice x = x, x, Nil
 #   mapFlat twice (seq 3) = 0, 0, 1, 1, 2, 2, Nil
 #   mapFlat seq (seq 5) = 0, 0, 1, 0, 1, 2, 0, 1, 2, 3, Nil
-export def mapFlat (f: a => List b): List a => List b =
+#   ```
+export def mapFlat (mapFn: a => List b): List a => List b =
     def loop = match _
         Nil  = Nil
-        h, t = f h ++ loop t
+        h, t = mapFn h ++ loop t
     loop
 
 # mapPartial: create a new List by applying a partial function to each element of a List.
 # Partial functions can return None, in which case the result is not included in the output.
 #
+# Parameters:
+#  - `mapFn`: The partial function to apply to each element
+#  - `list`: The list of elements to feed to `mapFn`
+#
+# Guarantees:
+#  - The resultant List is no longer than `list`
+#
+# Examples:
+#   ```
 #   mapPartial int ("3", "x", "44", Nil) = 3, 44, Nil
+#   ```
 export def mapPartial (f: a => Option b): List a => List b =
     def loop = match _
         Nil  = Nil
@@ -118,13 +151,29 @@ export def mapPartial (f: a => Option b): List a => List b =
 # An accumulator is updated from its initial value by combiningFn for each element.
 #
 # In C++ syntax, `foldl <combiningFn> <accumulateInitalValue> <List>` does:
+#   ```
 #   auto accumulator = <accumulateInitalValue>;
 #   for (element : <List>) accumulator = combiningFn(accumulator, element);
 #   return accumulator;
+#   ```
+# In python this is similar to functools.reduce
+#
+# Parameters:
+#  - `combiningFn`: The function to combine elements of the list with the accumulator
+#  - `initialValue`: The initial value of the accumulator
+#  - `list`: The list to combine the elements of
+#
+# Guarantees:
+#  - If `list` is Nil, the result is `initialValue`
 #
 # Examples:
+#   ```
+#   foldl f x Nil = x
 #   foldl f x (a, b, c, Nil) = f (f (f x a) b) c
 #   foldl (_+_) 0 (seq 6) = 15
+#   foldl (_*_) 0 l = 0
+#   foldl (_,_) Nil l = reverse l
+#   ```
 export def foldl (combiningFn: accumulator => element => accumulator): accumulator => List element => accumulator =
     def loop accumulator = match _
         Nil = accumulator
@@ -134,13 +183,26 @@ export def foldl (combiningFn: accumulator => element => accumulator): accumulat
 # scanl: create a new List from an accumulator run over the elements front-to-back.
 # The last element of the produced List is equivalent to the result of foldl.
 #
+# Parameters:
+#  - `combiningFn`: The function to combine elements of the list with the accumulator
+#  - `initialValue`: The initial value of the accumulator
+#  - `list`: The list to scan over
+#
+# Guarantees:
+#  - The resultant List is exactly one element longer than `list`
+#  - The first element of the resultant List will be `initialValue`
+#  - The last element of the resultant List will be equal to `foldl combiningFn initialValue list`
+#  - If `combiningFn` is a group operation, the difference between the resultants yields `list`
+#
 # Examples:
+#   ```
+#   scanl f z Nil = z
 #   scanl (_+_) 100 (seq 6) = 100, 100, 101, 103, 106, 110, 115, Nil
 #                                                           ^^^- 100+0+1+2+3+4+5
 #                                            ^^^- 100+0+1+2
 #                                  ^^^- 100+0
 #                             ^^^- 100
-
+#  ```
 export def scanl (combiningFn: accumulator => element => accumulator): accumulator => List element => List accumulator =
     def loop accumulator = match _
         Nil = accumulator, Nil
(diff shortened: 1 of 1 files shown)


share/wake/lib/core/list.wake
```diff
@@ -179,63 +268,99 @@ export def scanr (combiningFn: element => accumulator => accumulator): accumulat
...
 #   prepend 9 (seq 3) = 9, 0, 1, 2, Nil
 #   prepend 8 Nil     = 8, Nil
+#   ```
 export def prepend (item: a) (list: List a): List a =
     item, list
 
 # append: add `item` to the end of `list`.
-# This function must recreate the entire list to add the item.
+#
+# **NOTE**: This function must recreate the entire list to add the item.
```
ag-eitilt, line 338: It's probably something to do in a separate PR, but I wonder if we want to add these asterisks to the "Examples" headers as well, so they're bolded in the popups.  I'm also wondering whether it's better to have the colon inside or outside the formatting.

ag-eitilt (changes requested):
Really nice seeing the documentation being polished!

share/wake/lib/core/list.wake
```diff
@@ -23,46 +23,61 @@ package wake
...
 #   0, 1, 2, Nil = (0, (1, (2, Nil)))
+#   ```
 #
 # Since List has a type parameter, any type can be legally stored in a list
+#   ```
 #   0, 1, 2, Nil         # ok! This is a `List Integer`
 #   "0", "1", "2", Nil   # ok! This is a `List String`
+#   ```
 #
 # However, mixing types it NOT legal:
```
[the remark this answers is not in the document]
JakeSiFive: I'll make this edit. It looks better in the markdown your way.

share/wake/lib/core/list.wake
```diff
@@ -23,46 +23,61 @@ package wake
...
+#   ```
 #   0, "1", Nil          # bad! This will trigger a type-check failure.
+#   ```
 export data List a =
     Nil
     (head: a), (tail: List a)
 
 # empty: report if the list contains no elements.
```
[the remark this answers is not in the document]
JakeSiFive: I'll go through and remove them in this pass. There's a limit to the number of edits I care to make in this PR but I think this is an easy additional fix here.

share/wake/lib/core/list.wake
```diff
@@ -90,19 +106,36 @@ export def map (mapFn: a => b): List a => List b =
...
+#
+# Examples:
+#   ```
 #   def twice x = x, x, Nil
 #   mapFlat twice (seq 3) = 0, 0, 1, 1, 2, 2, Nil
 #   mapFlat seq (seq 5) = 0, 0, 1, 0, 1, 2, 0, 1, 2, 3, Nil
-export def mapFlat (f: a => List b): List a => List b =
+#   ```
+export def mapFlat (mapFn: a => List b): List a => List b =
     def loop = match _
```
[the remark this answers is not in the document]
JakeSiFive: Hmm. It looks like both the first and the second way get reported to the user. I like the `match _` idiom personally so I'm inclined towards the second.

That said the argument I see for the first is that if someone goes into this file to look at this, unless they know the `match _` idiom the first is going to look more readable. It sounds like "go to definition" is a common thing people do for better or worse. Maybe we ask some people working this week and see if there's a clear tie breaker.

share/wake/lib/core/list.wake
```diff
@@ -118,13 +151,29 @@ export def mapPartial (f: a => Option b): List a => List b =
 # An accumulator is updated from its initial value by combiningFn for each element.
 #
 # In C++ syntax, `foldl <combiningFn> <accumulateInitalValue> <List>` does:
+#   ```
```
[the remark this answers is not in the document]
JakeSiFive: Oh wow, I didn't know that vscode supported that but it does!

share/wake/lib/core/list.wake
```diff
@@ -118,13 +151,29 @@ export def mapPartial (f: a => Option b): List a => List b =
...
+# Guarantees:
+#  - If `list` is Nil, the result is `initialValue`
 #
 # Examples:
+#   ```
+#   foldl f x Nil = x
 #   foldl f x (a, b, c, Nil) = f (f (f x a) b) c
 #   foldl (_+_) 0 (seq 6) = 15
+#   foldl (_*_) 0 l = 0
+#   foldl (_,_) Nil l = reverse l
```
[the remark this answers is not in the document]
JakeSiFive: Good catch! I messed that up. And yeah since it isn't very clear I'll just remove it.

share/wake/lib/core/list.wake
JakeSiFive: Yeah good question. I vote we get the content of this PR into shape and make sure its at least consistent with itself and then we can make sweeping changes like that after a good-ol'-fashioned bikeshed later.

share/wake/lib/core/list.wake
```diff
@@ -23,46 +23,61 @@ package wake
...
 
 # tail: remove the first element from the List
 # If you find yourself using the function, consider using match instead.
 #
+# Examples:
+#   ```
 #   tail (seq 5) = 1, 2, 3, 4, Nil
 #   tail ("a", Nil) = Nil
 #   tail Nil = Nil
 #   tail (pi, 1.0, Nil) = 1.0
```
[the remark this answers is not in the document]
JakeSiFive: After being fixed it doesn't seem harmful so I left it in

share/wake/lib/core/list.wake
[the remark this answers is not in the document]
ag-eitilt: I agree, `match _` is definitely nice.  And yeah, I'm mainly just wanting the documentation to be clear whether we're looking at it after rendering or in the raw file.  (The third wasn't a serious suggestion, but just noting that the extension is smart enough to pull variable names from pointfree function invocations as well.)

share/wake/lib/core/list.wake
```diff
@@ -259,9 +387,12 @@ export def take (length: Integer) (l: List a): List a =
 
 # drop: discard the first `num` elements
 #
+# Examples:
+#   ```
 #   drop 4 (seq 6) = 4, 5, Nil
 #   drop 6 (seq 6) = Nil
 #   drop 0 (seq 3) = 0, 1, 2, Nil
+#   ```
```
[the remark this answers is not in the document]
ag-eitilt: I didn't even think about doing the same for `take`, but it might raise the same question.  Probably not a required example, though.

ag-eitilt (changes requested):
The `Nil` fix is the only change I think is required in this batch, the rest are just suggestions.

ag-eitilt (approved):
That's everything I was concerned about, thanks!

share/wake/lib/core/list.wake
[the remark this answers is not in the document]
terpstra: I vote for putting it in the top-line type signature (not the match).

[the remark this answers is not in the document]
terpstra: I want to clarify, we should NOT put the `list` as an argument to `mapFlat`, as in the first example. If you do that, you will degrade inlining.

Here are the relevant facts (as exploited in list.wake):
- wake never inlines recursive funcitons
- wake WILL inline a function that INCLUDES a recursive function
- wake WILL lift and CSE identical functions outside of irrelevant closures

This means that you want a style like this for important reused funtions:
```
def map mapFn =
  def loop list = match list
    ..
  loop
```

That way, if someone invokes `map str (seq 100)`, the `map` will be inlined, which then in turn allows `map` to inline the `mapFn`. This way you end up with an actual direct loop that invokes the `str` primitive directly. Furthermore, if someone else calls `map str (seq 101)`, wake will lift the `map str`'s `loop` function up and CSE the two functions, so both call sites will invoke the direct loop which applies `str`.

In contrast, if you had defined:
```
def map mapFn list = match list
  ...
```
... this would never be inlined, which means that `str` would be passed as a closure to `map` and called indirectly.

share/wake/lib/core/list.wake
```diff
@@ -23,46 +23,61 @@ package wake
...
 #   0, 1, 2, Nil         # ok! This is a `List Integer`
 #   "0", "1", "2", Nil   # ok! This is a `List String`
+#   ```
 #
-# However, mixing types it NOT legal:
+# However, mixing types is *not* legal:
+#   ```
 #   0, "1", Nil          # bad! This will trigger a type-check failure.
+#   ```
```
terpstra, line 39: You should build and install the newest wake LSP when testing this PR. In particular, wake now adds pop-up text for `List`, `Nil`, and `,`. For the two latter constructors, it combines the comments for `List` with any specific comments for the constructor. Therefore, I think it may make sense (especially for the all important List type) to include specific documentation for the constructors, to set a good example.

share/wake/lib/core/list.wake
```diff
@@ -23,46 +23,61 @@ package wake
...
 export data List a =
     Nil
     (head: a), (tail: List a)
 
-# empty: report if the list contains no elements.
+# Report if the list contains no elements.
 # If you find yourself using the function, consider using match instead.
 #
+# Examples:
+#   ```
```
terpstra, line 48: I notice you're indenting the triple-backtics to line up with the code inside. In wake itself, you would indent the contents of the opening escape, for example:

```
def foo =
  """
    This is a multi-line string.
    Hello world.
  """
```

Or:
```
def foo = (
  def bar = foo + foo
  foo
  + bar
  + bar
)
```

... to be consistent with code, should we follow this policy also within the comments?

terpstra: So, I guess I am advocating for either:
```
# Examples:
# ```
#   foo x = True
# ```

or

# Examples:
#   ```
#     foo x = True
#   ```
```

share/wake/lib/core/list.wake
```diff
@@ -114,58 +147,113 @@ export def mapPartial (f: a => Option b): List a => List b =
...
+# In python this is similar to `functools.reduce`
+#
+# Parameters:
+#  - `combiningFn`: The function to combine elements of the list with the accumulator
+#  - `initialValue`: The initial value of the accumulator
+#  - `list`: The list to combine the elements of
+#
+# Guarantees:
+#  - If `list` is Nil, the result is `initialValue`
```
terpstra, line 167: I understand that the 'golden example' included similarly a redundant `Guarantee`, but that was mostly because I was trying to show how to format stuff in  markdown. Since the examples already shows this claim, I think maybe we should not include the guarantee. There is a limit to how much text you get on a pop-up for documentation hover. I think we should try to include the most important stuff within that limit.

share/wake/lib/core/list.wake
```diff
@@ -179,63 +267,100 @@ export def scanr (combiningFn: element => accumulator => accumulator): accumulat
 # Concatenate two lists into one List.
 # The ++ operator must be provided by the wake package for publish to work.
 #
+# Guarantees:
+#  - ++ is associtive
+#  - Nil is an identity of ++
```
terpstra, line 272: In contrast, this associative guarantee is super relevant. It might help to write `x ++ (y ++ z) == (x ++ y) ++ z`, though, for people who don't know what that word means.

share/wake/lib/core/list.wake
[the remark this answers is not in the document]
terpstra: OMG. A behaviour change based on sign? That's pure evil! What language does that?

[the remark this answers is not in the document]
terpstra: If other languages do this, I totally agree that both `take` and `drop` need an example with a negative number.

share/wake/lib/core/list.wake
```diff
@@ -499,51 +729,69 @@ export def tab (f: Integer => a): Integer => List a =
...
+# Guarantees:
+#  - Both output lists will be the same length
+#  - if `(len x) == (len y)` then `unzip (zip x y) = Pair x y`
+#  - `(\(Pair x y) zip x y) (unzip l) = l`
+#
+# Example:
+#   ```
 #   unzip (Pair 4 "a", Pair 7 "b", Nil) = Pair (4, 7, Nil) ("a", "b", Nil)
+#   unzip Nil = Pair Nil Nil
+#   ```
```
terpstra, line 771: In retrospect, I wonder if anyone actually uses `zip`/`unzip`. I added them simply because `list.wake` was modeled on the scala `Seq` type. Giving `Pair` special treatment seems rather odd, IMO.

share/wake/lib/core/list.wake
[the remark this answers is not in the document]
ag-eitilt: I know Python does, and GNU head might be conceptually close enough to count (even though it has somewhat different semantics).  Looks like Javascript might as well.  I feel like I've come across a language before that actually does that *and* has a specific `take`/`drop` pair rather than using substring-style functions, but those examples are probably related enough to generalize to this.

terpstra:
I see that CI has failed repeatedly on this PR. I'll fix that shortly.

Also: when you believe you've addressed the comments people leave on your PR, please click the 'resolved' button to differentiate open vs. closed issues. That way, if someone agrees with the solution, they don't feel compelled to write anything extra.

share/wake/lib/core/list.wake
JakeSiFive: Sure. I think of those two options I prefer the first (quotes unintended but code remains indented). I'll switch that over

terpstra:
If you rebase this PR onto the current master branch, it should fix your CI troubles.

share/wake/lib/core/list.wake
JakeSiFive: How do I install the latest one?

share/wake/lib/core/list.wake
JakeSiFive: I'm down to delete unzip honestly, let me know what you think (@ag-eitilt as well). I've seldom if ever had a reason to use it. When it comes up you can use a pair of maps for it instead I think.

Regardless lets maybe remove unzip in a different PR

ag-eitilt: Definitely a separate PR.  I agree that `zip`/`unzip` as part of the "List API" doesn't make as much sense (the main ways I use them in Haskell involve either `uncurry` or infinite lists, neither of which we have here), but it might fit a bit better as part of the "Pair API" -- conceptualized not as a special-casing lists which happen to contain pairs, but as an operation on pairs gathered in some collection.  Still probably not something we need to keep.

JakeSiFive: I use zip a fair but but not unzip. Also zipWith is way cooler than zip anyway and we seem to be lacking it. zip is just zipWith Pair then.

share/wake/lib/core/list.wake
terpstra: Probably easiest is to grab the 'vsix' from the circle CI artifacts on the 'release' green check mark of a PR.

terpstra (approved):
Looks good. Please merge!

PR #842, opened 2022-01-07
system: entirely remove `BadPath` constructor

ag-eitilt:
In preparation for the next release, this finally removes the last traces of `BadPath`.  As they are then redundant, `getPathResult` and `getPathError` have both also been removed from the live API.  Unfortunately, I've not been able to come up with any good ways to maintain compatibility with the previous interface without also blowing a hole through the type safety, so this removes the 0.19 and 0.22 compatibility layers.


ag-eitilt:
I just realized that `getPathParent` could potentially break type safety (taking the parent of `workspace` will still return a seemingly-valid `Path`) and should probably be removed; it's not used currently, and if anyone feels a need to do so, they can `getPathName` and try to `mkdir` or similar.  Since I was doing that already, I removed `getPathChild` which is poorly named as a counterpart to `getPathParent` and can be easily duplicated with `_.getPathName | basename`.

share/wake/lib/system/path.wake
```diff
@@ -96,33 +96,17 @@ export def basename (file: String): String =
...
-  BadPath _ = False
-  Path str = matches `0+` (hashcode str)
+export def isPathDir (path: Path): Boolean =
+    getPathHash path
+    | matches `0+`
 
 # Paths differ from Strings in that they have been hashed; their content is frozen
-from internals_wake export type Path
-from internals_wake import def Path BadPath # Path constructor stays private!
+export data Path =
```
terpstra, line 104: Please remove the `export`. We don't want the user able to construct `Path`.

share/wake/lib/system/path.wake
```diff
@@ -96,33 +96,17 @@ export def basename (file: String): String =
...
-
-export def getPathName (path: Path): String = match path
-  Path name = name
-  BadPath _ = "BADPATH"
-
-export def getPathChild (path: Path): String = match path
-  Path name = replace `^.*/` '' name
-  BadPath _ = "BADPATH"
-
-export def getPathParent (path: Path): Path = match path
```
terpstra, line 123: This cannot be replicated using `mkdir`, because the parent might have been created by a `Job`, resulting in two jobs with the same output file. If you run `getPathParent` on `Path "."`, you will get `Path ".."`, which I don't think is too terrible.

So, please don't remove this one.

share/wake/lib/system/path.wake
```diff
@@ -96,33 +96,17 @@ export def basename (file: String): String =
...
-
-export def getPathError (path: Path): Option Error = match path
-  Path _    = None
-  BadPath e = Some e
-
-export def getPathName (path: Path): String = match path
-  Path name = name
-  BadPath _ = "BADPATH"
-
-export def getPathChild (path: Path): String = match path
```
terpstra, line 119: I guess I don't really care about this function, now that `basename` is in the standard library.

terpstra (approved):
Assuming you make my two proposed changes, I approve.

share/wake/lib/system/path.wake
ag-eitilt: Ah, I must not have been looking when copying things out of `internals_wake`.  This should definitely not be exported!

share/wake/lib/system/path.wake
ag-eitilt: Isn't `mkdir` an exception to the same-output checks?  Anyway, if you don't see an issue with this (and since even `/..` is a valid absolute path) I'll add it back in.

terpstra: Multiple to `mkdir` do not conflict with each other, but would conflict with a `Job`.
Actually, `simplify "/.." = "/"`.

mmjconolly:
We should roll a new release after this is merged, as it's a change worthy of a version bump given that it breaks backwards compatibility

PR #849, opened 2022-01-27
Rename .fuse mountpoint to include user-id and group-id

mmjconolly:
Fuse mounts are viewable to linux processes that have the same uid:gid pair.
If someone invoke fuse-wake or wakbox with a different group-id then they're unable to contact an existing fuse-waked daemon.

This change creates different mountpoint names depending on the invoked fuse-wake/wakebox userid and groupid, so that it can create a separate fuse-waked daemon for its uid:gid, thus allowing it to continue even in the presence of an existing daemon of another id.

I've tested this manually by invoking wakebox twice each under different group ids, and by running it within wake via a job that invokes two sleep jobs where one changes the group before invoking wakebox, and then also by running a large build that changes job groups in another codebase.

Within the wake workspace, the user can observe the change if they run `ls -la`, the `.fuse` and `.fuse.log` are instead `.fuse-1000:1000` and `.fuse-1000:1000.log` (if their user-id was 1000). If the job used multiple ids there would be multiple log files present and they'd be owned by different ids.


mmjconolly:
Havent made sense of the CI failures yet

It also seems like there are more references to `.fuse` in the code that will need handling
```
tests/runtime/symlinks/pass.sh: rm -f datFile badLink goodLink .fuse.log   
src/runtime/sources.cpp:        if (name == ".build" || name == ".fuse") continue;
src/parser/wakefiles.cpp:       if (dirent.name != ".build" && dirent.name != ".fuse" && dirent.name != ".git")
src/parser/wakefiles.cpp:       if (name == ".build" || name == ".fuse" || name == ".git") continue;
```

src/wakefs/daemon_client.cpp
```diff
@@ -34,9 +34,16 @@
 #include "json/json5.h"
 #include "fuse.h"
 
+// The user-id and group-id are used so that fuse daemons with different uid:gid pairs
+// running within the same build can co-exist without trying to share. The kernel
+// prevents cross-user sharing when the 'allow_other' fuse mount option is not used.
+const std::string mount_path_id(const std::string& base_dir) {
+    return base_dir + "/.fuse-" + std::to_string(getuid()) + ":" + std::to_string(getgid());
+}
```
JakeSiFive, line 42: I didn't understand yesterday but I now get why you guys were laughing. This is painfully simple lol.

src/wakefs/daemon_client.cpp
```diff
@@ -34,9 +34,16 @@
 #include "json/json5.h"
 #include "fuse.h"
 
+// The user-id and group-id are used so that fuse daemons with different uid:gid pairs
+// running within the same build can co-exist without trying to share. The kernel
+// prevents cross-user sharing when the 'allow_other' fuse mount option is not used.
+const std::string mount_path_id(const std::string& base_dir) {
+    return base_dir + "/.fuse-" + std::to_string(getuid()) + ":" + std::to_string(getgid());
```
ag-eitilt, line 41: If we care about Windows compatibility, colons aren't allowed in filenames and we might want to choose a different separator.  Since Wake isn't tested on Windows that may be an unnecessary concern, but I just wanted to mention it anyway.

mmjconolly: ah ok, can change it

JakeSiFive: Totally unrelated to this review that's an interesting question I've been wondering. If we get wake into a better state it might be useful for a much wider audience but most large projects that might use wake (and then contribute back) require the big three to all work (linux, mac, windows). If supporting mac/windows becomes a real burden then we should make a choice about cutting support.

For now I suspect that just sandboxing less as needed is a viable mechanism to support mac/windows so simple things like this leave open the opportunity for future open source success.

mmjconolly: I don't think we try to support windows in general, just for the vscode plugin

mmjconolly: Though I guess if someone were willing to put in the work for it

terpstra: I would propose "/.fuse/{uid}.{gid}" so that all the fuse stuff remains in a directory called '.fuse', because some other parts of the codebase in wake give that directory special treatment.

terpstra: Currently, wake runs natively on OS/X and runs in WSL under windows.

mmjconolly: yep can try that out instead

mmjconolly:
When I run the tests locally it passes for me, so I'll need to test-in-CI so you may see some churn while I do various force-pushes

mmjconolly:
I've fixed the CI infra issues (in another PR), so this is ready for review again.

It uses a subdirectory of `.fuse/`, and the uid gid separator is now a `.`

JakeSiFive (approved):
Looks good to me!

src/wakefs/daemon_client.cpp
pxs7: > Totally unrelated to this review that's an interesting question I've been wondering. If we get wake into a better state it might be useful for a much wider audience but most large projects that might use wake (and then contribute back) require the big three to all work (linux, mac, windows). If supporting mac/windows becomes a real burden then we should make a choice about cutting support.
> 
> For now I suspect that just sandboxing less as needed is a viable mechanism to support mac/windows so simple things like this leave open the opportunity for future open source success.



> 
any way to fix  error that no fuse in wsl  when compiling wake src code？

mmjconolly: Please to not resurrect old pull requests to ask questions. Use the Github Issues mechanism instead.

Issue #852, opened 2022-02-01
centos 8 EOL


mmjconolly:
Centos 8 is now end-of-life, they announced some time ago that they're cutting support short.
They've now moved the package mirror to a vault URL, this breaks CI for us (thanks red hat..).

We should move to rocky linux 8, as that fills a similar role in being "RHEL8-ish"

PR #854, opened 2022-02-02
CI: use rocky linux 8 instead of centos 8

mmjconolly:
Both centos8 and rocky8 are RHEL8 clones.
Centos8 discontinued support and broke their mirror urls..
Use rocky8 instead, as that will be supported for some time (and using actual RHEL8 requires licences which isnt helpful).

Fixes https://github.com/sifive/wake/issues/852

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,23 +124,23 @@ jobs:
           paths:
             - release/centos_7_6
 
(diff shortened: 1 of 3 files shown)


mmjconolly:
and changed the macos version to 10.15 rather than latest, cause that's been broken the last few days

PR #857, opened 2022-02-03
release fetch: use rocky8

mmjconolly:
The fetch script for releases also needs to know about the `s/centos_8_3/rocky_8/`


ag-eitilt (approved):
I don't remember seeing any other references to Centos, so hopefully this is it all updated.

PR #858, opened 2022-02-09
Minor changes to timeline and licenses

elena-pudovkina:
Timeline code is cleaned and sped up.
Years and misprints in licenses are corrected.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,7 +2,7 @@ This source repository includes software from various sources.
 Each file includes a header which describes the relevant license.
 
 In broad strokes:
-  wake          LICENSE.Apache2   everthing not in vendor/
+  wake          LICENSE.Apache2   everything not in vendor/
 
   BLAKE2        LICENSE.CC0       vendor/blake2/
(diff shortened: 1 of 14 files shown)


JakeSiFive (approved):
LGTM

JakeSiFive:
wait no I take it back, we want 2022 not 2021...that always takes getting used to

elena-pudovkina:
> wait no I take it back, we want 2022 not 2021...that always takes getting used to

It does!

terpstra (approved):

JakeSiFive:
Uh this got merged without the requested change to 2022 btw

elena-pudovkina:
> Uh this got merged without the requested change to 2022 btw

Ah, I misinterpreted your comment. The lsp stuff was written in 2021

JakeSiFive:
Not a big deal but the advice that has been handed down to me from lawyers is that it should be the date of first publication not the date of authorship. So if these were previously published in 2021 then this is correct but generally when something is added in review in 2022 we want to use 2022....this probably won't matter again until 2023 but just FYI for next time!

PR #859, opened 2022-02-10
Added claim function for using edited files

JakeSiFive:
This change adds add_claim to the standard library where we have access to things like raw_sources. If we're happy with this I'll also use this as an excuse to do a minor release so I can learn that process.

The specific semantics I chose are to restrict `claim` to operating within wakeroot to limit its effectiveness. I couldn't think of a way to prevent people from using it instead of `source` but added that `source` is preferred in the documentation.

### DIFF
--- a/share/wake/lib/system/sources.wake
+++ b/share/wake/lib/system/sources.wake
@@ -66,7 +66,7 @@ export def source (file: String): Result Path Error =
 # all paths be produced by exactly one job, virtual or
 # otherwise.
 #
-# For source files please use `soruce`. For files outside
+# For source files please use `source`. For files outside
 # of wakeroot please use `claimFileAsPath` or `claimFileAsPathIn`.
 # `claim` is good for claiming files of previous builds that
 # are not being used again this build.
--- a/share/wake/lib/system/sources.wake
+++ b/share/wake/lib/system/sources.wake
@@ -58,6 +58,42 @@ export def source (file: String): Result Path Error =
         x, Nil = Pass x
         _      = unreachable "a quoted RegExp can never match more than one distinct String"
 
+# Allows claiming of a file so long as the file is within
+# wakeroot. Keep in mind that if another job produces this
+# file, claim will fail. Likewise if you claim a file then
+# any job that produces this file will fail. Claim should
+# be thought of as a "virtual" job and wake demands that
+# all paths be produced by exactly one job, virtual or
+# otherwise.
+#
+# For source files please use `soruce`. For files outside
+# of wakeroot please use `claimFileAsPath` or `claimFileAsPathIn`.
+# `claim` is good for claiming files of previous builds that
+# are not being used again this build.
+#
+# Parameters:
+#  - `file`: The string giving the path to the file to be claimed
+#
+# Example:
+# ```
+#   claim "build/my_flow/my_step/previous-output.txt" # Claims the file
+# ```
+export def claim (file: String): Result Path Error =
+    def claimRel relpath =
+        def isOutsideRoot = matches `\.\..*` relpath
+        if isOutsideRoot then
+            failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+        else
+            raw_source relpath
+
+    def simplePath = simplify file
+    def isAbs = matches `/.*` simplePath
+    require True = isAbs else claimRel simplePath
+
+    relative workspace file
+    | simplify
+    | claimRel
+
 # Find sources files
 export def sources (dir: String) (filterRegexp: RegExp): Result (List Path) Error =
     def scan dir regexp = prim "sources"
--- a/share/wake/lib/system/sources.wake
+++ b/share/wake/lib/system/sources.wake
@@ -84,7 +84,7 @@ export def claim (file: String): Result Path Error =
         def get_modtime file = prim "get_modtime"
         def time = get_modtime file
         if time == -1 then
-            failWithError "{file}: files does not exist. Maybe you haven't previouslly built this file?"
+            failWithError "{file}: files does not exist. Maybe you haven't previously built this file?"
         else
             # Its important that the command here is distinct from what `source` would use
             # so that claim and source do not overlap.
--- a/share/wake/lib/system/sources.wake
+++ b/share/wake/lib/system/sources.wake
@@ -105,10 +105,19 @@ export def claim (file: String): Result Path Error =
 
     # Fail if we're outside of wakeroot (relPath must be simplified)
     def isOutsideRoot = matches `\.\..*/.*` relPath
-    if isOutsideRoot then
-        failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
-    else
-        raw_claim relPath
+    require False = isOutsideRoot
+    else failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+
+    # Now that our path is uniform check if it could have been fetched with `source`
+    def base = basename relPath
+    def dir = dirname relPath
+    def scan dir regexp = prim "sources"
+    def canBeSourced = exists (_==~relPath) (scan dir base.quote)
+    require False = canBeSourced
+    else failWithError "{file}: Is a source file, please use `source` or `sources` instead of `claim`"
+
+    # Finally if the file is inside of wakeroot and can't be sourced we're in the clear
+    raw_claim relPath
 
 # Find sources files
 export def sources (dir: String) (filterRegexp: RegExp): Result (List Path) Error =
--- a/share/wake/lib/system/sources.wake
+++ b/share/wake/lib/system/sources.wake
@@ -79,20 +79,36 @@ export def source (file: String): Result Path Error =
 #   claim "build/my_flow/my_step/previous-output.txt" # Claims the file
 # ```
 export def claim (file: String): Result Path Error =
-    def claimRel relpath =
-        def isOutsideRoot = matches `\.\..*` relpath
-        if isOutsideRoot then
-            failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+    # Define the actual virtual job that produces the Path
+    def raw_claim file =
+        def get_modtime file = prim "get_modtime"
+        def time = get_modtime file
+        if time == -1 then
+            failWithError "{file}: files does not exist. Maybe you haven't previouslly built this file?"
         else
-            raw_source relpath
+            # Its important that the command here is distinct from what `source` would use
+            # so that claim and source do not overlap.
+            makeExecPlan ("<claim>", str time, file, Nil) Nil
+            | setPlanShare       False
+            | setPlanEcho        logDebug
+            | setPlanEnvironment Nil
+            | setPlanFnOutputs   (file, _)
+            | runJobWith virtualRunner
+            | getJobOutput
 
-    def simplePath = simplify file
-    def isAbs = matches `/.*` simplePath
-    require True = isAbs else claimRel simplePath
+    # Compute the relative path in simplist form
+    def isAbs = matches `/.*` file
+    def relPath = if isAbs then
+        relative workspace file
+    else
+        simplify file
 
-    relative workspace file
-    | simplify
-    | claimRel
+    # Fail if we're outside of wakeroot (relPath must be simplified)
+    def isOutsideRoot = matches `\.\..*/.*` relPath
+    if isOutsideRoot then
+        failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+    else
+        raw_claim relPath
 
 # Find sources files
 export def sources (dir: String) (filterRegexp: RegExp): Result (List Path) Error =
--- a/share/wake/lib/system/sources.wake
+++ b/share/wake/lib/system/sources.wake
@@ -59,17 +59,18 @@ export def source (file: String): Result Path Error =
         _      = unreachable "a quoted RegExp can never match more than one distinct String"
 
 # Allows claiming of a file so long as the file is within
-# wakeroot. Keep in mind that if another job produces this
-# file, claim will fail. Likewise if you claim a file then
-# any job that produces this file will fail. Claim should
-# be thought of as a "virtual" job and wake demands that
-# all paths be produced by exactly one job, virtual or
-# otherwise.
+# wake's workspace. Keep in mind that if another job
+# produced this file in the same run of wake, claim will
+# fail. Likewise if you claim a file then any job that
+# produces this file will fail. Claim should be thought
+# of as a "virtual" job and wake demands that all paths
+# be produced by exactly one job, virtual or otherwise.
 #
 # For source files please use `source`. For files outside
-# of wakeroot please use `claimFileAsPath` or `claimFileAsPathIn`.
-# `claim` is good for claiming files of previous builds that
-# are not being used again this build.
+# of wake's workspace please use `claimFileAsPath` or
+# `claimFileAsPathIn`. `claim` is good for claiming
+# artifacts of previous builds that should not be rebuilt
+# during this run.
 #
 # Parameters:
 #  - `file`: The string giving the path to the file to be claimed
@@ -84,7 +85,7 @@ export def claim (file: String): Result Path Error =
         def get_modtime file = prim "get_modtime"
         def time = get_modtime file
         if time == -1 then
-            failWithError "{file}: files does not exist. Maybe you haven't previously built this file?"
+            failWithError "{file}: file does not exist. Maybe you haven't previously built this file?"
         else
             # Its important that the command here is distinct from what `source` would use
             # so that claim and source do not overlap.
@@ -96,17 +97,17 @@ export def claim (file: String): Result Path Error =
             | runJobWith virtualRunner
             | getJobOutput
 
-    # Compute the relative path in simplist form
+    # Compute the relative path in simplest form
     def isAbs = matches `/.*` file
     def relPath = if isAbs then
         relative workspace file
     else
         simplify file
 
-    # Fail if we're outside of wakeroot (relPath must be simplified)
+    # Fail if we're outside of wake's workspace (relPath must be simplified)
     def isOutsideRoot = matches `\.\..*/.*` relPath
     require False = isOutsideRoot
-    else failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+    else failWithError "{file}: claim cannot be used on files outside of wake's workspace. Consider using claimFileAsPath instead."
 
     # Now that our path is uniform check if it could have been fetched with `source`
     def base = basename relPath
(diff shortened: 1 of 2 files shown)


share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,42 @@ export def source (file: String): Result Path Error =
...
 
+# Allows claiming of a file so long as the file is within
+# wakeroot. Keep in mind that if another job produces this
+# file, claim will fail. Likewise if you claim a file then
+# any job that produces this file will fail. Claim should
+# be thought of as a "virtual" job and wake demands that
+# all paths be produced by exactly one job, virtual or
+# otherwise.
+#
+# For source files please use `soruce`. For files outside
```
mmjconolly, line 69:
```suggestion
# For source files please use `source`. For files outside
```

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,42 @@ export def source (file: String): Result Path Error =
...
 
+# Allows claiming of a file so long as the file is within
+# wakeroot. Keep in mind that if another job produces this
+# file, claim will fail. Likewise if you claim a file then
+# any job that produces this file will fail. Claim should
+# be thought of as a "virtual" job and wake demands that
+# all paths be produced by exactly one job, virtual or
+# otherwise.
+#
+# For source files please use `source`. For files outside
```
mmjconolly, line 69: I'm wondering if this should error if they try to claim a file that source could have provided

terpstra:
Please set up a meeting to discuss this. I have lots of comments.

share/wake/lib/system/sources.wake
JakeSiFive: At the time I wasn't sure if that was possible but thinking about it now it looks like I could se `prim "sources"` to do that. I can't use `source` or `sources` because those claim the file but `prim "sources"` just tells you which sources match which is better.

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,42 @@ export def source (file: String): Result Path Error =
...
+# Parameters:
+#  - `file`: The string giving the path to the file to be claimed
+#
+# Example:
+# ```
+#   claim "build/my_flow/my_step/previous-output.txt" # Claims the file
+# ```
+export def claim (file: String): Result Path Error =
+    def claimRel relpath =
+        def isOutsideRoot = matches `\.\..*` relpath
```
terpstra, line 83: Add the the '/'... A file can be called "..hello" and would fail this test.

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,42 @@ export def source (file: String): Result Path Error =
...
+        if isOutsideRoot then
+            failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+        else
+            raw_source relpath
```
terpstra, line 87: Please make your own `raw_source` that uses "<claim>"

share/wake/lib/system/sources.wake
terpstra: This will be fixed by not using `raw_source` and "<source>".

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,42 @@ export def source (file: String): Result Path Error =
...
+
+    def simplePath = simplify file
+    def isAbs = matches `/.*` simplePath
+    require True = isAbs else claimRel simplePath
+
+    relative workspace file
+    | simplify
+    | claimRel
```
terpstra, line 95: Please unify the 'claimRel' invocations by doing absolute => relative as a prior common step.
Then you don't even need a helper method.

share/wake/lib/system/sources.wake
JakeSiFive: I think Matthew's point is that we're asking people to prefer `source` but not enforcing that. If we Instead used `prim "sources"` we could enforce it rather than leaving it for review.

JakeSiFive:
Wanted to add examples of this working so people could check the test cases in case the reasoning was a bit hard to follow. I tried to make the flow a bit more clear since it was confusing last time.

```
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "..test.txt"'
Pass (Path "..test.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "../..test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "/scratch/jakee/test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "/scratch/jakee/wake_build/test.txt"'
Pass (Path "test.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "test.txt"'
Pass (Path "test.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "../test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "foo/../../test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "foo/../test.txt"'
Pass (Path "test.txt")
```

These are all working as intended IMO and we've got some ok-ish edge case coverage there. `"foo/../../test.txt"` confirms that `relative` simplifies.

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,58 @@ export def source (file: String): Result Path Error =
...
+    # Define the actual virtual job that produces the Path
+    def raw_claim file =
+        def get_modtime file = prim "get_modtime"
+        def time = get_modtime file
+        if time == -1 then
+            failWithError "{file}: files does not exist. Maybe you haven't previously built this file?"
+        else
+            # Its important that the command here is distinct from what `source` would use
+            # so that claim and source do not overlap.
+            makeExecPlan ("<claim>", str time, file, Nil) Nil
```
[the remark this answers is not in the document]
JakeSiFive: Yes. You cannot both claim and source a file. You can however claim a file twice (which I had not intended initially but after speaking with Wes I guess thats how `source` works).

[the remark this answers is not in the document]
mmjconolly: I feel that the error message is probably going to confuse a user then, they wont view source and claim as jobs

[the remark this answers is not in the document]
JakeSiFive: I'll add that to that query, that query should not work that way most likely. What its trying to do I guess is identify virtual jobs. That indicates to me that we need a marker in the database for virtual jobs results.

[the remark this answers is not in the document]
JakeSiFive: I think I can use `prim "sources"` to make this work, it doesn't hash anything so I can add a check for this here; I'll throw it up real quick. One sec.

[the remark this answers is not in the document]
JakeSiFive: Ok I added some handling for that, here's the latest manual testing (I should add these as unit tests huh?)

```
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "..test.txt"'
Pass (Path "..test.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "../..test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "/scratch/jakee/test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "/scratch/jakee/wake_build/test.txt"'
Fail (Error "/scratch/jakee/wake_build/test.txt: Is a source file, please use `source` or `sources` instead of `claim`" Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "test.txt"'
Fail (Error "test.txt: Is a source file, please use `source` or `sources` instead of `claim`" Nil)
jakee@gamma06:/scratch/jakee/wake_build$ touch test2.txt
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "test2.txt"'
Pass (Path "test2.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "/scratch/jakee/wake_build/test2.txt"'
Pass (Path "test2.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "../test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "foo/../../test.txt"'
Fail (Error "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead." Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "foo/../test.txt"'
Fail (Error "foo/../test.txt: Is a source file, please use `source` or `sources` instead of `claim`" Nil)
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "foo/../test2.txt"'
Pass (Path "test2.txt")
jakee@gamma06:/scratch/jakee/wake_build$ ./bin/wake -x 'claim "test.txt"'
Fail (Error "test.txt: Is a source file, please use `source` or `sources` instead of `claim`" Nil)
```

With this `source` and `claim` will never succeed on each others turf so to speak. We can have a discussion later about a possible unification. I know that having to `git add` everything is annoying for people. I haven't heard the merits of using it quite.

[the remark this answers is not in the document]
mmjconolly: seems good to me

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+# wakeroot. Keep in mind that if another job produces this
+# file, claim will fail. Likewise if you claim a file then
+# any job that produces this file will fail. Claim should
+# be thought of as a "virtual" job and wake demands that
+# all paths be produced by exactly one job, virtual or
+# otherwise.
+#
+# For source files please use `source`. For files outside
+# of wakeroot please use `claimFileAsPath` or `claimFileAsPathIn`.
+# `claim` is good for claiming files of previous builds that
```
mmjconolly, line 71: > claiming files of previous builds

Is it worth trying to enforce that it did come from a previous build by checking if the file exists already in the database? Or is this intended to simply be a way to allow creation of `Path`s

mmjconolly (approved):
I'm good with this

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
         x, Nil = Pass x
         _      = unreachable "a quoted RegExp can never match more than one distinct String"
 
+# Allows claiming of a file so long as the file is within
+# wakeroot. Keep in mind that if another job produces this
+# file, claim will fail. Likewise if you claim a file then
```
ag-eitilt, line 63: "another job produced this file in the same run"?

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+# any job that produces this file will fail. Claim should
+# be thought of as a "virtual" job and wake demands that
+# all paths be produced by exactly one job, virtual or
+# otherwise.
+#
+# For source files please use `source`. For files outside
+# of wakeroot please use `claimFileAsPath` or `claimFileAsPathIn`.
+# `claim` is good for claiming files of previous builds that
+# are not being used again this build.
```
ag-eitilt, line 72: I know this is literally the main reason it's being written, but I feel this might be a bit of a dangerous recommendation as part of the docs.  Maybe add a sentence making it clear that we're talking about files which provide an explicit re-entry point, rather than anything which might not be being rebuilt simply due to caching?
```suggestion
# are not being built again this build.
```

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+# ```
+#   claim "build/my_flow/my_step/previous-output.txt" # Claims the file
+# ```
+export def claim (file: String): Result Path Error =
+    # Define the actual virtual job that produces the Path
+    def raw_claim file =
+        def get_modtime file = prim "get_modtime"
+        def time = get_modtime file
+        if time == -1 then
+            failWithError "{file}: files does not exist. Maybe you haven't previously built this file?"
```
ag-eitilt, line 87:
```suggestion
            failWithError "{file}: file does not exist. Maybe you haven't previously built this file?"
```

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+    def isAbs = matches `/.*` file
+    def relPath = if isAbs then
+        relative workspace file
+    else
+        simplify file
+
+    # Fail if we're outside of wakeroot (relPath must be simplified)
+    def isOutsideRoot = matches `\.\..*/.*` relPath
+    require False = isOutsideRoot
+    else failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
```
ag-eitilt, line 109:
```suggestion
    else failWithError "{file}: claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
```

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+
+    # Now that our path is uniform check if it could have been fetched with `source`
+    def base = basename relPath
+    def dir = dirname relPath
+    def scan dir regexp = prim "sources"
+    def canBeSourced = exists (_==~relPath) (scan dir base.quote)
```
ag-eitilt, line 115: These all just *barely* squeak through as exceptions to (my interpretation of) the newline style.  Really, I'm only all right with `canBeSourced` because it's hiding among everything else in the block which has legitimate reasons for being single lines, but that one in particular should probably be broken up.

ag-eitilt (approved):
Picture me giving you a dubious eye over the style in a place or two, but I'm otherwise happy with this.

share/wake/lib/system/sources.wake
terpstra, line 109: I think you meant "wake's workspace", not "wake root".

terpstra (approved):
I think this can go in, as is. We just adding some tech debt. As long as we plan to pay it down, that's fine.

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+            # so that claim and source do not overlap.
+            makeExecPlan ("<claim>", str time, file, Nil) Nil
+            | setPlanShare       False
+            | setPlanEcho        logDebug
+            | setPlanEnvironment Nil
+            | setPlanFnOutputs   (file, _)
+            | runJobWith virtualRunner
+            | getJobOutput
+
+    # Compute the relative path in simplist form
```
elena-pudovkina, line 99:
```suggestion
    # Compute the relative path in simplest form
```

elena-pudovkina (approved):
I don't really know wake, so I can't provide meaningful feedback. But I have nothing against this

share/wake/lib/system/sources.wake
JakeSiFive: I left a comment elsewhere about this but I'm willing to expend some social capital, even go in debt, to fight this standard. The current rules are not good for automation, learnability, and are too restrictive in many many cases IMO. Even the strictest standards you see elsewhere don't even come close to suggesting these be reformatted.

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,68 @@ export def source (file: String): Result Path Error =
...
+
+    # Compute the relative path in simplest form
+    def isAbs = matches `/.*` file
+    def relPath = if isAbs then
+        relative workspace file
+    else
+        simplify file
+
+    # Fail if we're outside of wake's workspace (relPath must be simplified)
+    def isOutsideRoot = matches `\.\..*/.*` relPath
```
ag-eitilt, line 108: I missed this regex in this version of the PR.  The first `.*` needs to be removed; as-is, it rejects `..x/y` which is a perfectly legal -- if not recommended -- file within the workspace, while allowing `..` without a trailing slash (until that gets caught by the "not a file" check) later.  The former is something we should allow regardless; for the latter I'd argue that `..` and `../` are conceptually identical and shouldn't be treated differently -- especially when `isOutsideRoot`, in isolation, will sometimes say that the parent directory is actually located within the root.

share/wake/lib/system/sources.wake
```diff
@@ -58,6 +58,67 @@ export def source (file: String): Result Path Error =
...
+    require False = isOutsideRoot
+    else failWithError "claim cannot be used on files outside of wake's root. Consider using claimFileAsPath instead."
+
+    # Now that our path is uniform check if it could have been fetched with `source`
+    def base = basename relPath
+    def dir = dirname relPath
+    def scan dir regexp = prim "sources"
+    def canBeSourced = exists (_==~relPath) (scan dir base.quote)
+    require False = canBeSourced
+    else failWithError "{file}: Is a source file, please use `source` or `sources` instead of `claim`"
```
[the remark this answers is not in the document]
ag-eitilt: Honestly, I'd love to remove the reliance on git.  My favorite VCS is Darcs, and as it is now, I wouldn't use Wake for personal projects.  I don't know exactly what went into designing `source`, but my guess at its merits would be not needing to walk the filesystem for basic file existence tests, ensuring that we can only `source` legal files (in the workspace, not a directory), and maybe something with filtering out `.gitignore` files.  Now that you've implemented those tests in Wake, I'm not sure we do need `source` any more, or rather that we could replace the behaviour of `source` with that of `claim`, though I'm certainly not the only one to ask.

share/wake/lib/system/sources.wake
JakeSiFive: Ah indeed ..x/y is rejected! This is why I don't like using regexes for this stuff! 

hmm whats the right regex now?

`\.\./.*` is what I think your suggesting now? I'll try that out (seems right at a glance to me). We should probably add better file path handling to wake as primitives.

ag-eitilt: Yeah, that's what would fix `claim`.  If you want to unify `..` and `../`, that would be `\.\.(/.*)?`, but I know you aren't as fond of that one.

share/wake/lib/system/sources.wake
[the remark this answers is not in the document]
mmjconolly: `export topic source: String` is an alternative to using `git`. You publish a list of strings of every file path that you would like to have be obtainable by the `sources` fn. I'm not suggesting we use that, but it exists if you want it for your Darcs usage you could create script that writes out a file containing what you want in your `publish source = ...`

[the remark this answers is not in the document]
ag-eitilt: Ah, thanks!  (The other reason I'm not going to use Wake for personal projects is because Haskell already has its own build systems, but I'll keep that in mind if I ever try doing something in a different language.)

PR #864, opened 2022-02-23
add truncate string function

mmjconolly:
Sometimes its nice to shorten a string, for example when printing a log line.

It feels inefficient to explode the string, is there a better way?

### DIFF
--- a/share/wake/lib/core/string.wake
+++ b/share/wake/lib/core/string.wake
@@ -56,6 +56,15 @@ export def explode (string: String): List String =
     def p s = prim "explode"
     p string
 
+# truncate: return only the first N Unicode code points from a given string.
+#
+#   truncate 7 "hello world" = "hello w"
+export def truncate (num: Integer) (string: String) : String =
+    string
+    | explode
+    | take num
+    | cat
+
 # strbase: convert an Integer into a String using a given base.
 #
 # For 2 <= base <= 36, the characters used for the encoding are:


share/wake/lib/core/string.wake
```diff
@@ -56,6 +56,15 @@ export def explode (string: String): List String =
     def p s = prim "explode"
     p string
 
+# truncate: return only the first N Unicode code points from a given string.
+#
+#   truncate 7 "hello world" = "hello w"
```
JakeSiFive, line 61: Needs more formatting, see list.wake for blessed examples. I realize you've matched the surrounding functions but I'd prefer to move things to the list.wake style now. Of particular importance is surrounding examples with triple back-ticks so that editors format it all nicely.

I'd also show what happens for edge cases like -10, the integer being over the string size, and 0

terpstra:
You could use a regular expression;
```
extract `.{1,len}` str
```

terpstra:
Final offer:
```
export def cut len =
  def len0 = if len < 0 then 0 else len
  def regexp = `(.{0,${len0.str.quote}}).*`
  def doit strToCut =
    require out, Nil = extract regexp strToCut
    else unreachable "infallable regular expression failed to match"
    out
  doit
```

That's the best I can come up with.

terpstra:
(If performance is actually needed, you would want/need a prim for this)

mmjconolly:
Mostly just looking to not explode a 5000 line string when I just want the first 50 chars

terpstra:
I'm not actually sure whether or not `re2` is smart enough to terminate the `extract` expression without scanning the whole string. Benchmarks could figure this out!

If `re2` can't do this, I don't think there exists a `prim` in wake that can operate on only part of a String.

terpstra:
Looks like my regexp `cut` is no better than `explode`. re2 let us down. :(

terpstra:
What this PR points out IMO is that we need an `extractPrefix` prim. If we had that, we could've implemented this better. It would also make it possible to write parsers in wake that don't suck.

JakeSiFive:
> What this PR points out IMO is that we need an `extractPrefix` prim. If we had that, we could've implemented this better. It would also make it possible to write parsers in wake that don't suck.

I wonder if we even want to open the pandoras box of people doing string manipulations in wake? I guess its already happening and we need to support it now.

mmjconolly:
Closing for https://github.com/sifive/wake/pull/884

PR #867, opened 2022-03-11
Workaround for building wake (x86_64) on MacOS ARM

ErikParawell-SiFive:
Steps I took to achieve this

1. Run `arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"` to install the x86_64 version of homebrew.
2. If a permissions error [occurs like this](https://github.com/Homebrew/discussions/discussions/600) then run `sudo chown -R $(whoami): /usr/local/share/zsh` to fix the permissions and then repeat step 1.
3. Optionally create a alias for the x86_64 version of Homebrew with `alias brow='arch --x86_64 /usr/local/Homebrew/bin/brew'` I picked brow where the o stands for old [like this person did](https://stackoverflow.com/a/64951025).
4. Run `arch --x86_64 /usr/local/Homebrew/bin/brew install gmp re2 pkgconfig dash` to get x86_64 versions of the required non cask packages.
5. Run `arch --x86_64 /usr/local/Homebrew/bin/brew install --cask macfuse` and reboot
6. Modified the Makefile to target x86_64 and then also figured out where the x86_64 version of re2 lived on my drive and modified the relevant portions of the Makefile to suit that as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ VERSION	:= $(shell if test -f manifest.wake; then sed -n "/publish releaseAs/ s/
 
 CC	:= cc -std=c11
 CXX	:= c++ -std=c++11
-CFLAGS	:= -Wall -O2 -DVERSION=$(VERSION)
+CFLAGS	:= -target x86_64-apple-darwin-macho -Wall -O2 -DVERSION=$(VERSION)
 LDFLAGS	:=
 DESTDIR ?= /usr/local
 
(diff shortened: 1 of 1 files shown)


mmjconolly (changes requested):
This changed the build mechanism for all targets, rather than just mac/arm

ErikParawell-SiFive:
> This changed the build mechanism for all targets, rather than just mac/arm

Should have made this a draft. @JakeSiFive requested a PR with the changes I made with a goal of making them generic.

terpstra:
You don't NEED the x86 intrinsics; wake has pure C fallback implementations for all of them.

Why is `__AVX2__` defined for an ARM mac? That's why it fails to compile natively.

JakeSiFive:
I think this specific PR isn't the solution. We also can't easily get an mac arm machine in CI yet: https://circleci.com/docs/2.0/arm-resources/

I think we should close this for now. We can come back to it later if needed.

PR #868, opened 2022-03-14
Make --failed display with colors

JakeSiFive:
I'm going to break this up into smaller PRs but it got so big I didn't even know exactly how I wanted to carve it up.

This change was all in the name of making --failed output something nicer for users. I found that --failed without color was quite hard to read so I decided I had to add colors. A rabbit hole ensued.

Changes:
* launch_job now creates psuedo terminals instead of pipes to child processes
* the colour and status functions now behave a bit differently, namely the status functions filter out escape sequences when given to ensure a consistently colored output
* the colour functions have been factored out into a separate source file; I'm not super fond of the current interface between colour and status but we can improve it.
* there is now a couple helper stream bufs for manipulating ostreams to filter out or alter escape sequences, as well as a file descriptor stream buf.
* The parsing and re-outputting of terminal escape sequences was a nightmare. Not too happy with the current code but I have some degree of confidence its correct now thanks the the unit tests
* I added a small gtest like unit testing framework to wake, My hope is that having a nice unit testing framework will encourage us to write more unit tests in wake going forward. I know it will encourage me to write more such tests which is probably all that really matters at this point.
* Wake now uses C++14 instead of C++11

This is like well over 1k LOC at this point so I'm going to have to split this up. Additionally each part needs some amount of fine tuning but I think this represents the hard part, and everything else is bike shedding from here.


JakeSiFive:
Likely breakup path

1) Update wake to C++14
2) Add unit testing framework
3) Break apart the existing colour api into its own file
4) Add a color-less version of the new less verbose --failed output
5) Add TermInfoBuf and FdBuf, along with unit tests but don't use them yet
6) Add color to --failed output (and other things in describe) by making job_launch create a psuedo terminal

5 and 6 are going to be kinda large but they should like a 3rd the size of this PR right now.

JakeSiFive:
I should clarify, by "wes is leaving us" I meant for paternity leave, not permanently.

JakeSiFive:
superseded by https://github.com/sifive/wake/pull/1120

PR #870, opened 2022-03-15
Use wakebox by default on linux

mmjconolly:
This makes wakebox the default sandbox on linux, with fallback to fuse-wake on other platforms.

The fuse-wake command line is adjusted such that it has the same `-p` and `-o` arguments as wakebox.
This allows wakebox to become a "drop-in" replacement for fuse-wake.

No attempt is made to use any additional functionality that wakebox provides.


tools/fuse-wake/fuse-wake.cpp
```diff
@@ -34,20 +34,20 @@
 
 int main(int argc, char *argv[])
 {
-	if (argc != 3) {
-		std::cerr << "Syntax: fuse-wake <input-json> <output-json>" << std::endl;
+	if (argc != 5 || 0 != strncmp(argv[1], "-p", 2) || 0 != strncmp(argv[3], "-o", 2)) {
```
JakeSiFive, line 37: Why do we need to add these markers? Just for clarity on the command line?

JakeSiFive (approved):

tools/fuse-wake/fuse-wake.cpp
mmjconolly: yeah otherwise someone might write `wakebox -o outputfile -p inputfile` and they'd be suprised that inputfile was written to

vendor/gopt/gopt-usage.c
```diff
@@ -1,83 +0,0 @@
-/* gopt-usage.c   PUBILC DOMAIN 2015   t.gopt@purposeful.co.uk */
```
terpstra, line 1: Why did you delete this file?

vendor/gopt/gopt.c
```diff
@@ -80,7 +80,7 @@ void add_argument_value(struct option options[], int option_index, char* value)
   int arg_index = options[option_index].count - 1;
 
   if (options[option_index].flags & GOPT_REPEATABLE_VALUE) {
-    if (arg_index < options[option_index].max_args) {
+    if (arg_index < (int)options[option_index].max_args) {
```
terpstra, line 83: You might consider checking if there is a newer "upstream" gopt that already fixed this. If so, we could just bump the library up.

vendor/gopt/gopt-usage.c
mmjconolly: It's unused, and gets caught by the `vendor/gopt/*.c`.
I could make the glob more specific if you prefer

mmjconolly:
> What is the rationale for this change?

Currently we support linux and mac to similar degrees in this repository, because wakebox isn't used at all.
This would be the first step to diverging the support, in future providing linux-only features like a squashfs runner

JakeSiFive:
I'm still sort of wondering what direction we want to commit to with wake. Do we want wake to be usable on all systems or just Linux? I think right now we're kinda of on the fence and just sort of barely committing to supporting multiple platforms. I think we should either take it fully seriously and get CI for the platforms we want to support, or we should just drop support for non-Linux systems.

The pro of multi-platform support in my eyes is eventual outside contributions, and outside users. The pro of only supporting Linux is a great simplification of certain things like this. The con to both is the opportunity cost of not doing the other.

As another note, sysname is sort of breaking with wake's general stance on reproducibility. I don't see any way to avoid it but while we can say "any supported version of the Linux kernel should produce the same results from wake" we can never say that wake behaves the same on Linux and Mac.

JakeSiFive:
Sure yeah, without a functioning wakebox my claim doesn't hold up. I guess I was calling those "unsupported". My claim actually just doesn't hold if preloadRunner is used period.

PR #872, opened 2022-03-17
fuse-wake becomes a subset of wakebox

mmjconolly:
This replaces the other two attempts to "remove" fuse-wake.

This co-locates the wakebox and fuse-wake CLIs in one file, so both are now called `wakebox` regardless of functionality or platform. They both already shared a conditionally-compiled backend via the `run_in_fuse()` entrypoint.

In this PR:
* The fuse-wake CLI is copied into wakebox.cpp, and tools/fuse-wake/* is removed
* all references to `fuse-wake` changed to `wakebox` or removed, but `fuse-waked` daemon is not renamed
* users on non linux systems now call `wakebox` instead of `fuse-wake`, though no internal behavior has changed
* non-linux wakebox has a new help message, so it looks somewhat similar to the linux version

Similar to the other PR versions:
* gopt-usage.c is removed as it's not needed
* a cast was added to gopt (in an part of the code that is not upstreamed)
* "-p" and "-o" args are added to the default makeJSONRunner

This is what the `wakebox -h` looks like on MacOS:
```
Usage: wakebox -p FILE -o FILE                                                   
                                                                                 
NOTE: Restricted functionality due to operating system limitations.              
                                                                                 
Batch options                                                                    
    -p FILE         Json file specifying input parameters.                       
    -o FILE         Json file written containing output results and return code. 
                                                                                 
Other options                                                                    
    -h --help       Print usage
```
and on linux it's unchanged:
```
Usage: wakebox [OPTIONS] [COMMAND...]

Interactive options
    -r --rootfs FILE         Use a squashfs file as the command's view of the root filesystem.
    -t --toolchain FILE      Make a toolchain visible on the command's view of the filesystem.
                             May be specified multiple times.
    -b --bind DIR1:DIR2      Place the directory (or file) at DIR1 within the command's view
                             of the filesystem at location DIR2.
                             May be specified multiple times.
    -x                       Shorthand for '--bind $PWD:$PWD'
    COMMAND                  The command to run.

Batch options
    -p --params FILE         Json file specifying input parameters. Above interactive options
                             will be ignored.
    -o --output-stats FILE   Json file written containing output results and return code.
    -s --force-shell         Run shell instead of command from params file.
                             Implies --allow-interactive.
                             Use 'eval $WAKEBOX_CMD' to run the command from params file.
    -i --allow-interactive   Use default stdin, ignoring the params json file's stdin value.
    -I --isolate-retcode     Don't allow COMMAND's return code to impact wakebox's return code.

Other options
    -h --help                Print usage
```


JakeSiFive (approved):
LGTM, this is a lot cleaner IMO than keeping both binary names around.

ag-eitilt (approved):
I like the reduced complexity!  Now that they're collapsed into a single file, I'd hope some work might (eventually) go into pulling out OS-independent features so that MacOS could use `-i` and such, and the programs would be less likely to diverge even further, but that's definitely a future concern.

PR #873, opened 2022-03-18
Add unit testing

JakeSiFive:
This change adds a basic unit testing framework to C++. It also flips wake to C++14 because this change needed that. I can easily remove that part of the change if we're not satisfied with that yet. We can remove the C++14 stuff from this change fairly easily, I just have to replace `std::quoted` and frankly since std::quoted doesn't escape things control characters its sort of not really the right tool anyway here.

Everything else is good for review. I'm not super satisfied with how I'm interacting with `term` here but there's a few things to keep in mind

I want to test the eventual escape sequence testing of the change I pulled this out of and that's quite tricky since it depends on weather the current process is attached to a terminal or not. This means I need to be able to run the unit tests as if they were in a terminal, but then output the test results without color. I still want to output more than than "passed" because it ensures that tests we expect to run are noted as running so that they can't silently fail to run and still have the tests pass. I could just drop all color from the unit testing framework but frankly I think it looks nice when using it. That personally encourages me to use this framework more which I think is a good thing. I might need to propagate --no-color into the TestLogger however as right now it doesn't do that.


Makefile
```diff
@@ -17,7 +17,7 @@ CORE_CFLAGS  := $(shell pkg-config --silence-errors --cflags sqlite3)	\
 FUSE_LDFLAGS := $(shell pkg-config --silence-errors --libs fuse    || echo -lfuse)
 CORE_LDFLAGS :=	$(shell pkg-config --silence-errors --libs sqlite3 || echo -lsqlite3)	\
 		$(shell pkg-config --silence-errors --libs gmp-6   || echo -lgmp)	\
-		$(shell pkg-config --silence-errors --libs re2     || echo -lre2)	\
+		$(shell pkg-config --silence-errors --libs re2     || echo -lre2) -std=c++14 \
```
terpstra, line 20: I'd like to see the c++14 bump be a separate PR.

src/util/colour.h
```diff
@@ -32,7 +32,9 @@
 #define TERM_DIM        (16*1)
 #define TERM_BRIGHT     (16*2)
 
+bool term_init(bool tty_);
 const char *term_colour(int code);
+const char *term_intensity(int code);
 const char *term_normal();
 
 #endif
```
terpstra, line 40: Please do not commit .fuse/xyz.log?

src/util/colour.h
```diff
@@ -32,7 +32,9 @@
 #define TERM_DIM        (16*1)
 #define TERM_BRIGHT     (16*2)
 
+bool term_init(bool tty_);
```
terpstra, line 35: This was previously defined in src/runtime/status.h. Perhaps remove the other definition.

tools/wake-unit/unit.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+TestRegister::TestRegister(const char* test_name, TestFunc test) {
+  tests__.emplace_back(test_name, test);
+}
+
+int main(int argc, char** argv) {
+  bool no_color = false;
+  if (argc > 2) {
+    std::cerr << "Too many arguments to be valid, --no-color is the only valid argument";
+    return -1;
```
terpstra, line 45: I try not to return negative numbers from main.

tools/wake-unit/unit.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+struct Test {
+  const char* test_name;
+  TestFunc test;
+  Test(const char* test_name_, TestFunc test_) : test_name(test_name_), test(test_) {}
+};
+
+// TODO: sequester this off to its own file and make it static
+// so that only TestRegister and the test runner can access it.
+std::vector<Test> tests__;
```
terpstra, line 35: Why not make this `static` and remove the trailing `__`?
You don't need it to necessarily have its own file; it could still live here while being static.

tools/wake-unit/unit.cpp
```diff
@@ -0,0 +1,114 @@
...
+  }
+  term_init(true);
+  TestLogger logger;
+  std::set<std::string> failed_tests;
+  std::set<std::string> passing_tests;
+  for (auto& test : tests__) {
+    // Set this logjump in case this test fails
+    size_t num_errors = logger.errors.size();
+    logger.test_name = test.test_name;
+    if(setjmp(logger.return_jmp_buffer)) {
```
terpstra, line 62: setjmp? really?
Even exceptions have less code smell than setjmp.
Is this just so you can use `assert()`?
Defining your own `wakeAssert()` seems like it would be a better move.

tools/wake-unit/unit.h
```diff
@@ -0,0 +1,165 @@
...
+  std::jmp_buf return_jmp_buffer;
+  const char* test_name = nullptr;
+
+private:
+  TestStream fail(ErrorMessage& err, bool assert) {
+    return TestStream(&err.user_error, assert ? &return_jmp_buffer : nullptr);
+  }
+
+  // TODO: Make this work with unicode
+  static std::string quote(std::string str) {
```
terpstra, line 73: Would `json_escape` have worked for this?

tools/wake-unit/unit.h
```diff
@@ -0,0 +1,165 @@
...
+
+template <class T>
+struct TestRegisterUnique {
+  static TestRegister unique_register;
+};
+
+#define TEST(name)\
+namespace {struct Test__Unique__ ## name {};}\
+void Test__ ## name (TestLogger&);\
+template<> TestRegister TestRegisterUnique<Test__Unique__ ## name>::unique_register(#name, Test__ ## name);\
```
terpstra, line 164: Why all the trickery with templates here? You could have defined a top-level `static` TestRegister.
Are you trying to somehow handle the case where two tests have the same name?

build.wake
```diff
@@ -161,10 +161,20 @@ export def static _: Result Path Error =
     | getJobOutput
 
 from test_wake import topic wakeTestBinary
+from test_wake import topic wakeUnitTestBinary
 publish wakeTestBinary = defaultWake, Nil
+publish wakeUnitTestBinary = defaultWakeUnit, Nil
```
terpstra, line 166: Why do we need this sort of pub-sub indirection here?
Couldn't `test_wake` just import `buildWakeUnit` from `build_wake`?

terpstra (dismissed):
This PR needs to be split into a c++14 PR and a unit testing PR.

build.wake
JakeSiFive: I was actually just cargo culting the pattern you used for wakeTestBinary until I got to a place where I understood what was happening honestly. Much happier to do it that way.

src/util/colour.h
JakeSiFive: Yeah sorry, that was a mistake. I'm gonna add .fuse/ to gitignore

tools/wake-unit/unit.cpp
JakeSiFive: So other files like wake-unit.cpp need to be able to add tests, it just means I need to add a function here to do that so I can make this static instead. I just forgot to address this TODO. This all used to be in wake-unit.cpp and unit.cpp *is* the aforementioned "other" file :)

tools/wake-unit/unit.cpp
JakeSiFive: Oh yeah. I forgot that 1 is the standard error return code, not sure why I though it was -1 for some reason. I think negative isn't specified by posix

tools/wake-unit/unit.cpp
JakeSiFive: It has to be a setjmp though (more on this later) and if you read gtest's docs it actully becomes clear that they're *also* using setjmp since they say they say destructors won't run after an assert.

The reason that setjmp is needed instead of exceptions is because I need to throw in the destructor to get the `<<` assert interface. I was coppying gtest's interface and I'm like 95% sure that this is what they're doing. You have to cause the test function to exit using code only in the destructor of whatever object is returned by an ASSERT_FOO macro, but destructors aren't allowed to throw and you can't return.

An alternative interface would be to now allow the `ASSERT_FOO(...) << "my message here"` interface and to make it something more like `ASSERT_FOO(..., "my message here")` then I can change the macro to use a return/throw instead of a setjmp from a destructor. But if I want that `<<` syntax I have to setjmp, it was unexpected I'm not gonna lie.

tools/wake-unit/unit.h
JakeSiFive: Yeah I think I'll do that instead, I dislike this function as is.

tools/wake-unit/unit.h
JakeSiFive: This allows two separate files to have the same test name in them because the unique type here is in an anon-namespace. If the same name is used in the same translation unit an error is triggered.

That said I didn't make the test functions themselves static soooo that would be a linker error right now actully. But the goal was to make it so that all errors were local to the translation unit rather than linker errors which this is part of a solution to, I just forgot the easy part.

JakeSiFive:
I can't fix the CI errors here until circle CI's testing code is fixed. Right now it creates a workspace inside the `tests` directory and thus the build_wake package files are not included. This means that when tests.wake goes to find the wake binary it can only find the wake binary made by `make` and not the one made by wake itself. I don't think that `wake-unit` should be bootstrapped by make however and I don't think we should be hard-coding the name of the wake-unit test binary in tests.wake.

I think we should just run the wake tests as `<wake> runTests` from the top level workspace and make runTests global which should fix the issue and allow referencing build_wake from tests. Where is the bash scripting for circleci?

JakeSiFive:
Ok this is ready for review again if someone can take a look.

.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ jobs:
       - run: sudo apt-get update && sudo apt-get install -y build-essential libfuse-dev libsqlite3-dev libgmp-dev libncurses5-dev pkg-config git g++ gcc libre2-dev python3-sphinx
       - run: make tarball
       - run: make test
+      - run: make unittest
```
mmjconolly, line 12: This runs the tests on ubuntu 20.04, is there value in running them within the other containers?

tests/tests.wake
```diff
@@ -77,6 +87,77 @@ export def runTests (cmdline: List String): Result String Error =
...
+        wakeToTest Unit
+
+    require Pass wakeUnitBin =
+        wakeUnitToTest Unit
+
+    def testJob =
+        makeExecPlan ("./{testScript.getPathName.inTestDir}", wakeUnitBin.inTestDir, Nil) (visibleFiles ++ wakeVisible)
+        | setPlanDirectory testDirectory
+        | setPlanLabel "testing: {testName}"
+        | setPlanLocalOnly True # On OS/X you cannot mount fuse within fuse
```
mmjconolly, line 133: `| setPlanLocalOnly True`
If this is a MacOS-only limitation, we could have an `if sysname == "Darwin"`.
If we have unconditional `setPlanLocalOnly` it causes the list of visible files to be ignored, which makes a few of the `require` statements above redundant

tests/tests.wake
```diff
@@ -77,6 +87,77 @@ export def runTests (cmdline: List String): Result String Error =
...
+        | setPlanStdout logNever
+        | setPlanStderr logNever
+        | runJob
+
+    require Pass jobStdout =
+        testJob.getJobStdout
+        | rmap (replace `\r` "")
```
mmjconolly, line 140: A named helper function like `removeCarriageReturns` could improve readability, as its used in a few places

.circleci/config.yml
JakeSiFive: At the moment no. This is intended algorithmic code not code that depends on details of the operating system, just the architecture and glibc really. We might read form files in these unit tests but that's about it. I don't anticipate that changing for a long time if ever. We can make a separate OS verification test suite later (might have to for the blob store alone)

mmjconolly (approved):
Minor comments, seems ok

(no idea about long jumps)

.circleci/config.yml
JakeSiFive: For that I think I might duplicate all the OSs here and then just rebuild wake on them instead of trying to use the package stuff since I don't want to package unit tests.

tests/tests.wake
JakeSiFive: I think I can remove that for this case. I cargo culted it from below.

JakeSiFive:
> Minor comments, seems ok
> 
> (no idea about long jumps)

I'm quite confident that this needs a long jump. gtest does the same thing for the same reason.

PR #875, opened 2022-03-22
circle-ci: newer base OS

mmjconolly:
Context: https://circleci.com/blog/ubuntu-14-16-image-deprecation/
Ubuntu 16.04 image types to disappear
```
The first brownouts of the affected images will take place on March 29, 2022, followed by another 
round of brownouts on April 26, 2022. Any pipeline runs that depend on these images will fail during
the specified brownout periods. The 14.04 and 16.04 images will be permanently removed on 
May 31, 2022, and pipelines using the affected images will fail permanently until a newer image is specified.
```

All of our circle ci configs used a VM with a Ubuntu 16.04 filesystem.
We would then start a docker container with whatever OS filesystem we actually wanted to test.

Since 16.04 is going to be removed, this PR bumps us to 20.04.

With any luck, this shouldn't change how our tests run

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 jobs:
   tarball:
     machine:
-      image: ubuntu-1604:202007-01
+      image: ubuntu-2004:202201-02
 #    resource_class: large # paywalled feature: 4 cores
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


JakeSiFive (approved):
Ok cool, So we're just changing the base image but the container we run in is still whatever we want. Sounds ideal to keep the base image up to date IMO.

PR #869, opened 2022-03-14
CI: remove older linux distros

mmjconolly:
We've carried some older linux distributions in CI for a while, it may be time to remove them.

Debian Jessie ends extended LTS support in few months, and has already ended regular LTS.
Ubuntu 14.04 & 16.04 are also outside LTS, though it's possible to pay for extended support.

If desired, we could add some newer distros to CI in a separate PR

PR #878, opened 2022-04-04
CI: Ubuntu 22.04 LTS

mmjconolly:
Adds Ubuntu 22.04 as a supported build target in CI

Follows on from the removals of older distros in https://github.com/sifive/wake/pull/869

Ubuntu 22.04 isn't actually officially released yet, but this is pointing at the docker image _tag_ that will become the official image in a few weeks (April 23). It's unlikely that much would change between now and then that would impact us

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,26 @@ jobs:
           paths:
             - release/ubuntu_18_04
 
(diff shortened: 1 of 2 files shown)


richardxia (approved):
LGTM, assuming CI passes.

PR #879, opened 2022-04-05
Remove preload runner

mmjconolly:
The preload runner tries to provide an isolated workspace area for wake jobs.

It does so by using a dynamically loaded library that intercepts other library calls for basic operations like read()/write().

This is not-ideal for two reasons:
* Not all executables that you'd want to run via wake are dynamically linked, so the process becomes un-sandboxed
* The absolute paths exposed to the running process have extra path components that do not resolve once the job has finished, which can break downstream jobs

The wakebox runner is able to handle those two cases properly, it should be used instead.

A downside of this change is that users without the ability to make a fuse mountpoint will be unable to use wake, but that should be an unusual situation


felixonmars:
This unfortunately breaks the Arch Linux packaging of wake because we don't have FUSE available in the build chroot. Do you have any suggestions about how we could properly build wake after this change?

PR #880, opened 2022-04-06
Add --timeline to help text

mmjconolly:


JakeSiFive (approved):
This is no where near high on my list but re-working wake options to use a proper options library would help resolve these sorts of issues from occurring in the first place if help text was required/localized.

PR #881, opened 2022-04-06
Bumping to 0.27-alpha

JakeSiFive:
After this lands I'll tag this as v0.27.0-alpha and the release will be complete.


JakeSiFive:
I've already committed to the v0.26 branch and added the v0.26.0 tag to a commit in that branch. This commit will bump the alpha version to v0.27.0-alpha in the master branch.

PR #883, opened 2022-04-07
Update release artifact fetch script for new/removed OS

mmjconolly:
also:
* removed some commented out parts from a yaml file that shouldn't be there
* added a top-level symlink to where the documentation lives, cause I always forget


scripts/fetch
```diff
@@ -13,7 +13,5 @@ curl -L -o wake-static_${version}.tar.xz               ${url}/alpine/wake-static
 curl -L -o centos-7-6-wake-${version}-1.x86_64.rpm     ${url}/centos_7_6/wake-${version}-1.x86_64.rpm
```
JakeSiFive, line 13: I thought I added this in another PR this morning but I guess not...whoops.

JakeSiFive (approved):

PR #884, opened 2022-04-08
substring prim

mmjconolly:
From discussion in https://github.com/sifive/wake/pull/864, I thought I'd try my hand at adding a `prim` for substrings.

It operates on unicode code points (like `explode`) rather than raw bytes.
The prim itself returns `Result String String` as it wasn't obvious how to create an `Error` type from C++ and I saw other prim functions not returning an `Error`, so let me know if that's the wrong direction.
In the wake stdlib, there is a wrapper function that adapts it to `Result String Error`


src/runtime/string.cpp
```diff
@@ -672,8 +672,70 @@ static PRIMFN(prim_shell_str) {
...
+  INTEGER_MPZ(len_arg, 2);
+
+  int start_pos = mpz_get_si(pos_arg);
+  int length = mpz_get_si(len_arg);
+  if (start_pos < 0 || length < 1) {
+    RETURN(allocate_result(runtime.heap, false, "substr: invalid position or length argument"));
+  }
+
+  std::string new_substr;
+  new_substr.reserve(length);
```
JakeSiFive, line 707: So you call `reserve` here but then later on all paths call `reserve` again. You only want to call it once. I'd just reserve the string space at the end with allocate_result instead. The general rule is that any path through a primitive should call reserve at most once.

reserve calls don't stack so you have to actually allocate between reserves or else the reserves are kind of pointless.

mmjconolly: I think I only call `reserve` once for the string, the other reserve() calls are for the runtime heap

JakeSiFive:
I'm pretty happy with this. I don't have a ton of experience with the sharp edges of using unicode strings this way but since you're iterating over runes I figure its all good.

mmjconolly:
I mostly wrote this as a learning exercise, but if it's useful I'm happy to continue with it

JakeSiFive:
I think its useful and the code looks 95% good to me. Seems useful. I agree explode is slow.

src/runtime/string.cpp
JakeSiFive: Oh I'm a numskull, standard library's string reserve here I see.

terpstra:
I generally find subvector operations to be more useful with a `[start, end)` indexing approach than `[start, length)`. Also, I don't think returning a `Result` is appropriate. Indexing too far left or right should just be capped in the logical way. In other words, start < 0, should be start=0. End > length should be end = length. Start > End should be start = end.

terpstra:
The C++ part looks good.

src/runtime/string.cpp
```diff
@@ -672,8 +672,70 @@ static PRIMFN(prim_shell_str) {
...
+  // number of variable width runes seen
+  int rune_count = 0;
+  // width of the rune
+  int rune_num_bytes = 0;
+
+  for (const char *ptr = input_str->c_str(); *ptr; ptr += rune_num_bytes) {
+    uint32_t rune;
+    rune_num_bytes = pop_utf8(&rune, ptr);
+    if (rune_num_bytes < 1) {
+      rune_num_bytes = 1;
```
terpstra, line 718: You probably don't want to push_utf8 here, as that would change the encoding.
Probably, you should just copy the byte over.

terpstra: In this scenario, an actual `Fail` might be appropriate. I'm not keen on making substr return `Result`, though, as you often want to use it in a larger expression, and needing a `require` adds a lot of friction, there.

src/runtime/string.cpp
```diff
@@ -672,8 +672,70 @@ static PRIMFN(prim_shell_str) {
...
+
+  std::string new_substr;
+  new_substr.reserve(length);
+
+  // number of variable width runes seen
+  int rune_count = 0;
+  // width of the rune
+  int rune_num_bytes = 0;
+
+  for (const char *ptr = input_str->c_str(); *ptr; ptr += rune_num_bytes) {
```
terpstra, line 714: This loop terminates when it hits a null byte. Null bytes are legal values in wake.

src/runtime/string.cpp
```diff
@@ -672,8 +672,70 @@ static PRIMFN(prim_shell_str) {
...
+
+// This operates on utf8 runes rather than bytes, to match prim_explode.
+static PRIMFN(prim_substr) {
+  EXPECT(3);
+  STRING(input_str, 0);
+  INTEGER_MPZ(pos_arg, 1);
+  INTEGER_MPZ(len_arg, 2);
+
+  int start_pos = mpz_get_si(pos_arg);
+  int length = mpz_get_si(len_arg);
```
terpstra, line 701: Erg. `int` is only 32 bits large. Strings in wake can be larger than 4GiB. You should use either `long` or `size_t`.

terpstra: Hmm. On windows, even `long` is 32 bits.

terpstra: GMP `mpz_get_si` returns `long`, so I guess that's the best you can hope for.
However, you definitely also need to check `mpz_fits_slong_p` in case the integer won't fit.

terpstra:
Come to think of it, there is a significantly faster way to implement this method. Just don't count bytes that start with `10xx xxxx`. A loop like that is also simple enough to vectorize well. One pass finds the array bounds, and then you construct the resulting string directly from the original input, thus also eliminating all intermediate copies.

terpstra:
If you have this method, you probably also need/want a `strlen` function that counts runes instead of bytes.

Also, ugh, `substr` as defined is potentially broken in that it can still cut the string in the middle of a character. In unicode, multiple codepoints can be combined to form a single character.

terpstra:
I've pushed my amendment to this function, which redefines it along the lines I pointed out earlier. You've really managed to nerd snipe me here, because there's an even faster way to do this, using broadword techniques... and now I am too far down the dark path to turn back.

share/wake/lib/core/string.wake
```diff
@@ -56,6 +56,27 @@ export def explode (string: String): List String =
...
+# Examples:
+# ```
+#   substr "ABCDE" 0 1    = "A"
+#   substr "ABCDE" 1 4    = "BCD"
+#   substr "ABCDE" 1 30   = "BCDE"
+#   substr "ABCDE" 50 3   = ""
+#   substr "ABCDE" (-1) 3 = "ABC"
+#   substr "ABCDE" 1 0    = ""
+# ```
+export def substr (string: String) (start: Integer) (end: Integer): String =
```
terpstra, line 76: Actually, do we want the string to be the last argument? That might make it more convenient to use in a String-processing pipeline.

src/runtime/string.cpp
terpstra: I made the function have well defined behaviour in this case, by carefully defining a codepoint.

src/runtime/string.cpp
terpstra: Fixed by using `+ size()` in an end pointer.

src/runtime/string.cpp
terpstra: Fixed by adding explicit bounds checks before conversion to an unsigned long.

terpstra:
Force-pushed after rebasing against master.

terpstra:
I ran a quick fuzz test to confirm that `substr x s e = explode x | take e | drop s | cat`.
Seems to be "correct".

share/wake/lib/core/string.wake
terpstra: Done.

mmjconolly:
@JakeSiFive says that we're not looking to merge this

JakeSiFive:
I'm happy to close this but for the record I actually said "In this light we're **leaning** towards not landing the substr PR at all" but frankly my lean has turned to desire at this point.

mmjconolly:
ah sorry, misinterpreted

PR #886, opened 2022-04-20
use isolate-retcode when running Jobs

mmjconolly:
The fuse-wake executable would default to not returning the exit code of the process as its own.
Wakebox does by default, but has a cli flag to use the other behavior.

When we removed the fuse-wake executable, it was missed to add the `-I` to isolate the return code so that wakelang runners would only observe the jobs return code via the output json


richardxia (approved):
LGTM

mmjconolly:
This probably needs to a 0.26.1 release

JakeSiFive (approved):

ag-eitilt (approved):

PR #889, opened 2022-05-03
Add --clean flag

JakeSiFive:
This change adds the ability to clean all tracked outputs of jobs (except for source and claim jobs as determined by their command line).

Not all files created during a wake build are tracked by wake so this doesn't clean up *everything* but it should under normal uses clean up most things.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -83,6 +83,9 @@ struct Database::detail {
   sqlite3_stmt *get_edges;
   sqlite3_stmt *get_job_visualization;
   sqlite3_stmt *get_file_access;
+  sqlite3_stmt *get_output_files;
+  sqlite3_stmt *remove_output_files;
+  sqlite3_stmt *remove_all_jobs;
 
   long run_id;
   detail(bool debugdb_)
@@ -433,6 +436,23 @@ std::string Database::open(bool wait, bool memory, bool tty) {
       " from filetree"
       " where access != 1"
       " order by file_id, access desc, job_id";
+  const char *sql_get_output_files =
+    "select f.path"
+    " from filetree ft join files f on f.file_id=ft.file_id join jobs j on ft.job_id=j.job_id"
+    " where ft.access = 2"
+    " and substr(cast(j.commandline as varchar), 1, 8) != '<source>'"
+    " and substr(cast(j.commandline as varchar), 1, 7) != '<claim>'";
+  const char *sql_remove_output_files =
+    "delete from files"
+    " where file_id in ("
+    "   select f.file_id"
+    "   from filetree ft join files f on f.file_id=ft.file_id join jobs j on ft.job_id=j.job_id"
+    "   where ft.access = 2"
+    "   and substr(cast(j.commandline as varchar), 1, 8) != '<source>'"
+    "   and substr(cast(j.commandline as varchar), 1, 7) != '<claim>'"
+    " )";
+  const char *sql_remove_all_jobs =
+    "delete from jobs";
 
 #define PREPARE(sql, member)                                                                     \
   ret = sqlite3_prepare_v2(imp->db, sql, -1, &imp->member, 0);                                   \
@@ -481,6 +501,9 @@ std::string Database::open(bool wait, bool memory, bool tty) {
   PREPARE(sql_get_edges, get_edges);
   PREPARE(sql_get_job_visualization, get_job_visualization);
   PREPARE(sql_get_file_access, get_file_access);
+  PREPARE(sql_get_output_files, get_output_files);
+  PREPARE(sql_remove_output_files, remove_output_files);
+  PREPARE(sql_remove_all_jobs, remove_all_jobs);
 
   return "";
 }
@@ -538,6 +561,9 @@ void Database::close() {
   FINALIZE(get_edges);
   FINALIZE(get_job_visualization);
   FINALIZE(get_file_access);
+  FINALIZE(get_output_files);
+  FINALIZE(remove_output_files);
+  FINALIZE(remove_all_jobs);
 
   close_db(imp.get());
 }
@@ -939,6 +965,16 @@ void Database::finish_job(long job, const std::string &inputs, const std::string
   if (fail) exit(1);
 }
 
+void Database::remove_outputs() {
+  const char *why = "Could not remove outputs";
+  single_step (why, imp->remove_output_files, imp->debugdb);
+}
+
+void Database::clear_jobs() {
+  const char *why = "Could not clear jobs";
+  single_step (why, imp->remove_all_jobs, imp->debugdb);
+}
+
 void Database::tag_job(long job, const std::string &uri, const std::string &content) {
   const char *why = "Could not tag a job";
   bind_integer(why, imp->tag_job, 1, job);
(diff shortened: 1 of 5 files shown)


JakeSiFive:
re: Wes on if nuking wake.db is useful....actually at first I wasn't going to delete it and then I sort of just decided to delete it anyway since the goal of a clean is to nuke as much as we can safely nuke I thought. I'm happy to go back since you pointed out that the `stats` table is still useful.

JakeSiFive:
Another thing I haven't thought about, how does this interact with incremental builds invoked by wake?

mmjconolly:
> Another thing I haven't thought about, how does this interact with incremental builds invoked by wake?

I think they will all be obliterated, given that it's a 'clean' function

tools/wake/main.cpp
```diff
@@ -381,6 +383,31 @@ int main(int argc, char **argv) {
...
+
+    // Delete all the files
+    for (const auto& file : files) {
+      if (unlink(file.c_str()) == -1) {
+        if (errno == EISDIR) continue;
+        if (errno == ENOENT) continue;
+        std::cerr << "error: unlink(" << file << "): " << strerror(errno) << std::endl;
+        return 1;
+      }
+    }
```
mmjconolly, line 400: I'm wondering if we should (after file deletes) loop through the directories that wake created and delete any that are empty, leaving the non-empty ones as an obvious 'here is stuff here' indicator

JakeSiFive:
> I think they will all be obliterated, given that it's a 'clean' function

That's what I *want* to happen but I'm not sure if we have incremental builds setup to be outputs. Wes showed me how that worked at one point but I've forgotten the details. The incremental inputs would have to be added as outputs in order for this to blow them away as written...I *think* that's how I would have implemented incremental builds but I could be wrong, I don't know.

tools/wake/main.cpp
JakeSiFive: That does seem more tidy

JakeSiFive:
I think I'm in favor of `git clean` instead honestly. Does anyone see an issue with people using `git clean`?

mwachs5:
> Does anyone see an issue with people using git clean?

I don't usually expect "make clean" to delete the very files I may have been working on and forgotten to commit to git, perhaps being the reason I was confused and running `wake clean` in the first place trying to get stuff to work...

mmjconolly:
Could be good to add 'clean' to `wake --help`

mmjconolly:
I think we have another case to consider: `setPlanFnOutputs`

It seems like if a job writes some files and then filters the outputs, then they're remaining on the filesystem after a `--clean`.

```
$ cat do.wake 
package top
from wake import _
global def do _ =
    "touch foo; touch baa"
    | makePlan "" Nil
    | setPlanFnOutputs (filter (matches `foo` _))
    | runJob
    | getJobOutputs

$ ls foo baa
ls: cannot access 'foo': No such file or directory
ls: cannot access 'baa': No such file or directory
$ wake do
Pass (Path "foo", Nil)
$ ls foo baa
baa  foo
$ wake --clean
$ ls foo baa
ls: cannot access 'foo': No such file or directory
baa
```

JakeSiFive:
> I think we have another case to consider: `setPlanFnOutputs`
> 
> If a job writes files and then filters the outputs, then those filtered-out are remaining on the filesystem after a `--clean`.
> 
> ```
> $ cat do.wake 
> package top
> from wake import _
> global def do _ =
>     "touch foo; touch baa"
>     | makePlan "" Nil
>     | setPlanFnOutputs (filter (matches `foo` _))
>     | runJob
>     | getJobOutputs
> 
> $ ls foo baa
> ls: cannot access 'foo': No such file or directory
> ls: cannot access 'baa': No such file or directory
> $ wake do
> Pass (Path "foo", Nil)
> $ ls foo baa
> baa  foo
> $ wake --clean
> $ ls foo baa
> ls: cannot access 'foo': No such file or directory
> baa
> ```
> 
> The usecases for I can think of are to avoid hashing large output files that we happen to know we don't care about, or to prevent visibility in a downstream job (though we could filter after getJobOutputs instead), or because multiple jobs produce the same file in a way that isn't of consequence (though I don't like that one)
> 
> Perhaps we could keep track of these files in the database too for the purposes of `--clean`? Though perhaps without hashing them?

I'm wondering if we should auto-delete such files after each job finishes? Or do we keep such files around for incremental builds?

mmjconolly:
> > I think we have another case to consider: `setPlanFnOutputs`
> > If a job writes files and then filters the outputs, then those filtered-out are remaining on the filesystem after a `--clean`.
> > ```
> > $ cat do.wake 
> > package top
> > from wake import _
> > global def do _ =
> >     "touch foo; touch baa"
> >     | makePlan "" Nil
> >     | setPlanFnOutputs (filter (matches `foo` _))
> >     | runJob
> >     | getJobOutputs
> > 
> > $ ls foo baa
> > ls: cannot access 'foo': No such file or directory
> > ls: cannot access 'baa': No such file or directory
> > $ wake do
> > Pass (Path "foo", Nil)
> > $ ls foo baa
> > baa  foo
> > $ wake --clean
> > $ ls foo baa
> > ls: cannot access 'foo': No such file or directory
> > baa
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The usecases for I can think of are to avoid hashing large output files that we happen to know we don't care about, or to prevent visibility in a downstream job (though we could filter after getJobOutputs instead), or because multiple jobs produce the same file in a way that isn't of consequence (though I don't like that one)
> > Perhaps we could keep track of these files in the database too for the purposes of `--clean`? Though perhaps without hashing them?
> 
> I'm wondering if we should auto-delete such files after each job finishes? Or do we keep such files around for incremental builds?

I'm not opposed to the auto-delete, though also I don't know if there was a reason why we didn't initially do it

mmjconolly:
@terpstra Do you know of any problems with auto-deleting filtered-out files when a job finishes?

JakeSiFive:
Let's punt on this issue for now so we can get something that works well but perhaps isn't quite perfect landed. I think not cleaning up files removed by setPlanFnOutputs can be almost safely ignored.

src/runtime/database.cpp
```diff
@@ -1091,6 +1126,20 @@ static std::vector<JobReflection> find_all(const Database *db, sqlite3_stmt *que
   return out;
 }
 
+std::vector<std::string> Database::get_outputs() const {
```
terpstra, line 1129: Maybe this should have been get_and_remove_outputs? So that the remove_output_files and get_output_files could be in the same begin/end_txn

test.cc
```diff
@@ -0,0 +1,9 @@
+#define _XOPEN_SOURCE 700
```
terpstra, line 1: Delete?

tools/fuse-waked/fuse-waked.cpp
```diff
@@ -20,7 +20,7 @@
 
 // Open Group Base Specifications Issue 7
 #define _XOPEN_SOURCE 700
-#define FUSE_USE_VERSION 26
+#define FUSE_USE_VERSION 39
```
terpstra, line 23: Maybe belongs in another PR?

tools/wake/main.cpp
```diff
@@ -381,6 +388,64 @@ int main(int argc, char **argv) {
...
+  // If the user asked us to clean the local build, do so.
+  if (clean) {
+    // Find all the file we need to delete.
+    auto files = db.get_outputs();
+
+    // Clean up the database of unwanted info. Jobs must
+    // be cleared before outputs are removed to avoid foreign key
+    // constraint issues.
+    db.clear_jobs();
+    db.remove_outputs();
```
terpstra, line 414: All three calls into one txn, please.

tools/wake/main.cpp
```diff
@@ -381,6 +388,64 @@ int main(int argc, char **argv) {
...
+          continue;
+        }
+        if (errno == ENOENT) continue;
+        std::cerr << "error: unlink(" << file << "): " << strerror(errno) << std::endl;
+        return 1;
+      }
+    }
+
+    // Sort them so that child directories come before parent directories
+    std::sort(dirs_to_remove.begin(), dirs_to_remove.end(), [&](const std::string& a, const std::string& b) -> bool {
```
terpstra, line 433: Just do your files erasure in the same pass post-sort.

terpstra (approved):
Easy fixes, but LGTM.

PR #893, opened 2022-05-13
Fix isNone

colinschmidt:
now this function does what is says on the tin


ag-eitilt (approved):
Good catch!  We should probably look into more extensive testing of the stdlib at some point to catch things like this.

JakeSiFive (approved):
Thank you!

PR #894, opened 2022-05-17
system job labels

mmjconolly:
This adds some `Plan` `Label`s to some of the `Job`s in `share/wake/lib/system`.
It's somewhat useful when making manual SQL queries on `wake.db`

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -37,6 +37,7 @@ target writeImp inputs mode path content =
         makeRunner "write" (\_ Pass 0.0) pre post virtualRunner
     makeExecPlan ("<write>", str mode, path, Nil) inputs
     | setPlanOnce        False
+    | setPlanLabel "write: {path}"
     | setPlanEnvironment Nil
     | runJobWith writeRunner
     | getJobOutput
@@ -68,6 +69,7 @@ export def installAs (dest: String) (file: Path): Result Path Error =
         sdest, Nil
     makeExecPlan cmd inputs
     | setPlanEnvironment Nil
+    | setPlanLabel "installAs: {dest}"
     | setPlanLocalOnly True
     | setPlanFnOutputs foutputs
     | runJob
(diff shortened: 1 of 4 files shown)


share/wake/lib/system/io.wake
```diff
@@ -68,6 +69,7 @@ export def installAs (dest: String) (file: Path): Result Path Error =
         sdest, Nil
     makeExecPlan cmd inputs
     | setPlanEnvironment Nil
+    | setPlanLabel "installAs: {dest}"
```
ag-eitilt, line 72: This one might want to record the source file as well.  Also, for `targz` and `claimFileAsPathIn`, you've used slightly more readable "job names", and I'd expect this to be `install-as:` to follow their pattern.

share/wake/lib/system/path.wake
```diff
@@ -124,7 +124,7 @@ target hashcode (f: String): String =
   if reuse !=* "" then reuse else
     def hashPlan cmd  =
       Plan "" cmd Nil Nil "." "" logNever logError logDebug ReRun True Nil (\_ True) hashUsage id id
-    def job = hashPlan ("<hash>", f, Nil) | runJobWith localRunner
+    def job = hashPlan ("<hash>", f, Nil) | setPlanLabel "hash: {f}" | runJobWith localRunner
```
ag-eitilt, line 127: I don't know the full logic; will these start showing up in logs/stdout (maybe with a `-v`, or a `--no-tty`)?  The others are relevant enough to the user that I wouldn't mind having them be shown -- though we want to be careful to not wind up with huge CI logs -- but hashing in particular seems like an internal job that we would kind of prefer to keep hidden.

share/wake/lib/system/sources.wake
```diff
@@ -155,5 +157,6 @@ export def claimFileAsPathIn (outputDirectory: Path) (existingFile: String) (des
         | setPlanLocalOnly   True
         | setPlanPersistence Once
         | setPlanFnOutputs   (\_ desiredWorkspacePath, Nil)
+        | setPlanLabel "claim-in: {existingFile}"
```
ag-eitilt, line 160: As a mirror, this might want to record the destination.

ag-eitilt (approved):
Looks good, conditional on the response to the second comment.

terpstra (commented):
test

terpstra (approved):

JakeSiFive:
Can we make this use Ashley's new `@!` feature and land this? I think this is a good change.

mmjconolly:
What does `@!` do?

ag-eitilt:
It's like `@here`, adding the line number: `@! ==~ "wake/build-graph.wake:226"`.

PR #895, opened 2022-05-31
tree: use the output type directly in initializing `tfoldmap`

ag-eitilt:
The `tfoldmap` parallel fold function previously took a seed value of the type used in the tree, however there are some cases where that is not expressive enough; for example, it was impossible to implement `treeToList` using `tfoldmap` since there was no way to get `Nil` for use as a seed value without either a `map Some` indirection step or by abusing some value which was assumed to never occur in the tree.  Instead, it seems better for that function were to take a seed value of the output type, and the resulting algorithm is even a strict superset of the previous, as seen in the compatibility implementation.  (I'll also note that the variable name implies that the latter was what was originally intended.)

The primary change of this PR is simply moving the application of the mapped function as below.  Since I was already working on the function, though, I updated it to follow the current style/documentation guidelines.
```patch
-    Tip = g a
-    Bin _ l x r = f (helper a l) (helper x r)
+    Tip = a
+    Bin _ l x r = f (helper a l) (helper (g x) r)
```


terpstra (approved):
Seems reasonable to me!

PR #901, opened 2022-06-15
Add c++ 14 build variants

V-FEXrt:


Makefile
```diff
@@ -3,7 +3,7 @@
 VERSION	:= $(shell if test -f manifest.wake; then sed -n "/publish releaseAs/ s/^[^']*'\([^']*\)'.*/\1/p" manifest.wake; else git describe --tags --dirty; fi)
 
 CC	:= cc -std=c11
-CXX	:= c++ -std=c++11
+CXX	:= c++ -std=c++14
```
JakeSiFive, line 6: $(shell pkg-config --silence-errors --cflags re2) below is adding c++11 back and overiding the C++14 flag here

vendor/re2.wake
```diff
@@ -18,7 +18,7 @@ from wake import _
 from gcc_wake import _
```
JakeSiFive, line 18: You'll need to add a thing to remove `-std=c++11` from pkg-config flags in here as well

JakeSiFive (commented):
There's a subtle issue, we should use a C++14 feature in wake with this change so that we know that it works

Makefile
```diff
@@ -3,7 +3,8 @@
 VERSION	:= $(shell if test -f manifest.wake; then sed -n "/publish releaseAs/ s/^[^']*'\([^']*\)'.*/\1/p" manifest.wake; else git describe --tags --dirty; fi)
 
 CC	:= cc -std=c11
-CXX	:= c++ -std=c++11
+CXX	:= c++
+CXX_VERSION := -std=c++14
 CFLAGS	:= -Wall -O2 -DVERSION=$(VERSION)
```
JakeSiFive, line 8: nit: I think just adding the CXX_VERSION onto the end of CFLAGS would be better

JakeSiFive (commented):
The reason for the failure is because centos 7 is *old old old*, we can just delete it from the CI list

Makefile
V-FEXrt: Happy to move it, but CORE_CFLAGS is listed after CFLAGS in the build rule so it would need to be pushed back as well. Is that ok?

V-FEXrt: Also CFLAGS is used for the C build rules

JakeSiFive: Ah I see. I guess a name like CXX_FLAGS would be better then in case we need more overrides in the future. All good as is.

JakeSiFive:
I'm happy with this but because we're removing centos 7.6 for lack of C++14 support I'd like to wait and get approval from someone else as well

mmjconolly:
I'm ok with the centos 7.6 removal.
At some point in the future we should replace it with a RHEL9-binary-compatible distro, like rocky9 or almalinux9

JakeSiFive:
can we replace with rocky9 now?

mmjconolly:
rocky hasn't updated to 9 yet but almalinux has a 9, which is just as good for what we're doing
https://hub.docker.com/layers/almalinux/library/almalinux/9.0/images/sha256-8be8c8b3b2262149e433de954fa349a1cbcad3ddaf02bf33327e53cdbdfd61b9

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -59,6 +60,9 @@ publish compileC = makeCompileC "native-c11-static"    (which "cc")  (c11Flags +
 publish compileC = makeCompileC "native-cpp11-debug"   (which "c++") (cpp11Flags ++ debugCFlags)
 publish compileC = makeCompileC "native-cpp11-release" (which "c++") (cpp11Flags ++ releaseCFlags)
 publish compileC = makeCompileC "native-cpp11-static"  (which "c++") (cpp11Flags ++ staticCFlags)
```
mmjconolly, line 62: Is it possible to remove all the cpp11 related code?

JakeSiFive (approved):
LGTM. Once we have rocky9 we can add that but for now this is the best we can do I think.

JakeSiFive:
Ah crap we didn't know about that. Let us discsuss.

PR #902, opened 2022-06-15
Add basic clang-format 'format all' target

V-FEXrt:


Makefile
```diff
@@ -51,6 +51,12 @@ wake.db:	bin/wake bin/wakebox lib/wake/fuse-waked lib/wake/shim-wake
 install:	all
 	$(WAKE_ENV) ./bin/wake install $(DESTDIR)
 
+# This assumes clang is available on the PATH
+# try: module load clang/12.0.1.1
```
mmjconolly, line 55: module load isn't available for open source users

V-FEXrt:
I wouldn't call this done nor a good final solution but it meets the requirements of a 'one time' format.

Ideally this would be a wake target but I was having troubles getting the target to run in a environment with everything availalble.

Thought this would be a good place to kick off discussion then flesh out from there.

mmjconolly:
> Ideally this would be a wake target

I think it make sense to have this outside of wake, as we usually don't allow modifying of input/source-control files via wake

mmjconolly:
It's probably worth spot-checking some of the formatted files to see if they look ok with the provided style

V-FEXrt:
Agreed! I wanted to do that in an isolated PR/branch.

See https://github.com/sifive/wake/compare/format-repo?expand=1 which was ran using `make formatAll` from this change

JakeSiFive (approved):

V-FEXrt:
Okay, I pinned down the format issue when there were no files to change. PTAL

JakeSiFive:
don't hate me but I *just now* thought of a better way to do this: https://unix.stackexchange.com/questions/15308/how-to-use-find-command-to-search-for-multiple-extensions

I'm so sorry

V-FEXrt:
Yeah, that doesn't work for format, but it does for formatAll. I went ahead and updated formatAll

JakeSiFive:
all good, please land!

V-FEXrt:
Part of my comment from the last comment got lost somehow. Fixed the comment, waiting for CI then will land!

mmjconolly:
I'm wondering if we want to avoid formatting code in the `vendor/` directory

V-FEXrt:
Seems like a reasonable idea, would just need to add some more filters to the files we match against.

All the h/cpp files under vendor/ are all external deps right? Probably a good idea to keep the close to the original to make future integration/updates easier

mmjconolly:
Yeah they're external that've been copied in

V-FEXrt:
Done! PTAL

Makefile
```diff
@@ -51,6 +51,26 @@ wake.db:	bin/wake bin/wakebox lib/wake/fuse-waked lib/wake/shim-wake
...
+	clang-format -i --style=file $(shell find . -not \( -path ./vendor -prune \) -type f \( -name "*.h" -o -name "*.cpp" -o -name "*.c" \))
+
+# Formats all changed or staged .h or .cpp files
+# It assumes clang is available on the PATH and will fail otherwise
+format:
+# Conditionally run clang-format based on the staged and changed files
+# || true is added after each if expression since they resolve with false when false
+# and we don't want make to report that as an error
+	@CHANGED_FILES=$$(git diff --name-only | grep '.h\|.cpp\|.c' | grep -v 'vendor/') && \
+	if [ ! -z $$CHANGED_FILES ]; then \
```
mmjconolly, line 66: I think `! -z` is the same as `-n` which is also the no-option
```suggestion
	if [ $$CHANGED_FILES ]; then \
```

mmjconolly (approved):
I wont claim to be able to read the .clang-format file, but glancing at the Makefile looks good to me

V-FEXrt:
> I wont claim to be able to read the .clang-format file

FWIW, it is 99.9% generated by the tool itself. The only thing I changed was to make the 80 character with 100

Makefile
V-FEXrt: Ah right!

PR #903, opened 2022-06-16
Format entire repo

V-FEXrt:
Runs clang-format on the entire repo.

For those interested in maintaining the git blame history  `git config --global blame.ignoreRevsFile .git-blame-ignore-revs` should be ran locally.

This PR shouldn't be squashed when merged in order to maintain the appropriate git hash

### DIFF
--- a/src/dst/expr.cpp
+++ b/src/dst/expr.cpp
@@ -30,23 +30,23 @@ CPPFile expr_h("src/dst/expr.h");
 static CPPFile cppFile(__FILE__);
 
 Expr::~Expr() {}
-const TypeDescriptor Prim ::type("Prim");
-const TypeDescriptor App ::type("App");
-const TypeDescriptor Lambda ::type("Lambda");
-const TypeDescriptor VarRef ::type("VarRef");
-const TypeDescriptor Literal ::type("Literal");
+const TypeDescriptor Prim::type("Prim");
+const TypeDescriptor App::type("App");
+const TypeDescriptor Lambda::type("Lambda");
+const TypeDescriptor VarRef::type("VarRef");
+const TypeDescriptor Literal::type("Literal");
 const TypeDescriptor DefBinding::type("DefBinding");
-const TypeDescriptor Get ::type("Get");
-const TypeDescriptor Construct ::type("Construct");
-const TypeDescriptor Destruct ::type("Destruct");
+const TypeDescriptor Get::type("Get");
+const TypeDescriptor Construct::type("Construct");
+const TypeDescriptor Destruct::type("Destruct");
 // these are removed by bind
-const TypeDescriptor Subscribe ::type("Subscribe");
-const TypeDescriptor Ascribe ::type("Ascribe");
-const TypeDescriptor Match ::type("Match");
-const TypeDescriptor DefMap ::type("DefMap");
+const TypeDescriptor Subscribe::type("Subscribe");
+const TypeDescriptor Ascribe::type("Ascribe");
+const TypeDescriptor Match::type("Match");
+const TypeDescriptor DefMap::type("DefMap");
 // these are just useful for dumping json ast
-const TypeDescriptor VarDef ::type("VarDef");
-const TypeDescriptor VarArg ::type("VarArg");
+const TypeDescriptor VarDef::type("VarDef");
+const TypeDescriptor VarArg::type("VarArg");
 
 Top::Top() : packages(), globals(), def_package(nullptr) {
   Package *builtin = new Package();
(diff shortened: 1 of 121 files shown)


V-FEXrt:
Ah, looks like there are some code files relying on import order for subsequent/indirect imports. I'll work through those and get them resolved

src/dst/expr.cpp
```diff
@@ -19,58 +19,64 @@
...
-const TypeDescriptor App       ::type("App");
-const TypeDescriptor Lambda    ::type("Lambda");
-const TypeDescriptor VarRef    ::type("VarRef");
-const TypeDescriptor Literal   ::type("Literal");
+Expr::~Expr() {}
+const TypeDescriptor Prim ::type("Prim");
+const TypeDescriptor App ::type("App");
+const TypeDescriptor Lambda ::type("Lambda");
+const TypeDescriptor VarRef ::type("VarRef");
+const TypeDescriptor Literal ::type("Literal");
```
JakeSiFive, line 37: Am I crazy or would clang-format not add a space if we removed the space here?

tools/wake/main.cpp
```diff
@@ -68,178 +68,171 @@
...
-    { 0,   "tag",                   GOPT_ARGUMENT_REQUIRED  },
-    { 0,   "export-api",            GOPT_ARGUMENT_REQUIRED  },
-    { 0,   "stdout",                GOPT_ARGUMENT_REQUIRED  },
-    { 0,   "stderr",                GOPT_ARGUMENT_REQUIRED  },
-    { 0,   "fd:3",                  GOPT_ARGUMENT_REQUIRED  },
-    { 0,   "fd:4",                  GOPT_ARGUMENT_REQUIRED  },
-    { 0,   "fd:5",                  GOPT_ARGUMENT_REQUIRED  },
-    { ':', "shebang",               GOPT_ARGUMENT_REQUIRED  },
-    { 0,   0,                       GOPT_LAST}};
+    {'p', "percent", GOPT_ARGUMENT_REQUIRED | GOPT_ARGUMENT_NO_HYPHEN},
```
JakeSiFive, line 153: something is bugged here

JakeSiFive (commented):
Overall I'm very happy with this. If there are some minor things we can fix their formatting as we go. I don't think we should be worrying too much about getting every little format issue taken care of. This is a large improvement IMO

tools/wake/main.cpp
V-FEXrt: Yeah, looks like the auto formatting is having issues with this block of code.

It's to be expected that we'll have some cases like this so I just turned off formatting for this section and manually formatted it.

We will like have several of these that look better hand formatted vs what the auto formatter does.

(While I was hear, I skipped formatting for the help message as well, makes it a little bit easier to read)

src/dst/expr.cpp
V-FEXrt: Would you like be to go ahead and remove the extra space while I'm here?

JakeSiFive: If you want to that's fine. I don't really think it makes sense for us to nit pick over issues like this. We can fix those as we edit the relevant files.

We just need a policy about allowing formatting fixes to be done as we go with other PRs

JakeSiFive:
The vscode failure makes me think you formatted some javascript here somehow?

JakeSiFive:
huh...maybe rerun CI to see if that vscode issue pops up again. I think that syntax error is in generated javascript so this all seems very strange to me.

src/dst/expr.cpp
V-FEXrt: Makes sense. I think we will have to write the policy as we explore the code change over the next several PRs

One to start with

Replace `X ::Y()` with `X::Y()`

V-FEXrt:
@elena-pudovkina are there any other JS macros ala https://github.com/sifive/wake/pull/903/commits/7f7f1f075aa02e0c92e18d45520da940c6c176a1 that should be excluded from the formatting?

V-FEXrt:
Okay! I think this is ready to land!

JakeSiFive (approved):
I think thankfully my in flight commits are not included here. Lets land this ASAP!

elena-pudovkina:
> @elena-pudovkina are there any other JS macros ala [7f7f1f0](https://github.com/sifive/wake/commit/7f7f1f075aa02e0c92e18d45520da940c6c176a1) that should be excluded from the formatting?
I see the formatting changed JS code in `readable.c` and `file.cpp`. Not sure why that didn't cause problems as well...

PR #905, opened 2022-06-17
Add format check to CI

V-FEXrt:


.circleci/config.yml
```diff
@@ -167,10 +167,24 @@ jobs:
           git -c user.email="ci-wake-build@sifive.com" -c user.name="ci-wake-build" commit -m "[skip ci] Deploy to GitHub Pages"
           git push --force "git@github.com:sifive/wake.git" master:gh-pages
 
+  check_format:
```
JakeSiFive, line 170: Can we do this in the tarball step instead?

.circleci/config.yml
```diff
@@ -167,10 +167,24 @@ jobs:
...
+    machine:
+      image: ubuntu-2004:202201-02
+    steps:
+      - run: |
+          sudo apt-get update && sudo apt-get install -y clang-format-12
+          SOURCE_FILES=$(find . -not \( -path ./vendor -prune \) -type f \( -name "*.h" -o -name "*.cpp" -o -name "*.c" \))
```
JakeSiFive, line 176: Maybe we should make a bash script that everything can reference here instead? Something like "find_files_to_format.sh" in our scripts directory?

.circleci/config.yml
```diff
@@ -23,6 +24,9 @@ jobs:
             - www
             - tests
             - scripts/sphinx/build/html
+# TODO: should be the line below but that'll generate a bunch of noise until the formatAll PR lands
+# clang-format --style=file --Werror -n ${SOURCE_FILES} 
+
```
JakeSiFive, line 29: I think this is done?

Makefile
```diff
@@ -55,22 +55,17 @@ install:	all
 # Formats all .h and .cpp file under the current directory
 # It assumes clang is available on the PATH and will fail otherwise
 formatAll:
-	clang-format -i --style=file $(shell find . -not \( -path ./vendor -prune \) -type f \( -name "*.h" -o -name "*.cpp" -o -name "*.c" \))
+	@clang-format -i --style=file $(shell ./scripts/which_clang_files all)
```
JakeSiFive, line 58: Whats the issue with echoing from clang-format? I kinda like it. Is it just too long in this case?

scripts/which_clang_files
```diff
@@ -0,0 +1,33 @@
+#! /bin/sh
```
JakeSiFive, line 1: Love this! Handles both cases! I'm also glad you're the writing bash and not me. Matthew is our resident bash expert so please have him review this.

V-FEXrt: I wouldn't say I'm doing a good job at writing the bash :)
Will certainly have Matthew take a look before landing

scripts/which_clang_files
```diff
@@ -0,0 +1,33 @@
...
+
+  # Flag if any files were found
+  if [ "${MERGED}" ]; then
+    exit 0
+  fi
+  exit 1
+fi
+
+echo "unknown parameter: '${1}'"
+exit 1
```
JakeSiFive, line 33: BTW, for whatever reason everyone else that works on wake has their vscode configured to always insert a newline at the end of the file. Consequently if they open a file to view it and it *doesn't* have a newline their editor adds it. Consequently we're all forced to do it 🤷 

TL;DR: Always have a newline at the end of your files. You can add this as a setting in vscode....why on earth we chose this convention I don't know.

Makefile
JakeSiFive: You added an `@` prefix which I understood to surpress echoing the command

V-FEXrt: ohhh right. Yeah it was just a bit long/inconsistent with the `make format` I don't have a strong preference either way

JakeSiFive: This is fine just wanted to know why. all good

JakeSiFive (approved):
Actually I looked over the bash and I'm pretty sure its not doing anything dangerous so happy to land this. Matthew can educate us later

scripts/which_clang_files
mmjconolly: A line of text in posix is defined as having an newline at the end, so some tools might expect it and may ignore the last "line" if it's just EOF rather than \nEOF

JakeSiFive: Maybe its just strange that this isn't the default in vscode?

PR #906, opened 2022-06-17
fuse-waked: allow directories that appear to be empty to rmdir()

mmjconolly:
If a directory existed on the underlying filesystem and was non-empty, but appeared to be empty within wakebox/fuse-waked, then fuse-waked would disallow the rmdir() even though it appeared to be empty. This could confuse a process that would think it's entitled to remove a writeable empty dir it had created (note: this means the directory is not in the 'visible' list).

This PR attempts to allow the _fiction_ of a successful rmdir() by letting the fuse client believe that the directory is gone, even though the underlying storage is left as-is.

Example of the issue:
```
$ cat bug.json
{
  "command": [
    "bash",
    "-c",
    "set -x; mkdir testdir/foo; ls -l -a testdir/foo; rm -rf testdir/foo"
  ],
  "environment": [ "PATH=/usr/bin:/bin" ],
  "directory": ".",
  "mount-ops": [
     {
      "type": "workspace",
      "destination": "."
     }
  ],
  "visible": [
      "testdir"
  ]
}
$ mkdir -p testdir/foo
$ touch testdir/foo/baa
$ wakebox -p bug.json
+ mkdir testdir/foo
+ ls -l -a testdir/foo
total 8
drwxrwxr-x. 2 root root 4096 Jun 10 18:43 .
drwxrwxr-x. 3 root root 4096 Jun 10 18:43 ..
+ rm -rf testdir/foo
rm: cannot remove 'testdir/foo': Directory not empty
$ echo $?
1
$ rm testdir/foo/baa
$ wakebox -p bug.json 
+ mkdir testdir/foo
+ ls -l -a testdir/foo
total 8
drwxrwxr-x. 2 matthewc matthewc 4096 Jun 10 18:51 .
drwxrwxr-x. 3 matthewc matthewc 4096 Jun 10 18:51 ..
+ rm -rf testdir/foo
$ echo $?
0
```

### DIFF
--- a/tools/fuse-waked/fuse-waked.cpp
+++ b/tools/fuse-waked/fuse-waked.cpp
@@ -621,6 +621,15 @@ static int wakefuse_unlink_trace(const char *path) {
   return out;
 }
 
+// Check if the directory has children that were written and not yet deleted.
+// Note: we don't check the 'files_visible' list as when the directory itself or
+// a file within that directory is visible then is_writable() would have already failed.
+static bool has_written_children(const std::string &dir, Job &job) {
+  auto i = job.files_wrote.lower_bound(dir + "/");
+  return i != job.files_wrote.end() && i->size() > dir.size() &&
+         0 == i->compare(0, dir.size(), dir);
+}
+
 static int wakefuse_rmdir(const char *path) {
   if (is_special(path)) return -ENOTDIR;
 
@@ -637,7 +646,14 @@ static int wakefuse_rmdir(const char *path) {
   if (!it->second.is_writeable(key.second)) return -EACCES;
 
   int res = unlinkat(context.rootfd, key.second.c_str(), AT_REMOVEDIR);
-  if (res == -1) return -errno;
+  if (res == -1) {
+    if ((errno == ENOTEMPTY) && !has_written_children(key.second, it->second)) {
+      // Let the fuse client process believe that the directory was unlinked,
+      // even though the underlying filesystem still has a populated directory.
+    } else {
+      return -errno;
+    }
+  }
 
   it->second.files_wrote.erase(key.second);
   it->second.files_read.erase(key.second);


tools/fuse-waked/fuse-waked.cpp
```diff
@@ -697,6 +697,19 @@ static int wakefuse_unlink_trace(const char *path)
 	return out;
 }
 
+// Check if the directory has children that were written and not yet deleted.
+// Note: we don't check the 'files_visible' list as when the directory itself or
+// a file within that directory is visible then is_writable() would have already failed.
+static bool has_written_children(const std::string& dir, Job& job) {
```
richardxia, line 703:
```suggestion
static bool has_written_children(const std::string& dir, const Job& job) {
```

I think this can be a const reference?

tools/fuse-waked/fuse-waked.cpp
```diff
@@ -697,6 +697,19 @@ static int wakefuse_unlink_trace(const char *path)
...
+	for (auto &written : job.files_wrote) {
+		if (written.size() > dir.size() && written[dir.size()] == '/' && written.find(dir) == 0) {
```
richardxia, line 705: `find()` could be slow because it'll attempt to continue searching past the 0th index, even though that you only care about whether `dir` is a proper prefix of `written`. It's too bad that only C++20 added [`starts_with()`](https://en.cppreference.com/w/cpp/string/basic_string/starts_with), but I think you can at least use [`compare()`](https://cplusplus.com/reference/string/string/compare/) and just specify the start position and length to compare.

richardxia (approved):
Thanks! I can confirm that this fixes the both the original bug in SiFive's internal codebase as well as the small test case I created to reproduce the issue.

Is there a way for us to add that as a test in Wake's CI tests? I think we could create an even simpler test case than the one I narrowed by just creating an empty directory prior to running a wake target that attempts to remove it.

JakeSiFive:
If a process removes a file what happens right now? Does it remove the file from `files_wrote`?

Also don't we still need to error out if there was a visible file in that directory? As is a bad process that tried to delete a folder that was populated even from its perspective would be allowed to think that it succeeded in deleting that file and then it could later read the directory and succeed when it shouldn't have.

tools/fuse-waked/fuse-waked.cpp
terpstra: Not to mention the outer loop is unnecessary. Try `lower_bound(dir + "/")` followed by check vs. `end` and a `compare`.

JakeSiFive (approved):
Happy with the code, the fact that it works for Richard, and Wes's explanation. Lets land it.

JakeSiFive (approved):

PR #908, opened 2022-06-21
Skip formatting on missed JS code

V-FEXrt:


JakeSiFive (approved):
LGTM but please wait on Elena since she's the expert here. We might have missed some other embedded javascript code.

elena-pudovkina (commented):
JS is used wherever `emscripten.h` is included. 
There's JS in `windows.c` since there's a string in single quotes: `if (process.platform == 'win32') {`
And in `whereami.c` with a require statement: `const path = require('path');`
That's all the usages I'm aware of

V-FEXrt:
> JS is used wherever `emscripten.h` is included. So other than what's been added, there's JS in `windows.c` since there's a string in single quotes: `if (process.platform == 'win32') {` And in `whereami.c` with a require statement: `const path = require('path');` That's all the usages I'm aware of

Great, thanks! I fixed `windows.c` though since the formatter didn't change anything there wasn't anything to revert. For `whereami.c` that file is under `vendor/` and we don't format vendored files so nothing to change!

elena-pudovkina (approved):

PR #910, opened 2022-06-22
[wake stdlib][list] Add unary comma operator

JakeSiFive:
This change adds the unary comma operator as the singleton list operation which can be used to improve list syntax.

### DIFF
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -49,13 +49,13 @@ export data List a =
 # ```
     (head: a), (tail: List a)
 
-# Create a singelton list from a value. This
+# Create a singleton list from a value. This
 # can be used to slightly increase the niceness
 # of lists by avoid the Nil at the end.
 #
 # Examples:
 # ```
-#   10,                      # Create a singelton list containg just 10
+#   10,                      # Create a singleton list containg just 10
 #   1, 2, 3,                 # Create a list of 3 elements without using Nil
 #   (1, 2, 3,) ++ (4, 5, 6,) # append two lists
 # ```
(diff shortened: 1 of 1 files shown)


mmjconolly (approved):

V-FEXrt (approved):

ag-eitilt (approved):
I personally think the biggest benefit to this is probably for vertical lists rather than horizontal ones, but the documentation for that isn't as nice.

share/wake/lib/core/list.wake
```diff
@@ -49,6 +49,18 @@ export data List a =
 # ```
     (head: a), (tail: List a)
 
+# Create a singleton list from a value. This
+# can be used to slightly increase the niceness
+# of lists by avoid the Nil at the end.
+#
+# Examples:
+# ```
+#   10,                      # Create a singleton list containg just 10
```
richardxia, line 58:
```suggestion
#   10,                      # Create a singleton list containing just 10
```

richardxia: Also, unrelated to the typo I just pointed out, I was wondering if it would make sense to _not_ actually use the unary comma for single-item lists as a style preference. Python has exactly this syntax for single-item tuples, and I think it's a very frequent accident to construct a single-item tuple when you weren't expecting to, especially when refactoring code and moving lines around.

Alternatively, maybe we could prefer the unary comma for singleton lists _but_ also recommend wrapping it with parentheses to make it more obvious? E.g. `(10,)`

JakeSiFive: Yeah I'm ambivalent. Eventually we want to have square bracket syntax per the poll, this is just a small tide over.

JakeSiFive: I agree we should request parens around it in review.

JakeSiFive:
> I personally think the biggest benefit to this is probably for vertical lists rather than horizontal ones, but the documentation for that isn't as nice.

good point. I added mention of this in the documentation and we can start pointing this out in review. People should like this.

PR #913, opened 2022-06-24
Job cache read

JakeSiFive:
This change is intended to add reading of jobs from the cache, with handling for directories and sandbox to real world redirection.

### DIFF
--- a/src/wcl/filepath.h
+++ b/src/wcl/filepath.h
@@ -46,6 +46,12 @@ class filepath_iterator {
     next();
   }
 
+  filepath_iterator operator++(int) {
+    auto out = *this;
+    next();
+    return out;
+  }
+
   filepath_iterator& operator++() {
     next();
     return *this;
--- a/tools/job-cache/job-cache.cpp
+++ b/tools/job-cache/job-cache.cpp
@@ -387,7 +387,7 @@ wcl::optional<std::pair<std::string, std::string>> parent_and_base(const std::st
     if (*rbegin == '/') {
       // Advance to the character past the slash
       rbegin++;
-      // Now return thw two strings
+      // Now return the two strings
       return {wcl::in_place_t{}, std::string(rend, rbegin), std::string(rbegin, str.end())};
     }
   }
@@ -647,7 +647,11 @@ struct AddJobRequest {
   }
 };
 
-// TODO: Ideally we'd have someting like std::filepath
+// join takes a sequence of strings and concats that
+// sequence with some seperator between it. It's like
+// python's join method on strings. So ", ".join(seq)
+// in python joins a list of strings with a comma. This
+// function is a C++ equivlent.
 template <class Iter>
 static std::string join(char sep, Iter begin, Iter end) {
   std::string out;
@@ -798,7 +802,7 @@ class Cache {
     // If we didn't link all the files over we need to return a failure.
     if (!success) return {};
 
-    // TODO: We should really return a different hting here
+    // TODO: We should really return a different thing here
     //       that mentions the *output* locations but for
     //       now this is good enough and we can assume
     //       workspace relative paths everywhere.
--- a/tools/job-cache/job-cache.cpp
+++ b/tools/job-cache/job-cache.cpp
@@ -289,7 +289,6 @@ class InputFiles {
 
   void insert(const std::string &path, Hash256 hash, int64_t job_id) {
     add_input_file.bind_string(1, path);
-    std::cerr << "Hash at bind site: " << hash.to_hex() << std::endl;
     add_input_file.bind_string(2, hash.to_hex());
     add_input_file.bind_integer(3, job_id);
     add_input_file.step();
@@ -492,8 +491,6 @@ class SelectMatchingJobs {
     while (find.step() == SQLITE_ROW) {
       std::string path = find.read_string(1);
       Hash256 hash = Hash256::from_hex(find.read_string(2));
-      std::cerr << "FromDB: hash = " << find.read_string(2) << std::endl;
-      std::cerr << "Checking: " << path << " = " << hash.to_hex() << std::endl;
       auto iter = find_job_request.visible.find(path);
       if (iter == find_job_request.visible.end() || hash != iter->second) {
         find.reset();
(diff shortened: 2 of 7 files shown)


src/wcl/filepath.h
```diff
@@ -0,0 +1,111 @@
+/*
```
JakeSiFive, line 1: This is part of another PR but I rebased it in slightly differently so the PR diff isn't correct. I'll land the other thing soon and rebase this to fix it. Just ignore this for now.

tests/wake-unit/stdout
```diff
@@ -1,4 +1,12 @@
 PASSED:
+  filepath_range_basic
```
JakeSiFive, line 2: Ignore these for now, they're a rebase issue from the other PR

tools/wake-unit/filepath.cpp
```diff
@@ -0,0 +1,231 @@
+/*
```
JakeSiFive, line 1: Again, please ignore this

tools/wake-unit/unit.h
```diff
@@ -93,6 +93,86 @@ struct TestLogger {
     return fail(*err, assert);
   }
 
+  TestStream expect_equal(bool assert, std::vector<std::string> expected,
```
JakeSiFive, line 96: Also please ignore

tools/job-cache/job-cache.cpp
```diff
@@ -354,6 +366,99 @@ class JobTable {
...
+  Hash256 hash;
+};
+
+struct MatchingJob {
+  int64_t job_id;
+  std::vector<CachedOutputFile> output_files;
+};
+
+// Returns the end of the parent directory in the path.
+wcl::optional<std::pair<std::string, std::string>> parent_and_base(const std::string &str) {
```
JakeSiFive, line 380: We could really use std::filepath and std::string_view honestly...uhg

tools/job-cache/job-cache.cpp
```diff
@@ -478,17 +712,99 @@ class Cache {
         input_dirs(db),
         output_files(db),
         transact(db),
+        matching_jobs(db),
         dir(std::move(_dir)),
         rng(wcl::xoshiro_256::get_rng_seed()) {}
 
+  // TODO: Unlike reading, we need to account for
+  // directory remappings here.
+  bool read(const FindJobRequest &find_request) {
```
JakeSiFive, line 721: This does more tricky filesystem work but it should all be fairly cheap hopefully.

JakeSiFive:
This doesn't work just yet but its ready for review I think. This allows reading a job from the cache, with redirects!

tools/job-cache/job-cache.cpp
```diff
@@ -354,6 +367,93 @@ class JobTable {
...
+wcl::optional<std::pair<std::string, std::string>> parent_and_base(const std::string &str) {
+  // traverse backwards but using a normal iterator instead of a reverse
+  // iterator.
+  auto rbegin = str.end() - 1;
+  auto rend = str.begin();
+  for (; rbegin >= rend; --rbegin) {
+    if (*rbegin == '/') {
+      // Advance to the character past the slash
+      rbegin++;
+      // Now return thw two strings
```
V-FEXrt, line 390: typo: thw -> the

JakeSiFive:
I've got this to where it can read back simple jobs, working on making sure that it writes back the files the right way now.

tools/job-cache/job-cache.cpp
```diff
@@ -442,12 +641,41 @@ struct AddJobRequest {
...
-      output.hash = output_file.second.get("hash").value;
+      output.hash = Hash256::from_hex(output_file.second.get("hash").value);
       outputs.emplace_back(std::move(output));
     }
   }
 };
 
+// TODO: Ideally we'd have someting like std::filepath
+template <class Iter>
+static std::string join(char sep, Iter begin, Iter end) {
```
V-FEXrt, line 652: Can you document what these functions are supposed to do?

Looks like this one adds `sep` between each item in `Iter`?

tools/job-cache/job-cache.cpp
```diff
@@ -478,17 +707,103 @@ class Cache {
...
+
+    // Return early if there was no match.
+    if (!result) return {};
+
+    // We need a tmp directory to put these outputs into
+    std::string tmp_job_dir = dir + "/tmp_outputs_" + rng.unique_name();
+    mkdir_no_fail(tmp_job_dir.c_str());
+
+    // We also need to know what directory we're reading out of
+    std::string job_dir = dir + "/" + std::to_string(result->job_id & 0xFF) + "/" +
```
V-FEXrt, line 733: nit: looks like a good candiate for `join`?

tools/job-cache/job-cache.cpp
```diff
@@ -244,6 +247,14 @@ class PreparedStatement {
     }
   }
 
+  int64_t read_integer(int64_t index) { return sqlite3_column_int64(query_stmt, index); }
+
+  std::string read_string(int64_t index) {
+    const char *str = reinterpret_cast<const char *>(sqlite3_column_text(query_stmt, index));
```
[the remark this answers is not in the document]
JakeSiFive: sqlite allocates it and its valid until the next sqlite call that invalidates it (such as step, or another column read etc...). One of these is always lingering around in the prepared statement until you finalize it or reset the query. It gets collected eventually just not here.

tools/job-cache/job-cache.cpp
JakeSiFive: not quite because its not an array of values.

tools/job-cache/job-cache.cpp
```diff
@@ -544,6 +859,19 @@ int main(int argc, char **argv) {
       JAST::parse(argv[3], std::cerr, job_result);
       AddJobRequest add_request(job_result);
       cache.add(add_request);
+      return 0;
+    }
+
+    if (std::string(argv[2]) == "read") {
+      JAST job_plan;
+      JAST::parse(argv[3], std::cerr, job_plan);
```
mmjconolly, line 867: What produces the json for the `add` operation?

JakeSiFive: Right now me by hand, but eventually wake would do that.

That said I plan on adding a third "read or run" operation which will have a very wakebox .in.json like interface, it will invoke wakebox with a .in.json that job-cache writes, and then it will read.out.json to do the write, and it will attempt to read first before it attempts to run. That's the next step to get a demo runner working.

JakeSiFive: I'll go work on a diagram to explain who invokes what and how now lol

JakeSiFive:
Update, I add to make some small changes but I have the latest version working on non-trivial jobs. I just need to test redirection now

tools/job-cache/job-cache.cpp
```diff
@@ -478,17 +709,111 @@ class Cache {
...
+      unlink_no_fail(to_copy.first.c_str());
+    }
+
+    // Lastly clean up the tmp dir itself
+    rmdir_no_fail(tmp_job_dir.c_str());
+
+    // If we didn't link all the files over we need to return a failure.
+    if (!success) return {};
+
+    // TODO: We should really return a different hting here
```
V-FEXrt, line 801: typo: hting -> thing

V-FEXrt (approved):

tools/job-cache/job-cache.cpp
V-FEXrt: I think this is still open too

JakeSiFive:
reading took me 26 days to finish but I think its done 😭

PR #915, opened 2022-06-27
Emit warning when using fake re2c

V-FEXrt:


V-FEXrt:
hmmmm this actually might create too much noise. Seems like the default flow is to not have it available. I'll see if I can find a way to print it only when it would be expected that you are using the real re2c

JakeSiFive:
Oh yeah like if the environment variable is set.

This all said it would only affect wake devs so the noise of one line per build isn't that high.

You can also make the function that checks if everything is ok a `target` and the effects will only run once (which is a subtle way in which `print` breaks the illusion that wake is "pure" in some sense)

V-FEXrt:
Superseded by https://github.com/sifive/wake/pull/916

PR #916, opened 2022-06-27
Update re2c to use environment PATH

V-FEXrt:
`re2c.wake` wasn't reading the env path so it was always returning the system `re2c`. This PR fixes that issue.

Also supersedes https://github.com/sifive/wake/pull/915 with a more detailed error message.

Ex:
```
[WARN] Requested re2c which is unsatisfiable. Using fake
  why: re2c version 0.14 is not 1.3 or higher
```

### DIFF
--- a/vendor/re2c.wake
+++ b/vendor/re2c.wake
@@ -16,10 +16,25 @@
 package build_wake
 from wake import _
 
-# Require re2c >= 1.0 for flag support
-def re2cOk Unit =
-    require Some x, Some y, Some z, _ =
-        makeExecPlan (which "re2c", "--version", Nil) Nil
+def re2cPath =
+    require Some path =
+        getenv "PATH"
+
+    def re2c =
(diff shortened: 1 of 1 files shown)


vendor/re2c.wake
```diff
@@ -16,10 +16,25 @@
...
+    require Some path =
+        getenv "PATH"
+
+    def re2c =
+        whichIn path "re2c"
+
+    if matches `/.*` re2c then
+        Some re2c
+    else
+        None
```
JakeSiFive, line 29: Use `re2c == ""` instead, so in the reverse of this order. That way you don't use a regex.

JakeSiFive (approved):

vendor/re2c.wake
JakeSiFive: Yeah I think that would be the best thing. Let's just do the `==~ "re2c"` thing for now

PR #917, opened 2022-06-27
[wcl] Introduce the Wake Core Library

JakeSiFive:
We've updated wake to use C++14 but we've decided we won't be updating to C++17 for quite some time. In the mean time however we want to be able to use modern looking C++ with proper error handling, and modern libraries. This library is the start of a modern set of useful tools and data structures that we can use to write highly optimized C++ that looks and feels a lot like real modern C++.

This library's interface is inspired by the current C++ standard library, absl, and llvm's library. Where possible we've chosen names that come from the C++ standard library so that migration to C++17 will be slightly eased in the future.

### DIFF
--- a/src/wcl/optional.h
+++ b/src/wcl/optional.h
@@ -15,6 +15,9 @@
  * limitations under the License.
  */
 
+#include <utility>
+#include <cstdint>
+
 namespace wcl {
 
 template <class T>
@@ -25,27 +28,27 @@ class alignas(T) optional {
 
  public:
   // Default construct is none
-  Optional() = default;
+  optional() = default;
 
-  Optional(const Optional& other) : none(other.none) {
+  optional(const optional& other) : none(other.none) {
     if (other.none) return;
     new (reinterpret_cast<T*>(data)) T(*other);
   }
 
-  Optional(Optional&& other) : none(other.none) {
+  optional(optional&& other) : none(other.none) {
     if (other.none) return;
     new (reinterpret_cast<T*>(data)) T(std::move(*other));
   }
 
   // Placement new, perfect forwarding construction
   template <class... Args>
-  Optional(Args&&... args) : none(false) {
+  optional(Args&&... args) : none(false) {
     new (reinterpret_cast<T*>(data)) T(std::forward<Args>(args)...);
   }
 
-  ~Optional() { reinterpret_cast<T*>(data)->~T(); }
+  ~optional() { reinterpret_cast<T*>(data)->~T(); }
 
-  Optional& operator=(const Optional& other) {
+  optional& operator=(const optional& other) {
     if (!none) reinterpret_cast<T*>(data)->~T();
     none = other.none;
     if (none) return *this;
@@ -53,7 +56,7 @@ class alignas(T) optional {
     return *this;
   }
 
-  Optional& operator=(Optional&& other) {
+  optional& operator=(optional&& other) {
     if (!none) reinterpret_cast<T*>(data)->~T();
     none = other.none;
     if (none) return *this;
--- a/src/wcl/optional.h
+++ b/src/wcl/optional.h
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2022 SiFive, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You should have received a copy of LICENSE.Apache2 along with
+ * this software. If not, you may obtain a copy at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace wcl {
+
+template <class T>
+class alignas(T) optional {
+ private:
+  uint8_t data[sizeof(T)];
+  bool none = true;
+
+ public:
+  // Default construct is none
+  Optional() = default;
+
+  Optional(const Optional& other) : none(other.none) {
+    if (other.none) return;
+    new (reinterpret_cast<T*>(data)) T(*other);
+  }
+
+  Optional(Optional&& other) : none(other.none) {
+    if (other.none) return;
+    new (reinterpret_cast<T*>(data)) T(std::move(*other));
+  }
+
+  // Placement new, perfect forwarding construction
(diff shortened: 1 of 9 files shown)


JakeSiFive:
FYI, the current set is mostly just what I wound up making as I was developing the job-cache. None of these libraries are tested yet so I'll need to add a bunch of unit tests for them all

Additional things I plan on adding in the future:
1) Random number generation
2) File path and filesystem functions
3) Defer and possibly Ashley's idea for UniqueResource
4) Bloom Filter
5) Logging
6) Maybe something like this: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0040r3.html

As we have new needs we might add more things.

JakeSiFive:
A note on wcl::optional, we already have an Optional in util.h but it isn't great. It uses dynamic memory allocation. Also the only place it's being used it's being used more as a pointer that automatically copies itself by reallocating than as an optional. wcl::optional can't be used as a replacement because it's being used in a recursive position.

The existing Optional can be trivially replaced with a smart pointer with only minor effort.

src/wcl/small_vector.h
```diff
@@ -0,0 +1,125 @@
...
+    return reinterpret_cast<const T(&)[_small_size]>(_data);
+  }
+
+  inline void grow() {
+    // Figure out the old and new caps
+    size_t old_cap = cap();
+    size_t new_cap = old_cap * 2;
+
+    // Get the new space.
+    T* new_data = reinterpret_cast<T>(malloc(new_cap * sizeof(T)));
```
V-FEXrt, line 63: Can you not use `realloc` instead?

src/wcl/small_vector.h
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+#include <algorithm>
+#include <vector>
+
+namespace wcl {
+
+// TODO: Use type traits to check if we can use realloc
+template <class T, size_t small_size>
+class alignas(T) small_vector {
```
V-FEXrt, line 25: not sure if you need extra code for it or not, but this should probably be iterable right? that way it can be used in `foreach` loops

src/wcl/trie.h
```diff
@@ -0,0 +1,76 @@
...
+  optional<size_t> matching_child(size_t pindex, const Key& value) {
+    if (nodes.size() <= pindex) return optional<size_t>();
+    trie_node* node = &nodes[pindex];
+    for (size_t child_index : node->child_indexes) {
+      if (value == nodes[child_index].value) return child_index;
+    }
+    return optional<size_t>();
+  }
+
+ public:
```
V-FEXrt, line 47: I think the public interface here needs to be fleshed out a bit more.

- move_insert takes a begin and end for the Key which isn't super clear
- There only functions to insert into the trie, I think you need some kind of lookup/retrieval to do anything with the DS once you've built it

src/wcl/unique_fd.h
```diff
@@ -0,0 +1,89 @@
...
+    int fd = ::open(str, flags);
+    if (fd == -1) {
+      return make_error<unique_fd, int>(errno);
+    }
+    return make_result<unique_fd, int>(unique_fd(fd));
+  }
+
+  // Opens a file just as posix's open would, returning
+  // either a unique_fd or the errno value.
+  static result<unique_fd, int> open(const char* str, int flags, mode_t mode) {
```
V-FEXrt, line 80: The duplication of these two functions has always bothered me a tiny bit. Can we lookup the "default mode" of `open` then have `static result<unique_fd, int> open(const char* str, int flags)` call this function with the "default open mode"?

V-FEXrt (approved):
Overall looks good, some of the C++ stuff is a bit complicated but its to be expected for stdlib type stuff.

I think the testing story is super important here to glad to see those rolling it. Will follow along for the test of the test cases as they come through

mmjconolly:
I'm not super familiar with what is/isn't provided by the various c++ versions, but from your initial description it sounds like some of this isn't needed if we use c++17.
Many of these types added in this PR look like fairly fundamental things, how much of this is roll-your-own compared to vendored in from elsewhere?
It makes me wonder if sticking with c++14 to support older distros is actually worth it

JakeSiFive:
I never thought sticking with C++14 was worth in the first place :)

But this is an ok compromise TBH and I want to be able to build wake on the rocky8 at the very least which I think doesn't quite support C++17.

Also small_vector, trie, unique_fd, and result would all be needed even if we went with C++17. small_vector is an optimization but it makes trie a lot more efficient so I think its worth it here.

This we might write in wcl that C++17 has:
* optional (already written here)
* span (already written here)
* string_view (already written here)
* File path and filesystem functions
* https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0040r3.html

Things we need regardless:
* small_vector (already here)
* result (already here)
* unique_fd (already here)
* trie (already here)

The filesystem library is the big win of going with C++17 IMO. I would be pro moving to C++17 and dropping older builds but I'm also ok sticking with C++14 since we need wcl regardless

mmjconolly:
> The filesystem library is the big win of going with C++17

Is the boost filesystem library close-enough to what c++17 would provide?

src/wcl/unique_fd.h
JakeSiFive: I don't really want to change it. The mode is only relevant when creating a file which you aren't always doing. This isn't really duplicating much code frankly and this ensure that we're a 1:1 mapping with posix where as if we pass a value (which I think is just 0) we're not quite the same. These are different overloads, not the same function.

src/wcl/trie.h
JakeSiFive: yeah it does, work in progress.

src/wcl/small_vector.h
JakeSiFive: You can if the type's move/copy constructor is "simple" meaning it can just be copied over. That is a good optimization to add later but it requires template magic I don't want to mess with just this moment.

src/wcl/optional.h
```diff
@@ -0,0 +1,75 @@
...
+
+template <class T>
+class alignas(T) optional {
+ private:
+  uint8_t data[sizeof(T)];
+  bool none = true;
+
+ public:
+  // Default construct is none
+  Optional() = default;
```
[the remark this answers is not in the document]
JakeSiFive: Yeah but that's outdated

src/wcl/unique_fd.h
V-FEXrt: ack, sgtm

JakeSiFive:
Optional split out over here: https://github.com/sifive/wake/pull/928

JakeSiFive:
This is going to get broken up, and this process has already started. We can come back this PR to get implementation code later

Issue #925, opened 2022-06-29
copy_file_range in job_cache won't build on Centos7


timsnyder-siv:
https://github.com/sifive/wake/blob/a9a14e99c24b53f712bd66c7b7b94aee2978b4ba/tools/job-cache/job-cache.cpp#L76

was added in #904 uses `copy_file_range()` to accelerate copying files.  However, I'm unable to find `copy_file_range()` when I build on Centos7.  Yes, I'm stuck on Centos7 until AWS adds support for something newer for their FPGA Development AMI.

I assume that since the new job_cache is work-in-progress, that I can just not build it and using the Federation wake flows won't even notice.

It would be greatly appreciated if the code could be updated to use `copy_file_range()` if it's available and fall back to a normal slow copy if it isn't.

JakeSiFive:
job-cache isn't used right now so you can manually delete that code and compile locally. I'll see if we can make that conditional. I might have to use dlsym which would kinda suck.

mmjconolly:
You can probably detect via `#ifdef __NR_copy_file_range`

JakeSiFive:
This was fixed using conditional compilation and we added Centos 7 back to our test set with an updated compiler and stdlib to get the features we want.

PR #926, opened 2022-06-29
Restore CentOS 7.6

V-FEXrt:


mmjconolly:
In https://github.com/sifive/wake/blob/master/.circleci/dockerfiles/centos-7.6
You could add something like:
```
yum install centos-release-scl
yum install devtoolset-9-gcc*
source scl_source enable devtoolset-9
```

mmjconolly:
I guess we could try directly using
```
source /opt/rh/devtoolset-9/enable
```
instead of 
```
source scl_source enable devtoolset-9
```

.circleci/config.yml
```diff
@@ -80,6 +80,24 @@ jobs:
...
+    machine:
+      image: ubuntu-2004:202201-02
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - run: docker build -f /tmp/workspace/.circleci/dockerfiles/centos-7.6 -t wake-centos-7.6 .
+      - run: |
+          cp /tmp/workspace/wake_*.tar.xz . && \
+          cp -a /tmp/workspace/tests . && \
+          docker run --rm --device /dev/fuse --cap-add SYS_ADMIN --security-opt apparmor:unconfined --mount type=bind,source=$PWD,target=/build wake-centos-7.6 rpmbuild -ta --define "_rpmdir /build" wake_*.tar.xz && \
```
mmjconolly, line 93:
```
docker run --rm --device /dev/fuse --cap-add SYS_ADMIN --security-opt apparmor:unconfined --mount type=bind,source=$PWD,target=/build wake-centos-7.6 $some_command
```
I was thinking that the `source` would need to be part of $some_command
When `docker build` is running I think it runs each command in a fresh shell and doesn't preserve the state, so when we do `docker run` the sourced variables are missing

mmjconolly:
What we see now is:
```
tools/job-cache/job-cache.cpp: In function 'void copy(int, int)':
tools/job-cache/job-cache.cpp:76:7: error: 'copy_file_range' was not declared in this scope; did you mean 'sync_file_range'?
```

V-FEXrt:
That aligns with https://github.com/sifive/wake/issues/925. We will need to add some `#ifdef`s for CentOS/job cache (wild how quickly things bit rot when not protected by CI

@JakeSiFive FYI

JakeSiFive:
I'll make a change tomorrow to do some ifdef magic.


1) We need to check for Linux 4,5 *OR* glibc version 2.27 per this: https://man7.org/linux/man-pages/man2/copy_file_range.2.html

2) For the FICLONE direct route we can just check to see if FICLONE is defined.


So we have the following cases for copy

* Apple
* Linux with no copy_file_range
* Linux with copy_file_range

Then for reflink_or_copy we have two cases

* FICLONE is defiend
* FICLONE is not defined

JakeSiFive (approved):

mmjconolly (approved):

timsnyder-siv:
Greatly appreciated.  Thank you!

PR #928, opened 2022-06-30
[wcl] Introduce the Wake Core Library

JakeSiFive:
This change just adds an optional type that functions the same way as std::optional more or less.
It doesn't respect assignment quite the same way right now which means that wcl::optional<std::vector<T>>
isn't as efficen't as it could be since if you assign one vector to another that already has capacity,
the current implementation deallocates the first one and reallocates this one despite that not being
needed.

### DIFF
--- a/tools/wake-unit/optional.cpp
+++ b/tools/wake-unit/optional.cpp
@@ -69,7 +69,7 @@ TEST(option_move) {
   // We want to make sure we can make optional move only things.
   wcl::optional<std::unique_ptr<int>> move_only1(wcl::in_place_t{}, std::make_unique<int>(10));
   wcl::optional<std::unique_ptr<int>> move_only2(std::move(move_only1));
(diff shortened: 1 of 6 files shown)


tools/wake-unit/optional.cpp
```diff
@@ -0,0 +1,230 @@
...
+  EXPECT_FALSE((bool)none);
+}
+
+TEST(option_some) {
+  wcl::optional<int> some(wcl::in_place_t{}, 10);
+  ASSERT_TRUE((bool)some);
+  EXPECT_EQUAL(10, *some);
+}
+
+TEST(option_inplace) {
```
V-FEXrt, line 37: this is the same as `option_some`.  Bad copy+paste?

V-FEXrt (approved):
Can we construct an optional without using `wcl::in_place_t{}`? seems a bit verbose just for construction

PR #929, opened 2022-07-01
Add a table of contents to the tutorial

ErikParawell-SiFive:
When reading through it, I thought it would be nice to have a table of contents to better sort through it so I added one.


ag-eitilt (approved):
The links don't work in the change preview, but that seems like it's just a PR diff; going to "View file" winds up with them scrolling the window correctly.

ag-eitilt:
I will point out that Github does add a table of contents automatically as the understated button in the top left of the rendered file view.  As I only just discovered that myself, it might still be nice to have this more visible table of contents, which has the benefit of working in more places.

mmjconolly (approved):

ErikParawell-SiFive:
I don't have permissions to merge, so someone else will have to do it.

PR #930, opened 2022-07-05
Support building on aarch64-apple-darwin

V-FEXrt:


Makefile
```diff
@@ -12,12 +12,12 @@ DESTDIR ?= /usr/local
...
 FUSE_CFLAGS  :=	$(shell pkg-config --silence-errors --cflags fuse)
 CORE_CFLAGS  := $(shell pkg-config --silence-errors --cflags sqlite3)	\
-		$(shell pkg-config --silence-errors --cflags gmp-6)	\
+		$(shell pkg-config --silence-errors --cflags gmp)	\
 		$(shell pkg-config --silence-errors --cflags re2)	\
 		$(shell pkg-config --silence-errors --cflags-only-I ncurses)
 FUSE_LDFLAGS := $(shell pkg-config --silence-errors --libs fuse    || echo -lfuse)
 CORE_LDFLAGS :=	$(shell pkg-config --silence-errors --libs sqlite3 || echo -lsqlite3)	\
-		$(shell pkg-config --silence-errors --libs gmp-6   || echo -lgmp)	\
+		$(shell echo -lgmp)	\
```
terpstra, line 20: Why do you need this change? If `--cflags` worked, wouldn't `--libs` work too?

V-FEXrt: This was just testing what would break without it. (surprisingly nothing)

but for Mac M1 this line actually needs to be 
`$(shell pkg-config --silence-errors --libs gmp   || echo -lgmp)`


ex:
```
> pkg-config --silence-errors --libs gmp-6

> pkg-config --silence-errors --libs gmp 
-L/opt/homebrew/Cellar/gmp/6.2.1_1/lib -lgmp
```

So I was trying to figure out if something like `gmp-6` --fallback--> `gmp` --fallback--> `echo` would work

V-FEXrt:
Leaving as a comment so I can delete the file. Install steps were

```
# install homebrew
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"

# install deps
brew install gmp
brew install pkg-config
brew install re2
brew install --cask macfuse
```

mmjconolly:
I guess we should refer to this as something non-product like aarch64-apple-darwin rather than M1

JakeSiFive (commented):
code looks good to me, builds on gamma and all CI, and has been tested locally on M1 macs

land it IMO

JakeSiFive:
wait stop the press, critical error...we assumed that x86 and arm were the only two architectures how dare we!!!

src/util/rank.cpp
```diff
@@ -22,7 +22,10 @@
 #include "rank.h"
 
 #include <assert.h>
+
+#ifndef __aarch64__
```
JakeSiFive, line 26: This should "if not x86_64"

JakeSiFive: with this change I think wake will compile on risc-v ;)

mmjconolly: We should file a task for that haha

V-FEXrt:
Okay I tested the most recent changes on M1 and gamma again. All looks good so I'm going to go ahead and land!

src/util/rank.cpp
JakeSiFive: Send me parts to get my unmatched working and I will do it lol.

PR #931, opened 2022-07-05
.gitignore: wake.db-shm and wake.db-wal

mmjconolly:
`wake.db-shm` and `wake.db-wal` shouldn't be tracked by git


terpstra:
Hmm. These files should only exist if wake crashed. I left them out of `.gitignore` to draw my attention to that.

PR #932, opened 2022-07-05
Basic Null Formatter

V-FEXrt:
First pass at the wake-format tool. This version strictly fleshes out the tool/api and just re-emits the source file unchanged.


tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "gopt/gopt-arg.h"
+#include "gopt/gopt.h"
+#include "parser/cst.h"
+#include "parser/syntax.h"
+#include "util/diagnostic.h"
+#include "util/file.h"
+
+#define VERSION "0.0.1"
+
+class TerminalReporter : public DiagnosticReporter {
```
V-FEXrt, line 35: This is copied directly from `tools/wake/main.cpp` not sure if there is an easily/clean way to pull that out/share it.

mmjconolly:
To run it over a large codebase, would we do something like:
```
wake-format -i $(find . -name '*.wake')
```

build.wake
```diff
@@ -38,7 +38,7 @@ export def install: List String => Result String Error = match _
 
 # Build all wake targets
 def targets =
-    buildWake, buildWakeBox, buildFuseDaemon, buildShim, buildBSP, buildLSP, buildJobCache, Nil
+    buildWake, buildWakeBox, buildFuseDaemon, buildShim, buildBSP, buildLSP, buildJobCache, buildWakeBox, buildWakeFormat, Nil
```
mmjconolly, line 41:
```suggestion
    buildWake, buildWakeBox, buildFuseDaemon, buildShim, buildBSP, buildLSP, buildJobCache, buildWakeFormat, Nil
```

tools/wake-format/wake-format.cpp
JakeSiFive: there is not but at least you can modify this to your needs now.

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,154 @@
...
+      std::cerr << diagnostic.getMessage() << std::endl;
+    }
+  }
+};
+
+void print_help(const char *argv0) {
+  // clang-format off
+  std::cout << std::endl
+    << "Usage: " << argv0 << " [OPTIONS] [<file> ...]" << std::endl
+    << "  --dry-run  -n     Check if formatting needed, but don't apply it"      << std::endl
```
JakeSiFive, line 62: love this option.

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,154 @@
...
+    << "  --help     -h     Print this help message and exit"                    << std::endl
+    << "  --in-place -i     Edit files in place. Default emits file to stdout"   << std::endl
```
JakeSiFive, line 64: FYI when you edit files in place, make sure to do the rename trick to make it atomic. There used to be a bug in clang-format where if you ended your program at the wrong time it deleted your code. Some people working on Fuchsia got very mad at the toolchain team for this lol.

Rename trick:
1) create a temp file somewhere
2) write the correct results into that temp file
3) rename the temp file into place

rename is the atomic swiss army knife of file systems.

JakeSiFive: Right now you're doing the "write file into its final destination" thing btw.

JakeSiFive: nevermind I'm wrong lol I misread your code.

V-FEXrt: Already fixed lol, wanna take a look at the most recent commits just to confirm?

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,157 @@
...
+  for (int i = 1; i < argc; i++) {
+    ExternalFile file = ExternalFile(*reporter, argv[i]);
+    CST cst = CST(file, *reporter);
+    if (terminalReporter.errors) {
+      std::cerr << argv[0] << ": failed to parse file: " << argv[i] << std::endl;
+      exit(EXIT_FAILURE);
+    }
+
+    std::streambuf *buffer;
+    std::ofstream stream;
```
JakeSiFive, line 131: Instead of doing stream buffer tricks, use an `std::unique_ptr<std::ofstream>` and an `std::ostream*` it uses no uncommonly used tricks at the cost of an extra memory allocation and indirection (which is meaningless when working with file IO)

```
std::unique_ptr<std::ofstream> output_file;
std::ostream* out;
...
if (in_place) {
  output_file = std::make_unique<std::ofstream>(tmp);
  out = output.get();
} else {
  out = &std::cout;
}
```

Now you just use `*out` and you don't need the second ostream

tools/wake-format/wake-format.cpp
mmjconolly: The dry-run option can be later used by CI systems so we can prevent unformatted code from getting into codebases

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    std::string name(argv[i]);
+    std::string tmp = name + ".tmp";
```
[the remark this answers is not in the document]
mmjconolly: stdlib.h has a `mktemp()` that accepts a template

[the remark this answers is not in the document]
JakeSiFive: Yeah mktemp() has issues too but I would accept it here

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,157 @@
...
+    std::ostream fout(buffer);
+
+    walk_cst(fout, cst.root());
+
+    // When editing in-place we need to copy the tmp file over the original
+    if (in_place) {
+      // It's implementation defined what rename
+      // does if the file already exists so it must
+      // be explictly deleted here
+      std::remove(name.c_str());
```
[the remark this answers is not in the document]
terpstra: You need the two files to be on the same filesystem, so usually in the same directory.

tools/wake-format/wake-format.wake
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package build_wake
+from wake import _
+
+target buildWakeFormat variant =
+    tool here Nil variant "bin/wake-format" (parser, gopt, util,)
```
mmjconolly, line 20: One thing to be aware of is that `tool` isn't creating `bin/wake-format` until `make install` is run, it's creating `wake-format.native-cpp14-release`.
I know that there are other opinions about this topic, but I find it nice when we're able to use the build output from the source tree without always having to do the install step.

V-FEXrt: yeah, makes sense. I've just been running `./bin/wake-format.native-cpp14-release ...` but I can see how that wouldn't be ideal.

I personally don't have a strong opinion but it sounds like some other folks do. Would be happy to take a look at that but seems better for a separate discussion/PR

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+void walk_cst(std::ostream *out, CSTElement node) {
+  for (CSTElement child = node.firstChildElement(); !child.empty(); child.nextSiblingElement()) {
+    if (child.isNode()) {
+      walk_cst(out, child);
+      continue;
+    }
+
+    if (child.id() == TOKEN_KW_MACRO_HERE) {
```
V-FEXrt, line 109: Added the `here` -> `@here` refactor because it was pretty trivial but I'm happy to revert if we want to delay the change.

tools/wake-format/wake-format.wake
JakeSiFive: I'm honestly not a big fan of the whole "native-cpp14-release" thing but I don't have strong opinions either. The only benefit is that you can do debug and non-debug builds at the same time. Currently `share` has to be relative to the wake binary in order for wake to find the standard library otherwise I'd suggest putting wake in `bin/native-cpp14-release/wake` and `bin/native-cpp14-debug/wake` but then we'd need to change how wake finds the standard library. We could solve that having `native-cpp14-debug/bin/wake` and `native-cpp14-release/bin/wake` and then symlink the share dir into `native-cpp14-debug` and `native-cpp14-release` I suppose. Then an "install" would just be a copy that resolved all symlinks.

I'm not 100% sure how wake should find the standard library but I'm not a fan of how its done now anyway.

tools/wake-format/wake-format.cpp
JakeSiFive: nope love it!

tools/wake-format/wake-format.cpp
```diff
@@ -0,0 +1,200 @@
...
+  TerminalReporter terminalReporter;
+  reporter = &terminalReporter;
+
+  // clang-format off
+  struct option options[] {
+    {'d', "debug", GOPT_ARGUMENT_FORBIDDEN},
+    {'n', "dry-run", GOPT_ARGUMENT_FORBIDDEN},
+    {'h', "help", GOPT_ARGUMENT_FORBIDDEN},
+    {'i', "in-place", GOPT_ARGUMENT_FORBIDDEN},
+    {'v', "version", GOPT_ARGUMENT_FORBIDDEN},
```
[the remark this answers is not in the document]
V-FEXrt:
```
> ./bin/wake --version                                                                                                                                                                             07/07/2022 05:47:37 PM
wake v0.27.0-alpha-74-ga80af46-dirty
```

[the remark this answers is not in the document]
mmjconolly: `#include "version.h"`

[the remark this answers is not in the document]
V-FEXrt: I'm pretty sure that file is generated based on the git tag. Do we want to couple the `wake-format` version to the `wake` version?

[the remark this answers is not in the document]
V-FEXrt: Okay, so I updated it to match what `wake` does, but somewhere in the pipeline wake adds `v` to the version (ex `wake v0.27.0-alpha-74-ga80af46-dirty` vs `wake 0.27.0-alpha-74-ga80af46-dirty` and even though I'm fairly confident I'm doing the exact same thing `wake-format` doesn't have the `v` (ex `wake-format 0.27.0~alpha.74+ga80af46~dirty`)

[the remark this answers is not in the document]
JakeSiFive: that's fine by me.

JakeSiFive (approved):

tools/wake-format/wake-format.cpp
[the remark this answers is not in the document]
JakeSiFive:
```
jakee@gamma06:/scratch/jakee/wake$ ./bin/wake --version
wake v0.27.0-alpha-62-ga69c43e-dirty
jakee@gamma06:/scratch/jakee/wake$ ./bin/wake.native-cpp14-release --version
wake 0.27.0~alpha.59+g536f5be
```

You can look in Makefile to see where the bootstrap version is printed.

tests/wake-format/basic/pass.sh
```diff
@@ -0,0 +1,3 @@
+#! /bin/sh
+
+../../../bin/wake-format.native-cpp14-release basic.wake
```
V-FEXrt, line 3: Anyone have any ideas on how to do this without making major changes to `tests/tests.wake`?

JakeSiFive: You can add a `$2` to the testing framework to be wake-fromat. That's the only real hack here. wake-unit is even more special because `wake-unit` doesn't wind up in the install like we want wake-format to.

Really we need to redesign the output API of wake's build I think. It would be better if `$1` was an install directory so you'd say `$1/bin/wake` and then we had `github.com/wake:/native-cpp14.release/bin/wake` as the default wake location so that each variant gave you a release and then `$1` could be gathered from the variant in the test code. That's a bit of major yak shave thoguh.

PR #933, opened 2022-07-07
Move Xoshiro256 to wcl

V-FEXrt:
Moves `Xoshiro256` to wcl. `Xoshiro256` is currently used to generate unique file names for `job-cache`. We also need to generate unique filenames for `wake-format` so it is being moved to wcl to make it generally available

### DIFF
--- a/src/wcl/xoshiro256.h
+++ b/src/wcl/xoshiro256.h
@@ -35,6 +35,17 @@ static std::string to_hex(const T *value) {
   return name;
 }
 
+// Use /dev/urandom to get a good seed
+static std::tuple<uint64_t, uint64_t, uint64_t, uint64_t> get_rng_seed() {
+  auto rng_fd = UniqueFd::open("/dev/urandom", O_RDONLY, 0644);
+  uint8_t seed_data[32] = {0};
+  if (read(rng_fd.get(), seed_data, sizeof(seed_data)) < 0) {
+    log_fatal("read(/dev/urandom): %s", strerror(errno));
+  }
+  uint64_t *data = reinterpret_cast<uint64_t *>(seed_data);
+  return std::make_tuple(data[0], data[1], data[2], data[3]);
+}
+
 // Adapted from wikipedia's code, which was adapted from
 // the code included on Sebastiano Vigna's website for
 // Xoshiro256**. Xoshiro256** is a modern, efficent, and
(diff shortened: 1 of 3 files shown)


src/wcl/xoshiro256.h
```diff
@@ -20,7 +20,7 @@
 #include <cstdint>
 #include <tuple>
 
-#pragma once
```
V-FEXrt, line 23: it's not shown in the diff, but `#pragma once` was listed twice

JakeSiFive: oh my bad lol

JakeSiFive: Xoshiro256 should be renamed `xoshiro_256` to fit the wcl naming convention.

JakeSiFive (approved):

src/wcl/xoshiro256.h
JakeSiFive: Also you should change `step` to be `operator()` instead so that it can be used with C++ <random> distributions. This is basically like mersene_twister but with a tiny fraction of the memory foot print, vastly improved speed, and improved statistical properties. It would be nice to use it with `std::uniform_int_distriution` and `std::uniform_real_distribution` in the future I think.

mmjconolly:
Please try to fill out the description fields for the PRs
(I don't know what Xoshiro256 is)

JakeSiFive:
Ah yee sorry, Xoshiro256 is a psuedo random number generator. It's part of how I create temporary files in job-cache and since we need temp file names in other places too we're moving it here to a common location.

Basically we give it a high quality seed using /dev/urandom, and then it's a very high quality PRNG from there. Basically /dev/urandom picks are starting point in a very long sequence defined by Xoshiro256 and Xoshiro256 just cycles through that sequence. This lets you call /dev/urandom once and then get fast random values from there. Admittedly it would be just as good to do a big integer increment after you get a single large random value from /dev/urandom but this has other uses and is just about as little code. Understanding it is not needed, it's copied from a reference implementation. The only real functionality requirement for temp file generation (which is its current use case) is that it have a *very* long period so that the sequences don't overlap. Because we have a 256-bit period, the odds of any two 2^64 bit sub-sequences overlapping is close to the chance of two 128-bit UUIDs colliding...e.g. it just ain't happening. This avoids all the issues of mktemp, all the atomicity issues, and means that we don't have to check the file system or make any syscalls etc... we just generate a unique file ID in our personal slice of the Xoshiro256 sequence.

JakeSiFive (approved):
This at least motiviates me to work on `result`

PR #934, opened 2022-07-07
CST: including trailing comments in the top node

terpstra:
The code was previously erroneously expanding the token range of
the last node, when it should have been expanding the top node.

### DIFF
--- a/src/parser/cst.cpp
+++ b/src/parser/cst.cpp
@@ -146,8 +146,8 @@ CST::CST(CSTBuilder &&builder) : token_starts(builder.token_starts) {
   }
 
   // parseWake filters out COMMENT and WS tokens from the stream passed to the lexer
-  // This means that any leading or trailing comments in the file are orphaned by the lexer parse tree.
-  // This expands the range of the top node to include those tokens.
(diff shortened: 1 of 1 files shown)


terpstra:
I hate your linter.

V-FEXrt:
I can run the format on my side if you want.

V-FEXrt:
```
git diff                                                                                                                                                                                         07/07/2022 01:41:38 PM
diff --git a/src/parser/cst.cpp b/src/parser/cst.cpp
index 53df507..40bb971 100644
--- a/src/parser/cst.cpp
+++ b/src/parser/cst.cpp
@@ -146,8 +146,8 @@ CST::CST(CSTBuilder &&builder) : token_starts(builder.token_starts) {
   }
 
   // parseWake filters out COMMENT and WS tokens from the stream passed to the lexer
-  // This means that any leading or trailing comments in the file are orphaned by the lexer parse tree.
-  // This expands the range of the top node to include those tokens.
+  // This means that any leading or trailing comments in the file are orphaned by the lexer parse
+  // tree. This expands the range of the top node to include those tokens.
   if (!nodes.empty()) {
     nodes.front().begin = 0;
     nodes.front().end = file->segment().size();
```

PR #937, opened 2022-07-08
Remove v and add '-bootstrap' to bootstrap build

V-FEXrt:
1) `wake --version` prepended `v` for the bootstrap build while the main build didn't. This brings them in alignment with no `v`

2) Adds `-bootstrap` to the bootstrap build so that it explicitly calls out that its the bootstrap build and not the main one


V-FEXrt:
I confirmed that `make install` uses the "non-bootstrap" build so should be good there.

Not sure which build the release process uses but I'm assuming its the "non-bootstrap"?

JakeSiFive (approved):

PR #939, opened 2022-07-11
test: Provide dir to bin/ instead of wake as $1

V-FEXrt:
Instead of setting `$1` to `<something>/bin/wake` set it to `<something>/bin`.
This allows tests to execute any of the tools installed by wake.

Also updates all wake tests to point to bin/wake.

It is mainly needed to support wake-format tests but may be used for
other tools in the future.

### DIFF
--- a/tests/tests.wake
+++ b/tests/tests.wake
@@ -35,11 +35,12 @@ export topic wakeTestBinary: Unit => Result (Pair String (List Path)) Error
 # the way we test currently.
 export topic wakeUnitTestBinary: (variant: Pair String String) => Result (List Path) Error
 
-def wakeToTestDir Unit =
-    require buildTestWake, Nil = subscribe wakeTestBinary
-    else failWithError "wake binary to test not found"
-    require Pass (Pair path visible) = buildTestWake Unit
-    Pass (Pair (simplify "{path}/..") visible)
+def wakeToTestDir Unit = match (subscribe wakeTestBinary)
+    buildTestWake, Nil =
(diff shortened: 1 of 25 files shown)


tests/tests.wake
```diff
@@ -202,11 +204,11 @@ def runTest (testScript: Path): Result Unit Error =
...
-        wakeToTest Unit
+    require Pass (Pair wakeDir wakeVisible) =
+        wakeToTestDir Unit
 
     def testJob =
-        makeExecPlan ("./{testScript.getPathName.inTestDir}", wakeBin.inTestDir, Nil) (visibleFiles ++ wakeVisible)
+        makeExecPlan ("./{testScript.getPathName.inTestDir}", wakeDir.inTestDir, Nil) (visibleFiles ++ wakeVisible)
         | setPlanDirectory testDirectory
         | setPlanLabel "testing: {testName}"
         | setPlanLocalOnly True # On OS/X you cannot mount fuse within fuse
```
JakeSiFive, line 214: I guess this line here is saving us because otherwise this is problematic. The issue is that we don't have all of the generated binaries listed here. It works fine because this is running without fuse so all files are visible.

Ideally the test code would consume the binary paths from somewhere.

Maybe we want two interfaces 1) one that uses claimFileAsPath and a directory and 2) one that takes a list of visible binaries and extracts the directory from that list

JakeSiFive (approved):
Ok, this is a tiny bit of tech debt but I don't think the yak is worth shaving this time.

PR #940, opened 2022-07-12
[core] add `Map` data structure

ag-eitilt:
A homebrew `List (Pair k v)` may work for most cases, so long as the set of values remains small, but for general use the time complexity becomes undesirable; a `Tree (Pair k v)` would seem to solve some issues, but since the type's not specialized for comparison functions which only reference the left-hand key, lookup and some other manipulations wind up no better than the `List`. Instead, providing a `Map k v` wrapper with access to the `Tree` internals where necessary allows a full and optimal implementation of the data structure.  I would appreciate people double-checking my evaluation of the `Tree` interface via map.wake, to both be sure I've not re-implemented something which was already optimal time, and to be sure I'm not missing anywhere that can be improved by only caring about guaranteed-unique keys.

To ensure consistent behaviour no matter which of potentially several existing values in the tree are encountered first, the `*With` functions need to pass the resolution function over *all* equal values.  Unfortunately, this winds up giving them a greater time complexity (especially `*unionWith` and `*intersectWith`, which I think approach exponential) than their collision-dropping counterparts.  That can't even be simplified away with `Map` enforcing unique keys, as the comparison function is tied to the specific map, not the type (`munionWith fn scmpMap scmpIdentifierMap` could result in several items -- even just from `scmpMap` -- collapsing into each other due to the simpler canonical forms of the result).  Still, the ability to determine what to do when a conflict exists can be helpful enough that the time complexity is worth it.

If we could bikeshed before this gets merged, "Map" is certainly the more common terminology for this, but I wonder if we'd rather use "Dictionary" for the data structure to avoid confusion with the `*map` category of functions.  The fact that they're both doing the same thing mathematically doesn't help people who internalized the imperative "a function is a collection of instructions operating on inputs" definition rather than the functional/algebraic "a function is a mapping from a set of inputs to a consistent output".  Also, `mmap` standing for `mapmap` seems a bit imprecise.

### DIFF
--- a/share/wake/lib/core/map.wake
+++ b/share/wake/lib/core/map.wake
@@ -1,0 +1,561 @@
+# Copyright 2022 SiFive, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You should have received a copy of LICENSE.Apache2 along with
+# this software. If not, you may obtain a copy at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package wake
+
+# A dictionary associating a key-value pair, providing fast lookup by key.
+# This is internally implemented by a balanced tree, so some total ordering must
+# be able to be produced for the key type.
+data Map k v = Map (cmp: k => k => Order) (root: Tree (Pair k v))
+from wake export type Map
+
+def makeCmpPair (cmpKey: k => k => Order) (left: Pair k v) (right: Pair k v): Order =
+    cmpKey left.getPairFirst right.getPairFirst
+
+
+# Initialize an empty `Map` which will use the given total order function.
+#
+# *Parameters:*
+#   * `cmpKey`:
+#       A function providing a total ordering over values of the key type.
+#
+# *Example:*
+#   ```
+#   mnew scmp | msize = 0
+#   ```
+export def mnew (cmpKey: k => k => Order): Map k v =
+    tnew (makeCmpPair cmpKey)
+    | Map cmpKey
+
+# Construct a `Map` from the pre-associated key-value pairs in the `List`.
+# If multiple `Pair`s have the same left value (key), then the resulting `Map`
+# will contain the right value of only the *first* occurrence.
+#
+# *Parameters:*
+#   * `cmpKey`:
+#       A function providing a total ordering over values of the key type.
+#   * `pairs`:
+#       The keys and values which the `Map` should contain.  This does not have
+#       to be sorted, but each left-hand value should be unique.
+#
+# *Examples:*
+#   ```
+#   listToMap scmp ("a" → True, "b" → False, "c" → False, Nil) | msize = 3
+#   listToMap scmp ("a" → 1, "a" → 2, Nil) | mlookup "a" = Some 1
+#   ```
+export def listToMap (cmpKey: k => k => Order) (pairs: List (Pair k v)): Map k v =
+    listToTree (makeCmpPair cmpKey) pairs
+    | Map cmpKey
+
+# Construct a `Map` from the pre-associated key-value pairs in the `Vector`.
+# If multiple `Pair`s have the same left value (key), then the resulting `Map`
+# will contain the right value of only the *first* occurrence.
+#
+# *Parameters:*
+#   * `cmpKey`:
+#       A function providing a total ordering over values of the key type.
+#   * `pairs`:
+#       The keys and values which the `Map` should contain.  This does not have
+#       to be sorted, but each left-hand value should be unique.
+export def vectorToMap (cmpKey: k => k => Order) (pairs: Vector (Pair k v)): Map k v =
+    vectorToTree (makeCmpPair cmpKey) pairs
+    | Map cmpKey
+
+
+# Count how many key-value associations are contained in the `Map`.
+#
+# *Examples:*
+#   ```
+#   mnew scmp | msize = 0
+#   listToMap scmp ("a" → True, "b" → False, "c" → False, Nil) | msize = 3
+#   ```
+export def msize (map: Map k v): Integer =
+    def Map _ tree = map
+    tlen tree
+
+# Test if the `Map` does not contain any elements.
+#
+# *Examples:*
+#   ```
+#   mnew scmp | mempty = True
+#   listToMap scmp ("a" → True, "b" → False, "c" → False, Nil) | mempty = False
+#   ```
+export def mempty (map: Map k v): Boolean =
+    def Map _ tree = map
+    tempty tree
+
+
+# Add a given value into the map under the key, if that key does not already exist.
+# Any pair with the same key which already exists in the map *remains unchanged*.
+#
+# For a similar function which uses the new value provided, see `minsertReplace`
+# or `minsertWith`.
+#
+# *Examples:*
+#   ```
+#   mnew scmp | minsert "a" 2 | mlookup "a" = Some 2
+#   listToMap scmp ("a" → 1, Nil) | minsert "a" 2 | mlookup "a" = Some 1
+#   ```
+export def minsert (key: k) (value: v) (map: Map k v): Map k v =
+    def Map cmpKey tree = map
+    tinsert (Pair key value) tree
+    | Map cmpKey
+
+# Add a given value into the map under the key, whether or not it already exists.
+#
+# For a similar function which preserves the original value contained in the
+# map, see `minsert` or `minsertWith`.
+#
+# *Examples:*
+#   ```
+#   mnew scmp | minsertReplace "a" 2 | mlookup "a" = Some 2
+#   listToMap scmp ("a" → 1, Nil) | minsertReplace "a" 2 | mlookup "a" = Some 2
+#   ```
+export def minsertReplace (key: k) (value: v) (map: Map k v): Map k v =
+    def Map cmpKey tree = map
+    tinsertReplace (Pair key value) tree
+    | Map cmpKey
+
+# Add a given value into the map under the key, resolving conflicts as specified.
+#
+# If either the original value or the new will always be used, see `minsert` or
+# `minsertReplace`.
+#
+# *Examples:*
+#   ```
+#   mnew scmp | minsertWith (_+_) "a" 2 | mlookup "a" = Some 2
+#   listToMap scmp ("a" → 1, Nil) | minsertWith (_+_) "a" 2 | mlookup "a" = Some 3
+#   ```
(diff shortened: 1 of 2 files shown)


share/wake/lib/core/tree.wake
```diff
@@ -94,6 +95,28 @@ export def tinsertMulti (y: a) (Tree cmp root: Tree a): Tree a =
       _  = balanceR l x (helper r)
   Tree cmp (helper root)
 
+# Insert y into the tree, or the value resulting from fn on a collision
+# `y` is passed as the left-hand value of `fn`.
+export def tinsertWith (fn: a => a => a) (y: a) (tree: Tree a): Tree a =
```
JakeSiFive, line 100: This is by far the API I use most often with Haskell's Map

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+#
+# For a similar function which preserves the original value contained in the
+# map, see `minsert` or `minsertWith`.
+#
+# *Examples:*
+#   ```
+#   mnew scmp | minsertReplace "a" 2 | mlookup "a" = Some 2
+#   listToMap scmp ("a" → 1, Nil) | minsertReplace "a" 2 | mlookup "a" = Some 2
+#   ```
+export def minsertReplace (key: k) (value: v) (map: Map k v): Map k v =
```
JakeSiFive, line 130: minor nit, I think replacing is more common so `minsertOnce` might be a better name for the above and this should be `minsert`, alternatively this could also be called `mset`

V-FEXrt:
Yak Shave Vote: I think Dictionary would be better than Map. Map is too overloaded in the context of functional programming and it makes it difficult to talk about when speaking. If Dictionary is too long Dict is a reasonable shortening

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+    def Map _ leftTree = left
+    # tintersect uses the comparison function of the right tree, so similarly
+    # attach the raw key comparison function of the right map to the output
+    def Map cmpKey rightTree = right
+    tintersect leftTree rightTree
+    | Map cmpKey
+
+# Remove all keys from the left map which do not occur in the right, joining values.
+# If the key comparison function differs between the two inputs, then the one
+# from the `right` is used.
```
JakeSiFive, line 527: I think this comment and the examples are for regular mintersect?

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+#   ```
+export def munion (left: Map k v) (right: Map k v): Map k v =
+    def Map _ leftTree = left
+    # tunion uses the comparison function of the right tree, so similarly attach
+    # the raw key comparison function of the right map to the output
+    def Map cmpKey rightTree = right
+    tunion leftTree rightTree
+    | Map cmpKey
+
+# Collect all key-value associations in maps with the given conflict resolultion.
```
JakeSiFive, line 451: I think this comment is for munion and not munionWith?

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+export def mfilter (fn: k => v => Boolean) (map: Map k v): Map k v =
+    def pairFn (Pair k v) = fn k v
+    editMapData (tfilter pairFn) map
+
+
+# Collect all key-value associations in either of two maps into a single one.
+# If the same key occurs in both, the value from `left` is kept and the one from
+# `right` is discarded.  However, if the key comparison function differs between
+# the two inputs, then the one from the `right` is used.
+#
```
JakeSiFive, line 433: Maybe you meant to use munionWith in the comment here?

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+# *Examples:*
+#   ```
+#   def left  = listToMap scmp ("a" → 1, "b" → 2, Nil)
+#   def right = listToMap scmp ("b" → 11, "f" → 15, Nil)
+#
+#   munionWith (\_\lv\rv lv + rv) left right | mlookup "a" = Some 1
+#   munionWith (\_\lv\rv lv + rv) left right | mlookup "f" = Some 15
+#   munionWith (\_\lv\rv lv + rv) left right | mlookup "b" = Some 13
+#   ```
+export def munionWith (fn: k => v => v => v) (left: Map k v) (right: Map k v): Map k v =
```
JakeSiFive, line 467: The value type should be allowed to change to a v2 here I think?

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+# *Examples:*
+#   ```
+#   def left  = listToMap scmp ("a" → 1, "b" → 2, Nil)
+#   def right = listToMap scmp ("b" → 11, "f" → 15, Nil)
+#
+#   mintersectWith left right | mlookup "a" = None
+#   mintersectWith left right | mlookup "f" = None
+#   mintersectWith left right | mlookup "b" = Some 13
+#   ```
+export def mintersectWith (fn: k => v => v => v) (left: Map k v) (right: Map k v): Map k v =
```
JakeSiFive, line 538: The value type should be allowed to change to a v2 here I think?

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+#
+# *Examples:*
+#   ```
+#   mnew scmp | mfoldmap (_+_) 0 (\_\v v) = 0
+#   listToMap scmp ("a" → 1, "b" → 2, Nil) | mfoldmap (_+_) 0 (\_\v v) = 3
+#   listToMap scmp ("a" → 1, "b" → 2, Nil) | mfoldmap ("{_} {_}") "k=v:" ("{_}={str _}") = "k=v: a=1 b=2"
+#   ```
+export def mfoldmap (combineFn: a => a => a) (base: a) (transformFn: k => v => a) (map: Map k v): a =
+    def pairFn (Pair k v) = transformFn k v
+    tfoldmap combineFn base pairFn map.getMapData
```
JakeSiFive, line 228: I wonder if a nicer interface is `combineFn: k => v => a => a`, I can see myself using `Pair` the `transformFn` a lot with this current interface.

Your interface does have a classic mapReduce style to it though which is nice.

ag-eitilt:
Other than the tuple usage in the second commit and the documentation, this is almost exactly what I had written in staging; the only logical differences I'm seeing are that `munion` directly uses `tunion` rather than `munionWith`, that the `tunionWith` implementation has been rewritten since the original didn't consider multi-trees, and that `*insertWith` and `*intersectWith` have both been added following the model of (the new) `tunionWith`.

share/wake/lib/core/map.wake
```diff
@@ -0,0 +1,548 @@
...
+#   mnew scmp | minsertReplace "a" 2 | mlookup "a" = Some 2
+#   listToMap scmp ("a" → 1, Nil) | minsertReplace "a" 2 | mlookup "a" = Some 2
+#   ```
+export def minsertReplace (key: k) (value: v) (map: Map k v): Map k v =
+    editMapData (tinsertReplace (Pair key value)) map
+
+# Add a given value into the map under the key, resolving conflicts as specified.
+#
+# If either the original value or the new will always be used, see `minsert` or
+# `minsertReplace`.
```
JakeSiFive, line 136: This comment needs some work I think.

JakeSiFive (commented):
I *really* like this Map interface. Very complete and more or less what I would have done. It has the nice functions like minsert and mlookup but it also has the really helpful powerful functions I like such as minsertWith. It even has all the set-wise operations!

share/wake/lib/core/map.wake
ag-eitilt: I would definitely agree, but this is the pattern we already had in the `Tree` interface, so I wanted the consistency.

share/wake/lib/core/map.wake
ag-eitilt: Anything which only occurs in either the left or the right but not both isn't going to trigger the function, so the type needs to remain the same, unless we add a function taking one less argument to cover those cases.

share/wake/lib/core/map.wake
ag-eitilt: This one could indeed become a `v2`, though.  My only argument would be for consistency with the others, but maybe the abstraction is better.

JakeSiFive:
> Bikeshed Vote: I think Dictionary would be better than Map. Map is too overloaded in the context of functional programming and it makes it difficult to talk about when speaking. If Dictionary is too long Dict is a reasonable shortening

I think I disagree. Map is what's used for this in ocaml and Haskell. The other other thing is a map function but a Map data structure always means this I think. Happy to be out-voted but I think `Map` or `OrderedMap` are best. `Dictionary` sounds like a hash table to me which this isn't.

V-FEXrt:
`Dictionary` sounds like a Key -> Value lookup table to me
`HashMap` sounds like a hash table :)

I'm fine with `Map` or `OrderedMap` as well. I just sometimes have difficulty in *verbal* communication when talking about `map` vs `Map`. Its not an issue in textual communication.

share/wake/lib/core/map.wake
JakeSiFive: I don't see a lot of use cases for being more abstract but I figure why limit it here. Up to you, minor issue.

share/wake/lib/core/map.wake
JakeSiFive: I guess foldl/foldr are just more generic than foldmap. You could implement this as follows right?

```
def fn k v a = combineFn (transformFn k v) a
mfoldr fn base map
```

So I guess this is fine, its just a decomposition some people might prefer.

JakeSiFive (approved):
Please make sure those comments are corrected but I'm happy with this! This is a huge win for wake users!

share/wake/lib/core/map.wake
ag-eitilt: The main difference is in execution order; you could certainly reimplement the results and signature of foldmap, but it has a more branched behaviour where it executes the `combineFn` as soon as all neighbors have been completed, rather than adding one element at a time.

share/wake/lib/core/map.wake
ag-eitilt: I'm not quite sure what part you're specifically pointing at as needing work.  Mind giving a bit more detail?

ag-eitilt (commented):
At this point, I'm (somewhat regretfully) thinking we should stick with `Map` over `Dictionary`; it might be a bit harder to talk about the data structure vs the mathematical concept, but it seems like Python and C# are the only major languages which call them dictionaries.

share/wake/lib/core/map.wake
JakeSiFive: It took me like 4 reads to parse it, my initial impression was that it wasn't grammatical.  Maybe

"If just replacing or keeping the original, consider using `minsert` or `minsertReplace` instead"

JakeSiFive (approved):

share/wake/lib/core/map.wake
ag-eitilt: I like that wording, thanks!

PR #946, opened 2022-07-14
wake-format: Implement basic tree walk/formatting API

V-FEXrt:
This PR steps up a basic framework for walking the tree while formatting it. Before going off and actually waking the tree/generating outputs I'd like to get the API reviewed and landed. This will hopefully minimize rewrites later.


src/wcl/rope.h
```diff
@@ -126,6 +126,7 @@ class rope_builder {
 
  public:
   void append(const char* str) { r = r.concat(rope::lit(str)); }
+  void append(std::string str) { r = r.concat(rope::lit(str)); }
```
JakeSiFive, line 129: you can move that string here, you might as well delete the `const char*` above when you do that since at's what happens here anyway. Later on we made make an `append_static` as an optimization for string_views  that have a static lifetime...again another thing rust does better...sigh.

tools/wake-format/emitter.h
```diff
@@ -0,0 +1,78 @@
...
+
+  struct nest_t {
+    ctx_t* ctx;
+
+    nest_t(ctx_t* ctx) : ctx(ctx) { ctx->nest_level++; }
+
+    ~nest_t() { ctx->nest_level--; }
+  };
+
+  Emitter(std::ostream* ostream) : ostream(ostream) { assert(ostream != nullptr); }
```
JakeSiFive, line 50: Now that we're returning ropes, can we just avoid the ostream? Like we compute the correct rope and then that rope can write to the output stream later. We always have a place to write to and we memoize so I think it decouples ostream creation from the emitter. The thing I didn't like previously was something (I've forgotten what now) that you were forced to do because of needing the ostream.

tools/wake-format/emitter.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "parser/cst.h"
+
+class Emitter {
+ public:
+  struct ctx_t {
+    int nest_level = 0;
+    bool is_flat = false;
+
+    ctx_t flat() {
```
JakeSiFive, line 35: Maybe instead of having a `nest_t` we instead just make an `nest()` method like `flat()`, that way its completely functional and we don't need nest_t. That said maybe you want to use `ctx` almost everywhere and temporarily mutating a ctx is a nicer interface. Up to you.

tools/wake-format/emitter.h
```diff
@@ -0,0 +1,78 @@
...
+  // Walks the CST, formats it, and writes it to ostream
+  void layout(CST cst);
+
+ private:
+  // Top level tree walk. Dispatches out the calls for various nodes
+  wcl::optional<wcl::rope> walk_top(ctx_t ctx, CSTElement node);
+
+  // Returns a nest_t instance that increases the indent level
+  // until the instance goes out of scope
+  nest_t nest(ctx_t ctx);
```
JakeSiFive, line 61: How does this work if it doesn't take a pointer or reference?

tools/wake-format/emitter.h
```diff
@@ -0,0 +1,78 @@
...
+
+  // Emits a newline and any indentation needed
+  wcl::rope newline(ctx_t ctx);
+
+  // Emits a space character `count` times
+  wcl::rope space(ctx_t ctx, uint8_t count = 1);
```
JakeSiFive, line 67: does this need a context?

tools/wake-format/emitter.h
```diff
@@ -0,0 +1,78 @@
...
+
+  // TODO: memoize this function
+  // returns the subtree with all newlines and indentation removed
+  // if possible, None otherwise
```
JakeSiFive, line 71: minor nit: I think this doesn't care about indentation, its just that indentation follows newlines of which there are none here.

tools/wake-format/wake-format.cpp
```diff
@@ -169,35 +168,34 @@ int main(int argc, char **argv) {
...
+    auto output_file = std::make_unique<std::ofstream>(tmp);
+    Emitter emitter(output_file.get());
+    emitter.layout(cst);
+
+    output_file.reset();
 
-    // When editing in-place we need to copy the tmp file over the original
     if (in_place) {
-      output_file.reset();
+      // When editing in-place we need to copy the tmp file over the original
```
JakeSiFive, line 192: nit: "rename it over"

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,106 @@
...
+  wcl::rope_builder builder;
+  for (uint8_t i = 0; i < count; i++) {
+    builder.append(" ");
+  }
+  return std::move(builder).build();
+}
+
+wcl::optional<wcl::rope> Emitter::flat(ctx_t ctx, CSTElement node) {
+  return walk_top(ctx.flat(), node);
+}
```
JakeSiFive, line 51: I'm wondering now if we're going to have several things like top, expression, etc... that we need to memorize for separately. No request here just a thought. It might mean that we don't have a single `flat` method just we always use `ctx.flat()` to request the flat version of something.

This would be needed if a CST element needs to be contextually formatted a certain way that isn't apt for encoding in the context. I can't think of any examples now but worth keeping in mind that we have that out if we need it.

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,106 @@
...
+  ctx_t ctx;
+  wcl::optional<wcl::rope> r = walk_top(ctx, cst.root());
+  r->write(*ostream);
+}
+
+wcl::optional<wcl::rope> Emitter::walk_top(ctx_t ctx, CSTElement node) {
+  wcl::rope_builder builder;
+  for (CSTElement child = node.firstChildElement(); !child.empty(); child.nextSiblingElement()) {
+    switch (child.id()) {
+      case CST_CASE: {
```
JakeSiFive, line 63: Can we make this a pure dispatch? That way each CST node type gets its own method.

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,106 @@
...
+          assert(full_child);
+          builder.append(*full_child);
+        } else {
+          builder.append(child.fragment().segment().str());
+        }
+        break;
+      }
+    }
+  }
+  return wcl::optional<wcl::rope>(wcl::in_place_t{}, std::move(builder).build());
```
JakeSiFive, line 105: This is so stupidly efficient and I love it....of course rust makes this sort of thing trivial but this is hard to do without modern C++

tools/wake-format/emitter.h
V-FEXrt: I'm not sure I follow, we need to remove the indentation for flat to work no?

tools/wake-format/emitter.h
JakeSiFive: was this resolved? I still see a context here.

tools/wake-format/emitter.h
JakeSiFive: Oh yeah I guess when a soft newline is emitted you don't emit the indentation true. I guess I think of indentation and newlines are being related things so the entire `newline()` call would just return when the context is falt. But you're right that also means indentation is ignored too.

tools/wake-format/emitter.cpp
V-FEXrt: Agreed :)

tools/wake-format/emitter.cpp
JakeSiFive: Yeah exactly. I think we probably only need the things that dispatch to memoize. Like `expression` will memoize but that might call `match` which calls `case` which then calls `expression` again and I think only expression needs to memoize there. The cool thing is that memoizing is fairly simple since the CSTElements contain an index and the context is easily hashable/orderable and ropes are very cheap to copy construct and move.

tools/wake-format/emitter.h
V-FEXrt: It is resolved if you look at the full diff. for some reason it's not showing here though

src/wcl/rope.h
```diff
@@ -58,6 +60,9 @@ class rope_impl_pair : public rope_impl_base {
     pair.first->write(ostream);
     pair.second->write(ostream);
   }
+  void drop_last() override {
```
JakeSiFive, line 63: I think this would be just as efficient if it returned an std::shared_ptr instead. The method on the rope can still be a mutation but I think the impls should be immutable.

But also I'm wondering if this is really what's needed? Why do we want this method?

tools/wake-format/emitter.cpp
```diff
@@ -28,7 +28,7 @@
 wcl::rope Emitter::newline(ctx_t ctx) {
   wcl::rope_builder builder;
 
-  builder.append("\n");
+  builder.append("⏎\n");
```
JakeSiFive, line 31: What's up with this? Is this a carriage return or something? Is just only for debug?

tests/wake-format/basic/stdout
```diff
@@ -1,168 +1,12 @@
...
-from test_wake import topic wakeUnitTestBinary
-publish wakeTestBinary = defaultWake, Nil
-publish wakeUnitTestBinary = buildWakeUnit, Nil
-
-def defaultWake Unit =
-    require Pass wakeVisible =
-        doInstall "tmp" "default"
-
-    Pass (Pair "tmp/bin/wake" wakeVisible)
+def·name·Unit·=⏎
```
JakeSiFive, line 1: Is the plan to eventually go back to spaces and regular newlines?

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,217 @@
...
+  return std::move(builder).build();
+}
+
+wcl::optional<wcl::rope> Emitter::flat(ctx_t ctx, CSTElement node) {
+  return walk_node(ctx.flat(), node);
+}
+
+wcl::rope Emitter::try_flat(ctx_t ctx, CSTElement node) {
+  auto flat_node = flat(ctx, node);
+  if (flat_node && flat_node->size() + ctx.nest_level * space_per_indent < 80) {
```
JakeSiFive, line 53: Shouldn't this be worried about the current column? I think your ctx needs to know the current column in order to figure this out.

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+wcl::rope Emitter::layout(CST cst) {
+  ctx_t ctx;
+  return walk_top(ctx, cst.root()).concat(newline(ctx));
+}
+
+wcl::rope Emitter::walk_top(ctx_t ctx, CSTElement node) {
+  wcl::rope_builder builder;
+
+  for (CSTElement child = node.firstChildElement(); !child.empty(); child.nextSiblingElement()) {
```
JakeSiFive, line 73: hmm...does it not matter what type of node this is? It seems to me like you should look at the top of `node` before traversing its children right?

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    if (child.isNode()) {
+      builder.append(try_flat(ctx, child));
+      builder.append(newline(ctx));
+    } else {
+      builder.append(walk_token(ctx, child));
+    }
+  }
+
+  // probably shouldn't do this unconditionally
+  builder.drop_last();
```
JakeSiFive, line 83: Very concerned about what this is doing, makes me think we got part of this design wrong.

src/wcl/rope.h
V-FEXrt: API convenience mostly. There are several cases where you have loops of things with a common separator that shouldn't have a trailing instance of the separator

ie

`a <space> b <space> c <space>`

`drop_last` drops the last rope, which in every use case has been the last separator

We could avoid it but partly unrolling the loop but I didn't really want to write the code for that

tools/wake-format/emitter.cpp
V-FEXrt: just debugging, I'll put it behind the debug flag. Can I do that in the next PR though?

tests/wake-format/basic/stdout
V-FEXrt: Yeah, the formatter is in this weird middle state where some spaces/newlines are emitted by the formatter and some are from the original source. This helps to tell which is which. Once the formatter is 'done' all the spaces/newlines should be emitted by the formatter and none from the original source but we aren't quite there yet.

In the next PR, I'm moving these behind the debug flag

tools/wake-format/emitter.cpp
V-FEXrt: Right, this isn't going to correctly determine column width. this is mostly just to test that _really_ long lines are forced to be split. This will need to be fixed but I think it can be deferred for the moment

src/wcl/rope.h
JakeSiFive: I see. Let's only use it when we have code like this then

```
for (...) {
...
  space(builder);
}
builder.drop_last();
```

or

```
for (...) {
...
  builder.emit(",");
}

builder.drop_last();
```

I think `drop_last` only works correctly right now on builder because builder concats using pair in a certain way. Maybe builder needs a stack of ropes and an "undo" instead?

We could also make a special function for doing these kinds of loops instead but you likely need a couple since the CST doesn't traverse as nicely as a vector.

tools/wake-format/emitter.cpp
V-FEXrt: Eh yes, kinda but not this case. The follow up PR is much more explict though.

In this case,  this is only every called on `TOP` and `BLOCK` nodes. Both of which are just a list of other nodes, no specially handling required.

tools/wake-format/emitter.cpp
JakeSiFive: The above can't know that a newline/space was just output I think right? I think all of these things, all the walk functions that is, should not output trailing newline/slashes. They never know if they need to do that or not.

tools/wake-format/emitter.cpp
JakeSiFive: hmm, ok lets chat about your more complete local code today. I think this is a fundamental issue we need to make sure we solve.

src/wcl/rope.h
V-FEXrt: yeah, it's only being done like that so far. We can keep an eye out for it in review and kinda feel it out for a better API over time. I wasn't a big fan of it either but it seemed like the most sensical way to do it

tools/wake-format/emitter.cpp
JakeSiFive: I see. That makes more sense. Can we name this 'walk block' instead then? I think of `top` as just being a top level block anyway. Also comments saying which node/token types each walk function handles would be nice when they handle more than one.

tools/wake-format/emitter.cpp
V-FEXrt: Line 76 is the newline getting dropped here so we aren't assuming some recursive call added it. My local code doesn't have the if/else anymore so it is always dropping the `newline()`

tools/wake-format/emitter.cpp
JakeSiFive: Please make the column changes we discussed to make rope track column and fix this line

nit: Also let's make the line length a variable and set it to 100 not 80. I did a survey of wake code a while back and found that 120 is probably better for wake code though since we hit that today in practice even with heavy indentation since somethings like the first line of a def can't be broken up right now.

JakeSiFive (commented):
Ok looks good, just had a couple things we had to work out. I'm super happy with how this API looks in practice! Thanks for showing me that.

V-FEXrt:
Done, PTAL

tools/wake-format/emitter.cpp
V-FEXrt: There is actually a subtle difference between walking `top` and `block` in the follow up PR "walk top" is just called `walk` and "walk block" is called `walk_block`.

JakeSiFive (approved):
I think I'll have some time today to write up a PR for how I think builder.column() should look. This looks good for now. Let the onslaught of formatting boilerplate roll in now.

tools/wake-format/emitter.cpp
```diff
@@ -0,0 +1,217 @@
...
+  return std::move(builder).build();
+}
+
+wcl::optional<wcl::rope> Emitter::flat(ctx_t ctx, CSTElement node) {
+  return walk_node(ctx.flat(), node);
+}
+
+wcl::rope Emitter::try_flat(ctx_t ctx, CSTElement node) {
+  auto flat_node = flat(ctx, node);
+  if (flat_node && flat_node->column() <= max_column_width) {
```
JakeSiFive, line 53: This isn't the correct condition. I don't think `try_flat` is ever possible since you need to know the context its being inserted into. I guess this can curtail the search space somewhat but the maximum search space isn't that big once we memoize.

JakeSiFive (approved):

tools/wake-format/emitter.cpp
V-FEXrt: This function is a bit rewritten in the next PR (though still wrong if this is wrong) can we address there?

JakeSiFive: Yeah this is approved. I'm confident that the fix to this bug is nothing catastrophic which is all I was concerned about with the column stuff in the first place. This PR is about getting the API right.

PR #950, opened 2022-07-20
wake-format: Add tree walk boilerplate code

V-FEXrt:
Adds all the boilerplate code required to walk the CST. These changes are pulled out of of much larger set of changes for ease of review.

Formatting output is currently messed up but will be fixed in the follow up PR

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -25,10 +25,30 @@
 
 #include "parser/parser.h"
 
+#define ASSERT_TOKEN(node, token) \
+  assert(!node.empty());          \
+  assert(node.id() == token);
+
+#define APPEND_OR_BUBBLE(func, ctx, node) \
+  {                                       \
+    auto subtree = func(ctx, node);       \
+    if (!subtree) {                       \
+      return {};                          \
+    }                                     \
+    builder.append(*subtree);             \
+  }
+
+#define CONSUME_WS_NL(node)                                                     \
+  {                                                                             \
+    while (!node.empty() && (node.id() == TOKEN_WS || node.id() == TOKEN_NL)) { \
+      node.nextSiblingElement();                                                \
+    }                                                                           \
+  }
+
 wcl::rope Emitter::newline(ctx_t ctx) {
   wcl::rope_builder builder;
 
-  builder.append("⏎\n");
+  builder.append("\n");
   for (int i = 0; i < ctx.nest_level; i++) {
     builder.append(space(space_per_indent));
   }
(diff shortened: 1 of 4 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -39,123 +59,75 @@ wcl::rope Emitter::newline(ctx_t ctx) {
...
-  wcl::rope_builder builder;
-
-  for (CSTElement child = node.firstChildElement(); !child.empty(); child.nextSiblingElement()) {
-    if (child.isNode()) {
-      builder.append(try_flat(ctx, child));
-      builder.append(newline(ctx));
-    } else {
-      builder.append(walk_token(ctx, child));
-    }
+  if (flat_node->column() > max_column_width) {
```
V-FEXrt, line 74: @JakeSiFive  Happy to this condition here if it is still incorrect in the scope of the function changes here

tools/wake-format/emitter.cpp
```diff
@@ -25,10 +25,30 @@
...
+  assert(!node.empty());          \
+  assert(node.id() == token);
+
+#define APPEND_OR_BUBBLE(func, ctx, node) \
+  {                                       \
+    auto subtree = func(ctx, node);       \
+    if (!subtree) {                       \
+      return {};                          \
+    }                                     \
+    builder.append(*subtree);             \
```
JakeSiFive, line 38: This assume `builder` is the name of the builder which is asking for trouble as soon as someone uses a different name. For instance you didn't do this with 'ctx' despite using that name consistently (probably because of ctx.flat() and such though).

tools/wake-format/emitter.cpp
JakeSiFive: You can't get the correct condition in this context. You always have to look at the context that you're appending it into to know if we can you do this or not. You always have to add the size (or column) with the current builder column to know what the new last column will be. What you have now is a context free approximation.

tools/wake-format/emitter.cpp
```diff
@@ -39,123 +59,75 @@ wcl::rope Emitter::newline(ctx_t ctx) {
...
-
-  wcl::rope_builder builder;
-
-  CSTElement child = node.firstChildElement();
-  ASSERT_TOKEN(child, CST_ID);
-  {
-    auto node = walk_node(ctx, child);
-    assert(node);
-    builder.append(*node);
+wcl::optional<wcl::rope> Emitter::flat_or(ctx_t ctx, CSTElement node) {
```
JakeSiFive, line 81: If you had `builder` ref here you could possible make this work for some context but you really need that + a function to call. Not sure you can make a flat_or function with this API but you can still write "flat or" code

```
auto flat_foo = walk_foo(ctx.flat(), foo_node);
if (flat_foo && builder.column() + flat_foo.size() <= max_column) {
  builder.append(*flat_foo);
} else {
  ctx_t sub_ctx = ctx.indent(); // Sometimes you might not want to indent here
  builder.append(newline(sub_ctx));
  builder.append(walk_foo(sub_ctx));
}
```

The things that change are the `walk_foo` and weather you want to indent or not. Two macros is probably sufficient, one for indent and one for not indenting.

tools/wake-format/emitter.cpp
V-FEXrt: actually, I lied, fixing this in the next PR, but the change there is just
```  
if (flat_node->column() + (ctx.nest_level * space_per_indent) > max_column_width) {
  return {};
}
```
which accounts for the missing indentation at the start of the line

V-FEXrt: Ah, hmm, that's not good enough either. I think I'm starting to see why we need the `builder->column()`

JakeSiFive (approved):
Only worried about the builder thing, happy to have the column stuff fixed next.

tools/wake-format/emitter.cpp
```diff
@@ -139,118 +139,118 @@ wcl::optional<wcl::rope> Emitter::walk_node(ctx_t ctx, CSTElement node) {
...
-      APPEND_OR_BUBBLE(walk_topic, ctx, node);
+      APPEND_OR_BUBBLE(builder, walk_topic, ctx, node);
       break;
     case CST_TUPLE:
-      APPEND_OR_BUBBLE(walk_tuple, ctx, node);
+      APPEND_OR_BUBBLE(builder, walk_tuple, ctx, node);
       break;
     case CST_TUPLE_ELT:
-      APPEND_OR_BUBBLE(walk_tuple_elt, ctx, node);
+      APPEND_OR_BUBBLE(builder, walk_tuple_elt, ctx, node);
```
JakeSiFive, line 247: sorry for this 😭 

Its more correct now though

JakeSiFive (approved):

PR #951, opened 2022-07-21
CI: Update MacOS build image

ag-eitilt:
According to actions/virtual-environments#5583, the MacOS 10.15 image is deprecated; we're currently in the middle of a brownout which is keeping PRs from merging for another 24 hours.  Per Zoom discussion, this probably also needs changing the `osxfuse` to `macfuse`, but I'm trying first without it just to see if it works.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,7 +6,7 @@ on:
 jobs:
     build:
         name: "MacOS build"
-        runs-on: macos-10.15
+        runs-on: macos-12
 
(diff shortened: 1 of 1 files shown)


JakeSiFive:
I'll take a look into seeing if I can setup macos CI on circleci unless github responds with a solution but it looks like the answer to that is "no" as near as I can tell.

ag-eitilt:
Closing this until we figure out which way we want to go, since whatever we do will be a very different PR.

PR #952, opened 2022-07-21
Convert rope into a document and add geometry functions

V-FEXrt:
Coverts `rope` into `doc` and adds document geometry functions.

Fundamentaly a `doc` is a superset of the previously designed rope with functions to track the shape and size of various features of the `doc`. Specifically the following

- `first_width`: the width of the first line in the document
- `last_width`: the width of the last line in the document
- `max_width`: the width of the widest line in the document
- `character_count`: the number of characters in the document
- `newline_count`: the number of newlines in the document
- `height`: The number of rows in a document. Same as newline count + 1


JakeSiFive:
One other "nice to have" bit of geometry would be the width since in the case that no possible document is within the column limit you want to choose the one that's closest.

src/wcl/doc.h
```diff
@@ -0,0 +1,253 @@
...
+#include <cassert>
+#include <memory>
+#include <sstream>
+#include <vector>
+
+namespace wcl {
+
+// Internal state for a doc object. It is intentionally not exposed in the
+// public API and should not be used directly.
+struct doc_state {
```
JakeSiFive, line 29: This is nice because now the builder can use this! Smart! Now we can add more geometry in the future and we just update it all in one place.

src/wcl/doc.h
```diff
@@ -0,0 +1,253 @@
...
+      state.last_width++;
+    }
+    return state;
+  }
+};
+
+class doc_impl_base {
+ protected:
+  // On construction this has to be set
+  size_t size_;
```
JakeSiFive, line 77: size can be added to doc_state...if you wanted to be cheecky you could even call it "area" instead of size ;) But probably don't call it area lol.

src/wcl/doc.h
```diff
@@ -0,0 +1,253 @@
...
+  virtual ~doc_impl_base() = default;
+
+  // methods that all RopeImpl share
+  virtual void write(std::ostream&) const = 0;
+  size_t size() const { return size_; }
+  size_t first_width() const { return state_.first_width; }
+  size_t last_width() const { return state_.last_width; }
+  size_t newline_count() const { return state_.newline_count; }
+  bool has_newline() const { return state_.newline_count > 0; }
+  doc_state state() const { return state_; }
```
JakeSiFive, line 93: you can leave this out and state_ can just be accessed directly.

src/wcl/doc.h
```diff
@@ -0,0 +1,253 @@
...
+  void write(std::ostream& ostream) const override {
+    pair.first->write(ostream);
+    pair.second->write(ostream);
+  }
+};
+
+// Forward declare doc_builder so it can be named as a friend of doc
+class doc_builder;
+
+// `doc` is a very efficient data structure for editing strings. It is ideal to use when
```
JakeSiFive, line 128: I realize this is an old comment but I just re-read it and it isn't true, its an efficient data structure for *constructing* strings. It's actually immutable.

src/wcl/doc.h
```diff
@@ -0,0 +1,253 @@
...
+class doc {
+ private:
+  std::shared_ptr<doc_impl_base> impl;
+
+  doc_state state() const { return impl->state(); }
+
+  explicit doc(std::unique_ptr<doc_impl_base> impl) : impl(std::move(impl)) {}
+
+ public:
+  // O(1) but constructing the string takes O(n)
```
JakeSiFive, line 152: This became O(n) because we loop over the characters now.

JakeSiFive (approved):
Dig the tests and the doc_state a lot! None of my comments are blockers.

src/wcl/doc.h
V-FEXrt: I don't actually think this works because I need to do something like `left->state_` in the pair impl. Even though I'm inside of a subclass, I think I can only see _my_ protected members

JakeSiFive: nope, you can see any protected member from the same class.

src/wcl/doc.h
V-FEXrt: yee :)

src/wcl/doc.h
V-FEXrt: going with `character_count` matches `newline_count` and is more explicit than `size`

src/wcl/doc.h
V-FEXrt: Well its not compiling :)

```
In file included from tools/wake-format/wake-format.cpp:23:
src/wcl/doc.h: In constructor 'wcl::doc_impl_pair::doc_impl_pair(std::shared_ptr<wcl::doc_impl_base>, std::shared_ptr<wcl::doc_impl_base>)':
src/wcl/doc.h:128:29: error: 'wcl::doc_state wcl::doc_impl_base::state_' is protected within this context
       : doc_impl_base(left->state_.merged(right->state_)),
                             ^~~~~~
src/wcl/doc.h:93:13: note: declared protected here
   doc_state state_;
             ^~~~~~
src/wcl/doc.h:128:50: error: 'wcl::doc_state wcl::doc_impl_base::state_' is protected within this context
       : doc_impl_base(left->state_.merged(right->state_)),
                                                  ^~~~~~
src/wcl/doc.h:93:13: note: declared protected here
   doc_state state_;
```

also doc.h needs to forward `state` to rope_builder, so even if it worked above it wouldn't work for doc which isn't a subclass

V-FEXrt: I'll just go ahead and merge this. Will revisit later if there is a way to do it without the function.

JakeSiFive: oh huh...today I was one of the lucky 10k...and was wrong lol

src/wcl/doc.h
```diff
@@ -0,0 +1,274 @@
...
+
+    merged.first_width = first_width;
+    // If the left side doesn't have any newlines, then other side needs to be accounted for
+    if (newline_count == 0) {
+      merged.first_width += other.first_width;
+    }
+
+    merged.max_width = std::max(max_width, other.max_width);
+    // It is possible that the join point is a longer line than any other in either doc
+    merged.max_width = std::max(merged.max_width, last_width + other.first_width);
```
JakeSiFive, line 57: yep this was the only edge case, let me make sure you added a test case for it and I think we're kosher

tools/wake-unit/doc.cpp
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  {
+    wcl::doc_builder builder;
+    builder.append("He\nllo");
+    builder.append("Worl\nd!");
+
+    EXPECT_EQUAL(2u, builder.first_width());
+    EXPECT_EQUAL(2u, builder.last_width());
+    EXPECT_EQUAL(7u, builder.max_width());
```
JakeSiFive, line 175: yep here's the test. looks like your merge without my approval gets my seal of approval ;)

PR #954, opened 2022-07-21
Implement ctx width, basic def & block

V-FEXrt:
This add support for `ctx.width` and implements `def` and `block` using it. Seems like this may actually be the final API. _knocks on wood_

I still need to go back and delete `flat` and make the walk functions not return `optional` but I wanted to get this up first before another major refactor.

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -45,23 +45,35 @@
     }                                                                           \
   }
 
-wcl::doc Emitter::newline(ctx_t ctx) {
-  wcl::doc_builder builder;
+bool Emitter::fits(const wcl::doc_builder& bdr, ctx_t ctx, wcl::doc doc) {
+  if (bdr.has_newline()) {
+    return bdr.last_width() + doc.first_width() <= max_column_width;
+  } else {
+    return bdr.last_width() + doc.first_width() + ctx.width <= max_column_width;
+  }
+}
 
-  builder.append("\n");
+bool Emitter::requires_nl(ctx_t ctx, CSTElement node) {
+  return node.id() == CST_BLOCK && node.firstChildElement().id() != CST_MATCH;
+}
+
+wcl::doc Emitter::newline(ctx_t ctx) {
+  wcl::doc_builder bdr;
+
+  bdr.append("\n");
   for (size_t i = 0; i < ctx.nest_level; i++) {
-    builder.append(space(space_per_indent));
+    bdr.append(space(space_per_indent));
   }
 
-  return std::move(builder).build();
+  return std::move(bdr).build();
 }
 
 wcl::doc Emitter::space(uint8_t count) {
-  wcl::doc_builder builder;
+  wcl::doc_builder bdr;
   for (uint8_t i = 0; i < count; i++) {
-    builder.append(" ");
+    bdr.append(" ");
   }
-  return std::move(builder).build();
+  return std::move(bdr).build();
 }
 
 wcl::optional<wcl::doc> Emitter::flat(ctx_t ctx, CSTElement node) {
@@ -101,7 +113,7 @@ wcl::doc Emitter::layout(CST cst) {
 }
 
 wcl::doc Emitter::walk(ctx_t ctx, CSTElement node) {
-  wcl::doc_builder builder;
+  wcl::doc_builder bdr;
 
   for (CSTElement child = node.firstChildElement(); !child.empty(); child.nextSiblingElement()) {
     // remove optional whitespace
@@ -110,12 +122,12 @@ wcl::doc Emitter::walk(ctx_t ctx, CSTElement node) {
     }
 
     if (child.id() == TOKEN_NL) {
-      builder.append(newline(ctx));
+      bdr.append(newline(ctx));
       continue;
     }
 
     if (child.id() == TOKEN_COMMENT) {
-      builder.append(walk_token(ctx, child));
+      bdr.append(walk_token(ctx, child));
       continue;
     }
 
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -403,7 +426,72 @@ wcl::optional<wcl::doc> Emitter::walk_data(ctx_t ctx, CSTElement node) {
...
+  // optional ws/nl
+  CONSUME_WS_NL(child);
+
+  // rhs
+  {
+    auto new_child = walk_node(ctx.sub(builder), child);
+    assert(new_child);
+    wcl::doc new_doc = *new_child;
+    new_doc = space().concat(new_doc);
+    if (builder.last_width() + new_doc.first_width() + ctx.width <= max_column_width) {
```
JakeSiFive, line 477: If the builder has a newline this isn't the right calculation, we probably need a helper function to get this correct....but all of the info is *here* finally which is fucking amazing.

tools/wake-format/emitter.cpp
```diff
@@ -403,7 +426,72 @@ wcl::optional<wcl::doc> Emitter::walk_data(ctx_t ctx, CSTElement node) {
...
+      builder.append(*full_child);
+    }
+  }
+  child.nextSiblingElement();
+
+  // optional ws/nl
+  CONSUME_WS_NL(child);
+
+  assert(child.empty());
+  return wcl::optional<wcl::doc>(wcl::in_place_t{}, std::move(builder).build());
```
JakeSiFive, line 493: Do we ever need options now?

tools/wake-format/emitter.cpp
```diff
@@ -403,7 +426,72 @@ wcl::optional<wcl::doc> Emitter::walk_data(ctx_t ctx, CSTElement node) {
...
 wcl::optional<wcl::doc> Emitter::walk_def(ctx_t ctx, CSTElement node) {
-  return walk_placeholder(ctx, node);
+  ASSERT_TOKEN(node, CST_DEF);
+  wcl::doc_builder builder;
+  CSTElement child = node.firstChildElement();
+
+  // def shouldn't follow anything other than indentation.
+  if (ctx.width > ctx.nest_level * space_per_indent) {
+    builder.append(newline(ctx));
+  }
```
JakeSiFive, line 436: This line is suspicious, I think this should be an assert not a fix right?

tools/wake-format/emitter.h
```diff
@@ -29,7 +29,8 @@
 class Emitter {
  public:
   struct ctx_t {
-    int nest_level = 0;
+    size_t width = 0;
+    size_t nest_level = 0;
     bool is_flat = false;
```
JakeSiFive, line 34: I think you can delete this now?

JakeSiFive (commented):
Yeah this is starting to pass the smell test. I want to see `match` and `import` working all at once but I'd wager we're on to something here. After we're happy with match and import (And def obvi) we should try to get '|' and '$' and lists using ',' working correctly.

tools/wake-format/emitter.cpp
V-FEXrt: No, they can go away. but I'm waiting until we are happy with the API for another refactor (may have mentioned that in the PR description 😉 )

tools/wake-format/emitter.cpp
V-FEXrt: how so? if builder has a newline then `last_width()` is 0

tools/wake-format/emitter.cpp
JakeSiFive: I *might* be guilty of not reading those things sometimes 😬  sorry

tools/wake-format/emitter.cpp
V-FEXrt: Agreed that is it weird, and I'd prefer it to be added by the parent call but not sure how to spell it out. There are plenty of `defs` that don't exceed the line width rule like `def a = b` so something like `def c = def a = b` doesn't get split. 

Either, `def` needs to check _every_ case that forces a newline after an `=` with lookahead or each case that forces a newline after an `=` needs to insert the it themselves. The solution may still be to have a `walk_rhs` function under the new API but it is unclear to me

tools/wake-format/emitter.cpp
JakeSiFive: It might have a newline in the *middle*....ok to be fair we know there's no newline in this case but I'm trying to think more generically about it. We want a function that computes "will adding this doc overflow this builder's last line?" and this gets that wrong for some builders since if the builder has a newline you don't want to include `ctx.width`

```
bool fits(const wcl::doc_builder& bdr, ctx_t ctx, wcl::doc new_doc) {
  if (bdr.has_newline()) {
    return bdr.last_width() + new_doc.first_width() <= max_column_width;
  } else {
    return bdr.last_width() + new_doc.first_width() + ctx.width <= max_column_width
  }
}
```

tools/wake-format/emitter.cpp
JakeSiFive: The only case is block I think.

JakeSiFive: And you only have to do this in `def`, `require`, and `case`

tools/wake-format/emitter.cpp
V-FEXrt: ah right, thanks!

tools/wake-format/emitter.cpp
V-FEXrt: I came up with a reasonable solution I think. Will push it up in a second.

tools/wake-format/emitter.cpp
```diff
@@ -45,23 +45,35 @@
...
+bool Emitter::fits(const wcl::doc_builder& bdr, ctx_t ctx, wcl::doc doc) {
+  if (bdr.has_newline()) {
+    return bdr.last_width() + doc.first_width() <= max_column_width;
+  } else {
+    return bdr.last_width() + doc.first_width() + ctx.width <= max_column_width;
+  }
+}
+
+bool Emitter::requires_nl(ctx_t ctx, CSTElement node) {
+  return node.id() == CST_BLOCK && node.firstChildElement().id() != CST_MATCH;
```
JakeSiFive, line 57: Question not a request: Do matches get wrapped in blocks?

tools/wake-format/emitter.cpp
```diff
@@ -403,7 +436,66 @@ wcl::optional<wcl::doc> Emitter::walk_data(ctx_t ctx, CSTElement node) {
...
+  ASSERT_TOKEN(child, TOKEN_KW_DEF);
+  bdr.append("def");
+  child.nextSiblingElement();
+
+  // ws
+  ASSERT_TOKEN(child, TOKEN_WS);
+  bdr.append(space());
+  child.nextSiblingElement();
+
+  // CST_ID, CST_APPLY, TODO: others?
```
JakeSiFive, line 453: Not requesting a change just giving info:

string and integer literals are the others but you can do those later. `def 1 = 2` parses, it just fails to pass semantic checks because `1` is not an exhaustive pattern. You can also have parens as well in patterns which I think you need to handle eventully.

Also if binary/unary ops aren't handled by CST_APPLY then you need to handle those as well.

Another thing to check is that `def (if x then y else z) = 10` is a *parse* error and not just a semantic error, otherwise you actually have to handle any arbitrary expression

JakeSiFive (approved):
I dig this. `require_nl` and `fits` plus `sub` and the new naming convention are all good IMO.

tools/wake-format/emitter.cpp
V-FEXrt: Oh, interesting point. Looking at my old code, the implication is no. I'll confirm that and fix this bit of code when I get to implementing walk_match

tools/wake-format/emitter.cpp
V-FEXrt: yeah, makes sense! Thanks for the context

PR #955, opened 2022-07-22
Refactor: remove flat() and wcl::optional from format API

V-FEXrt:
Now that `flat()` and `wcl::optional` are no longer part of the wake-format internal API, remove them


JakeSiFive (approved):
Yep looks good. Really happy we got rid of this actually. This is a much better API

PR #956, opened 2022-07-22
Introduce fluent formatting api

V-FEXrt:


tools/wake-format/emitter.cpp
```diff
@@ -372,56 +372,27 @@ wcl::doc Emitter::walk_data(ctx_t ctx, CSTElement node) { return walk_placeholde
...
+  return formatter()
+      .token(TOKEN_KW_DEF)
+      .ws()
+      .walk({CST_ID, CST_APP}, [this](ctx_t ctx, CSTElement node) { return walk_node(ctx, node); })
+      .ws()
+      .token(TOKEN_P_EQUALS)
+      .consume_wsnl()
+      .escape([this](wcl::doc_builder& bdr, ctx_t ctx, CSTElement& node) {
+        wcl::doc doc = space().concat(walk_node(ctx.sub(bdr), node));
+        if (fits(bdr, ctx, doc) && !requires_nl(ctx, node)) {
```
JakeSiFive, line 385: It would be a bit more efficent to do the requires_nl check first but I vote we leave this like this for now until we hit performance issues. If everything goes to plan with memoization we should be fine I think.

tools/wake-format/formatter.h
```diff
@@ -0,0 +1,243 @@
...
+
+  template <class Walker>
+  Formatter<SeqAction<Action, WalkAction<Walker>>> walk(uint8_t node_type, Walker texas_ranger) {
+    return {
+        SeqAction<Action, WalkAction<Walker>>(action, WalkAction<Walker>{texas_ranger, node_type})};
+  }
+
+  template <class Walker>
+  Formatter<
+      SeqAction<Action, WalkPredicateAction<Walker, std::function<bool(const CSTElement& node)>>>>
```
JakeSiFive, line 203: These should not be std::functions, define a function object and use that for this case. std::function can't be inlined.

tools/wake-format/formatter.h
```diff
@@ -0,0 +1,243 @@
...
+  walk(std::vector<uint8_t> types, Walker texas_ranger) {
```
JakeSiFive, line 204: use an init list instead of a vector so its all on the stack. This will cause everything to inline nicely without allocation. The compiler can't optimize the allocation away if you use a vector but the init list will just vanish as the list gets unrolled during optimization. It's also just nicer because you don't have to move it, it can be copied.

tools/wake-format/emitter.cpp
```diff
@@ -418,15 +416,26 @@ wcl::doc Emitter::walk_ideq(ctx_t ctx, CSTElement node) { return walk_placeholde
...
+      .walk(CST_ID, WALK(walk_identifier))
+      .ws()
+      .token(TOKEN_KW_IMPORT)
+      .ws()
+      .fmt_if(CST_KIND, formatter().walk(CST_KIND, WALK(walk_kind)).ws())
+      .fmt_if(CST_ARITY, formatter().walk(CST_ARITY, WALK(walk_arity)).ws())
+      // clang-format off
+      .fmt_if_else(
+          TOKEN_P_HOLE,
+          formatter().walk(TOKEN_P_HOLE, WALK(walk_token)),
```
JakeSiFive, line 432: not something to do here but I wonder if it would make more sense to make the WALK macro into like `WALK(TOKEN_P_HOLE, walk_token)` and you could use it in the fluent API as like a macro method?

tools/wake-format/formatter.h
```diff
@@ -238,6 +303,12 @@ struct Formatter {
     assert(node.empty());
     return std::move(builder).build();
   }
+
+  wcl::doc compose(ctx_t ctx, CSTElement& node) {
```
JakeSiFive, line 307: Is this just to avoid a friend function?

V-FEXrt: This shares the node reference with the composed formatter. I don't think its sufficient to make a friend function because you have to advance the node reference of the `parent` class and you can't just do simple assignment since the formatter class doesn't actually store its node, instead it only manipulates the reference created from the top level format call.

(that said, I do think minimizing friend classes is a preferable thing to do)

tools/wake-format/emitter.cpp
V-FEXrt: I actually prefer this, but `WALK({CST_ID, CST_APP}, walk_node)` doesn't seem to work :(

JakeSiFive: Well you'd have to change the macro but it should work

```
#define NEW_WALK(pred, method) walk(pred, WALK(method))
```

tools/wake-format/formatter.h
```diff
@@ -0,0 +1,320 @@
...
+  SpaceAction(uint8_t count) : count(count) {}
+  ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
+    space(builder, count);
+  }
+};
+
+struct NewlineAction {
+  ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
+    builder.append("\n");
+    space(builder, 4 * ctx.nest_level);
```
V-FEXrt, line 76: Not sure the best way to pass `4` here from emitter.cpp without breaking the niceness of the api

tools/wake-format/emitter.cpp
V-FEXrt: The issue is the `,` `{CST_ID, CST_APP}`. It messes up the expansion.

There might be a way to escape the comma, that that would be a weird public api

JakeSiFive: oh shit I didn't think about that, what if we flip it around and we make the second argument a vararg and then we expand it?

tools/wake-format/formatter.h
JakeSiFive: I think you should pass it in or make it a macro :/

tools/wake-format/emitter.cpp
```diff
@@ -446,7 +415,28 @@ wcl::doc Emitter::walk_ideq(ctx_t ctx, CSTElement node) { return walk_placeholde
...
+          formatter().fmt_while(
+              CST_IDEQ,
+              formatter().walk(CST_IDEQ, WALK(walk_ideq)).fmt_if(TOKEN_WS, formatter().ws())))
```
JakeSiFive, line 435: Can you define a local called something like "id_list_fmt" above this and pass it in? I think it will make the formatting better

tools/wake-format/formatter.h
```diff
@@ -0,0 +1,320 @@
...
+  ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
+    assert(node.id() == node_type);
+    auto doc = walker(ctx.sub(builder), const_cast<const CSTElement&>(node));
+    builder.append(doc);
+    node.nextSiblingElement();
+  }
+};
+
+template <class F, class P>
+struct WalkPredicateAction {
```
JakeSiFive, line 135: This is strictly more generic than the above really. I think you can just subsume the above into this one.

tools/wake-format/formatter.h
```diff
@@ -0,0 +1,320 @@
...
+
+  ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
+    if (node.id() == node_type) {
+      builder.append(formatter.compose(ctx, node));
+    }
+  }
+};
+
+template <class IFMT, class EFMT>
+struct IfElseAction {
```
JakeSiFive, line 164: This subsumes the above. You can just use an EpslionAction formatter when you only want the `if` in the Formatter API

JakeSiFive (approved):
No blockers, looks good!

V-FEXrt:
Didn't see this was approved and I added another non-trivial commit :(

Can you PTAL at https://github.com/sifive/wake/pull/956/commits/407586ea9373ea2e032c84a5d8aa2be38dc4e3c1

src/wcl/doc.h
```diff
@@ -263,7 +263,17 @@ class doc_builder {
...
+    if (docs.size() == 0) {
+      return doc::lit("");
+    }
+
+    if (docs.size() == 1) {
+      doc copy = std::move(docs[0]);
+      docs = {};
+      state = {};
+      return copy;
+    }
```
JakeSiFive, line 275: These edge cases should be handled in `merge` if merge is written correctly.

tools/wake-format/emitter.cpp
V-FEXrt: Ack

tools/wake-format/emitter.cpp
```diff
@@ -337,16 +335,45 @@ wcl::doc Emitter::walk_token(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_rhs(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
 
-wcl::doc Emitter::walk_apply(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
+wcl::doc Emitter::walk_apply(ctx_t ctx, CSTElement node) {
+  assert(node.id() == CST_APP);
+
+  return formatter()
+      .walk({CST_ID, CST_APP}, WALK(walk_node))
```
JakeSiFive, line 342: You need an "is expression" predicate It can just be ID and APP for now but that's the set of nodes actully expected here, and you'll wind up needing that predicate over and over again. Make it a static function

tools/wake-format/emitter.cpp
V-FEXrt: TBH I'm partial to having it be the first parameter. Lets hold on this change for a bit and feel it out while implementing subtrees

tools/wake-format/emitter.cpp
```diff
@@ -337,16 +335,45 @@ wcl::doc Emitter::walk_token(ctx_t ctx, CSTElement node) {
...
 wcl::doc Emitter::walk_arity(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
 
 wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
 
-wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
+wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
+  assert(node.id() == CST_BINARY);
+
+  return formatter()
+      .walk(WALK(walk_node))
```
JakeSiFive, line 357: This should have an is expression predicate

tools/wake-format/emitter.cpp
```diff
@@ -337,16 +335,45 @@ wcl::doc Emitter::walk_token(ctx_t ctx, CSTElement node) {
...
+      .consume_wsnl()
+      .escape([this](wcl::doc_builder& bdr, ctx_t ctx, CSTElement& node) {
+        assert(node.id() == CST_OP);
+        // only add a space if the binop is not a comma
+        // a+b -> a + b
+        // a,b -> a, b
+        if (node.firstChildElement().id() != TOKEN_OP_COMMA) {
```
JakeSiFive, line 364: In this PR don't special case anything I think. There's several special cases for operators. the "." operator for instance has no surrounding spaces.

tools/wake-format/emitter.cpp
```diff
@@ -366,62 +393,54 @@ wcl::doc Emitter::walk_block(ctx_t ctx, CSTElement node) {
...
+      .escape([this](wcl::doc_builder& bdr, ctx_t ctx, CSTElement& node) {
+        wcl::doc doc = space().concat(walk_node(ctx.sub(bdr), node));
+        if (fits(bdr, ctx, doc) && !requires_nl(ctx, node)) {
+          bdr.append(doc);
+        } else {
+          ctx_t n_ctx = ctx.nest();
+          bdr.append(newline(n_ctx));
+          bdr.append(walk_node(n_ctx.sub(bdr), node));
+        }
+        node.nextSiblingElement();
```
JakeSiFive, line 414: If this is identical to `def` maybe it *should* be factored out into a seperate function like you had it before. I'm happy to leave it separate and dedup later though. require will also have this chunk of code I guess.

tools/wake-format/emitter.cpp
```diff
@@ -458,23 +503,82 @@ wcl::doc Emitter::walk_lambda(ctx_t ctx, CSTElement node) { return walk_placehol
 
 wcl::doc Emitter::walk_literal(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
 
-wcl::doc Emitter::walk_match(ctx_t ctx, CSTElement node) { return walk_placeholder(ctx, node); }
+wcl::doc Emitter::walk_match(ctx_t ctx, CSTElement node) {
```
JakeSiFive, line 506: fucking sick!

tools/wake-format/emitter.cpp
```diff
@@ -458,23 +503,82 @@ wcl::doc Emitter::walk_lambda(ctx_t ctx, CSTElement node) { return walk_placehol
...
+
+  return formatter()
+      .token(TOKEN_KW_REQUIRE)
+      .ws()
+      .walk(WALK(walk_node))
+      .consume_wsnl()
+      .space()
+      .token(TOKEN_P_EQUALS)
+      .consume_wsnl()
+      .escape([this](wcl::doc_builder& bdr, ctx_t ctx, CSTElement& node) {
```
JakeSiFive, line 554: Aight I'm convinced, break it out into its own thing.

tools/wake-format/formatter.h
```diff
@@ -0,0 +1,321 @@
...
+};
+
+inline void space(wcl::doc_builder& builder, uint8_t count) {
+  for (uint8_t i = 0; i < count; i++) {
+    builder.append(" ");
+  }
+}
+
+struct ConsumeWhitespaceAction {
+  ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
```
JakeSiFive, line 58: oh hey thanks for adding these!

JakeSiFive (changes requested):

src/wcl/doc.h
V-FEXrt: The `0` edge case is needed so that we can have "epsilon else" in the IfElseAction since we try to build an empty builder. The `1` case is probably not needed. I just added it without really thinking about it

JakeSiFive: I didn't say it wasn't needed, I said it should be handled in merge ;) merge should be able to merge any length of array, 0, 1 or higher. Shame on me for not checking on that sooner.

V-FEXrt: Yeah, the `1` isn't needed, but `0` is since you can't merge from `0` to `-1` and even if you could, there is nothing in the array to return

tools/wake-format/emitter.cpp
V-FEXrt: yeah, I removed it. I was just porting over the other code I wrote sitting in an abandoned branch so it could go away.

tools/wake-format/formatter.h
V-FEXrt: ye, might double check that I have the macro right. I haven't inlined all that much stuff before

src/wcl/doc.h
JakeSiFive: you can define merge to return the empty doc in that case

JakeSiFive (approved):

PR #962, opened 2022-07-28
Simplify formatter.h via more template magic and lift more code into the formatter API

V-FEXrt:
- Predicates are now more flexible and will pick the appropiate signature type 
- fits is moved into formatter
- space per indent is moved into formatter
- emitter.cpp has been updated to remove those bits
- no more (current) usage of escape is left
- `{}` syntax is used in formatter when possible simplifying the functions


JakeSiFive (approved):
I know I said I'd review this later but this all looks good to me. It's all stuff we went over already.

tools/wake-format/emitter.cpp
```diff
@@ -27,38 +27,19 @@
...
 }
 
-void Emitter::walk_rhs(wcl::doc_builder& bdr, ctx_t ctx, CSTElement& node) {
-  auto nested_fmt = fmt().nest(fmt().walk(WALK(walk_node)));
-  if (requires_nl(ctx, node)) {
-    bdr.append(nested_fmt.compose(ctx, node));
-    return;
-  }
-
+decltype(auto) Emitter::rhs_fmt() {
```
JakeSiFive, line 37: Actually fix this: I learned this should be `auto` instead. When its `decltype(auto)` we open ourselves up to some issues.

JakeSiFive (changes requested):

JakeSiFive (approved):

PR #967, opened 2022-08-02
Implement `# wake-format off`

V-FEXrt:
Implements `# wake-format off` with the following semantics.

1) For a `# wake-format off` comment at top level, the next sibling node and all its children will be re-emitted exactly as they are. Including whitespace, indents, and newlines
2) For all other `# wake-format off` comments, the next sibling node and all its children will be emitted exactly as they are except that indentation will be overwritten to match the indentation for the rest of the containing block. Since wake is whitespace sensitive, this is required to maintain a semantically valid program.

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -254,6 +254,17 @@ wcl::doc Emitter::walk_no_edit(ctx_t ctx, CSTElement node) {
 
   wcl::doc_builder bdr;
   for (CSTElement child = node.firstChildElement(); !child.empty(); child.nextSiblingElement()) {
+    // The last nl of a *tagged* "no format" CST_DEF node shouldn't be emitted.
+    // The nominal formtting for the larger program structure will ensure the correct NLs are
+    // emitted.
+    if (node.id() == CST_DEF && child.id() == TOKEN_NL && traits[node].format_off) {
+      CSTElement next = child;
+      next.nextSiblingElement();
+      if (next.empty()) {
+        continue;
+      }
+    }
+
     bdr.append(walk_no_edit(ctx, child));
   }
 
@@ -575,16 +586,16 @@ wcl::doc Emitter::walk_match(ctx_t ctx, CSTElement node) {
           .ws()
           .walk(WALK_NODE)
           // clang-format off
-               .nest(fmt()
-                   .consume_wsnl()
-                   .fmt_while(
-                       {CST_CASE, TOKEN_COMMENT}, fmt()
-                       .newline()
-                       .fmt_if_else(
-                         TOKEN_COMMENT,
-                         fmt().token(TOKEN_COMMENT),
-                         fmt().walk(WALK_NODE))
-                       .consume_wsnl()))
+          .nest(fmt()
+              .consume_wsnl()
+              .fmt_while(
+                  {CST_CASE, TOKEN_COMMENT}, fmt()
+                  .newline()
+                  .fmt_if_else(
+                    TOKEN_COMMENT,
+                    fmt().token(TOKEN_COMMENT),
+                    fmt().walk(WALK_NODE))
+                  .consume_wsnl()))
           // clang-format on
           .format(ctx, node.firstChildElement()));
 }
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -281,8 +281,9 @@ void Emitter::mark_no_format_nodes(CSTElement node) {
       // you can never format *just* the first block item
       if (child.id() == CST_BLOCK) {
         CSTElement block_item = child.firstChildElement();
-        for (; !(child.empty() || child.isNode()); child.nextSiblingElement())
-          ;
+        while (!(child.empty() || child.isNode())) {
+          child.nextSiblingElement();
+        }
 
         // This shouldn't be possible, but assert anyways just in case
         assert(!child.empty());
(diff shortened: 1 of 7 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -64,15 +68,22 @@ static bool is_expression(uint8_t type) {
 }
 
 auto Emitter::rhs_fmt() {
```
JakeSiFive, line 70: I see what this is doing now and I like it!

tools/wake-format/emitter.cpp
```diff
@@ -83,7 +94,12 @@ wcl::doc Emitter::walk(ctx_t ctx, CSTElement node) {
   auto extra_fmt = fmt()
       .fmt_if(TOKEN_WS, fmt().next())
       .fmt_if(TOKEN_NL, fmt().next().newline())
-      .fmt_if(TOKEN_COMMENT, fmt().walk(WALK(walk_token)));
+      .fmt_if([](wcl::doc_builder& builder, ctx_t ctx, CSTElement& node){
```
JakeSiFive, line 97: Should these two predicates be given names? Like "is_format_off_token" and "is_inert_comment"?

Actully you don't need the `is_inert_token` just `TOKEK_COMMENT` is fine. That way when you make this a match thing you can still assert if you see a bad token

tools/wake-format/emitter.cpp
```diff
@@ -392,6 +464,16 @@ wcl::doc Emitter::walk_case(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
   assert(node.id() == CST_CASE);
 
+  if (traits[node].format_off) {
```
JakeSiFive, line 467: hmm can we eliminate this somehow? I don't want to repeat this in every walk function. Seems like something walk_node can do by itself right? The sub walkers that walk_node dispatches to don't need these right? Or do we dispatch to these directly?

Either way I want to think about how to eliminate these

tools/wake-format/emitter.cpp
```diff
@@ -517,21 +613,25 @@ wcl::doc Emitter::walk_match(ctx_t ctx, CSTElement node) {
...
-          .format(ctx, node.firstChildElement()));
+  MEMO_RET(fmt()
+               .token(TOKEN_KW_MATCH)
+               .ws()
+               .walk(WALK(walk_node))
+               // clang-format off
+               .nest(fmt()
+                   .consume_wsnl()
+                   .fmt_if(TOKEN_COMMENT, fmt().newline().token(TOKEN_COMMENT))
+                   .consume_wsnl()
```
JakeSiFive, line 624: I think you can have white-space and newlines between multiple comments right? So this only handles one comment before each match, not multiple.

tools/wake-format/formatter.h
```diff
@@ -356,6 +369,7 @@ struct Formatter {
   Formatter(Action a) : action(a) {}
 
   Formatter<SeqAction<Action, ConsumeWhitespaceAction>> consume_wsnl() { return {{action, {}}}; }
+  Formatter<SeqAction<Action, CopyWhitespaceAction>> copy_wsnl() { return {{action, {}}}; }
```
JakeSiFive, line 372: Not now but I think we need to rethink our handling of white space and newlines (and probably comments) a bit. I think the different types are proliferating and its becoming unwieldy.

JakeSiFive (commented):
I want us to think about how to get rid of the boilerplate for handling the no-fmt case in each walker

tools/wake-format/emitter.cpp
```diff
@@ -93,18 +80,16 @@ wcl::doc Emitter::layout(CST cst) {
...
-   .pred(TOKEN_NL, fmt().next().newline())
-   .pred(TOKEN_COMMENT, fmt().walk(WALK(walk_token)))
-   .pred([](wcl::doc_builder& builder, ctx_t ctx, CSTElement& node){
-        return node.id() == TOKEN_COMMENT && node.fragment().segment().str() != FORMAT_OFF_COMMENT;
-   }, fmt().token(TOKEN_COMMENT))
-   .pred([](wcl::doc_builder& builder, ctx_t ctx, CSTElement& node){
-        return node.id() == TOKEN_COMMENT && node.fragment().segment().str() == FORMAT_OFF_COMMENT;
-   }, fmt().token(TOKEN_COMMENT).consume_wsnl().newline().walk(WALK(walk_no_edit)))
-   .otherwise(fmt().walk(WALK(walk_node)).newline()));
+   .pred(TOKEN_NL, fmt().next())
```
JakeSiFive, line 88: This is so much cleaner it makes me want to cry

tools/wake-format/emitter.cpp
```diff
@@ -281,9 +271,27 @@ void Emitter::mark_no_format_nodes(CSTElement node) {
...
       }
+
+      // Instead of marking the entire block as format off
+      // only the first non-token child should be marked as format off
+      // this allows turing off formatting for the first block item, otherwise
+      // you can never format *just* the first block item
+      if (child.id() == CST_BLOCK) {
+        CSTElement block_item = child.firstChildElement();
+        for (; !(child.empty() || child.isNode()); child.nextSiblingElement())
+          ;
```
JakeSiFive, line 285: Is this how clang-format did this? Can you make that empty '{}' instead?

JakeSiFive: actully do this as a while loop, that will fix it and is more idiomatic.

JakeSiFive (approved):
This is so clearly better, glad we had that talk!

tools/wake-format/emitter.cpp
V-FEXrt: they are both gone now :)

tools/wake-format/emitter.cpp
V-FEXrt: ahh, that's actually a really good callout. Turns out this line isn't actually needed though and this is fully covered by the `fmt_while` just below. It was been removed

JakeSiFive: Awesome, that's better.

PR #968, opened 2022-08-04
Support unicode with in wcl::doc

V-FEXrt:
`wcl::doc` was incorrectly reporting width values when passed in unicode strings. As unicode is common place in wake code, we need to also support it here. This change does the thing.

### DIFF
--- a/src/wcl/doc.h
+++ b/src/wcl/doc.h
@@ -75,16 +75,13 @@ struct doc_state {
 
     state.byte_count = str.size();
 
-    utf8proc_int32_t codepoint;
-    utf8proc_ssize_t pos = 0;
-    while (true) {
-      pos += utf8proc_iterate(reinterpret_cast<const unsigned char*>(str.c_str()) + pos, -1,
-                              &codepoint);
+    const utf8proc_uint8_t* iter = reinterpret_cast<const unsigned char*>(str.c_str());
+    const utf8proc_uint8_t* iter_end = iter + str.size();
+    while (iter < iter_end) {
+      utf8proc_int32_t codepoint;
+      iter += utf8proc_iterate(iter, -1, &codepoint);
 
-      assert(codepoint >= 0);  // < 0 means error parsing utf8
-      assert(pos >= 0);        // < 0 means pos was overflown
-
-      if (codepoint == 0) break;
+      assert(codepoint > 0);  // < 0 means error parsing utf8
 
       int charwidth = utf8proc_charwidth(codepoint);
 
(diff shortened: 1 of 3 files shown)


src/wcl/doc.h
```diff
@@ -62,10 +73,22 @@ struct doc_state {
   static doc_state from_string(const std::string& str) {
     doc_state state;
 
-    state.character_count = str.size();
+    state.byte_count = str.size();
+
+    utf8proc_int32_t codepoint;
+    utf8proc_ssize_t pos = 0;
+    while (true) {
+      pos += utf8proc_iterate(reinterpret_cast<const unsigned char*>(str.c_str()) + pos, -1,
```
JakeSiFive, line 81: reinterpret_cast to `const utf8_proc_uint8_t*` once above and call it iter, then call

```
iter += utf8proc_iterate(iter, -1, &codepoint);
```

Also nit: declare `codepoint` inside the loop and the loop should be `while (iter < iter_end)` where iter_end is `iter + str.size()`. Both pos and codepoint are made negative so you can rely on codepoint to check for the error case.

src/wcl/doc.h
```diff
@@ -62,10 +73,22 @@ struct doc_state {
   static doc_state from_string(const std::string& str) {
     doc_state state;
 
-    state.character_count = str.size();
+    state.byte_count = str.size();
+
+    utf8proc_int32_t codepoint;
+    utf8proc_ssize_t pos = 0;
+    while (true) {
```
JakeSiFive, line 80: I don't think we need to do this but an adapter range for std::string to iterate over code points would be nice. Too much work for right now I think.

JakeSiFive (approved):

src/wcl/doc.h
V-FEXrt: yeah, that would be a good future improvement. this is a common enough use case

src/wcl/doc.h
```diff
@@ -62,10 +73,19 @@ struct doc_state {
   static doc_state from_string(const std::string& str) {
     doc_state state;
 
-    state.character_count = str.size();
+    state.byte_count = str.size();
+
+    const utf8proc_uint8_t* iter = reinterpret_cast<const unsigned char*>(str.c_str());
```
JakeSiFive, line 78: The cast should be utf8proc_uint8_t not unsinged char. They're the same but we wan't to use the libraries type. The declared type can be `auto` for all I care.

JakeSiFive (approved):

src/wcl/doc.h
V-FEXrt: My bad. Fixed here https://github.com/sifive/wake/pull/969

PR #971, opened 2022-08-08
Add `match` syntax to the formatter API

V-FEXrt:
Adds `match` to the formatting API. Used as shown below

```
fmt().match(match()
  .pred(TOKEN_ID, fmt())
  .pred({TOKEN_ID, TOKEN_ID}, fmt())
  .pred(some_predicate, fmt())
  .fits(fmt())
  .otherwise(fmt())
.format(..., ...)
```

- `pred` -> the `case` statement. if predicate returns true then fmt is used, if false the the next entry is considered
- `fits` -> fmt is used if it fits the size limit, otherwise the next entry is considered
- `otherwise` -> the `default` statement. fmt is always used and no other case is considered. Should be the last item in a match

if no statement is accepted then an assertion is thrown; therefore a `match` should always end in an `otherwise` unless it is known that a `pred` will always pass 

Cases are evaluated top down, the first to be accepted is the one that is used.

### DIFF
--- a/tools/wake-format/formatter.h
+++ b/tools/wake-format/formatter.h
@@ -395,8 +395,9 @@ struct MatchAction {
 
   MatchAction(Case c) : c(c) {}
 
+  // Predicate case that is accepted if FMT passes the FitsPredicate
   template <class FMT>
-  MatchAction<MatchSeq<Case, PredicateCase<FitsPredicate<FMT>, FMT>>> fits(FMT formatter) {
+  MatchAction<MatchSeq<Case, PredicateCase<FitsPredicate<FMT>, FMT>>> pred_fits(FMT formatter) {
     return {{c, {FitsPredicate<FMT>(formatter), formatter}}};
   }
 
--- a/tools/wake-format/formatter.h
+++ b/tools/wake-format/formatter.h
@@ -347,10 +347,6 @@ struct FmtPredicate<std::initializer_list<T>> {
   bool operator()(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return set[node.id()]; }
 };
 
-struct EpsilonCase {
-  ALWAYS_INLINE bool run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return false; }
-};
-
 template <class Predicate, class FMT>
 struct PredicateCase {
   Predicate predicate;
@@ -558,6 +554,10 @@ struct Formatter {
 
 inline Formatter<EpsilonAction> fmt() { return {{}}; }
 
-inline MatchAction<EpsilonCase> match() { return {{}}; }
+template <class Predicate, class FMT>
+inline MatchAction<PredicateCase<FmtPredicate<Predicate>, FMT>> pred(Predicate predicate,
+                                                                     FMT formatter) {
+  return {{predicate, formatter}};
+}
 
 #undef ALWAYS_INLINE
--- a/tools/wake-format/formatter.h
+++ b/tools/wake-format/formatter.h
@@ -347,6 +347,85 @@ struct FmtPredicate<std::initializer_list<T>> {
   bool operator()(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return set[node.id()]; }
 };
 
+struct EpsilonCase {
+  ALWAYS_INLINE bool run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return false; }
+};
+
+template <class Predicate, class FMT>
+struct PredicateCase {
+  Predicate predicate;
+  FMT formatter;
+
+  PredicateCase(Predicate predicate, FMT formatter) : predicate(predicate), formatter(formatter) {}
+
+  ALWAYS_INLINE bool run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
+    if (!predicate(builder, ctx, node)) {
+      return false;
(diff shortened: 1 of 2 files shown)


tools/wake-format/formatter.h
```diff
@@ -347,6 +347,85 @@ struct FmtPredicate<std::initializer_list<T>> {
   bool operator()(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return set[node.id()]; }
 };
 
+struct EpsilonCase {
+  ALWAYS_INLINE bool run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return false; }
+};
+
+template <class Predicate, class FMT>
+struct PredicateCase {
```
JakeSiFive, line 355: Is this different from fmt_if?

tools/wake-format/formatter.h
```diff
@@ -347,6 +347,85 @@ struct FmtPredicate<std::initializer_list<T>> {
   bool operator()(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) { return set[node.id()]; }
 };
 
+struct EpsilonCase {
```
JakeSiFive, line 350: Is this different from an epsilon formatter?

tools/wake-format/formatter.h
```diff
@@ -347,6 +347,85 @@ struct FmtPredicate<std::initializer_list<T>> {
...
+    if (!predicate(builder, ctx, node)) {
+      return false;
+    }
+    builder.append(formatter.compose(ctx.sub(builder), node));
+    return true;
+  }
+};
+
+template <class FMT>
+struct OtherwiseCase {
```
JakeSiFive, line 371: I think anytime you use this you can instead just return the FMT underlying it. This is basically `IDFmt`

tools/wake-format/formatter.h
```diff
@@ -347,6 +347,85 @@ struct FmtPredicate<std::initializer_list<T>> {
...
+  OtherwiseCase(FMT formatter) : formatter(formatter) {}
+
+  ALWAYS_INLINE bool run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
+    builder.append(formatter.compose(ctx.sub(builder), node));
+    return true;
+  }
+};
+
+template <class Case1, class Case2>
+struct MatchSeq {
```
JakeSiFive, line 383: Is this different from `SeqFmt`?

tools/wake-format/formatter.h
```diff
@@ -347,6 +347,85 @@ struct FmtPredicate<std::initializer_list<T>> {
...
+  }
+
+  template <class FMT>
+  MatchAction<MatchSeq<Case, PredicateCase<FmtPredicate<std::initializer_list<uint8_t>>, FMT>>>
+  pred(std::initializer_list<uint8_t> ids, FMT formatter) {
+    return {{c, {ids, formatter}}};
+  }
+
+  template <class Predicate, class FMT>
+  MatchAction<MatchSeq<Case, PredicateCase<FmtPredicate<Predicate>, FMT>>> pred(Predicate predicate,
```
JakeSiFive, line 414: Ok yeah that's the ticket, I think you *can* get the order right. You just have to use SeqFmt to do it correctly. You have all the tools I just didn't see how to do it.

I think having an `end` method that returns an AssertFmt for the default case would be good to end it.

JakeSiFive:
I'm happy with this all I think, I think if you stick with this instead of using `fmt().match().pred(...)` you should add a `pred` function so you don't have to call `match()` to make the syntax a bit nicer...in fact actually I like that best

Yeah just replace `match()` with a `pred()` so you don't need an EpsilonMatch

tools/wake-format/formatter.h
V-FEXrt: run returns a `bool` instead of `void`

tools/wake-format/formatter.h
V-FEXrt: yeah, Epsilon doesn't have `return false`

tools/wake-format/formatter.h
V-FEXrt: We need the `return true` part

tools/wake-format/formatter.h
V-FEXrt: yep, `SeqFmt` _always runs case1 and case2, this only runs case2 if case1 returns false

V-FEXrt:
> I'm happy with this all I think, I think if you stick with this instead of using `fmt().match().pred(...)` you should add a `pred` function so you don't have to call `match()` to make the syntax a bit nicer...in fact actually I like that best
> 
> Yeah just replace `match()` with a `pred()` so you don't need an EpsilonMatch

Done!

tools/wake-format/formatter.h
```diff
@@ -347,6 +347,81 @@ struct FmtPredicate<std::initializer_list<T>> {
...
+};
+
+template <class Case>
+struct MatchAction {
+  Case c;
+
+  MatchAction(Case c) : c(c) {}
+
+  template <class FMT>
+  MatchAction<MatchSeq<Case, PredicateCase<FitsPredicate<FMT>, FMT>>> fits(FMT formatter) {
```
JakeSiFive, line 399: This was confusing to me, can you rename it to `pred_fits` and add a comment?

JakeSiFive (approved):

PR #972, opened 2022-08-09
Implement topic and paren subtrees

V-FEXrt:


tools/wake-format/emitter.cpp
```diff
@@ -553,7 +554,13 @@ wcl::doc Emitter::walk_package(ctx_t ctx, CSTElement node) {
...
 wcl::doc Emitter::walk_paren(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  assert(node.id() == CST_PAREN);
+
+  MEMO_RET(fmt()
+               .token(TOKEN_P_POPEN)
+               .walk(is_expression, WALK(walk_node))
+               .token(TOKEN_P_PCLOSE)
+               .format(ctx, node.firstChildElement()));
```
JakeSiFive, line 563: white-space and comments between tokens?

tools/wake-format/emitter.cpp
```diff
@@ -619,7 +651,40 @@ wcl::doc Emitter::walk_package(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_paren(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  assert(node.id() == CST_PAREN);
+
+  auto comment_fmt =
```
JakeSiFive, line 656: Handling comments makes me want to cry lol

JakeSiFive (commented):
Lets think about our comment handling strategy a bit. This is feeling very complicated to me right now.

tools/wake-format/emitter.cpp
```diff
@@ -71,6 +90,18 @@ auto Emitter::rhs_fmt() {
...
+  // ws COMMENT -> ws COMMENT nl
+  // ws nl COMMENT -> nl COMMENT nl
+  return fmt().fmt_if(has_comment_next,
+                      fmt()
+                          .fmt_if(TOKEN_WS, fmt().next())
+                          .fmt_if_else(TOKEN_NL, fmt().next().nest(fmt().newline()), fmt().space())
+                          .consume_wsnl()
+                          .token(TOKEN_COMMENT)
+                          .consume_wsnl());
+}
```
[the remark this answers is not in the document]
V-FEXrt: Yeah sorry, the comment explains it if you've  bashed your head against it a lot.

So we can have a comment 

```
( # comment
something
)
```

or 

```
(
# comment 
something
)
```

and we want to maintain the location of the comment.


So the condition here is 

if TOKEN_WS, TOKEN_COMMENT then emit TOKEN_WS, TOKEN_COMMENT, TOKEN_NL
else if  TOKEN_WS, TOKEN_NL, TOKEN_COMMENT then emit TOKEN_NL TOKEN_COMMENT TOKEN_NL

the reason this is a function is because I've already seen other other case outside this PR where this is needed and there is very likely to be more


The `nest` in there _does_ cause some weirdness that needs to be worked out. In this I did a weird thing in walk_paren but it could be solve by passing a "nest formatter" in or maybe a bool "should nest"

tools/wake-format/emitter.cpp
V-FEXrt: yeeeeeep, it is very hard to get right and it is going to make everythign 5x harder.

Easy solution, we should just remove comments from wake.

JakeSiFive: lol

tools/wake-format/emitter.cpp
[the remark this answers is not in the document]
JakeSiFive: I'm a bit confused why the `nest` is needed? Can you explain that?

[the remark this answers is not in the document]
V-FEXrt: yeah sorry.

Consider 

```
def x = (
    # comment
   1 + 2
)
```

without `nest()`

```
def x = (
# comment
    1 + 2
)
```
would be emitted.

Basically `place_comment()` is assuming that if it inserts a newline it also needs to nest. This turns out to be a bad assumption generally though so I'm thinking it may be better to have a "comment_should_be_newlined" predicate that then leaves the callsite formatter to determine what to do if a newline is required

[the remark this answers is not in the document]
JakeSiFive: Something isn't right here. This would cause multiple levels of nesting which would result in a parse error. You shouldn't be nesting inside of parens ever. You only nest for things like def/require/match/case/publish.

The problem here isn't with parens its with `def`. The case where the `def` tries not to nest causes this issue and you've smoothed that over using a bunch of hacks. This is frustrating though because we're back to the "failed flat" thing we got rid of.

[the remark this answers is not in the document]
V-FEXrt: I'm not sure I fully understand.

From my perspective I wasn't smoothing over anything. I just wanted a multiline `paren` node to have its contents nested in by one. Is that not the correct goal?

Assuming is parses, doing 

```
def x = ( 
# comment 
1 + 2
)
```

is also very easy. I just don't like the way to looks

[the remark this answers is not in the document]
V-FEXrt: If the goal is 

```
def x =
    (
    # comment
    1 + 2
    )
```

then I think I see what your saying. That a different issue inside def though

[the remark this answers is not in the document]
JakeSiFive: I just wasn't clear on how wake was parsing something until we talked, I think you're outputting things correctly as discussed but I'm not happy with the un-nest stuff. Looking forward to your solution!

tools/wake-format/emitter.cpp
```diff
@@ -47,11 +47,30 @@
...
+static inline bool requires_nl(cst_id_t type) { return type == CST_BLOCK || type == CST_REQUIRE; }
 
-static bool is_expression(cst_id_t type) {
+static inline bool is_expression(cst_id_t type) {
   return type == CST_ID || type == CST_APP || type == CST_LITERAL || type == CST_HOLE ||
-         type == CST_BINARY;
+         type == CST_BINARY || CST_PAREN;
+}
+
+static inline bool has_comment_next(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
```
JakeSiFive, line 57: you can make the `node` here a const ref and everything will still work

tools/wake-format/emitter.cpp
```diff
@@ -47,11 +47,30 @@
...
+  CSTElement copy = node;
+  while (!copy.empty() && (copy.id() == TOKEN_WS || copy.id() == TOKEN_NL)) {
+    copy.nextSiblingElement();
+  }
+  return copy.id() == TOKEN_COMMENT;
+}
+
+static inline bool has_comment(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
```
JakeSiFive, line 65: ditto

tools/wake-format/emitter.cpp
```diff
@@ -71,6 +90,16 @@ auto Emitter::rhs_fmt() {
   // clang-format on
 }
 
+auto position_comment() {
+  // ws COMMENT -> ws COMMENT nl
+  // ws nl COMMENT -> nl COMMENT nl
```
JakeSiFive, line 95: This comment is a bit misleading now I think, This positions the whitespace so that you can place a comment right? It doesn't touch the comment it just looks ahead at at it. I'd explain that out in a full comment.

JakeSiFive (approved):
yup this is a lot better

tools/wake-format/emitter.cpp
```diff
@@ -71,6 +90,16 @@ auto Emitter::rhs_fmt() {
...
+  return fmt().fmt_if(has_comment_next,
```
JakeSiFive, line 96: You currently duplicate the `has_comment_next` here and it use site, pick one and stick with it. I think here is better than there.

tools/wake-format/emitter.cpp
```diff
@@ -619,7 +649,35 @@ wcl::doc Emitter::walk_package(ctx_t ctx, CSTElement node) {
...
 wcl::doc Emitter::walk_paren(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  assert(node.id() == CST_PAREN);
+
+  // clang-format off
+  auto comment_fmt =
+      fmt().nest(fmt()
+             .token(TOKEN_P_POPEN)
+             .fmt_if(has_comment_next, position_comment().join(fmt().token(TOKEN_COMMENT)))
```
JakeSiFive, line 658: I think the `join(fmt().token(TOKEN_COMMENT)))` can be dropped perhaps? You'll have to make comments have trailing newlines instead of proceeding newlines.

tools/wake-format/emitter.cpp
V-FEXrt: This is needed at the callsite since we should only do `join(fmt().token(TOKEN_COMMENT))` if `position_comment()` evaluates.

I also agree that `position_comment` should be the place where the check lives.

What do you think about having position_comment take `fmt().token(TOKEN_COMMENT)` as a parameter?

Then the callsite becomes

```
fmt()
  ...
  .position_comment(fmt().token(TOKEN_COMMENT))
  ...
```

it also gives a slightly stronger enforcement of the fact that the caller needs to consume the comment

JakeSiFive: Yeah that's better...that said if this is always how this works position_comment should just consume the token

tools/wake-format/emitter.cpp
V-FEXrt: we talked about it in DM, posting here for posterity

Moving/deleting it would be ideal but that causes an extra trailing newline right before the expression and I'm trying very hard to not use  builder->undo so we have to keep it like this to not generate the extra newline

tools/wake-format/emitter.cpp
V-FEXrt: yeah. there was a reason that I split it out, but I'm not seeing it anymore. Maybe that was part of a previous issue.
I'll lift it back into the function and we can always pull it back out later if needed.

PR #973, opened 2022-08-09
 wake-format documentation

V-FEXrt:
This is the first draft of the high level documentation for `wake-format`. It should include most things needed to jump in and contribute to `wake-format`

### DIFF
--- a/tools/wake-format/README.md
+++ b/tools/wake-format/README.md
@@ -18,17 +18,17 @@ The information below should be helpful to contribute to `wake-format`
 
 ## Building
 
-`wake-format` is built in the same manner as `wake`. Running `make` or `./bin/wake build default` from within the wake dir will produce the `wake-format` binary.
+`wake-format` is built in the same manner as `wake`. Running `make` from within the wake dir will produce the `wake-format` binary. `./bin/wake build default` may also be used after running `make` the first time.
 
-The locally build `wake-format` can be found at `./bin/wake-format.native-cpp14-release`
+The locally build `wake-format` can be found at `./bin/wake-format.native-cpp14-release` or at `$DIR/bin/wake-format` after running `DEST_DIR="$DIR" make install`
 
 ## wake-format Fluent API/DSL
 
-`wake-format` relies heavily on a custom fluent API/DSL to generate the formatting rules. Usage of the API is documented below. See [here](TODO) for design decisions, invariants, and engineering tradeoffs considered.
+`wake-format` relies heavily on a custom fluent API/DSL to generate the formatting rules. Usage of the API is documented below. See [here](#design-invariants-and-tradeoffs) for design decisions, invariants, and engineering tradeoffs considered.
 
 ### Usage
 
-All formatting rules start by creating a basic formatting object using `fmt()`. From there functions are added to describe the expectations of the current [CST node](TODO) and to return the appropiate text.
+All formatting rules start by creating a basic formatting object using `fmt()`. From there functions are added to describe the expectations of the current [CST node](#cstelement) and to return the appropiate text.
 
 ```c++
 auto basic_fmt = fmt()
@@ -38,7 +38,7 @@ auto basic_fmt = fmt()
 
 creates a formatter that returns a space and a newline.
 
-Once a formatter has been defined, the `format()` function is used to run the formatter and return the formated source. The source is returned as a [wcl::doc](TODO) for efficient concatenation and geometry lookup. `format()` takes a `ctx_t` and a `CSTElement`. The `ctx_t` (context) tracks the context (such as nest level) that all subnodes should be formatted in, and the `CSTElement` is the current node being formatted.
+Once a formatter has been defined, the `format()` function is used to run the formatter and return the formated source. The source is returned as a [wcl::doc](#document-wcldoc) for efficient concatenation and geometry lookup. `format()` takes a `ctx_t` and a `CSTElement`. The `ctx_t` (context) tracks the context (such as nest level) that all subnodes should be formatted in, and the `CSTElement` is the current node being formatted.
 
 ```c++
 ctx_t ctx;
@@ -57,12 +57,13 @@ std::string s = doc.as_string(); // " \n"
 ### API Functions & Examples
 
 **CSTElement Manipulation**
+---
 
 Manipulation of the current state of the formatter relating to CSTElement nodes.
 
-**`token(cst_id_t id)`**
+#### **`token(cst_id_t id)`**
 
-`assert()`s the current node.id() is `id` then emits the string representation of the node.
+Asserts the current node.id() is `id` then emits the string representation of the node.
 
 ```c++
 CSTElement node = "def";
@@ -73,9 +74,9 @@ auto doc = fmt()
 doc.as_string(); // "def "
 ```
 
-**`token(cst_id_t id, const char* str)`**
+#### **`token(cst_id_t id, const char* str)`**
 
-`assert()`s the current node.id() is `id` then emits `str`.
+Asserts the current node.id() is `id` then emits `str`.
 
 ```c++
 CSTElement node = "def";
@@ -86,7 +87,7 @@ auto doc = fmt()
 doc.as_string(); // "foo "
 ```
 
-**`next()`**
+#### **`next()`**
 
 Moves the current node forward by one element.
 
@@ -102,10 +103,11 @@ auto doc = fmt() // -> def
 ```
 
 **Control Flow**
+---
 
 Conditional formatting based on the state at a given point in the formatter.
 
-**`fmt_if(Predicate, Formatter)`**
+#### **`fmt_if(Predicate, Formatter)`**
 
 Applies `Formatter` if `Predicate` returns `true`. See [Predicate](#predicate) for more Predicate examples.
 
@@ -118,7 +120,7 @@ auto doc = fmt()
 doc.as_string(); // "foo"
 ```
 
-**`fmt_if_else(Predicate, Formatter if_fmt, Formatter else_fmt)`**
+#### **`fmt_if_else(Predicate, Formatter if_fmt, Formatter else_fmt)`**
 
 Applies `if_fmt` if `Predicate` returns `true`, `else_fmt` otherwise. See [Predicate](#predicate) for more Predicate examples.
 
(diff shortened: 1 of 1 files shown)


tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+Running `wake-format` without `-i` will emit the file to `stdout` instead of overwritting it.
+
+# Documentation
+
+The information below should be helpful to contribute to `wake-format`
+
+## Building
+
+`wake-format` is built in the same manner as `wake`. Running `make` or `./bin/wake build default` from within the wake dir will produce the `wake-format` binary.
```
mmjconolly, line 21: > or `./bin/wake build default` from within

This implies that they've already run `make` at least once, it might be nice to add a Makefile target for the formatter even if it just calls wake under the hood

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
+# wake-format
+
+`wake-format` is a tool for automatically applying the wake style guide to wake source files. It is intentionally opinionated to create a singular correct formatting standard.
+
+# Usage
```
mwachs5, line 5:
```suggestion
## Usage
```

should there be some heading levels?

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+# Documentation
+
+The information below should be helpful to contribute to `wake-format`
+
+## Building
+
+`wake-format` is built in the same manner as `wake`. Running `make` or `./bin/wake build default` from within the wake dir will produce the `wake-format` binary.
+
+The locally build `wake-format` can be found at `./bin/wake-format.native-cpp14-release`
```
mmjconolly, line 23: We should probably have the build system produce a normalized name for it, either via the makefile, or via wake's `installAs`

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+## wake-format Fluent API/DSL
+
+`wake-format` relies heavily on a custom fluent API/DSL to generate the formatting rules. Usage of the API is documented below. See [here](TODO) for design decisions, invariants, and engineering tradeoffs considered.
+
+### Usage
+
+All formatting rules start by creating a basic formatting object using `fmt()`. From there functions are added to describe the expectations of the current [CST node](TODO) and to return the appropiate text.
```
mmjconolly, line 31: > current CST node

It may be helpful to spell out the acronym as individual words the first time it's used

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+`assert()`s the current node.id() is `id` then emits `str`.
+
+```c++
+CSTElement node = "def";
+auto doc = fmt()
+  .token(TOKEN_KW_DEF, "foo")
+  .space()
+  .format(ctx, node);
+doc.as_string(); // "foo "
```
mmjconolly, line 86: is `"foo "` intended to have the trailing space char? when I look at `next()` the `id` seems to be separated from spaces around it

V-FEXrt: The trailing space comes from `space()` on line 84. I think I was trying to make the example slightly less trivial however if it is confusing it should just be removed

tools/wake-format/README.md
V-FEXrt: Yeah, I need to figure out the correct nesting. I hit `#####` at one point which I couldn't tell apart from normal text so I un-nested everything by one, I don't think that was the correct choice though. Will work on it! Thanks :)

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+`format()` expects to consume the entire `CSTElement` node and will `assert()` if that is not the case. Asserting may seem heavily handed, however this is a very powerful method for identifying and fixing nodes with unexpected contents. 
+
+### API Functions & Examples
+
+**CSTElement Manipulation**
+
+Manipulation of the current state of the formatter relating to CSTElement nodes.
+
+**`token(uint8_t id)`**
```
mmjconolly, line 63: Is it important that the user knows that it's a `uint8_t`, wondering if a typedef (or other opaque) would make it more self describing

tools/wake-format/README.md
mmjconolly: oh somehow my eyes entirely slid over the `space()`

tools/wake-format/README.md
V-FEXrt: Jake and I talked about this previously. I'm actually very pro adding a typedef here. Would make it easier to bump the size if we ever have more that 256 IDs as well. 

It's one of those things where I'd want to change it everywhere if we added it though. I'll look into at least making the change for `wake-format` but I think `wake` would greatly benefit from it

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+Applies `Formatter` if `Predicate` returns `true`. See [Predicate](#predicate) for more Predicate examples.
+
+```c++
+CSTElement node = "def id = 5";
+auto doc = fmt()
+  .fmt_if(TOKEN_KW_DEF, fmt().token(TOKEN_KW_DEF, "foo"))
+  .fmt_if(TOKEN_COMMENT, fmt().token(TOKEN_COMMENT, "bar"))
+  .format(ctx, node);
+doc.as_string(); // "foo"
+```
```
mmjconolly, line 119: When we do `doc.as_string()` and get `foo`, is that in replacement for the entierity of `id = 5`, or just `id`, or all of `def id = 5`?

tools/wake-format/README.md
JakeSiFive: Yeah we just add it to the installs with shim-wake, job-cache, and wakebox

tools/wake-format/README.md
JakeSiFive: It should be noted that this documentation is largely intended for wake devs at this time, we'll add user documentation once we have a user interface :)

tools/wake-format/README.md
JakeSiFive: Well this example would actually crash because it didn't consume all the nodes.  Ignoring that this would only output "foo" because only the `def` was consumed.

tools/wake-format/README.md
mmjconolly: Yep I can see that, mostly I myself had forgotten what CST stood for

tools/wake-format/README.md
JakeSiFive: There's a partial analogy for this being dual to a parser/regex if that's helpful for you. `format` implicitly has an ending anchor at the end which is why this would fail. But ignoring that anchor, this would only match and consume the `def` and would output `foo` in its place.

tools/wake-format/README.md
```diff
@@ -0,0 +1,481 @@
...
+#### FitsPredicate
+
+Determines if a given formatter `fits` if appended to the current builder. Fitting is defined as being less than/equal to the max column width for the line of text. Used internally for `fmt_if_fits`
+
+```c++
+// FMT needs to be derived via a template function
+auto pred = FitsPredicate<FMT>(fmt())
+```
+
+## Fits
```
mmjconolly, line 413: Is this regarding things like line length?

tools/wake-format/README.md
V-FEXrt: Yeah I was actually debating how to handle the fact that most of these examples actually crash. Fully consuming the input makes the example tedious, and giving a partial node is unclear.

maybe an imaginary `consume_rest()` function

JakeSiFive: walk is also annoying to deal with in examples like this but unavoidable I think

tools/wake-format/README.md
V-FEXrt: https://github.com/sifive/wake/pull/975 should fix that, I'll update the documentation once it lands

tools/wake-format/README.md
V-FEXrt: If you run `make install` then it _does_ get installed/renamed appropriately

```
> DESTDIR=../install make install
WAKE_PATH=/usr/bin ./bin/wake build default
Pass "BUILD"
WAKE_PATH=/usr/bin ./bin/wake install ../install
Pass "INSTALL"
> ls ../install/bin 
job-cache  wake  wakebox  wake-format
```

Is that what you mean, or do we also want to copy from `./bin/wake-format.native-cpp14-release` to `./bin/wake-format`

tools/wake-format/README.md
JakeSiFive: I think users that would prefer `make wake-format` can just call `make` instead.

tools/wake-format/README.md
JakeSiFive: That's all that's needed. I think you've satisfied the requirement.

tools/wake-format/README.md
```diff
@@ -0,0 +1,506 @@
...
+- `pred_fits(FMT)`: `pred()` where `Predicate = FitsPredicate`. See [Fits](#fits) for details.
+- `otherwise(FMT)` unconditionally applies `FMT` used as the default/else case of a match.
+
+```c++
+CSTElement node = "def id = 5";
+auto doc = fmt()
+  .match(
+     pred(TOKEN_COMMENT, fmt().token(TOKEN_COMMENT, "bar"))
+    .pred(TOKEN_KW_DEF, fmt().token(TOKEN_KW_DEF, "foo"))
+    .fits(fmt().space())
```
JakeSiFive, line 166: pred_fits

tools/wake-format/README.md
```diff
@@ -0,0 +1,506 @@
...
+---
+
+Dispatch formatting to something else.
+
+`Walker`s must have the type signature `(ctx_t ctx, CSTElement node) -> wcl::doc` The typical use case is to recursivly call into some other function that _walks_ a specific type of `CSTElement`.
+
+
+#### **`walk(Walker)`**
+
+Unconditionally applies the `Walker`
```
JakeSiFive, line 196: Maybe mention the WALK macro here?

JakeSiFive (approved):
I'm quite happy with this but maybe ask for an overview from Matthew/Sam or Wes to get a second pair of eyes. I'll try to write at least one of those TODO sections a week from here.

JakeSiFive:
Ashley said I could close this

PR #974, opened 2022-08-10
Create explicit type for CSTElement ids

V-FEXrt:
Creates an explicit type for CSTElement Ids. This makes the documentation in `wake-format` more clear and serves as a safety net if we ever expand past 256 CST ids

I know I missed some. `lexer.h` and `lexer.re` has a lot of suspicious `uint8_t`s, but this is a start down the correct path.

### DIFF
--- a/src/parser/cst.h
+++ b/src/parser/cst.h
@@ -69,6 +69,8 @@
 #define CST_UNARY 164
 #define CST_ERROR 255
 
+typedef uint8_t cst_id_t;
+
 class FileContent;
 class CSTElement;
 class DiagnosticReporter;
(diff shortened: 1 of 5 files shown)


src/parser/cst.h
```diff
@@ -69,6 +69,8 @@
 #define CST_UNARY 164
 #define CST_ERROR 255
 
+typedef uint8_t cst_id_t;
```
JakeSiFive, line 72: Don't use `typedef` in modern C++, use `using cst_id_t = uint8_t`

JakeSiFive (approved):

PR #975, opened 2022-08-10
build: allow specific targets to be built

V-FEXrt:
Enables building specific targets by specifying another parameter to `wake build`

 Examples:
```
wake build default
wake build debug
wake build wake-format default
wake build job-cache debug
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,9 @@ vscode:		wake.db
 static:	wake.db
 	$(WAKE_ENV) ./bin/wake static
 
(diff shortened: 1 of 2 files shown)


JakeSiFive (commented):
This isn't bad but it isn't good either. It's kind of neutral. As I understand it this was done to get a `make` target, if you want that just run `make` instead of `make wake-format`

I think the build already has enough crufty features that we don't need to add this frankly. It's a slight negative to do this I'd argue.

PR #976, opened 2022-08-10
which: Add whichInEnvPath

V-FEXrt:
The repeated code here has been bothering me for a bit, had a minute so I decided to knock it out.

There are a few places inside the wake build where we depend on a cli program present in the users path. Each of them thus far implemented the same helper function to get it. This change makes a function for them all to share


JakeSiFive (approved):
Thanks!

ag-eitilt (approved):

PR #977, opened 2022-08-11
Alert user with wake-format fails

V-FEXrt:
There are certain cases where using `# wake-format off` outputs code that won't successfully parse. In the long term these cases will be handled better with appropriate warnings emitted but in the shorter term we need to collect real world examples of when/how/how often this happens before committing resources to pin down a hairy heuristic. For now, we fall over and ask the user to manually fix the issue and submit an example of the problematic code.


tools/wake-format/wake-format.cpp
```diff
@@ -167,7 +174,9 @@ int main(int argc, char **argv) {
     exit(EXIT_FAILURE);
   }
 
-  wcl::xoshiro_256 rng(wcl::xoshiro_256::get_rng_seed());
+  auto seed = (no_rng) ? std::tuple<uint64_t, uint64_t, uint64_t, uint64_t>({0, 0, 0, 0})
```
JakeSiFive, line 177: Oh interesting we need this because the tmp file is a part of the error message and we need to be deterministic but we *also* need running two of the same wake-formats to not conflict....interesting....Yeah I don't see a good way around this.

JakeSiFive (approved):

PR #978, opened 2022-08-11
Sketch of diff algorithm

JakeSiFive:
This needs some work (tests, pretty printing, etc...) but I think using this a base for something a lot better than just calling `diff`

### DIFF
--- a/src/wcl/diff.h
+++ b/src/wcl/diff.h
@@ -19,10 +19,10 @@
 
 #include <algorithm>
 #include <memory>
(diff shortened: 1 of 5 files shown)


V-FEXrt (commented):
LGTM other than a few small things

1) needs tests
2) needs formatted
3) there are a couple debug prints left in the code that should be removed

V-FEXrt (approved):

V-FEXrt (approved):

PR #979, opened 2022-08-15
shell: always escape empty strings

V-FEXrt:
Bug Fix: When calling `shell_escape()` over the components of a shell command and collecting it into a string [as done here](https://github.com/sifive/wake/blob/master/tools/wakebox/wakebox.cpp#L179-L186) an empty string gets re-emitted as an empty string effectively erasing it as an argument.
 
Ex: 
```
0: bash
1: -c
2: echo $#
3: test0
4: 
5: test1
6: test2
```
becomes `bash -c 'echo $#' test0 test1 test2` instead of the expected `bash -c 'echo $#' test0 '' test1 test2`

Fix this by always escaping empty string.

### DIFF
--- a/tests/wake-unit/stdout
+++ b/tests/wake-unit/stdout
@@ -24,6 +24,10 @@ PASSED:
   option_some
   sanity_check1
   sanity_check2
+  shell_escape_empty_string
+  shell_escape_nominal
+  shell_escape_spaces
+  shell_escape_special
   trie_basic
   trie_basic_const
   trie_char
--- a/tools/wake-unit/shell.cpp
+++ b/tools/wake-unit/shell.cpp
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2022 SiFive, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You should have received a copy of LICENSE.Apache2 along with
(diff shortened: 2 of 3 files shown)


tests/wake-unit/stdout
```diff
@@ -24,6 +24,10 @@ PASSED:
   option_some
   sanity_check1
   sanity_check2
+  shell_escape_empty_string
+  shell_escape_nominal
+  shell_escape_spaces
+  shell_escape_special
```
JakeSiFive, line 30: bless

JakeSiFive (approved):
Solid debugging. Thanks! Glad we added tests for this!

tools/wake-unit/shell.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "util/shell.h"
+
+#include "unit.h"
+
```
JakeSiFive, line 21: Wait actually can you add a test for leading/trailing spaces?

JakeSiFive (approved):
thanks for writing theses test! I'm glad *you* added them ;)

All good to go, land it

PR #980, opened 2022-08-16
Fix width tracking in formatter

V-FEXrt:
Adds ctx.sub() inside of all calls to compose. Before this width wasn't being correctly tracked so the formatted wouldn't newline when the width was too high.

### DIFF
--- a/tools/wake-format/formatter.h
+++ b/tools/wake-format/formatter.h
@@ -174,7 +174,7 @@ struct NestAction {
   NestAction(FMT formatter) : formatter(formatter) {}
 
   ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
-    builder.append(formatter.compose(ctx.nest(), node));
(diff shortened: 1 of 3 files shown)


tools/wake-format/formatter.h
```diff
@@ -174,7 +174,7 @@ struct NestAction {
   NestAction(FMT formatter) : formatter(formatter) {}
 
   ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node) {
-    builder.append(formatter.compose(ctx.nest(), node));
+    builder.append(formatter.compose(ctx.sub(builder).nest(), node));
```
JakeSiFive, line 177: Oh shit right lol...whoopsy

The rule stated was that if you throw out the builder you have to sub....I somehow turned a blind eye to the nest case lol.

I think if `nest` had the affect of also adding a newline we would have been right...oh well.

JakeSiFive (approved):
100% a good change

PR #983, opened 2022-08-18
Enable wake to build against wasm target

V-FEXrt:


V-FEXrt:
I wasn't able to test this since I still can't build without gmp.h, but this removed the curses build error

JakeSiFive (approved):

JakeSiFive:
> I wasn't able to test this since I still can't build without gmp.h, but this removed the curses build error

Yep, I'll fix this tommarow

JakeSiFive (approved):
Not sure if this is the working version but I'm happy with all the code here

.circleci/config.yml
```diff
@@ -212,6 +223,9 @@ workflows:
       - vscode:
           requires:
             - tarball
+      - wasm:
+          requires:
+            - tarball
```
JakeSiFive, line 228: Oh I got this wrong. I assumed vscode didn't depend on tarball. Still given that it does this is the right outcome.

JakeSiFive (approved):

PR #986, opened 2022-08-23
vscode: make available in web

elena-pudovkina:
Node and web implementations of both client and server are now separated.

Instead of executing the server binary, typescript calls exposed wasm functions.

LSPServer::callMethod() returns the response to a request, as well as diagnostics and a notification (instead of dumping messages to stdout).

The extension is compiled with webpack.


tools/lsp-wake/lsp.cpp
```diff
@@ -538,40 +516,105 @@ class LSPServer {
...
+#include <emscripten/emscripten.h>
+
+void instantiateServerInternal(const std::string &stdLib) {
+  // clang-format off
+  int isNode = EM_ASM_INT({
+    if (ENVIRONMENT_IS_NODE) {
+      return 1;
+    } else {
+      return 0;
+    }
```
V-FEXrt, line 554: does `return ENVIRONMENT_IS_NODE` not work?

tools/lsp-wake/lsp.cpp
```diff
@@ -538,40 +516,105 @@ class LSPServer {
...
+  if (isNode) {
+    if (is_readable((stdLib + "/core/boolean.wake").c_str())) {
+      lspServer = new LSPServer(true, stdLib);
+    } else {
+      lspServer = new LSPServer(false, stdLib);
+    }
+  } else {  // no need to check stdlib validity in web, since it can't be customized there
+    lspServer = new LSPServer(true, stdLib);
   }
+}
```
V-FEXrt, line 566: nit:

```js
let readable = !isNode || is_readable((stdLib + "/core/boolean.wake").c_str());
lspServer = new LSPServer(readable, stdLib);
```

V-FEXrt: or maybe 

```js
var readable = false;
if (isNode) {
  readable = is_readable((stdLib + "/core/boolean.wake").c_str());
} else {
  readable = true;
}

lspServer = new LSPServer(readable, stdLib);

File: tools/lsp-wake/lsp.cpp [hunk 3]
```diff
@@ -538,40 +516,105 @@ class LSPServer {
...
+
+  // Turn the json into a string
+  std::stringstream stringstream;
+  stringstream << jsonResult;
+  std::string str = stringstream.str();
+  const char *c = str.c_str();
+  size_t length = strlen(c);
+
+  // Malloc the string for typescript to use. Typescript will free it
+  char *result = static_cast<char *>(malloc(length + 1));
```
V-FEXrt, line 593: who frees  result?

tools/lsp-wake/lsp.cpp
```diff
@@ -538,40 +516,105 @@ class LSPServer {
...
+
+#else
+
+void instantiateServer(const std::string &stdLib) {
   if (is_readable((stdLib + "/core/boolean.wake").c_str())) {
-    lsp = LSPServer(stdLib);
+    lspServer = new LSPServer(true, stdLib);
+  } else {
+    lspServer = new LSPServer(false, stdLib);
+  }
```
V-FEXrt, line 607:
```js
lspServer = new LSPServer(is_readable((stdLib + "/core/boolean.wake").c_str()), stdLib);
```

extensions/vscode/lsp-client/src/clientWeb.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
```
JakeSiFive, line 4: Does MIT require an author's license?

extensions/vscode/lsp-client/src/common.ts
```diff
@@ -0,0 +1,72 @@
...
+export function registerFsMethods(client: CommonLanguageClient, stdLib: string): void {
+    client.onReady().then(() => {
+        client.onRequest('getStdLib', async (): Promise<string> => {
+            return stdLib;
+        });
+
+        const decoder = new TextDecoder();
+
+        client.onRequest('isReadable', async (path: string): Promise<number> => {
+            try {
```
JakeSiFive, line 39: I think promises have some sort of exception handling thing that they can do if you want to check on that. That's likely better and doesn't require a universal reduction of the error to just '0' and probably lets us display a better error message to the user at the site of use.

https://javascript.info/promise-error-handling

extensions/vscode/lsp-client/src/common.ts
```diff
@@ -0,0 +1,72 @@
...
+                let uri = Uri.parse(path);
+                await workspace.fs.stat(uri);
+                return 1;
+            } catch {
+                return 0;
+            }
+        });
+
+        client.onRequest('readFile', async (path: string): Promise<string | any> => {
```
JakeSiFive, line 48: `string | any` is the same thing as `any` I believe but read the thing below about error handling

extensions/vscode/lsp-client/src/common.ts
```diff
@@ -0,0 +1,72 @@
...
+            try {
+                let uri = Uri.parse(path);
+                const content = await workspace.fs.readFile(uri);
+                return decoder.decode(content);
+            } catch (err) {
+                return { message: err.toString() };
```
JakeSiFive, line 54: Ditto here, you're really using `string | any` to document that if you don't return a string its an error (it could just as easily been the other way though). It would be better to make this a `Promise<string>` and then have the promise handle the exceptions.

extensions/vscode/lsp-client/src/common.ts
```diff
@@ -0,0 +1,72 @@
...
+            try {
+                let uri = Uri.parse(path);
+                const files = await workspace.fs.readDirectory(uri);
+                let strippedFiles: [string, boolean][] = []; // [fileName, isDirectory]
+                for (const file of files) {
+                    strippedFiles.push([ file[0], !!(file[1] & FileType.Directory) ]);
+                }
+                return strippedFiles;
+            } catch (err) {
+                return { message: err.toString() };
```
JakeSiFive, line 68: ditto here again.

src/compat/readable.c
```diff
@@ -25,20 +25,35 @@
...
+      try {
+        const fs = require('fs');
+        fs.accessSync(UTF8ToString($0), fs.constants.R_OK);
+        return 1;
+      } catch (err) {
+        return 0;
+      }
+    } else {
       return 0;
     }
```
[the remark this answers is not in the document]
JakeSiFive: +1

extensions/vscode/lsp-server/wasm/lsp-wake.pre.js
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference types="emscripten" />
+
+var Module = {
+    'locateFile': function (wasmBinaryFile, prefix) {
+        if (ENVIRONMENT_IS_NODE) {
+            return prefix + "/../wasm/" + wasmBinaryFile; // TODO: understand why a hash file is needed
```
JakeSiFive, line 22: Just noting that we still haven't solved this. I'm happy to come back to this later though, no need to change it here. It looks like with `ENVIRONMENT_IS_NODE` you made it work in both situations and everything is still using relative paths which is good.

src/compat/readable.c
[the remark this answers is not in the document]
JakeSiFive: Personally for me, I loose track of what conditions are true and the nesting makes it hard for me to look it up and/or remember. But I'd argue is generally good practice and is widely done as Ashley said. Also a lot of times what you really doing is eliminating non-happy-path conditions so your brain really doesn't have to think about what conditions are true/false, it just does what it expects.

When you get really good at reviewing code this sort of high level of nesting visually stands out but the code that Ashley wrote looks a like less "noisy" and *feels* cleaner. It's a subjective feeling but I think it's a widely held subjective feeling.

And heck, I'm here to impart my style of programming on others even if that style was imparted on me by others :)

src/parser/wakefiles.cpp
```diff
@@ -45,38 +45,42 @@
...
-          if (dirent.isDirectory()) {
-            if (dirent.name != ".build" && dirent.name != ".fuse" && dirent.name != ".git")
-              walkTree(absolute);
-          } else {
-            files.push(absolute);
-          }
-      });
+  async function walkTree(dir) {
+    const dirFiles = await wakeLspModule.sendRequest('readDir', dir);
+    if (dirFiles.hasOwnProperty('message')) { // readDir request resulted in an error
```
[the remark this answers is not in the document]
JakeSiFive: We should use promise error handling here I think. We saw the same issue but commented on it in different spots :)

src/util/file.cpp
```diff
@@ -90,33 +90,41 @@ StringFile::StringFile(const char *filename_, std::string &&content_)
...
+// clang-format off
+EM_ASYNC_JS(int, getBase, (int *length, const char *filename), {
+  try {
+    const content = await wakeLspModule.sendRequest('readFile', UTF8ToString(filename));
+    if (content.hasOwnProperty('message')) { // readFile request resulted in an error
+      throw content;
+    }
+    const encoder = new TextEncoder();
+    const bytes = encoder.encode(content);
+    const wasmPointer = Module._malloc(bytes.length + 1);
```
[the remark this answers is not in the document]
JakeSiFive: The contract here is that you have to free it when its returned. This sort of always has to happen when interfacing with certain types between javascript and C

Module._malloc == C malloc and Moudle._free == free so we're safe and in the code being used below that's what we get.

src/util/file.cpp
```diff
@@ -90,33 +90,41 @@ StringFile::StringFile(const char *filename_, std::string &&content_)
 #ifdef __EMSCRIPTEN__
 #include <emscripten/emscripten.h>
 
+// clang-format off
+EM_ASYNC_JS(int, getBase, (int *length, const char *filename), {
```
JakeSiFive, line 94: Why does this return `int`? Why not make it return uint8_t* instead?

src/util/file.cpp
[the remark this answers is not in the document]
JakeSiFive: Specifically ExternalFile takes ownership in this case and frees it in the destructor.

JakeSiFive (commented):
I haven't reviewed lsp.cpp yet but I'm going to come back to that later

extensions/vscode/.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
+// A launch configuration that compiles the extension and then opens it inside a new window
```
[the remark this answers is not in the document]
richardxia: A lot of [Microsoft's tools](https://github.com/microsoft/vscode/issues/100688) use [JSONC](https://github.com/json-c/json-c), which is basically JSON with comments. Many of their tools generate a default config file with comments embedded in them, so comments are supported, even if I can't find an official description of it in the VS Code docs.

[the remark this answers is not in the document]
JakeSiFive: I'm pro comments here then. It doesn't seem like we'll ever need to work around this and if we do its a few deletes away.

extensions/vscode/lsp-server/src/common.ts
```diff
@@ -0,0 +1,208 @@
...
+                // the default stdLib is if the custom lib is invalid. in that case the user sees the notification and
+                // can act accordingly.
+                initializeRequest = '';
+            }
+            // Process all subsequent requests normally.
+            return await lspModule.processRequest(request);
+        });
+
+        if (lspModule == null) {
+            return new ResponseError(-32603, 'Failed to load wake lsp module.'); // InternalError
```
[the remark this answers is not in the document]
elena-pudovkina: Done

src/compat/readable.c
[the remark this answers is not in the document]
elena-pudovkina: Done

tools/lsp-wake/lsp.cpp
elena-pudovkina: Done

tools/lsp-wake/lsp.cpp
elena-pudovkina: Done

src/util/file.cpp
elena-pudovkina: Done

src/parser/wakefiles.cpp
```diff
@@ -45,38 +45,42 @@
...
+    }
+    for (const dirent of dirFiles) {
+      const absolute = Path.join(dir, dirent[0]); // dirent.name
+      if (dirent[1]) { // dirent.isDirectory
+        if (dirent[0] !== ".build" && dirent[0] !== ".fuse" && dirent[0] !== ".git") {
+          await walkTree(absolute);
+        }
+      } else {
+        files.push(absolute);
+      }
```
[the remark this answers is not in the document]
elena-pudovkina: Done

src/util/file.cpp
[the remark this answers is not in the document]
elena-pudovkina: Added a clarifying comment

tools/lsp-wake/lsp.cpp
```diff
@@ -111,10 +65,49 @@ const char *term_normal() { return ""; }
...
+                                                              "Must request initialize first");
+        return MethodResult(errorMessage);
+      } else if (isShutDown && (method != "exit")) {
+        JAST errorMessage = JSONConverter::createErrorMessage(
+            request, InvalidRequest,
+            "Received a request other than 'exit' after a shutdown request.");
+        return MethodResult(errorMessage);
+      } else if (!method.empty()) {
+        return callMethod(method, request);
+      }
```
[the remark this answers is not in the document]
elena-pudovkina: Done

tools/lsp-wake/lsp.cpp
elena-pudovkina: Done

tools/lsp-wake/lsp.cpp
elena-pudovkina: Added a clarifying comment

tools/lsp-wake/lsp.cpp
```diff
@@ -111,10 +65,49 @@ const char *term_normal() { return ""; }
...
+  };
+
+  MethodResult processRequest(const std::string &requestString) {
+    // Parse that requestString as JSON
+    JAST request;
+    std::stringstream parseErrors;
+    if (!JAST::parse(requestString, parseErrors, request)) {
+      JAST errorMessage = JSONConverter::createErrorMessage(ParseError, parseErrors.str());
+      return MethodResult(errorMessage);
+    } else {
```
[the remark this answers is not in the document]
elena-pudovkina: Done

src/parser/wakefiles.cpp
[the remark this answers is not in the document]
elena-pudovkina: Using promise error handling now :)

extensions/vscode/lsp-client/src/common.ts
elena-pudovkina: We don't display the error in this specific case, but I modified it so that we could if we wanted.

extensions/vscode/lsp-client/src/clientWeb.ts
elena-pudovkina: Made some changes to the licensing, please check if things are looking correct now

elena-pudovkina: Also, you can now use the Wake nightly extension to review the pull request in github.dev!

JakeSiFive (approved):
This looks good to me! Thanks Elena!

V-FEXrt (approved):

extensions/vscode/vscode.wake
```diff
@@ -29,28 +29,29 @@ def installRO path =
...
     | getJobOutput
 
+def createStdlibJSON (stdlibSources: List Path): Result Path Error =
+    def formatPath path =
+        cat ("\t\t\"", (getPathName path | replace `share/wake/lib` ""), "\"",)
+    def formattedPaths = 
+        map formatPath stdlibSources
+    def content = 
+        cat ("\{\n\t\"files\": [\n", (catWith ",\n" formattedPaths), "\n\t]\n\}",)
+    write "{here}/stdlib.json" content
```
ag-eitilt, line 39: I should probably point out that we do have native JSON handling so you don't have to deal with the pretty-printing yourself:
```suggestion
    def formatPath path =
        getPathName path | replace `share/wake/lib` ""
    def formattedPaths =
        map formatPath stdlibSources
    def content =
        JObject ("files" :-> JArray (map JString formattedPaths), Nil)
    write "{here}/stdlib.json" (prettyJSON content)
```

ag-eitilt (commented):
Looks good (with one note), works well, and is worth a release!

extensions/vscode/vscode.wake
elena-pudovkina: Thanks!

PR #989, opened 2022-08-29
Extract command line options into dedicated class

V-FEXrt:
Moves all the CLI options into a dedicated class. This cleans up main/run_wake but the main purpose is to enable launching wake from web assembly. We can't repeatedly call `main()` from wasm so we need some other manner to enable things like

```
wake --init .
wake -x "5"
```

so from wasm we will now have something like

```
CommandLineOptions clo = //... set init = true
run_wake(clo)
CommandLineOptions clo = //... set command = "5"
run_wake(clo)
```


tools/wake/main.cpp
```diff
@@ -647,27 +717,29 @@ int main(int argc, char **argv) {
...
-  status_set_bulk_fd(3, fd3);
-  status_set_bulk_fd(4, fd4);
-  status_set_bulk_fd(5, fd5);
+  if (noexecute && !clo.fd1) clo.fd1 = "error";
+  if (clo.debug && !clo.fd1) clo.fd1 = "debug,info,echo,report,warning,error";
+  if (clo.verbose && !clo.fd1) clo.fd1 = "info,echo,report,warning,error";
+  if (clo.quiet && !clo.fd1) clo.fd1 = "error";
+  if (!clo.tty && !clo.fd1) clo.fd1 = "echo,report,warning,error";
+  if (!clo.fd1) clo.fd1 = "report,warning,error";
+  if (!clo.fd2) clo.fd2 = "error";
```
ag-eitilt, line 730: There's only a few places where we're editing the supposed "command-line" options so it's probably not worth setting up a separate stage for populating `clo` with default values, but this use of it as a mutable store (where it's less obvious it's just assigning defaults) feels a bit weird given the name.

ag-eitilt (approved):
I'm maybe not the best person to review this, but in service of getting it moving, I do like the idea -- we have a *lot* of potential CLI options and centralizing them in a way which obviously marks their use would have really helped months ago when I was tracing how the `fd*` streams were set.

PR #990, opened 2022-09-01
impl freshline() and fix formatting

V-FEXrt:
This pins down and solves the last major ws errors/issues we've been seeing. The following changes are made

1) newline() now  only emits a single NL
2) freshline() emits a single NL if the last line doesn't have one and brings the total whitespace up to the value required by the nest level
3) 'before bound' comments emit a freshline()
4) 'after bound' comments emit a newline()

### DIFF
--- a/tools/wake-format/formatter.h
+++ b/tools/wake-format/formatter.h
@@ -117,6 +117,29 @@ inline void newline(wcl::doc_builder& builder, uint8_t space_count) {
   space(builder, space_count);
 }
 
+inline void freshline(wcl::doc_builder& builder, ctx_t ctx) {
+  auto goal_width = SPACE_PER_INDENT * ctx.nest_level;
+  auto merged = ctx.sub(builder);
+
+  // there are non-ws characters on the line, thus a nl is required
+  if (merged->last_width() > merged->last_ws_count()) {
+    newline(builder, goal_width);
+    return;
+  }
+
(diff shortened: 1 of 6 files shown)


tools/wake-format/formatter.h
```diff
@@ -117,6 +117,29 @@ inline void newline(wcl::doc_builder& builder, uint8_t space_count) {
...
+  if (merged->last_width() < goal_width) {
+    space(builder, goal_width - merged->last_width());
+    return;
+  }
+
+  // If there are too many spaces, then a freshline() was used instead
+  // of newline(). Assert to ensure it is fixed.
+  if (merged->last_width() > goal_width) {
+    assert(false);
+  }
```
JakeSiFive, line 140: Ah yeah because if its too much and we newline then that's the same as leaving trailing space huh? So we should probs assert in that case...plus since we're not hitting it for right now best to assert.

JakeSiFive (approved):
Yus yus yus it's all coming togethor!

This is a nice clean short PR which is a nice break from the previous two lol

tools/wake-format/formatter.h
V-FEXrt: yeah, I don't think it should ever be a valid case to hit this but we can always remove the assert if I am wrong. I want to wait for a valid use case before doing that though

PR #993, opened 2022-09-09
Wake format binop full choice

V-FEXrt:
Implements 'full choice' for binary op formatting. Explores six different formatting options and then chooses the one that best optimizes for formatting, character width, and height.

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -655,8 +655,11 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
 
   // WARNING: I'm using Node() here which is differnt than everywhere else
   // This is probably not right since I'm loosing comments
-  CSTElement op = node.firstChildNode();
+  CSTElement lhs = node.firstChildNode();
+  CSTElement op = lhs;
   op.nextSiblingNode();
+  CSTElement rhs = op;
+  rhs.nextSiblingNode();
 
   assert(op.id() == CST_OP);
   CSTElement op_token = op.firstChildElement();
@@ -667,12 +670,37 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
   std::vector<std::pair<int, wcl::doc>> lte_fmt = {};
   std::vector<std::pair<int, wcl::doc>> gt_fmt = {};
 
+  size_t allowed_extra_nl = token_traits[lhs.firstChildElement()].before_bound.size();
+
+  CSTElement last_rhs;
+  {
+    IsWSNLCPredicate is_wsnlc;
+    CSTElement curr_rhs = rhs.firstChildElement();
+    CSTElement next_rhs = curr_rhs;
+    next_rhs.nextSiblingElement();
+
+    while (!next_rhs.empty()) {
+      while (!next_rhs.empty() && is_wsnlc(next_rhs)) {
+        next_rhs.nextSiblingElement();
+      }
+      if (next_rhs.empty()) {
+        last_rhs = curr_rhs;
+      } else {
+        curr_rhs = next_rhs;
+        next_rhs.nextSiblingElement();
+      }
+    }
+    last_rhs = curr_rhs;
+  }
+
+  allowed_extra_nl += token_traits[last_rhs].after_bound.size();
+
   // Maybe nest here
 
   // 1
   wcl::doc flat = combine_flat(op_token, ctx, parts);
   std::cerr << "#1 <<<:" << flat.as_string() << ":>>>" << std::endl;
-  if (!flat->has_newline()) {
+  if (flat->newline_count() == allowed_extra_nl) {
     if (flat->max_width() > MAX_COLUMN_WIDTH) {
       gt_fmt.push_back({1, flat});
     } else {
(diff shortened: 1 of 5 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -524,19 +526,335 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+            collect_over.fragment().segment().str())) {
+    return {node};
+  }
+
+  auto left_collect = collect_binary_parts(collect_over, left);
+  auto right_collect = collect_binary_parts(collect_over, right);
+
+  size_t size = left_collect.size();
+  left_collect.resize(size + right_collect.size());
+  std::move(right_collect.begin(), right_collect.end(), left_collect.begin() + size);
```
JakeSiFive, line 553: You should use `insert` to do this to avoid the default construction from `resize`

tools/wake-format/emitter.cpp
```diff
@@ -524,19 +526,335 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+  right.nextSiblingNode();
+
+  if (!(op.id() == CST_OP && op.firstChildElement().id() == collect_over.id() &&
+        op.firstChildElement().fragment().segment().str() ==
+            collect_over.fragment().segment().str())) {
+    return {node};
+  }
+
+  auto left_collect = collect_binary_parts(collect_over, left);
+  auto right_collect = collect_binary_parts(collect_over, right);
```
JakeSiFive, line 549: This is correct for a kind of strange reason IMO. This algorithm wouldn't work with an AST for instance. It only works becasue a PAREN will get in the way and the parser will save you from having to deal with the case where both the left and the right side are the collect_over operator. I would instead suggest separate left collect and right collect implementations personally but I'm willing to be convinced otherwise.

V-FEXrt: I'm down to fix this, but can I delay until after the next PR to minimize merge conflicts / simplify the impl?

V-FEXrt: nvm, I managed to pull it in

tools/wake-format/emitter.cpp
```diff
@@ -524,19 +527,307 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+  return allowed;
+}
+
+bool min_doc_height(const wcl::doc& lhs, const wcl::doc& rhs) {
+  return lhs->height() < rhs->height();
+}
+
+bool min_doc_max_width(const wcl::doc& lhs, const wcl::doc& rhs) {
+  return lhs->max_width() < rhs->max_width();
+}
```
JakeSiFive, line 630: nit: I'd call these compare_height and compare_width, also these can be static.

tools/wake-format/emitter.cpp
```diff
@@ -524,19 +527,307 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+
+// TODO: this is far from fully correct
+bool is_op_left_assoc(const CSTElement& op) { return op.id() == TOKEN_OP_OR; }
```
JakeSiFive, line 633: nit: static, also check if this function already exists in the parser somewhere.

tools/wake-format/emitter.cpp
```diff
@@ -524,19 +527,307 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+
+wcl::doc select_best_choice(std::vector<wcl::optional<wcl::doc>> choices) {
```
JakeSiFive, line 635: static, also please document that it is assumed that at least one of the choices is viable.

tools/wake-format/emitter.cpp
```diff
@@ -524,19 +527,307 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+}
+
+wcl::optional<wcl::doc> Emitter::combine_flat(CSTElement over, ctx_t ctx,
+                                              const std::vector<CSTElement>& parts) {
+  wcl::doc_builder builder;
+  for (size_t i = 0; i < parts.size() - 1; i++) {
+    builder.append(walk_node(ctx.sub(builder), parts[i]));
+    builder.append(fmt().space().compose(ctx.sub(builder), over, token_traits));
+    builder.append(walk_token(ctx.sub(builder), over));
+    builder.append(fmt().space().compose(ctx.sub(builder), over, token_traits));
```
JakeSiFive, line 678: Maybe this can be done with one formatter instead?

edit: hmm it looks like you do this a lot. I'm not overly tied to this really though. It feels a bit verbose though.

tools/wake-format/emitter.cpp
```diff
@@ -524,19 +527,307 @@ wcl::doc Emitter::walk_ascribe(ctx_t ctx, CSTElement node) {
...
+  return *min_width;
+}
+
+wcl::doc post_binop_spacing(CSTElement over, const token_traits_map_t& traits,
+                            const wcl::doc& fmt_binop, const wcl::doc_builder& builder, ctx_t ctx) {
+  if (fmt_binop->has_newline()) {
+    return fmt().freshline().compose(ctx.sub(builder), over, traits);
+  }
+  return fmt().space().compose(ctx.sub(builder), over, traits);
+}
```
JakeSiFive, line 669: Can you document what this is doing?

Also please static all the static thingies

JakeSiFive (commented):
I'm 99% cool with this. Please give the small things I asked about a look first.

tools/wake-format/emitter.cpp
V-FEXrt: That would be nice, but I can't seem to find one

tools/wake-format/emitter.cpp
V-FEXrt: I'm giving this a try but there are some fiddly details making it difficult. One being that `parts[i]` and `over` aren't necessarily siblings so it requires at least two `fmt()`s. That is still nicer though. I'm getting some subtle detail wrong somewhere and having difficulties pinning it down

JakeSiFive: No worries. Like I said I wasn't overly tied to it. It can be cleaned up later if you figure it out.

JakeSiFive (approved):

PR #1001, opened 2022-09-19
Resolve 'Grows NL' issues

V-FEXrt:
Several previous formatting cases could 'grow NLs' with the change in how comments are handled. These are now updated to do the correct thing in those cases.


JakeSiFive (approved):
I'm still not super happy with counting newlines but I don't have a better solution right now. I've forgotten our discussion on the matter unfortunately but I remember thinking something of this sort had to be done. The code otherwise looks fine.

V-FEXrt:
Yeah, I'm not a huge fan either, but I don't know of any alternatives either. Will keep mulling on it though

PR #1003, opened 2022-09-20
Format more wake subtrees 

V-FEXrt:
Formats the following

- multiline strings
- publish
- target
- require else

Adds the following improvements
- `-d` output is now sent to stderr to separate from the formatting output
- `is_op_left_assoc()` is now done
- Adds a "to string" for CSTElement IDs to make debug printing nicer


JakeSiFive:
So uh...I see 4 PRs here that all look very separable 👉 👈

PR #1004, opened 2022-09-20
Fix is_op_left_assoc TODO

V-FEXrt:


tools/wake-format/emitter.cpp
```diff
@@ -70,8 +70,21 @@ static bool compare_doc_width(const wcl::doc& lhs, const wcl::doc& rhs) {
...
-static bool is_op_left_assoc(const CSTElement& op) { return op.id() == TOKEN_OP_OR; }
+static bool is_op_left_assoc(const CSTElement& op) {
+  switch (op.id()) {
+    case TOKEN_OP_DOT:
+    case TOKEN_OP_QUANT:
+    case TOKEN_OP_MULDIV:
+    case TOKEN_OP_ADDSUB:
+    case TOKEN_OP_COMPARE:
+    case TOKEN_OP_AND:
+    case TOKEN_OP_OR:
```
JakeSiFive, line 81: Docs suggest there should be a "Test" and "InEqual" but here we just have "compare". I'm not sure what compare maps to but I would expect `!=` and `==` to be lower precedence than '<` or `>=` for instance but to still be left associative. Can you check if there is one more case? One of the quirks of wake is that `!` binds very loosely so I kind of expect that to be here already.

Those might have just been merged though. It's unlikely anyone was using equality operators in conjunction with comparision operators.

JakeSiFive (approved):
If there's no additional comparae/test/inequal case this is good to land. If there is just add it and land without further review.

tools/wake-format/emitter.cpp
V-FEXrt: https://github.com/sifive/wake/blob/master/src/parser/parser.y.m4#L214 is where I got it from

JakeSiFive: Looks like you missed OP_INEQUAL then?

JakeSiFive: oh no wait I'm wrong that's right associtive

JakeSiFive: And indeed the docs match and that's the expected associativity. I just got that wrong. TIL

Approved.

PR #1006, opened 2022-09-20
Format several subtrees

V-FEXrt:
Formats the following subtrees
 - multiline strings
 - publish
 - target
 - require else

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -54,7 +54,7 @@
 
 static inline bool requires_nl(cst_id_t type) { return type == CST_BLOCK || type == CST_REQUIRE; }
 static inline bool requires_fits_all(cst_id_t type) {
-  return type == CST_APP || type == CST_BINARY;
+  return type == CST_APP || type == CST_BINARY || type == CST_LITERAL || type == CST_INTERPOLATE;
 }
 
 static inline bool is_expression(cst_id_t type) {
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -1030,7 +1030,42 @@ wcl::doc Emitter::walk_lambda(ctx_t ctx, CSTElement node) {
...
+
+  // clang-format off
+  MEMO_RET(fmt().match(
+    // TODO: starting 'pred()' function doesn't allow init lists
+    pred(ConstPredicate(false), fmt())
+   .pred({TOKEN_MSTR_BEGIN, TOKEN_MSTR_RESUME}, multiline_str_fmt)
+   .pred([](wcl::doc_builder&, ctx_t, CSTElement& node,
+                  const token_traits_map_t&){ return node.isNode(); }, node_fmt)
+   .otherwise(token_fmt))
+   .format(ctx, node.firstChildElement(), token_traits));
```
JakeSiFive, line 1067: What are all the literal cases that are being handled here? multi-line string and token are two...what are the others?

Also does this handle interpolation? It doesn't need to in this version I'm just curious if somehow it already does.

tools/wake-format/emitter.cpp
```diff
@@ -1102,13 +1137,33 @@ wcl::doc Emitter::walk_prim(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_publish(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  assert(node.id() == CST_PUBLISH);
+
+  MEMO_RET(fmt()
+               .token(TOKEN_KW_PUBLISH)
+               .ws()
+               .walk(WALK_NODE)  // identifier
```
JakeSiFive, line 1145: I wonder if we want a WALK_ID to be a bit more pedantic than just WALK_NODE?

JakeSiFive (approved):
This looks good to me actully, no need to split it up. I read it all.

tools/wake-format/emitter.cpp
V-FEXrt: > What are all the literal cases that are being handled here?
You got it, right now its just multi-line string and token. There are likely more cases but I'll circle back around to them when I hit examples of them. Right now I'm just trying to format increasingly more complicated wake source files.

tools/wake-format/emitter.cpp
V-FEXrt: Yeah, we had something like that previously but removed it when we said "every dispatch needs to go through walk_node" 

I'd like to find that change and "revert" it now that we have DISPATCH

PR #1007, opened 2022-09-21
Improve binop separation

V-FEXrt:
Binop separation was moved to a single function/source of truth. This enabled things like moving the `,` to be a suffix instead of a prefix and had the bonus benefit of cleaning up the combine_* functions significantly

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -86,6 +86,16 @@ static bool is_op_left_assoc(const CSTElement& op) {
   }
 }
 
+static bool is_op_suffix(const CSTElement& op) {
+  switch (op.id()) {
+    case TOKEN_OP_COMMA:
+      return true;
+
+    default:
+      return false;
+  }
+}
+
 static size_t count_leading_newlines(const token_traits_map_t& traits, const CSTElement& node) {
   CSTElement token = node;
   while (token.isNode()) {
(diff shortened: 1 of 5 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -167,30 +177,23 @@ static wcl::doc select_best_choice(std::vector<wcl::optional<wcl::doc>> choices)
...
+  switch (op.id()) {
+    case TOKEN_OP_DOT:
+    case TOKEN_OP_COMMA:
+      return wcl::doc::lit("");
+    default:
+      return wcl::doc::lit(" ");
+  }
+}
+
+static wcl::doc binop_rhs_separator(const CSTElement& op) {
```
JakeSiFive, line 190: I like these

tools/wake-format/emitter.cpp
```diff
@@ -1229,6 +1207,67 @@ wcl::doc Emitter::walk_error(ctx_t ctx, CSTElement node) {
   MEMO_RET(walk_placeholder(ctx, node));
 }
 
+wcl::doc Emitter::place_binop(CSTElement op, bool is_flat, ctx_t ctx) {
```
JakeSiFive, line 1210: I like this as well

JakeSiFive (approved):
Yep I actually just think this is 100% right. Good job.

PR #1008, opened 2022-09-21
Fix getmtime for symlinks

V-FEXrt:


terpstra (approved):

terpstra:
You should probably get @ag-eitilt to confirm this fixes the problem he ran into before you merge, though.

terpstra (approved):
You should get @ag-eitilt to confirm this fixes his issue before merging.

V-FEXrt:
@ag-eitilt confirmed the fix in a call

mmjconolly:
It helps to have a description in the PR, or in the git commit, that describes for future contributors why this change was made

V-FEXrt:
Right, sorry. I didn't have all that much context myself.

`getmtime()` when called on a symlink was returning the modified time of the thing the symlink was pointing to, not the symlink's modified time. Which was causing some sort of bug. @ag-eitilt will need to comment on the actual behavior of the bug though

ag-eitilt:
`source`, `claim`, and related functions were ultimately using file modification time as a proxy for a file existence check.  This had been implemented with a syscall which checked the modification time of a symlink's target, rather than that of the link itself.  In normal operation, that's perfectly fine (if we follow the link somewhere, it obviously existed to be followed), but if the symlink was broken it would still *try* to follow the link, not find the target, and then report that the target doesn't exist -- which actually still ultimately results in similar enough behaviour in many cases.  However, if the link would be later fixed by the sandbox (say, it points to some absolute `/workspace` path), then the earlier rejection based on the non-sandboxed paths wound up being incorrect once both the link and the target got remapped to their in-sandbox locations.  With this change now no longer following the link, it doesn't care whether the symlink is broken or not during the `claim`, and happily passes it to the sandboxed job to take care of.

PR #1013, opened 2022-09-26
split long def types over multiple lines

V-FEXrt:
Very long types for `def ` will be split across several lines in an (ugly) manner. Unfortunately this is unavoidable as moving the parens changes the semantic meaning of the type

### DIFF
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -585,3 +585,19 @@ def defaultWake Unit =
     Pass (Pair "tmp/bin/wake" wakeVisible)
 
 
+def (
+    tool
+    (dir: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+    (addHeaders: String)
+) = ""
+
+
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -59,7 +59,8 @@ static inline bool requires_fits_all(cst_id_t type) {
 
 static inline bool is_expression(cst_id_t type) {
   return type == CST_ID || type == CST_APP || type == CST_LITERAL || type == CST_HOLE ||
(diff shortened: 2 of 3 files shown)


tests/wake-format/basic/stdout
```diff
@@ -585,3 +585,19 @@ def defaultWake Unit =
     Pass (Pair "tmp/bin/wake" wakeVisible)
 
 
+def (
```
JakeSiFive, line 588: One thing to watch out for since we're changing the CST here in a non-whitespace way, is that we're idempotent. That is if we see a long paren with a series of apply we format it the same way as a long series of applies in a pattern without this. I think we're consistent on this front right now but we should add a two tests

1) `def (tool (addHeaders: String) ... (addHeaders: String))`
2)
```
def (
  tool
  (addHeaders: String)
  ...
  (addHeaders: String)
)
```

This is to smoke test that we're not going to wind up adding parens to something that already gas parens

tools/wake-format/emitter.cpp
```diff
@@ -1284,7 +1285,7 @@ wcl::doc Emitter::walk_type(ctx_t ctx, CSTElement node) {
                .nest(fmt().freshline().walk(WALK_NODE).consume_wsnlc())
                .freshline()
                .lit(wcl::doc::lit(")"))
-               .format(ctx, node, token_traits));
+               .compose(ctx, node, token_traits));  // TODO: find a way to make this not-compose
```
JakeSiFive, line 1288: What is this it terrifies me.

JakeSiFive (commented):
holy crap this was super short. I have questions though

tools/wake-format/emitter.cpp
V-FEXrt: Yeah, I don't like it either. Its a similar problem to the "walk_children" problem mentioned before.

Basically, since `walk_type` isn't looking into node (via `node.firstChildElement`) it is possible (and likely) that there is some unconsumed sibling that should be consumed somewhere else. Doing `format` causes an assert

And it isn't acceptable to use `node.firstChildElement` since node is the type we are formatting and we loose the outer apply/paren/whatever if we look at the child.

V-FEXrt:
Superceded by https://github.com/sifive/wake/pull/1018

PR #1014, opened 2022-09-26
Format if subtree

V-FEXrt:
Formats the `if` subtree. Short expressions are on a single line

```
def x = if true then 5 else 3
```

while expressions with NLs or expressions that don't fit are split across several 

```
def linkKind =
    if verrrrrrrry loooooooooooog naaaaaaaaaaaame then
        verrrrrrrrrrrrrrrrrrrrrrry
    else
        looooooooooooooooong
```

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -1013,8 +1013,30 @@ wcl::doc Emitter::walk_if(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
   assert(node.id() == CST_IF);
 
-  // TODO: support flat case?
-  // Ex: def x = if true then 5 else 6
+  auto fits_no_nl =
+      fmt()
+          .fmt_if_fits_all(
+              fmt()
+                  .token(TOKEN_KW_IF)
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -978,7 +979,47 @@ wcl::doc Emitter::walk_ideq(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_if(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  assert(node.id() == CST_IF);
+
+  auto fits_no_nl =
```
JakeSiFive, line 984: I didn't see a test for this case up above.

JakeSiFive (approved):
Approved with addition of single line case

tools/wake-format/emitter.cpp
V-FEXrt: It was already in there, I only noticed to fix this because there was a regression. I'll add an explicit one though

PR #1015, opened 2022-09-26
Format data subtree

V-FEXrt:
Formats the `data` subtree

Ex:

```
export data JValue =
  JString  String
  JInteger Integer
  JDouble  Double
  JBoolean Boolean
  JNull
  JObject  (List (Pair String JValue))
  JArray   (List JValue)
```


JakeSiFive (approved):
Yep looks good to me. Unconditionally approved. This is awesome!

I think this conveniences me that we have our API pretty much down. This is all go so smoothly.

PR #1018, opened 2022-09-27
Implement cat() only subformatter

V-FEXrt:
Implements `cat()` format subsystem to be used when formatting without a CSTElement. This enables making `compose` internal/private ensuring that it is used correctly.

A follow up PR is needed to convert all the uses of `compose` to `cat()...`


tools/wake-format/catters.h
```diff
@@ -0,0 +1,99 @@
...
+struct FormatCatter {
+  FMT formatter;
+  CSTElement node;
+  const token_traits_map_t& traits;
+
+  FormatCatter(FMT formatter, CSTElement node, const token_traits_map_t& traits)
+      : formatter(formatter), node(node), traits(traits) {}
+
+  ALWAYS_INLINE void cat(wcl::doc_builder& builder, ctx_t ctx) {
+    builder.append(formatter.compose(ctx.sub(builder), node, traits));
```
JakeSiFive, line 97: You need to make a copy here or it will mutate the node member and that's going to be a major bitch. To protect against this you should make the `cat` methods const since they shouldn't mutate any internal state....actully maybe `run` should be const too for that matter?

tools/wake-format/catters.h
```diff
@@ -0,0 +1,100 @@
...
+  ALWAYS_INLINE void cat(wcl::doc_builder& builder, ctx_t ctx) { builder.append(std::move(lit)); }
+};
+
+template <class Catter1, class Catter2>
+struct SeqCatter {
+  Catter1 catter1;
+  Catter2 catter2;
+  SeqCatter(Catter1 c1, Catter2 c2) : catter1(c1), catter2(c2) {}
+
+  ALWAYS_INLINE void cat(wcl::doc_builder& builder, ctx_t ctx) {
```
JakeSiFive, line 59: Can we make all these const? I think it would improve the API

JakeSiFive (approved):

tools/wake-format/catters.h
V-FEXrt: Please see DMs :) 

tl;dr: I'm having some SFINAE issues, and the const correctness can't really be done iteratively, just needs to be right on the first try do to how things are interdependent

V-FEXrt: chatting in DM, forgoing this change for now

PR #1019, opened 2022-09-27
wake: check the hash of a symlink itself, not of the target

ag-eitilt:
When a symlink was included as part of a job's input or output files, wake had previously been hashing the *target* rather than the link itself; when the symlink was broken on-disc (e.g. referring to an absolute `/workspace` sandboxed path), this could cause the job to rerun unnecessarily as it saw the symlink as having been "deleted".  By using the flag-aware `faccessat` rather than the simpler `access`, we can instruct wake to just look at but not follow the link for the proper behaviour.  This is a similar but distinct bug to that solved by #1008.

This does break compatibility with some older versions of MacOS; current/recent versions do implement the required features.

### DIFF
--- a/.circleci/dockerfiles/alpine
+++ b/.circleci/dockerfiles/alpine
@@ -1,4 +1,4 @@
-FROM alpine:3.11.5
+FROM alpine:3.14.0
 
 RUN apk add m4 g++ make pkgconf git tar xz gmp-dev re2-dev sqlite-dev fuse-dev ncurses-dev dash sqlite-static ncurses-static linux-headers
 
--- a/scripts/fetch
+++ b/scripts/fetch
@@ -9,7 +9,6 @@ version="$2"
 
 curl -L -o wake_${version}.tar.xz                      ${url}/wake_${version}.tar.xz
 curl -L -o wake-${version}.vsix                        ${url}/vscode/wake-${version}.vsix
-curl -L -o wake-static_${version}.tar.xz               ${url}/alpine/wake-static_${version}.tar.xz
 curl -L -o centos-7-6-wake-${version}-1.x86_64.rpm     ${url}/centos_7_6/wake-${version}-1.x86_64.rpm
 curl -L -o rocky_8-wake-${version}-1.x86_64.rpm        ${url}/rocky_8/wake-${version}-1.x86_64.rpm
 curl -L -o debian-bullseye-wake_${version}-1_amd64.deb ${url}/debian_bullseye/wake_${version}-1_amd64.deb
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -21,6 +21,7 @@
 
 #include "database.h"
 
+#include <fcntl.h>
 #include <sqlite3.h>
 #include <string.h>
 #include <time.h>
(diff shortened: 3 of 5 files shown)


V-FEXrt:
Looks like this caused the symlinks test to fail. Can you TAL?

ag-eitilt:
Yeah, trying to figure out what's going on -- Wake is seeing two different hashes for the `badLink` which makes me think there might be something in the job output hashing which needs to be fixed, but I've not zeroed in quite yet.

JakeSiFive:
For what its worth, having missed the discussions I agree that his is the right behavior. I think if a test is failing the test is probably wrong or we have additional bad assumptions in the wake code base about how symlinks are to be handled. They should always be handled the way this PR changes us to handle them.

ag-eitilt:
@terpstra Do you remember what you had in mind for the `ln -s whatever xxLink` lines in the symlink test?  They don't seem to fit the behaviour we're wanting from symlinks (a different target gives a different hash), and the `faccessat` fix winds up breaking them.  I've rewritten it in https://github.com/sifive/wake/pull/1019/commits/fe1d59636f5bf3ad32c4dd83421c12b40344482a with something that seems more like what we'd be looking for, but I wanted to be sure I've not dropped some corner case you were wanting to exercise.

JakeSiFive:
To follow up on this for the record: Thanks to Sam's debugging and some rubber ducking from me we figured out the issue. The issue as Matthew suspected was musl having an implementation that didn't support the NOFOLLOW flag. Interestingly we would not have caught this if not for another bug! I think I was wrong about the fuse-deamon.cpp faccessat call, that should remain without a flag. That function is meant to behave like `access` not like `faccessat`. I'm not sure what happens in the kernel if a process calls `SYS_faccessat2` on a path mapped to a fuse fs but it doesn't pass the flags along to fuse since there's no argument for that. So passing zero should be correct.

Sam found that Alpine 3.14 has a sufficiently recent implementation of musl that it supports this flag. We might want to consider adding a rebuild test to the wake repository to better exercise the wake implementation. Plus that's good practice anyhow. I think we might have hit this edge case had we had that (not sure)

ag-eitilt:
> I think we might have hit this edge case had we had that (not sure)

No, you were right originally and I searched for the wrong thing: we do use `AT_SYMLINK_NOFOLLOW` elsewhere, but none of the existing uses were in `faccessat`.  Presumably musl had a complete implementation of the flag in the other functions.

JakeSiFive:
oh perhaps 3.11 fails the symlink test already even with the current PR?

ag-eitilt:
Reverting the change in `fuse-waked.cpp` according to the discussion that this *seems to* expect `access`-like behaviour in following symlinks, and with the assumption that we can add `AT_SYMLINK_NOFOLLOW` to the invocation later if we determine that it should be the other way around.

> oh perhaps 3.11 fails the symlink test already even with the current PR?

I'd expect that to have been the case, but 3.11 and 3.13 both pass the test.  I'll give a quick look at the logic flow before merging just in case I see an obvious test case to add which would trigger a failure (for the record: to indicate a lack of libc support), but otherwise I'll make note of the minimum supported versions and merge this without a regression test to be able to release the bugfix.  On that: the rebuilding use case we're targeting is successfully retrieved without being rebuilt.

PR #1023, opened 2022-09-29
Add fmt_assert macro for detailed debugging

V-FEXrt:
It's useful to know exactly where and which CSTElement fails an assert. This add a macro that adds that debug info as part of the assert. If it proves to be too expensive in real usage then it can be `ifdef`d away for release builds

### DIFF
--- a/tools/wake-format/actions.h
+++ b/tools/wake-format/actions.h
@@ -80,12 +80,9 @@ struct TokenAction {
   TokenAction(cst_id_t token_id) : token_id(token_id) {}
   ALWAYS_INLINE void run(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node,
                          const token_traits_map_t& traits) {
-    if (node.id() != token_id) {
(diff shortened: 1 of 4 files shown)


JakeSiFive (approved):
Something else we could do perhaps is have hidden global stack of structs that pushes the current context so that when the FMT_ASSERT fails it can give you the latest context. That would let you do something like add a macro that tells you what the current context is and then this error can print out something really deeply useful (it can even look at the stack of walks and the entire tree if need be)

V-FEXrt:
Interesting idea. Where would the upkeep go? We could maybe to the push in FMT_ASSERT but not sure where the pop would happen. 

I'm going to go ahead a land this and I'll do a separate PR if we do the context stack.

aside: if there is a way to check for an "in scope context" then we could also just use that for a majority of the FMT_ASSERT calls as well. It wouldn't be super common to use it in a case where context isn't available

PR #1024, opened 2022-09-30
Make fmt_fits generic via try_else

V-FEXrt:
In the formatting API there is this use case to say "use this formatter if it fits" otherwise use this other formatter. Since fitting can only be checked after actually running the formatter we previously ran the formatter on a copy, checked if it fit, then threw out the copy. 

This idea of asking a question about an already formatted bit of code before continuing to format actually comes up in other formatting cases and is a precursor to any kind of "full choice" formatting. 

This change makes the idea of "try to format something and make a decision on how to proceed based on the format" generic so that it can be used in any formatter. This genercism also simplifies our previous implementations of `fits_first` and `fits_all`

A common case is to format condidtionally on if something previously emitted a newline. Ex:

```
  return fmt().fmt_try_else(
      [](const wcl::doc_builder& builder, ctx_t ctx, wcl::doc doc) { return doc->has_newline(); },
      fmt().freshline(), fmt().space());
```

This gives the formatting API two types of conditional questions. `fmt_if_else(pred, if_fmt, else_fmt)` which calls `pred` _before_ calling `if_fmt` and `fmt_try_else(pred, try_fmt, else_fmt)` which calls `pred` _after_ calling `try_fmt` (on a fake copy)

### DIFF
--- a/tools/wake-format/actions.h
+++ b/tools/wake-format/actions.h
@@ -326,16 +326,16 @@ struct MatchAction {
 
   // Predicate case that is accepted if FMT passes the FitsFirstPredicate
   template <class FMT>
-  MatchAction<MatchSeq<Case, PredicateCase<FitsFirstPredicate<FMT>, FMT>>> pred_fits_first(
-      FMT formatter) {
-    return {{c, {FitsFirstPredicate<FMT>(formatter), formatter}}};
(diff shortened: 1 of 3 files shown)


JakeSiFive (approved):
This is for sure a good change

PR #1026, opened 2022-10-03
Format multi-match subtree

V-FEXrt:
Formats a subtree that has multiple variables in the match pattern

Ex
```
def f0 = match _ _
  (x: Integer) (True:  Boolean) = x+1
  (x: Integer) (False: Boolean) = x+0
```

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -213,6 +213,36 @@ auto Emitter::rhs_fmt() {
   // clang-format on
 }
 
+auto Emitter::pattern_fmt(cst_id_t stop_at) {
+  auto first = fmt().walk(is_expression, WALK_NODE).consume_wsnlc();
+  auto rest_flat = fmt().space().walk(is_expression, WALK_NODE).consume_wsnlc();
+  auto all_flat =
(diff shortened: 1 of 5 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -213,6 +213,36 @@ auto Emitter::rhs_fmt() {
...
+  //
+  // #1 is the ideal and most common use, so flat is returned
+  // #2 is a somewhat rare case where the pattern is very wide without the influence of comments.
+  //    return the explode representation
+  // #3/#4 when flat has a newline, it must have been grown from a comment. At that point there
+  //    isn't a good way to determine if it fits since the NL can be in many places. Instead
+  //    just return flat, accepting in some very rare cases the output will exceed the max width
+
+  return fmt().fmt_try_else(
+      [](const wcl::doc_builder& builder, ctx_t ctx, wcl::doc doc) { return doc->has_newline(); },
```
V-FEXrt, line 242: Not sure if we should make a dedicated "HasNewline" TryPredicate, but I wanted to show off how the feature is used with a custom predicate anyways

JakeSiFive (approved):
Yeah this looks good to me. We might want to come back later for some cases not being thought of and tweak this but I think this looks nice. I see no downside

PR #1027, opened 2022-10-03
Support patterns in targets

V-FEXrt:
Support formatting `target` subtrees with patterns on the LHS

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -61,8 +61,7 @@ static inline bool requires_fits_all(cst_id_t type) {
 static inline bool is_expression(cst_id_t type) {
   return type == CST_ID || type == CST_APP || type == CST_LITERAL || type == CST_HOLE ||
          type == CST_BINARY || type == CST_PAREN || type == CST_ASCRIBE || type == CST_SUBSCRIBE ||
-         type == CST_LAMBDA || type == CST_UNARY || type == CST_BLOCK || type == CST_IF ||
-         type == CST_TARGET_ARGS;
+         type == CST_LAMBDA || type == CST_UNARY || type == CST_BLOCK || type == CST_IF;
 }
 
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -61,7 +61,8 @@ static inline bool requires_fits_all(cst_id_t type) {
 static inline bool is_expression(cst_id_t type) {
   return type == CST_ID || type == CST_APP || type == CST_LITERAL || type == CST_HOLE ||
          type == CST_BINARY || type == CST_PAREN || type == CST_ASCRIBE || type == CST_SUBSCRIBE ||
-         type == CST_LAMBDA || type == CST_UNARY || type == CST_BLOCK || type == CST_IF;
+         type == CST_LAMBDA || type == CST_UNARY || type == CST_BLOCK || type == CST_IF ||
+         type == CST_TARGET_ARGS;
```
JakeSiFive, line 65: Target args is *not* an expression. It needs a seperate handler. I think you can handle it with just `walk_all` or something.

tools/wake-format/emitter.cpp
```diff
@@ -214,7 +215,15 @@ auto Emitter::rhs_fmt() {
...
 
 auto Emitter::pattern_fmt(cst_id_t stop_at) {
-  auto part_fmt = fmt().walk(is_expression, WALK_NODE).consume_wsnlc();
+  // clang-format off
+  auto part_fmt = fmt().match(
+   pred([](wcl::doc_builder&, ctx_t, CSTElement& node, const token_traits_map_t&){
+     return node.isNode();
+    }, fmt().walk(is_expression, WALK_NODE))
+   .otherwise(fmt().walk(WALK_TOKEN)))
+   .consume_wsnlc();
```
JakeSiFive, line 224: The pattern format part should just be a single node so you don't have to do this. By splitting it up into the slash and non-slash part you can eliminate this entire thing.

JakeSiFive (approved):
Awesome!

PR #1030, opened 2022-10-05
Remove extra whitespace after a data node

V-FEXrt:
CST_DATA was emitting `<nl><space><space><space><space>` at the end of the node incorrectly. This PR fixes that

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -951,7 +951,8 @@ wcl::doc Emitter::walk_data(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
   FMT_ASSERT(node.id() == CST_DATA, node, "Expected CST_DATA");
 
-  auto fmt_members = fmt().walk_all(fmt().walk(WALK_NODE).freshline().consume_wsnlc());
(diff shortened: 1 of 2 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -951,7 +951,8 @@ wcl::doc Emitter::walk_data(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
   FMT_ASSERT(node.id() == CST_DATA, node, "Expected CST_DATA");
 
-  auto fmt_members = fmt().walk_all(fmt().walk(WALK_NODE).freshline().consume_wsnlc());
+  auto fmt_members = fmt().walk(WALK_NODE).consume_wsnlc().walk_all(
```
JakeSiFive, line 954: Ok so as I understand it, what you're doing here is assuming that there will always be at least one node (seems correct), and emitting one less freshline by emitting a leading freshline on everything else instead of a trailing newline. Got it.

JakeSiFive (approved):
Yeah nothing bad here

tools/wake-format/emitter.cpp
V-FEXrt: Yep, exactly. That is generally the policy (there are other cases of this) when there are extra "separators" at the end of a sequence of things.

PR #1031, opened 2022-10-05
Fix linkO args in tutorial.md

sifive-benjamin-morse:
linkO's signature changed in df1b122
Also fixed a missing "
Also fixed the code snip that covers curl

### DIFF
--- a/share/doc/wake/tutorial.md
+++ b/share/doc/wake/tutorial.md
@@ -695,7 +695,7 @@ export def buildSimple _ =
         source "main.cpp"
     require Pass main =
         compileC variant ("-I.", Nil) Nil mainSrc
-    linkO variant Nil main "simple"
+    linkO variant Nil main "simple" Nil
 ```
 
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):

share/doc/wake/tutorial.md
```diff
@@ -1055,7 +1055,7 @@ In `buildHeaders`, we've used the `sources` command to find all the header
 files in the same directory and pass them as legal inputs to gcc -- the
 keyword `here` expands to the directory of the `.wake` file.  The second
 argument to `sources` is a regular expression to select which files to
-return. We've used ``` `` ```s here which define regular expression literals
+return. We've used ``` `` ```s (backticks) here which define regular expression literals
```
ag-eitilt, line 1058: This would be the very simple way to clarify the markdown mess -- it looked good in the rendered view, at least...

ag-eitilt (approved):
Good catches, thanks!

PR #1033, opened 2022-10-06
Add match and interpolate as expressions

V-FEXrt:
Mark INTERPOLATE and MATCH as expressions. This brings the formatter to the point that _all_ wake files in the wake repo format without an abort. 

Also adds a regression test to ensure that it continues to not abort on all wake files.


JakeSiFive (approved):

V-FEXrt:
Okay, so it seems like the non-tarball builders don't actually have access to the source code. For now, I'm going to remove the `regression` test and land the final fix so we can keep progressing forward. Once we are running the formatter for every PR that will serve the same purpose as the regression test here and I don't think its worth spending more time on to fix/change the infra

PR #1040, opened 2022-10-11
Capture newlines internal to comments

V-FEXrt:
Captures the newlines internal to a comment block

Ex:
```
<nl 1>
# comment1 <nl 2>
<nl 3>
<nl 4>
 # comment2 <nl 5>
<nl 6>
def a = 5
```
Captures comment1, nl3, nl4, comment2, nl6, and def.


tests/wake-format/basic/stdout
```diff
@@ -701,9 +702,13 @@ def loop aaaaa sssss eeeee =
 # A dictionary associating a key-value pair, providing fast lookup by key.
 # This is internally implemented by a balanced tree, so some total ordering must
 # be able to be produced for the key type.
+
```
ag-eitilt, line 705: There's definitely arguments both ways, but if we've got some blank-line count we're enforcing between top-level `def`s (currently two, but I don't know if that's deliberate or just an artifact from something at the end of the blocks which is meant to eventually be fixed) I'd expect that same number of blank lines to be applied between "floating" top-level comments and ones more obviously attached to their token.  Then again, I'm not sure how you'd handle things like the below if you do go that way, especially with any consistency within blocks.  Maybe just enforce `^\s*#\s*$` as being the only way to connect distinct comment paragraphs, and otherwise they're treated as separate with whatever spacing is currently active (but minimum one)?
```wake
# A dictionary associating a key-value pair, providing fast lookup by key.
# This is internally implemented by a balanced tree, so some total ordering must
# be able to be produced for the key type.

# Second paragraph of the same documentation comment.
global export tuple Map k v =
```

ag-eitilt (commented):
I'm not quite as sure about this one -- I see blank lines as conceptually divorcing a comment from anything following, rather than as being paragraph separators.  Also, it feels like someone could accidentally mess up the formatting output by putting a comment at the end of some blank-delimited section (it's bad for other reasons, but the point stands):
```wake
def foo = unit

# -------------
# End constants
# -------------








# ---------------
# Begin functions
# ---------------

# Documentation for function.
#
# Second paragraph of documentation.
def firstFunction x = x
```

ag-eitilt:
Meant to say, I know that my suggestion is incompatible with how the comments are currently stored internally.  If you don't want to overhaul the internals, the PR as-is is certainly not bad.  It just doesn't match my mental model of what should be happening.

tests/wake-format/basic/stdout
V-FEXrt: > but if we've got some blank-line count we're enforcing between top-level defs 

ATM it's intentionally forcing 2 based on the style guide doc, but its easily changeable

> I'd expect that same number of blank lines to be applied between "floating" top-level comments and ones more obviously attached to their token.

Unfortunately there isn't a method to know if a comment is floating vs bound. We could make some rule like "any comment surrounded by newlines" is floating and then we could force whatever rules about spacing. That has the tradeoff that authors can no longer space comments with newlines and must instead space them with empty comments

```
# one
# 
# two
# three
def x = y
```
vs 
```
# one
 
# two
# three
def x = y
```

I'm not opposed to this, but we will have to make a decision now. We can't punt/bikeshed because its a foundational/API level decision.

I'm mildly partial to the thing that doesn't require a rewrite but this is a now or never kind of thing.

V-FEXrt: Though the more I think about it, the more I think floating comments are the right solution.

Let me sleep on it and see how I feel tomorrow morning

ag-eitilt: It's also probably something worth asking at the Wake Dev meeting, which aligns nicely with sleeping on it.

V-FEXrt: hmm. I thought of a tricky case here. Top level floating comments are fine but weird nested ones aren't so nice.

```
def thing =  (
    a
    b
    c

    # comment

    d
    e
    f
)
```

is very problematic. That `# comment` really needs to be a bound comment otherwise we are back the the core problem of checking for comments on every line of the formatter. So maybe we could change the rule to `any *top level* comment surrounded by newlines is floating` but that is a much harder thing to tag

ag-eitilt: Ah, yeah, that's not a friendly case.  We've already got an output difference between top-level and internal tokens in `def` spacing so *conceptually* I don't think treating top comments differently would be too jarring, but that definitely doesn't help anything on the implementation side.

V-FEXrt:
This is ready for another look @ag-eitilt when you have time

ag-eitilt (approved):
I'm happy with the top-level comments and mostly happy with the internal ones.  There still might be some potential improvements we can do for the latter, but I don't expect them to be as foundational a change, so I think we're at a spot where we can merge this and come back to it later if we want to.

tests/wake-format/basic/stdout
```diff
@@ -708,9 +741,14 @@ def loop aaaaa sssss eeeee =
...
+
 global export tuple Map k v =
     # The ordering function in use, over the key type only.  The `Tree` also
+    
+    
     # stores a version of this over the full `Pair` type, but some of the data
+
```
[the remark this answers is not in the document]
V-FEXrt: > how folding a floating-comment check into consume_wsnlc and similar won't be sufficient

So we could have `consume_wsnlc()` _stop_ at a floating comment but then any time we have a `consume_wsnlc()` we'd have to add a condition for "are we at a floating comment or the thing we expected.

Alternatively it might be possible to have `consume_wsnlc()` re-emit the floating comment but I'm not sure that it would have enough information to do so.

Mainly though I think my stance is that non-top-level floating comment feel fairly uncommon and less likely to be messed up so it doesn't feel worth the code/API complications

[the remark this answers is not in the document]
ag-eitilt: Makes sense, thanks!

tools/wake-format/emitter.cpp
```diff
@@ -65,6 +65,21 @@ static inline bool is_expression(cst_id_t type) {
          type == CST_INTERPOLATE || type == CST_MATCH;
 }
 
+// a floating comment is a comment bound to another comment
```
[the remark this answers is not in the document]
V-FEXrt: Ah, yeah. I don't think it's possible to remove the comment from the tree without a significant change to the underlying data structure (CSTElement). It would certainly be a nicer design to just remove the bound comments if we could. Though it would only be an improvement assuming floating comments are top level only. As long as comments can appear anywhere in the tree we will have to handle them specifically. 

That said, binding a "floating" comment to itself is fundamentally a hack but its a shorter term hack that can be unwound fairly easily if a better solution manifests itself in the future

tools/wake-format/emitter.cpp
```diff
@@ -466,7 +493,99 @@ void inorder_collect_tokens(CSTElement node, std::vector<CSTElement>& items) {
...
+// Input: The top level CST node
+// Output: token_traits[t].bound_to == t' forall t where t is a floating block comment and t' is the
+// *first* comment token in that floating block.
+//
+// Ex:
+//  token_traits[1a].bount_to == 1a
+//  token_traits[2a].bount_to == 2a
+//  token_traits[2b].bount_to == 2a
+//  token_traits[2c].bount_to == 2a
+void Emitter::bind_top_level_comments(CSTElement node) {
```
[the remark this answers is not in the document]
V-FEXrt: Yeah, fair enough. Its on the more complicated side. I'll make sure Jake takes a look at it when he gets an opportunity

PR #1042, opened 2022-10-12
Update `minsertWith` examples and label arguments in type signature

bmitc:
It looks like `minsertWith` was updated at some point to include the key in the update `fn` function, but the examples were out of date. So I updated the examples. I went with `\_key \old \new old + new` as I couldn't figure out how to make it more concise with Wake's syntax. I did try `(\_key _+_)` but this didn't work as expected. For example:

```bash
$ wakex '(\_key _+_)'
(\_key _+_): Integer => Integer => (_key: a) => Integer
```

But as you can see, this is the wrong type signature. I do find this a bit non-intuitive, as I might have expected (or maybe I should say hoped for to make the examples cleaner) `(_key: a) => Integer => Integer => Integer`.

Also, when using this function, it wasn't clear to me which value in the update function `(fn: k => v => v => v)` was the old or existing value and what the new incoming value to compute the result were. So I labeled the arguments to (hopefully) make this more clear with `(fn: k => (old: v) => (new: v) => v)`. Maybe this is even more clear, but potentially overkill, with `(fn: (key: k) => (old: v) => (new: v) => (result: v))`? I couldn't think of a better word for `new` other than maybe `incoming`.

Thoughts are welcome of course!

### DIFF
--- a/share/wake/lib/core/map.wake
+++ b/share/wake/lib/core/map.wake
@@ -138,10 +138,10 @@ export def minsertReplace (key: k) (value: v) (map: Map k v): Map k v =
 #
 # *Examples:*
 #   ```
-#   mnew scmp | minsertWith (_+_) "a" 2 | mlookup "a" = Some 2
-#   listToMap scmp ("a" → 1, Nil) | minsertWith (_+_) "a" 2 | mlookup "a" = Some 3
+#   mnew scmp | minsertWith (\_key \old \new old + new) "a" 2 | mlookup "a" = Some 2
+#   listToMap scmp ("a" → 1, Nil) | minsertWith (\_key \old \new old + new) "a" 2 | mlookup "a" = Some 3
 #   ```
-export def minsertWith (fn: k => v => v => v) (key: k) (value: v) (map: Map k v): Map k v =
+export def minsertWith (fn: k => (old: v) => (new: v) => v) (key: k) (value: v) (map: Map k v): Map k v =
     def pairFn (Pair k l) (Pair _ r) = Pair k (fn k l r)
     editMapData (tinsertWith pairFn (Pair key value)) map
 
--- a/share/wake/lib/core/map.wake
+++ b/share/wake/lib/core/map.wake
@@ -138,10 +138,10 @@ export def minsertReplace (key: k) (value: v) (map: Map k v): Map k v =
 #
 # *Examples:*
 #   ```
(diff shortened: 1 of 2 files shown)


ag-eitilt (changes requested):
Testing it out, it looks like this and `tinsertWith` both have `new` on the *left* side of the function.  I don't know why I did it like that (maybe as an analogy to pushing new values onto a list?  `tinsertWith` does use `foldl`), but since that's not how either of us expected the order to go, it's definitely something which should be documented.

bmitc:
Well, shoot. I'm not entirely sure why I missed it despite looking at it, but I didn't look into the code. I just tested it with some values, but I must have chosen poorly and picked a symmetric operation without thinking, or something. Whoops. Thanks for catching that! Only thing worse than no documentation is incorrect documentation. 🙃 

I'm still not crazy about my original choice of `new` because the *return* of the function is really the "new" value. What I called `new` is really `incoming`, but I feel that may be slightly too long for an argument label. I am tempted though. Also, `existing` seems more specific than `old`, but also maybe too long.

Here are some options I came up with:

1. ```
   tinsertWith (fn: (new: a) => (old: a) => a) (y: a) (tree: Tree a): Tree a
   minsertWith (fn: k => (new: v) => (old: v) => v) (key: k) (value: v) (map: Map k v): Map k v
   ```

3. ```
   tinsertWith (fn: (incoming: a) => (existing: a) => a) (y: a) (tree: Tree a): Tree a
   minsertWith (fn: k => (incoming: v) => (existing: v) => v) (key: k) (value: v) (map: Map k v): Map k v
   ```

3. ```
   tinsertWith (fn: (y: a) => (old: a) => a) (y: a) (tree: Tree a): Tree a
   minsertWith (fn: k => (value: v) => (old: v) => v) (key: k) (value: v) (map: Map k v): Map k v
   ```

What are your thoughts?

ag-eitilt:
I like the clear parallelism of 3, but I worry it might make the type signature a bit hard to read to have the `y`/`value` repeated like that.  My vote's probably 2, but not strongly.

ag-eitilt (approved):
Much more readable, thanks for the fix!

bmitc:
I liked (2) as well, despite the length, and agree on your thoughts for (3). And I also fixed the examples with the current behavior of `_` and my better understanding of it.

PR #1044, opened 2022-10-12
parser: treat \x ... as a block

terpstra:
Previously, wake treated lambda as a prefix operator. Since we switched to lemon, that hasn't been true anymore. Given that is the case, we might as well have `\x` introduce a block.

That makes these all the same:
```
(\x _ +. _): (x: a) => Double => Double => Double
\x (_ +. _): (x: a) => Double => Double => Double
(\(x: Integer) _ +. _): Integer => Double => Double => Double
(\(x: Integer) (_ +. _)): Integer => Double => Double => Double
```

### DIFF
--- a/src/dst/todst.cpp
+++ b/src/dst/todst.cpp
@@ -1398,7 +1398,7 @@ static Expr *dst_expr(CSTElement expr) {
       CSTElement child = expr.firstChildNode();
       AST ast = dst_pattern(child, nullptr);
       child.nextSiblingNode();
-      Expr *body = dst_expr(child);
+      Expr *body = relabel_anon(dst_expr(child));
       Lambda *out;
       FileFragment l = expr.fragment();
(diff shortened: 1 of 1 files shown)


terpstra:
I'll note that none of our large wake codebase appears impacted by this change.
It also means we don't have to fix this bug:
```
$ wake -x '(\(x: Integer) _ +. _)'
reference to undefined variable '_'
```

ag-eitilt (approved):
Technically a breaking change, but like you said, there's nothing which should be impacted by this.  (Maybe it's just saying we should release 0.29.0 next rather than 0.28.6.)  I do like the consistency; I've had to work around the previous behaviour in some of my code.

bmitc:
Added others more familiar with and responsible for the Wake processes and internals as reviewers.

I'll give my thumbs up for this feature.

V-FEXrt (approved):

PR #1045, opened 2022-10-13
emit flat paren node when NL from a comment

V-FEXrt:
Comments were forcing paren nodes to be multi-lined. This was changing the semantics of the `data` subtree and was also just wrong formatting. This PR fixes it so they are only multi-lined when appropriate.

### DIFF
--- a/tests/wake-format/basic/basic.wake
+++ b/tests/wake-format/basic/basic.wake
@@ -599,3 +599,15 @@ export tuple a +++ b =
 
 export tuple (z: Banger) =
   export X: Integer
+
+export data List a =
+    # The empty list. Nil represents a list with no elements.
+    Nil
+    # The prepend constructor. This is often called "cons".
+    #
+    # Examples:
+    # ```
+    #   10, list       # Add a number to `list`
+    #   1, 2, 3, Nil   # Create a list of 3 elements
+    # ```
+    (head: a), (tail: List a) # comment
--- a/tests/wake-format/basic/basic.wake
+++ b/tests/wake-format/basic/basic.wake
@@ -610,4 +610,15 @@ export data List a =
     #   10, list       # Add a number to `list`
     #   1, 2, 3, Nil   # Create a list of 3 elements
     # ```
+    (head: a), (tail: List a)
+
+export data List a =
+    # comment
+    (head: a), (tail: List a)
+
+export data List a =
+    (head: a), (tail: List a) # comment
+
+export data List a =
+    # comment
     (head: a), (tail: List a) # comment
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -1251,7 +1257,7 @@ wcl::doc Emitter::walk_paren(ctx_t ctx, CSTElement node) {
                    .token(TOKEN_P_PCLOSE)
                    .format(ctx, node.firstChildElement(), token_traits);
 
-  if (!no_nl->has_newline()) {
+  if (no_nl->newline_count() == count_allowed_newlines(token_traits, node)) {
```
ag-eitilt, line 1260: Is there ever going to be a case where `no_nl` will have *dropped* some of the allowed newlines?

tests/wake-format/basic/basic.wake
```diff
@@ -599,3 +599,15 @@ export tuple a +++ b =
...
+    # The empty list. Nil represents a list with no elements.
+    Nil
+    # The prepend constructor. This is often called "cons".
+    #
+    # Examples:
+    # ```
+    #   10, list       # Add a number to `list`
+    #   1, 2, 3, Nil   # Create a list of 3 elements
+    # ```
+    (head: a), (tail: List a) # comment
```
ag-eitilt, line 613: We don't have any test for a data constructor with a trailing comment but *not* a leading one, but I don't expect that to affect anything.

tests/wake-format/basic/stdout
```diff
@@ -717,3 +706,16 @@ export tuple a +++ b =
...
+    # The prepend constructor. This is often called "cons".
+    #
+    # Examples:
+    # ```
+    #   10, list       # Add a number to `list`
+    #   1, 2, 3, Nil   # Create a list of 3 elements
+    # ```
+    (head: a), (tail: List a) # comment
+
+
```
ag-eitilt, line 721: I'm sure you've got it on a list somewhere, but we'll probably want to figure out avoiding additional lines at the end of files.  Doesn't look like it's (purely) related to this PR, though, so that's probably something for future work.

ag-eitilt (approved):

tools/wake-format/emitter.cpp
V-FEXrt: nope! The "allowed newlines" are essentially the newlines emitted from comments immediately before and after the subtree being emitted. Those comments are automatically emitted via the API and can't be skipped.

tests/wake-format/basic/stdout
V-FEXrt: yeah, its something I've been thinking about. I don't have a solution yet but it'll be taken care of

tests/wake-format/basic/basic.wake
V-FEXrt: added more cases!

PR #1046, opened 2022-10-14
re-emit author provided newlines inside of BLOCKs

V-FEXrt:
non-comment "top level" newlines inside of a CST_BLOCK or CST_REQUIRE should be re-emitted as they represent spacing requested by the author. This implements that change.

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -65,6 +65,18 @@ static inline bool is_expression(cst_id_t type) {
          type == CST_INTERPOLATE || type == CST_MATCH;
 }
 
+static inline bool contains_req_else(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node,
+                                     const token_traits_map_t& traits) {
+  CSTElement copy = node;
+  while (!copy.empty()) {
+    if (copy.id() == CST_REQ_ELSE) {
+      return true;
+    }
+    copy.nextSiblingElement();
+  }
+  return false;
+}
+
 static bool compare_doc_height(const wcl::doc& lhs, const wcl::doc& rhs) {
   return lhs->height() < rhs->height();
 }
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -918,9 +930,12 @@ wcl::doc Emitter::walk_block(ctx_t ctx, CSTElement node) {
...
   auto body_fmt = fmt().match(
-    // TODO: starting 'pred()' function doesn't allow init lists
-    pred(ConstPredicate(false), fmt())
-   .pred({TOKEN_WS, TOKEN_NL, TOKEN_COMMENT}, fmt().next())
+    // Block level newlines should be re-emitted as these are the
+    // user creating 'pseudo-blocks'
+    pred(TOKEN_NL, fmt().newline().newline().next())
+   .pred(TOKEN_WS, fmt().next())
+   // Comments are following by a block level newline that shouldn't be emitted
+   .pred(TOKEN_COMMENT, fmt().next().next())
```
ag-eitilt, line 938: Is this meant to be "followed by"?  Either way, it looks like none of the in-block comments being tested have an author-provided newline next to them (neither before nor after), so that might be something we want to add to the test input.

ag-eitilt (commented):
Unless I'm missing where the highlighted change is being tested, I'd like to see a case or two added to `basic.wake` before approving this.

V-FEXrt:
> Unless I'm missing where the highlighted change is being tested, I'd like to see a case or two added to basic.wake before approving this.

Its a bit hidden

```

def buildRE2WASM Unit =
    require Pass outputs =
        job.getJobOutputs

    Pass (SysLib "1.0" headers objects cflags lflags)
```
would be 
```

def buildRE2WASM Unit =
    require Pass outputs =
        job.getJobOutputs
    Pass (SysLib "1.0" headers objects cflags lflags)
```
if the test failed. Also all the extra newlines in `stdout` are newlines that were previously getting deleted when they shouldn't be. 

The easiest way I've found so far is to diff `basic.wake` and `stdout` and see if the formatting rules makes sense. Personally I use VSCodes "compare file with" command but `diff` or similar should work. (`wake-format -n` also gives a nice diff)

tools/wake-format/emitter.cpp
V-FEXrt: >Is this meant to be "followed by"?

yeah, was a typo, nice catch.

>  Either way, it looks like none of the in-block comments being tested have an author-provided newline next to them 

That isn't the case being covered here. Every comment must be followed by a single newline to end the comment

```
def a = b # comment <nl>
def c = d # comment <nl>
```
since we always emit a newline when emitting a comment, we need to skip it here so that it isn't duplicated. In the case of a author added newline we'd have

```
def a = b # comment <nl>
<nl>
def c = d # comment <nl>
```

which would get covered by the line#935 predicate

> so that might be something we want to add to the test input.

yeah, that's a good idea. I'll add it

ag-eitilt: Ah, yeah, that makes sense.  I think I was getting thrown off by the "block level newline" terminology.  And with the new test (and associated fix it seems to have revealed), this all looks good!

ag-eitilt (approved):

PR #1047, opened 2022-10-17
Re-emit top level newlines to maintain author separation

V-FEXrt:
Re-emits author provided newlines for everything except defs which should be followed by 2 newlines. When a comment is encountered it and all wsnl is skipped as that is managed by the comment subsystem.

### DIFF
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -267,6 +267,7 @@ from wake import myMap=map foldl myFoldr=foldr
 from wake import def unary + -
 from wake import type Pair Result
 from wake import source,Nil
+
 # comment
 def linkKind =
     # comment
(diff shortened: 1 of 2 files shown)


ag-eitilt (approved):
This clearly successfully removes extra spaces which had been added between `publish` lines, but I don't see anything in the test input which checks whether it *retains* newlines between lists which had been deliberately separated rather than just at the edges.  (It does, but that might be a nice thing to be testing).
```patch
 from test_wake import topic wakeTestBinary
+
 from test_wake import topic wakeUnitTestBinary
 publish wakeTestBinary = defaultWake, Nil
+
 publish wakeUnitTestBinary = buildWakeUnit, Nil
```

ag-eitilt:
The test cases in added in #1048 also fill the gap I mentioned above.

V-FEXrt:
Rebased on #1048 so that test got picked up

PR #1048, opened 2022-10-18
correctly format when comments follow the rhs =

V-FEXrt:
```
publish compileC = # comment
  emccFn (makeCompileC "wasm-c11-release"   _ ("-std=gnu11", emscriptenCFlags))
```

should format as 

```
publish compileC = # comment
    emccFn (makeCompileC "wasm-c11-release" _ ("-std=gnu11", emscriptenCFlags))
```

not 

```
publish compileC = # comment
 emccFn (makeCompileC "wasm-c11-release" _ ("-std=gnu11", emscriptenCFlags))
```

This was happening because the `# comment` is invisible to `rhs_fmt()` since it is bound to `=` which is emitted before `rhs_fmt()` and so `flat_fmt` is used when `full_fmt` should be actually used. The PR adds a check for that case then forces `full_fmt`


ag-eitilt (approved):

tests/wake-format/basic/stdout
```diff
@@ -717,3 +717,26 @@ export tuple a +++ b =
 export tuple (z: Banger) =
     export X: Integer
 
+publish compileC = # comment
```
V-FEXrt, line 720: note to self: this publish is duplicated and can be deleted

PR #1049, opened 2022-10-19
job: include lookup id in job error messages

ag-eitilt:
Debugging what went wrong with a Job sometimes/often requires looking it up in the database, and if another `wake` invocation has been run since the failed job, or if we're looking into a reported issue rather than one we've triggered ourselves, `wake --last`/`wake --failed` can sometimes be less than helpful.  Printing the job ID directly in the error message doesn't introduce much extra noise but makes any `wake --job` calls to zero in on the info much simpler to construct.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -355,7 +355,7 @@ def guardPath job = match _
     Pass l if job.isJobOk =
         findFailFn treeOk l
     _ =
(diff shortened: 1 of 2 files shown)


mmjconolly (approved):
I like this

colinschmidt (approved):
LGTM. Thanks

V-FEXrt (approved):
LGTM

PR #1050, opened 2022-10-21
Remove 0.24 and 0.27 stdlib versions

mmjconolly:
Removes stdlib adapters in wake packages `v0_24_wake` and `v0_27_wake`


V-FEXrt:
Looks like we might be in another MacOS brownout?

V-FEXrt (approved):

mmjconolly:
After discussion, we'll leave these library adapters available for users of the LSP in vscode/neovim so that they don't need to change versions in their editor quite so much.

At some point in the future, supporting these old-version adapters could become a burden and at that point we can delete them and break compatibility.

PR #1053, opened 2022-10-24
Fix memo bug & def formatting

V-FEXrt:
1. Memoizing across several files was causing bad code to be re-emitted (memo clashes), fix this by resetting the memo tables for each file
2. `def` were being emitted multi-line wrapped in parens when they were too long. This parses but fails to type check. Temporarily disable that until a good solution is found

After these changes, I can format the entire wake repo and then use the formatted code to actually do a wake build


tools/wake-format/emitter.cpp
```diff
@@ -52,6 +56,14 @@
     return v;                                      \
   }
 
+#define MEMO_RESET()                         \
+  {                                          \
+    for (memo_map_t * map : __memo_maps__) { \
+      map->clear();                          \
+    }                                        \
+    __memo_maps__.clear();                   \
+  }
```
ag-eitilt, line 65: I'm sure there's a good C++ reason , but this only seems to be used the once and feels like it doesn't need to be a `#define`.

ag-eitilt (approved):
Seems like a good spot and as far as I can tell is good code.

tools/wake-format/emitter.cpp
ag-eitilt: Me either, and if there's others following this pattern, it's probably good to match.

V-FEXrt: SG!

PR #1054, opened 2022-10-25
[Bikeshed] Format the entire repo

V-FEXrt:
Time to start the bikeshedding :)

There are still a few more changed top of mind that need to be fixed but nothing that should significantly change discussion here. This is your opportunity to bikeshed before things get finalized

Things yet to fix:
- Extra Trailing newlines at the end of files
- Unary ops emitted with a space
- Certain inputs in the internal repo still crash the formatter


mmjconolly:
I could probably with live with _anything_ if we roll this out for consistent usage, but I'll throw some opinions out there. I know these are not universal preferences so this is more of discussion-generation reply

* a newline after every `=`, including when a `match` is involved
* lines that contain `|` get a newline before each `|` so the pipelines go down rather than right
* when a `def` contains `def`s, that the final value to return has a newline between it and the last inner-def
* if any match-branch has a new line after the `=`, that they all do

share/wake/lib/core/json.wake
```diff
@@ -13,161 +13,196 @@
...
-    _ =
-      if filterFn node then node, acc else acc
-  helper root Nil | JArray
+    def helper node acc = match node
+        JArray l =
+            def tail = foldr helper acc l
+            if filterFn node then node, tail else tail
+        JObject l =
+            def tail = foldr (helper _.getPairSecond _) acc l
+            if filterFn node then node, tail else tail
```
ag-eitilt, line 159: See what I mean with the single line?  The `then` is hard enough to spot, and the only reason the `else` isn't automatically read as just another object name is because `tail` is repeated on both sides (which is even more insidious in this case because `tail` is indeed normally a function, if it wasn't being shadowed here).

share/wake/lib/core/json.wake
```diff
@@ -13,161 +13,196 @@
...
+        JString _ = tail
+        JInteger _ = tail
+        JDouble _ = tail
+        JBoolean _ = tail
+        JNull = tail
+        JObject l =
+            def flatten v tail = match v.getPairSecond
+                JArray l = l ++ tail
+                w = w, tail
+            foldr flatten tail (filter (matches y _.getPairFirst) l)
```
ag-eitilt, line 183: Matching `match` case line breaks does mean sections like this will have a lot of very short lines, but the parallel structures are probably still a net positive.

share/wake/lib/core/json.wake
```diff
@@ -413,60 +416,54 @@ def mergeValueList (describeCompat: Boolean) (values: List JValue): Result JValu
...
     def requireEqual eqFn wrapFn values =
         def eqOrFail x y =
             if eqFn x y then
                 Pass x
             else
-                compatibilityMessage "N" "n" "ot every type can combine multiple different values. The incompatible values are: {map wrapFn values | formatExamples}"
+                compatibilityMessage
+                "N"
+                "n"
+                "ot every type can combine multiple different values. The incompatible values are: {map wrapFn values | formatExamples}"
```
ag-eitilt, line 436: Note to self: this probably deserves a `# wake-format off`

share/wake/lib/core/json.wake
```diff
@@ -500,39 +490,35 @@ def mergeValueList (describeCompat: Boolean) (values: List JValue): Result JValu
...
-            | JArray
-            | Pass
+            flatten objects | deduplicateJObjectKeys (mergeValueList describeCompat) | rmap JObject
+        _ None None None None None (Some arrays) = flatten arrays | JArray | Pass
         _ _ _ _ _ _ _ =
-            compatibilityMessage "V" "v" "alues of different types may not be combined. The incompatible values are: {formatExamples values}"
+            compatibilityMessage
+            "V"
+            "v"
+            "alues of different types may not be combined. The incompatible values are: {formatExamples values}"
```
ag-eitilt, line 499: Note to self: ...and here.

share/wake/lib/core/list.wake
```diff
@@ -380,11 +395,13 @@ export def append (item: a) (list: List a): List a =
 export def splitAt (index: Integer) (listToDivide: List a): Pair (List a) (List a) =
     if index <= 0 then
         Pair Nil listToDivide
-    else match listToDivide
-        Nil  = Pair Nil Nil
-        h, t =
-            def Pair u v = splitAt (index-1) t
-            Pair (h, u) v
+    else
+        match listToDivide
```
ag-eitilt, line 399: Do we want to give `else match` the same sort of exception as allows `else if`?  (Also, how does `require` play into that exception?)

ag-eitilt (commented):
Only gotten through `order.wake` and need to get ready for a meeting, but here's my thoughts so far.

build.wake
```diff
@@ -150,19 +158,26 @@ export def static _: Result Path Error =
...
 publish wakeUnitTestBinary = buildWakeUnit, Nil
 
 def defaultWake Unit =
-    require Pass wakeVisible =
-        doInstall "tmp" "default"
+    require Pass wakeVisible = doInstall "tmp" "default"
 
     Pass (Pair "tmp/bin/wake" wakeVisible)
+
+
```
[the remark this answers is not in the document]
V-FEXrt: Right, no plan to merge this yet. Its just for bikeshedding

mmjconolly:
I'm not sure if this PR exposes this, does the formatter move `package` decls to the first non-comment line?

share/wake/lib/core/boolean.wake
```diff
@@ -28,11 +30,8 @@ export data Boolean =
 # !True  = False
 # !False = True
 # ```
-export def !(x: Boolean): Boolean =
-    if x then
-        False
-    else
-        True
+export def !(x: Boolean): Boolean = if x then False else True
```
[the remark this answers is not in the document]
bmitc: I also tend to be (or at least I think I am) strict about only using `if __ then __ else __` on a single line for very simple expressions.

As for the formatting between the way it was (before the change here) and Sam's suggested formatting, I'm not sure I have a strong opinion. At least in F#, I bounce between the two. I even sometimes use:

```wake
if ___
then ___
else ___
```

🙈

share/wake/lib/core/integer.wake
```diff
@@ -13,78 +13,80 @@
...
+
 
 # Binary division operator for Integer values.
 # 12 / 3 =  4
 # 13 / 3 =  4
 # -8 / 4 = -2
-export def (x: Integer) / (y: Integer): Integer =
-    (\_\_ prim "div") x y
+export def (x: Integer) / (y: Integer): Integer = (\_\_ prim "div") x y
+
```
bmitc, line 62: Incoming bikeshed. 🚲 🚳 Haha.

I think I would prefer the formatter to not unformat newlines, particularly in cases like this (i.e., function definitions). The Elixir formatter *generally* abides by this, but the F# formatter (called Fantomas) does not, where I believe it is contentious.

For a simple function, it's probably okay, but I prefer the function, arguments, and type annotations line to be just that except for very simple functions or bindings. I probably have broken my own rule several times (although if I was code reviewing myself, I'd probably ask myself to actually follow the rule), but when I don't, I think I'd prefer the formatter not to break it for me.

I'm thinking my rubric is:

* If a top-level function definition, leave function name, arguments, and type annotations on a line by themselves, with the function expression starting on a newline. I.e.,
    ```
    def <functionName> <args> : <returnType> =
        <expression>
    ```
* If the top-level function body expression is *very* simple, allow it on the same line.
* For local function and value definitions, leave it to the discretion of the writer, particularly since these aren't as likely to have full type annotations.

share/wake/lib/core/integer.wake
```diff
@@ -13,78 +13,80 @@
...
 
 # Binary remainder operator for Integer values.
 # 11 % 5 = 1
 #  4 % 5 = 5
 #  7 % 5 = 2
-export def (x: Integer) % (y: Integer): Integer =
-    (\_\_ prim "mod") x y
+export def (x: Integer) % (y: Integer): Integer = (\_\_ prim "mod") x y
```
bmitc, line 68: I think I'd prefer `(\_\_ prim "mod") x y` to be formatted as `(\_ \_ prim "mod") x y`, with a space between the lambdas. I think that's particularly so for actual named arguments, like `\x \y (x + y)` (not sure if the formatter already picks that up and puts a space in between), and to maintain consistency between the cases.

share/wake/lib/core/result.wake
```diff
@@ -95,18 +102,21 @@ export def rmap (fn: a => b): Result a fail => Result b fail = match _
...
 export def rmapPass (fn: a => Result b fail): Result a fail => Result b fail = match _
     Pass a = fn a
     Fail f = Fail f
 
+
 # Applies a fallible function to Fail value or propogates Pass
 # If you find yourself using the function, consider using require instead.
 export def rmapFail (fn: a => Result pass b): Result pass a => Result pass b = match _
     Pass a = Pass a
     Fail f = fn f
```
bmitc, line 117: Although I have been trying to abide by this convention since a lot of Wake code uses it and I know this has probably been discussed a lot before, it would be my personal preference to put `match` on a newline. (I think this is what @mmjconolly comment also alluded to, but I could be mistaken.)

```suggestion
export def rmapFail (fn: a => Result pass b): Result pass a => Result pass b =
    match _
        Pass a = Pass a
        Fail f = fn f
```

If `match` didn't require the additional indentation, even better. :)

Aside from personal preference, I feel like in a lot of these cases, I don't notice the `match` as it just kinda hangs out over there ---->. It also breaks some consistency of my preferred formatting of top-level functions:

```wake
def <functionName> <args> : <returnType> =
    <expression>
```

bmitc (commented):
Thanks for the work on the formatter @V-FEXrt! And thank you for inviting us to throw our hats into the ring!

I've added some comments which, for the most part, cover a lot of the things I've had in mind as I've been writing and reading Wake code. I might come back and do another pass later.

share/wake/lib/core/boolean.wake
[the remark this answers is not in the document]
ag-eitilt: I do like having all three keywords on separate lines as well, unless the `then` and `else` bodies are indented, in which case putting `then` at the end of the `if` line makes sense.

share/wake/lib/core/double.wake
```diff
@@ -208,11 +212,11 @@ export def (x: Double) !=. (y: Double): Boolean =
...
 # dmin 3.0 1.0 = 1.0
-export def dmin (x: Double) (y: Double): Double =
-    match (x <=>. y)
-        None    = nan
-        Some LT = x
-        _       = y
+export def dmin (x: Double) (y: Double): Double = match (x <=>. y)
+    None = nan
+    Some LT = x
+    _ = y
```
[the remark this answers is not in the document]
bmitc: This case is really tough to let go of but also tough to implement. Haha. I also made a comment about alignment, and I caught a case as I submitted my comments as well that I planned to go back to and add some commentary. I'll add some thoughts there, but I think this is a case where I *really* like alignment in some cases but I'm not sure how to create a rubric that a formatter could follow.

One way could be that the formatter ignores extra whitespace if it notices alignment of certain characters or tokens or keywords, like `import` or `=`, on even a single adjacent line. Only the whitespace *before* the special token/keyword is left alone. If it doesn't, then boom, get rid of the extra whitespace and format away. Those are probably the two most common cases of manual alignment.

share/wake/lib/system/job.wake
```diff
@@ -168,245 +193,336 @@ export def getPlanHash (plan: Plan): Integer =
...
+            env.implode
+            cmd.implode
+            hash
+            visStrings.implode
+            (
+                if keep then
+                    1
+                else
+                    0
+            )
```
ag-eitilt, line 300: This is one of the times the `if` should not be broken up.

share/wake/lib/system/job.wake
```diff
@@ -417,131 +533,142 @@ export def getJobStatus (job: Job): Status = match (getJobReport job)
...
+    def pre = match _
+        Fail f = Pair (Fail f) ""
+        _ if ! ok = Pair (Fail (makeError "Runner {script} is not executable")) ""
+        Pass (RunnerInput label command visible environment directory stdin res prefix record) =
+            def Usage status runtime cputime membytes inbytes outbytes = record
+            def pmkdir m p = prim "mkdir"
+            def pwrite m p d = prim "write"
+            def json =
+                JObject
+                (
```
ag-eitilt, line 601: This is probably more generalizable, but I do prefer the opening parentheses be on the same line (the old C `{` debate again).
```suggestion
                JObject (
```

share/wake/lib/system/job.wake
```diff
@@ -417,131 +533,142 @@ export def getJobStatus (job: Job): Status = match (getJobReport job)
...
+                    "environment" :-> environment | map JString | JArray,
+                    "visible" :-> visible | map (_.getPathName.JString) | JArray,
+                    "directory" :-> JString directory,
+                    "stdin" :-> JString stdin,
+                    "resources" :-> res | map JString | JArray,
+                    "version" :-> JString version,
+                    "mount-ops" :-> JArray (JObject ("type" :-> JString "workspace", "destination" :-> JString ".", Nil), Nil),
+                    "usage"
+                    :-> JObject
+                    (
```
ag-eitilt, line 613: Especially if we do go with the single-line format, can the formatter keep these all together?
```suggestion
                    "usage" :-> JObject (
```

share/wake/lib/system/sources.wake
```diff
@@ -90,58 +96,54 @@ export def claim (file: String): Result Path Error =
...
             | runJobWith virtualRunner
             | getJobOutput
 
     # Compute the relative path in simplest form
     def isAbs = matches `/.*` file
-    def relPath = if isAbs then
-        relative workspace file
-    else
-        simplify file
+    def relPath = if isAbs then relative workspace file else simplify file
```
ag-eitilt, line 108: Where are the `then`/`else` in this sentence? 😉

tests/tests.wake
```diff
@@ -82,59 +83,45 @@ export def runTests (cmdline: List String): Result String Error =
...
+    require Pass wakeUnitBin = wakeUnitToTest Unit
 
     def testJob =
-        makeExecPlan ("./{testScript.getPathName.inTestDir}", wakeUnitBin.getPathName.inTestDir, Nil) (wakeUnitBin, visibleFiles)
+        makeExecPlan
+        ("./{testScript.getPathName.inTestDir}", wakeUnitBin.getPathName.inTestDir, Nil)
+        (wakeUnitBin, visibleFiles)
         | setPlanDirectory testDirectory
         | setPlanLabel "testing: {testName}"
         | setPlanStdout logNever
```
ag-eitilt, line 127: I can certainly see how this would be *legal* but it's unclear enough that it's probably a good PR comment to ask people to put the command on a separate line at this point.

tests/type-system/extract-pass/test.wake
```diff
@@ -1,9 +1,21 @@
-def Pair (_: Integer) (_: String)                      = Pair 44 "as"
+def Pair (_: Integer) (_: String) = Pair 44 "as"
+
+
 def Pair (_: Integer) (_: String): Pair Integer String = Pair 44 "as"
+
+
 def Pair _ _:                      Pair Integer String = Pair 44 "as"
```
ag-eitilt, line 7: This should probably not keep quite so many spaces, especially since we've gotten rid of other places which had been aligned.

vendor/emscripten.wake
```diff
@@ -21,40 +22,45 @@ def emccFn fn = match (whichInEnvPath "emcc")
...
-    "-s", "ASYNCIFY",
-    "-s", "MODULARIZE=1",
-    "-s", "EXPORT_ES6=1",
-    "-s", "SINGLE_FILE",
+    "-s",
+    "ALLOW_MEMORY_GROWTH",
+    "-s",
+    "EXIT_RUNTIME",
+    "-s",
+    "ASSERTIONS",
```
ag-eitilt, line 44: For things like this where the two-per-line was very deliberate, will we just have to fall back on a `# wake-format off`?

vendor/git.wake
```diff
@@ -28,15 +29,20 @@ def buildAs Unit = match (subscribe releaseAs)
...
+
+
+# v0.15.0-alpha-4-g6efe8e9-dirty ... for a tag v0.15.0-alpha with +4 commits
+
+
+#  0.15.0~alpha.4+g6efe8e9~dirty ... which sorts correctly for both rpm+deb
+
+
+# ^      ^     ^ ^        ^
+
```
ag-eitilt, line 45: Something went wrong here.

vendor/gmp.wake
```diff
@@ -13,49 +13,50 @@
...
-  require Pass buildDir = vendorBuildDir Unit
+    require Pass buildDir = vendorBuildDir Unit
 
-  def job =
-    """
+    def job =
+        """
       cd .build
       wget https://ftp.gnu.org/gnu/gmp/gmp-%{version}.tar.xz
       tar xmf gmp-%{version}.tar.xz
```
ag-eitilt, line 40: How can we handle this?

share/wake/lib/system/job.wake
bmitc: I agree that this is a case where the `if then else` should all be on one line. But we should probably also have a built-in function for this.

```wake
def booleanToInteger (b: Boolean) : Integer =
    if b then 1 else 0
```

or

```wake
def booleanToInteger (b: Boolean) : Integer = match b
    True = 1
    False = 0
```

The latter function is my preference, because it's easier to read in my opinion with the direct mapping between the boolean and integer values.

V-FEXrt:
> I'm not sure if this PR exposes this, does the formatter move package decl to be the first thing after the comments at the start of a file

It does not at the moment, its a feature we can add but it is something I'd request to punt for awhile

build.wake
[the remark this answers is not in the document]
V-FEXrt: no, its just a quirk/bug right now. Getting rid of it is a bit of work so I've been punting

share/wake/lib/core/boolean.wake
```diff
@@ -13,10 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 package wake
 
 # This Boolean type must exist in package wake for if/then/else syntax to work.
 
+
```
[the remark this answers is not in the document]
V-FEXrt: Can you clarify? Not sure I follow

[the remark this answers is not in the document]
ag-eitilt: There's one space between the `package` and the comment, and two spaces between that comment and the next (outside of the displayed section).  If we're going with two spaces at the top level, we should add an extra newline after `package`/`from` declarations, and if we want to reverse that decision we'll want to remove the second newline after floating comments/code blocks.

share/wake/lib/core/boolean.wake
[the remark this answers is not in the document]
V-FEXrt: So we are somewhat limited in what we can determine about the expression. The easiest is total line width but that a constant for all code. 

In the short term I think we need to pick between "allow non-optimal single line if/then/else" or "never allow single line if/then/else". In the long term, we can try to dial in hyper focused rules but I want to avoid those kinds of rules until we hit MVP.

Re: multi-line if/then/else, I have a pretty strong preference for 
```
if ______ then
    ______ 
else
    ______ 
```
over
```
if _____
then ______ 
else ______ 
```
lol

share/wake/lib/core/double.wake
```diff
@@ -13,65 +13,67 @@
...
 # The Double type is an IEEE 754 64-bit precision number.
 # Any wake literal which include a decimal point or an exponent is parsed into a Double.
 
+
 # Unary absolute value operator for a Double.
 # dabs (+. 2.5) = 2.5
 # dabs (-. 2.5) = 2.5
-export def dabs (x: Double): Double =
-    (\_ prim "dabs") x
+export def dabs (x: Double): Double = (\_ prim "dabs") x
```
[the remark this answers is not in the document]
V-FEXrt: For the first pass (and probably the forseeable future) I'd like to avoid rules that only apply to top-level as much as possible. When emitting the code there currently isn't a great way to know what level you are on (you can guess via nest level but that is a bit hacky). This top-level stuff only worked for comments since they have a separate dedicated subsystem and even that has a small hack in it

share/wake/lib/core/double.wake
```diff
@@ -13,65 +13,67 @@
...
+
 
 # Computes x*y + z with rounding only at the end.
 # The fused-multiply-add operation is useful in numeric algorithms.
 # It is necessary in order compute values with full precision.
 # dfma 2.0 3.0 1.0 = 7.0
 # dfma 1.0 1.0 1.0 = 3.0
-export def dfma (x: Double) (y: Double) (z: Double): Double =
-    (\_\_\_ prim "dfma") x y z
+export def dfma (x: Double) (y: Double) (z: Double): Double = (\_\_\_ prim "dfma") x y z
```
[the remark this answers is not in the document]
V-FEXrt: I haven't yet written any lambda formatting rules, right now it is just spitting out the input unedited. This would be a good place to make your case on what the rule _should_ be :)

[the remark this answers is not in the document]
V-FEXrt: Though I am working on the lambda rule right not to fix the internal repo, so you should post your suggestions soon :)

share/wake/lib/core/double.wake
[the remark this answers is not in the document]
V-FEXrt: Yeah, so for the first pass loosing alignment is fairly necessary.

This is something Jake and I have talked about quite a bit and certainly something we want to support long term but there is a challenging dynamic problem in 1) figuring out how many spaces to insert so every case is aligned and 2) taking and aligned and unaligned version and picking "which one a human would prefer"

My guiding light, is that I don't want the first pass of the formatter to throw away any information that can't be recovered later. I'm pretty confident that we can recalculate the spacing required to put something in alignment so I don't consider throwing away the alignment here to be a "permanent loss"

share/wake/lib/core/list.wake
V-FEXrt: easy enough to do. Just hadn't considered it.

Not sure I see the difference between `def x = match` and `else match` though

build.wake
```diff
@@ -13,100 +13,103 @@
...
-    "static"  = Pass (Pair "native-cpp14-static"  "native-c11-static")
-    "debug"   = Pass (Pair "native-cpp14-debug"   "native-c11-debug")
-    "wasm"    = Pass (Pair "wasm-cpp14-release"   "wasm-c11-release")
+    "static" = Pass (Pair "native-cpp14-static" "native-c11-static")
+    "debug" = Pass (Pair "native-cpp14-debug" "native-c11-debug")
+    "wasm" = Pass (Pair "wasm-cpp14-release" "wasm-c11-release")
     s = Fail "Unknown build target ({s})".makeError
 
+
 export def build: List String => Result String Error = match _
```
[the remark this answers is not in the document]
V-FEXrt: I'm actually in the opposite camp 😬 

There are certainly going to be small preference things like this that folks disagree on. I'm planning on collecting the more controversial rules into a survey to see what the users prefer.

build.wake
```diff
@@ -13,100 +13,103 @@
...
     def releaseBin exe = installAs "{dest}/{replace `\.[^.]*$` '' exe.getPathName}" exe
-    require Pass datinstall =
-        findFailFn (installIn dest ".") datfiles
-    require Pass readmeSource =
-        source "README.md"
-    require Pass readme =
-        installIn "{dest}/share/doc/wake" "." readmeSource
+    require Pass datinstall = findFailFn (installIn dest ".") datfiles
+    require Pass readmeSource = source "README.md"
+    require Pass readme = installIn "{dest}/share/doc/wake" "." readmeSource
```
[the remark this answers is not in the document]
V-FEXrt: They would get erased. In fact, even in the case where we fix alignment for `match`'s `=` this would still not be aligned. These `require` statements aren't associated to each other so it would be very challenging to do collective formatting. Basically anything that could be the common parent would have to check "do I have a list of requires as children" and then use some magic to see if those requires should be "connected"

extensions/vscode/vscode.wake
```diff
@@ -13,100 +13,79 @@
...
+    def dest = "{@here}/{path.getPathName}"
+    require Pass destDir = simplify "{dest}/.." | mkdir
+    def inputs = destDir, path, Nil
     "install -m 444 {path.getPathName} {dest}"
     | makePlan "install: {dest}" inputs
     | runJob
     | getJobOutput
 
+
 def createStdlibJSON (stdlibSources: List Path): Result Path Error =
```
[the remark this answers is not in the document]
V-FEXrt: Ah interesting. I see the difference here. Though I'm not sure if the formatter has enough info to tell these apart. I'll have to double check

build.wake
```diff
@@ -13,100 +13,103 @@
...
 
 # Useful build variants (arguments to all)
 def toVariant = match _
     "default" = Pass (Pair "native-cpp14-release" "native-c11-release")
-    "static"  = Pass (Pair "native-cpp14-static"  "native-c11-static")
-    "debug"   = Pass (Pair "native-cpp14-debug"   "native-c11-debug")
-    "wasm"    = Pass (Pair "wasm-cpp14-release"   "wasm-c11-release")
+    "static" = Pass (Pair "native-cpp14-static" "native-c11-static")
+    "debug" = Pass (Pair "native-cpp14-debug" "native-c11-debug")
+    "wasm" = Pass (Pair "wasm-cpp14-release" "wasm-c11-release")
```
terpstra, line 25: We can't line up the `=` signs in a group of `=`s?

build.wake
```diff
@@ -13,100 +13,103 @@
...
-        source "README.md"
-    require Pass readme =
-        installIn "{dest}/share/doc/wake" "." readmeSource
+    require Pass datinstall = findFailFn (installIn dest ".") datfiles
+    require Pass readmeSource = source "README.md"
+    require Pass readme = installIn "{dest}/share/doc/wake" "." readmeSource
     require Pass binfiles = binfilesResult
-    require Pass bininstall =
-        findFailFn releaseBin binfiles
+    require Pass bininstall = findFailFn releaseBin binfiles
```
terpstra, line 71: So, our style guide previously recommended that every `def/require` body starts on a new line.
I think that was too annoying, but I do think there are two cases we should use for format:

One-line style / collection, line up the `=`:
```
def abc = 345
def d    = 4546
def ef   = 123
```

Multi-line style, don't line up the `=`, but indent every body
```
def abc =
    ...
def d =
    ...
def ef =
    ... one of these is really long, which is why they all got dropped
```

terpstra: As a motivating example to do the work in the formatter: in scala/chisel it ended up really important to readability that "ready" and "valid" are both 5 characters long. When we switched to a third signal, that heavily motivated the choice to use "offer", "abort", and "ready". If you don't do the line up in the formatter, people will pick variable names to make things line up, sometimes impacting understandability by bad name choices.

build.wake
```diff
@@ -13,100 +13,103 @@
...
 
+
 # Replace @VERSION@ with 'release'
 def setVersion release file =
-    require Pass in =
-        source "{file}.in"
-    def script = "%
+    require Pass in = source "{file}.in"
+    def script =
+        "%
```
terpstra, line 80: The content of a multiline string should be indented wrt. the `"%` and `%"`.

build.wake
```diff
@@ -13,100 +13,103 @@
...
         | filter (! matches `(debian/|wake.spec).*` _.getPathName) # omit packaging files
         | filter (! matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
         | filter (! matches `(.*/)*\.gitignore` _.getPathName) # omit git files
     def Pair testPaths wakePaths = splitBy (matches `tests/.*` _.getPathName) srcs
-    def wakeSourcesString =
-        wakePaths
-        | map (_.getPathName.format)
-        | foldr (_, ",\n  ", _) Nil
-        | cat
+    def wakeSourcesString = wakePaths | map (_.getPathName.format) | foldr (_, ",\n  ", _) Nil | cat
```
terpstra, line 105: We might want to consider a special exemption for `|` where we prefer to do it multiline even if it isn't super long. Perhaps just decrease the threshold in this case. (So really simple ones are still one line, but the example above still ends up multiline)

extensions/vscode/vscode.wake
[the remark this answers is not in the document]
terpstra: I prefer no space. That's how both scala (heavily used at SiFive) and SML (despite the claim to the contrary above) do it.

share/wake/lib/core/boolean.wake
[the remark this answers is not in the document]
terpstra: I think a short ternary `if then else` is ok. When it's longer, I agree with @V-FEXrt 's multi-line format.

share/wake/lib/core/json.wake
```diff
@@ -13,161 +13,196 @@
...
-  JNull
-  JObject  (List (Pair String JValue))
-  JArray   (List JValue)
+    JString String
+    JInteger Integer
+    JDouble Double
+    JBoolean Boolean
+    JNull
+    JObject (List (Pair String JValue))
+    JArray (List JValue)
```
terpstra, line 27: Probably want to align the first constructor argument.

share/wake/lib/core/list.wake
```diff
@@ -38,15 +39,15 @@ package wake
...
-#   10, list       # Add a number to `list`
-#   1, 2, 3, Nil   # Create a list of 3 elements
-# ```
+    # The prepend constructor. This is often called "cons".
+    #
+    # Examples:
+    # ```
+    #   10, list       # Add a number to `list`
+    #   1, 2, 3, Nil   # Create a list of 3 elements
+    # ```
```
terpstra, line 50: Cool. I like this. One of the first places I think the formatter IMPROVED the existing code instead of degrading it.

share/wake/lib/core/list.wake
```diff
@@ -811,7 +822,8 @@ export def unzip (list: List (Pair a b)): Pair (List a) (List b) = match list
 export def groupBy (cmp: a => a => Boolean) (list: List a): List (List a) =
     def sorted = sortBy cmp list
     def combo elem acc = match acc
-        (head, rest), tail if ! cmp elem head =
-            (elem, head, rest), tail
+        (head, rest), tail if ! cmp elem head = (elem, head, rest), tail
         _ = (elem, Nil), acc
     foldr combo Nil sorted
+
+
```
terpstra, line 829: Where did these newlines come from?

terpstra (commented):
I made comments as a review by accident. Clicking "Comment" to publish them.

share/wake/lib/core/integer.wake
V-FEXrt: > If a top-level function definition, leave function name, arguments, and type annotations on a line by themselves, with the function expression starting on a newline. I.e., 

Rules just for top-level are not super trivial (at least until we add some kind of "you are top level" signal) and I want to punt on adding them. If we do this for _all_ defs then that is quite easy to do.

> If the top-level function body expression is very simple, allow it on the same line.

Can you give a formal definition for "very simple" :)

>  For local function and value definitions, leave it to the discretion of the writer, particularly since these aren't as likely to have full type annotations.

I disagree. We should be opinionated, otherwise there isn't a point in having the formatter. The more discretion we give the more inconsistency the codebase has and the more opportunity for formatting bikesheds in random PRs

share/wake/lib/core/integer.wake
V-FEXrt: There just isn't a lambda formatting rule right now. Taking suggestions. @ag-eitilt brought this up in another thread as well

share/wake/lib/core/print.wake
```diff
@@ -57,56 +59,63 @@ export data Intensity =
...
-            Yellow  = 3
-            Blue    = 4
-            Magenta = 5
-            Cyan    = 6
-            White   = 7
+        None = 0
+        Some c =
+            8
+            + match c
+                Black = 0
```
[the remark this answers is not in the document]
V-FEXrt: Yeah, agreed. you can also flip it to be 
```
match ...
    ...
    ...
+ 8
```

share/wake/lib/core/tree.wake
```diff
@@ -226,264 +250,307 @@ export def tfoldmap (combineFn: b => b => b) (base: b) (transformFn: a => b) (tr
...
+    def helper i = match _
+        Tip = Pair Tip Tip
+        Bin _ l x r =
+            def sizeL = size l
+            if i > sizeL then
+                match (helper (i - sizeL - 1) r)
+                    Pair rl rr = Pair (join3 l x rl) rr
+            else
+                match (helper i l)
+                    Pair ll lr = Pair ll (join3 lr x r)
```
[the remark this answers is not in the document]
V-FEXrt: I think the info and primitives are there but wake-format shouldn't be doing semantic rewrites. This kind of thing should be done with a `wake-tidy`/static analyzer thing.

FWIW spinning up that tool with just this one rule wouldn't be too bad based on all the primitives from wake-format

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -13,79 +13,112 @@
...
+    Pair
+    String
+    (
+        (extraFlags: List String)
+        => (objects: List Path)
+        => (targ: String)
+        => (extraFiles: List Path)
+        => Result (List Path) Error
+    )
+)
```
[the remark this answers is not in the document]
V-FEXrt: no, I'd consider this "correct" in terms of the currently written rules. How would you prefer this to be output?

tests/type-system/extract-pass/test.wake
V-FEXrt: Ah yeah, this is a bug. Probably an edge case where there currently isn't a rule

vendor/emscripten.wake
V-FEXrt: Yeah, I don't know how the formatter would ever know that this set of string parameters are special but not some other set of string parameters

vendor/git.wake
V-FEXrt: Uhh I think this might be one of the cases where we have to change the code.

These comments are "floating"  since they don't bind to `Pass` (can only bind up/left if no newlines) and the don't bind down to `topic` because there is a newline and they "are" top-level. It maybe be intentional or a quick of the parser but if you put a comment at the end like this they actually get parsed one level up. So these comments are parsed at top-level instead of being nested by 1

vendor/gmp.wake
V-FEXrt: We can't really. There is no safe edit we can make inside of the `"""` that doesn't change the users code. In _most_ places this is fine but it _is_ a semantic change.

I think the best/correct option is to detect this in static analysis and emit a warning ala "the indentation of a multi-line string doesn't match the quote indentation, is this correct?"

build.wake
V-FEXrt: discussed in the meeting but doing this inside of `match`es is 100% plan of record. I was just planning on punting to post "MVP" since no info is lost in not aligning today. My hand might be forced here though :)

build.wake
V-FEXrt: Jake and I are going to explore options here based on discussions in the meeting and come back with more info later

build.wake
V-FEXrt: I'm not too familiar with `"%` strings, does adding indentation not change the real value of the string literal?

V-FEXrt: Looking at the quickref, it seems like adding indentation here would definitely change the value of the multiline string. I don't think the formatter should be changing the value of the string even if its probably harmless

build.wake
V-FEXrt: Yeah, seems like that's pretty universally agreed upon. Def something we can do

share/wake/lib/core/list.wake
V-FEXrt: mentioned in the meeting, its just a bug that hasn't been fixed yet

vendor/gmp.wake
bmitc: I learned a new thing. Did I miss this `"""` syntax somewhere in the documentation? Is it *just* a multiline string or is it something more special?

build.wake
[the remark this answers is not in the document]
bmitc: Gotcha! Thanks for that. Was just curious if I didn't know something about file endings. Haha. (And also you mentioned this in the meeting.)

share/wake/lib/core/integer.wake
bmitc: Ah, gotcha. I'll look for Sam's comment on this, but I'm wondering now if there should be a delineation between the two. This is because even when I was writing my comment, I was wondering if no space between the lambda arguments is fine for `\_\_` but then having the space is preferred for `\x \y`. I went with my comment because I sided on spaces for `\x \y` and consistency.

Some of the formatting rules for lambdas may depend on the result of https://github.com/sifive/wake/pull/1044.

share/wake/lib/core/tree.wake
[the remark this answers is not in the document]
bmitc: I agree with @V-FEXrt on the static analyzer thing. I actually had a comment for this typed up prior to Ashley's, but I apparently never submitted it. The gist of it was that I think having the formatter do rewriting, even for obviously equivalent things, is probably a slippery slope to go down. In that, probably best keeping the formatter as a pure formatter and leaving other suggestions to a static analyzer. (I.e., what Ashley said. Haha.)

If I am not mistaken, the Elixir formatter does not rewrite and only strictly formats, and [Credo](https://github.com/rrrene/credo) is used for catching things like this. (It's a little easier for a language like Elixir to have a thing like Credo since the AST is available in the language itself.)

share/wake/lib/core/double.wake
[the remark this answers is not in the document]
bmitc: Apologies for duplicating this commentary, but I made a similar observation elsewhere. I definitely prefer `\x \y \z` but am somewhat ambivalent what's chosen for `\_\_\_` since it's a matter of consistency (space) vs conciseness (no space) in that case.

bmitc:
Matthew mentioned this:

> I'm not sure if this PR exposes this, does the formatter move `package` decl to be the first thing after the comments at the start of a file

I'm curious about this as well. I'm wondering if we should come up with a standard for commenting packages such that the VS Code extension's "intellisense" can read off a package's documentation upon hovering over the package name in `from <packageName>`.

My initial preference would be that `package` is at the top of the file. But if we want to allow package documentation, then maybe the easiest way is to allow comments above `package`. Otherwise, `package` should be the top level thing.

As in:

```wake
# This is package "whatever"'s documentation that gets
# displayed by the VS Code extension on hover
package whatever

# This is just some commentary on the package's use and
# usefulness not displayed by intellisense.
```

build.wake
```diff
@@ -13,100 +13,103 @@
...
-    require Pass changelog =
-        setVersion release "debian/changelog"
-    require Pass spec =
-        setVersion release "wake.spec"
+    require Pass changelog = setVersion release "debian/changelog"
+    require Pass spec = setVersion release "wake.spec"
     # Identify those sources files to include in the tarball
-    require Pass allSources =
-        sources "." `.*`
+    require Pass allSources = sources "." `.*`
```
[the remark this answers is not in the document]
V-FEXrt: one side effect of this rule:

```
global export target fib n =
    match n
        0 =
            1
        1 =
            1
        _ =
            fib (n - 1) + fib (n - 2)
```

build.wake
```diff
@@ -13,100 +13,103 @@
...
+    buildJobCache,
+    buildWakeFormat,
+    Nil
+
 
 def all variant =
-    require Pass x =
-        map (_ variant) targets
-        | findFail
+    require Pass x = map (_ variant) targets | findFail
```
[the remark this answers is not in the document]
bmitc: I will admittedly sometimes put a pipeline on a single line, but for the most part, I don't think I mind a universal rule of being split over multiple lines. If a pipeline can fit on a single line, it can often just be a function application (though not always). However, there are cases where that doesn't work as well, such as the small unit testing DSL I was using: `triple 3 | should equalInteger 3`. But it wouldn't be the worst thing if that was on multiple lines.

It's also possible that pipelines should not be adjusted, or they follow what we've discussed for other situations, such as for `if then else`.

share/wake/lib/gcc_wake/gcc.wake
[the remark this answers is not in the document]
ag-eitilt: I think the discussion about different-behaviours-depending-on-context nixes this, but I'd be writing it with the entire passed function on a single line.
```export topic linkO: (
    Pair
    String
    ((extraFlags: List String) => (objects: List Path) => (targ: String) => (extraFiles: List Path) => Result (List Path) Error)
)```
(Side note, I don't currently like the way we need to surround the entire type signature in parentheses in order to break it into multiple lines, but I suspect that's something I'll wind up appreciating once I get used to it.)

File: vendor/emscripten.wake (hunk 18 above)
Comment (ag-eitilt) [reply vendor/emscripten.wake:44]: I've written/am updating an in-Wake arg parser; it's non-trivial to figure out when you're in control of the entire parser itself.  There's absolutely no way we can figure it out for arbitrary programs which may not even look like programs at the point of list definition.

File: vendor/git.wake (hunk 19 above)
Comment (ag-eitilt) [reply vendor/git.wake:45]: All right, two issues then.  First, they don't bind where the author expected (that might just be a "move your comments" requirement).  Second, even at the top level, these should be kept together rather than treated as three separate single-line comments.

**Comment by ag-eitilt** [comment]
> I'm curious about this as well. I'm wondering if we should come up with a standard for commenting packages such that the VS Code extension's "intellisense" can read off a package's documentation upon hovering over the package name in `from <packageName>`.

The problem with Wake is that there's no requirement for one-file-per-package; even though the LSP can trace any given symbol to what package it's from and where the symbol was defined, there's nothing saying that the symbol right next to it from the same package is also from the same place (I actually slightly abused the packaging when developing the Map interface by writing a file as `package wake` but saving it in a completely separate repository).  When you have a `from x import a b`, where `a` and `b` are from two different files, what do you use if the packaged documentation conflicts between them?

File documentation, sure, that's a good thing to consider (though each project using Wake could easily have a different format to the comments).  Package documentation, that's probably not possible.  Then again, do C++/C#/etc. have any discoverable documentation for namespaces (does C# even have namespaces)?

File: build.wake (hunk 29 above)
Comment (terpstra) [reply build.wake:80]: It would not change the value of the string.

Comment (terpstra) [reply build.wake:80]: These format the same and are the only two ways I think we should accept formatting multi-line strings:
```
export def hax1 = "%
  fun day.
  happy day.
%"
export def hax2 =
  "%
    fun day.
    happy day.
  %"
```

Comment (V-FEXrt) [reply build.wake:80]: SG, this should be an easy change. I was just being cautious around editing the stuff inside the quotes

File: extensions/vscode/vscode.wake (hunk 26 above)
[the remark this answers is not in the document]
Comment (bmitc) [reply extensions/vscode/vscode.wake:30]: Given the input `let triple (x: int): int = 3 * x`:

* The [F# formatter, Fantomas](https://github.com/fsprojects/fantomas), will format this as
    ```fsharp
    let triple (x: int) : int = 3 * x
    ```
* The [OCaml formatter, OCamlFormatter](https://github.com/ocaml-ppx/ocamlformat), will format this as
    ```ocaml
    let triple (x : int) : int = 3 * x
    ```
    I will say that while this is arguably more consistent (in that a space is always present), it spreads things out too horizontally, in my opinion.
* The [Scala formatter, Scalafmt](https://scalameta.org/scalafmt/), will format the equivalent as
    ```scala
    def triple(x: Int): Int = {
      3 * x
    }
    ```

* SML is less definitive. For one, I couldn't get the only formatter I could find, [SMLFormat](https://github.com/jluningp/smlformat), to format anything. As I mentioned, I did look through some online code as well as several SML books I have, and it did seem to me that the space was more common.
    * MLton does have a published [stye guide](http://mlton.org/SyntacticConventions) which does differ and denotes no space to the left of the `:`.
    * There are several other style guides for SML, mainly published by universities who use it for teaching such as [Tufts' style guide](https://www.cs.tufts.edu/comp/105-2017f/handouts/mlstyle.pdf), which prefer the space.
    * And like I said, the books I browsed through did seem to prefer the space overall, but confirmation bias could have been at work there and it wasn't universal. For example, I did note that one book used the horrid `let triple (x :int) ...` syntax for parameters. Haha.

All that being said, this is somewhat of a toss-up. Although, I would personally side on following F# and OCaml due to being the more popular MLs in use (which is obviously what I am used to), but that is an "appeal to authority" style argument in a sense. Wake breaks away from F#, OCaml, SML, and Scala in pretty distinct ways at times. And I also suppose that the fact that Scala is the other main language in use at SiFive tips the scale that way.

I think the primary decision should probably center around the acceptability of the use case `def triple x: Integer = 3 * x` and whether that is clear or not.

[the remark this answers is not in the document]
Comment (ag-eitilt) [reply extensions/vscode/vscode.wake:30]: For what it's worth, Haskell uses `::` which as far as I've seen is almost always spaced, but is typically used slightly differently:
```haskell
triple :: Int => Int
triple x = 3 * x

-- Rarer, but closer to what we're talking about
tripleLet :: Int => Int
tripleLet x =
    let y = 3 :: Int
    in  y * x
```
I don't have much preference here when it comes to Wake; I'm used to no space before and lean slightly in that direction purely through familiarity, but don't feel any animosity toward having a space on both sides.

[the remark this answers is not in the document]
bmitc: `let y = 3 :: Int` is interesting in Haskell given that it occurs on the right of the value, whereas in F#, OCaml, SML it occurs on the left of the equals sign like `let y : Int = 3`. Elixir's typespecs also use `::` to separate the argument types and return type, although that is a pretty wildly different language and type system. (`::` is cons in F#, OCaml, and SML.)

Actually, I just tried this in F#:

```fsharp
> let x = 3 : int;;
val x: int = 3

> let x : int = 3;;
val x: int = 3
```

Today I learned that that is even supported.

Given that Wake does already print out the function type as `triple: (x: Integer) => Integer` on the command line when doing `wake -vx 'triple'`, I suppose I am swimming upstream here in that Wake already has a formatting standard set by that, as well as Scala's formatter goes with the no space as well and is the other functional language in use at SiFive. I think I stand behind it breaking tradition (if we can call it tradition) with F# and OCaml and muddles the example I gave a little bit. However, I suppose that case can be mitigated by encouraging argument type annotations.

share/wake/lib/core/tree.wake
```diff
@@ -13,94 +13,115 @@
...
-      GT = balanceL (helper l) x r
-      EQ = t
-      LT = balanceR l x (helper r)
-  Tree cmp (helper root)
+    def helper t = match t
+        Tip = Bin 1 Tip y Tip
+        Bin _ l x r = match (cmp x y)
+            GT = balanceL (helper l) x r
+            EQ = t
+            LT = balanceR l x (helper r)
```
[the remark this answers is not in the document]
bmitc: I'll admit that I feel a little lost in the woods here with complex match cases like this, because there's some jolting of my F#/OCaml and even Elixir senses with `=` replacing `->`, there being no `|` separators, and `match` not being able to be aligned with the match clauses. I know those aren't coming in at all, so not arguing that, but all that to say that I don't think I have much opinion here. For example, despite me knowing Wake syntax, I actually had to read this `Bin _ l x r if cmp x y | isGT =` a few times to re-orient myself that `|` is not a separator. Haha.

@ag-eitilt Could you write out what you were thinking the `match`s could be formatted as? That might help to have something to look at. I think I'm fine with the original match formatting (without thinking about the logic too heavily) given Wake's syntax constraints/choices. If the `match` statements get put on new lines (justifiable), then that unfortunately pushes everything to the right with two additional indentation levels.

share/wake/lib/system/job.wake
bmitc: Looking at the surrounding code, if this `if then else` can't be kept on a single line, I think I'd recommend in code review to not have a `if then else` here and rather either a variable name or simple function application (like `booleanToInteger keep`). I would possibly recommend this regardless. However, this one of those cases where I really wish

```
(
    <stuff>
)
```

could be avoided.

bmitc:
> The problem with Wake is that there's no requirement for one-file-per-package

That's a good point @ag-eitilt, and one I know but neglected to think of. That does ruin the documentation for package idea. I think now my opinion is that the package declaration should come at the very top of the file, at least without thinking about it more. (Is it true that there can be only one `package` declaration per file? I haven't tried.) The reasoning would be that it's something you'd look for and has a static height (single line) versus comments/documentation/descriptions having variable line lengths. So, if the `package` declaration came after comments, then that could variably place it near the top of the file. That's just an initial thought though.

C# does have namespaces, and F# has namespaces (by virtue of being a .NET language) and modules. That's basically what I was thinking of (while I was forgetting Wake's unique package feature).

In F#, you can document a module with:

```fsharp
/// This documents the module ModuleName
module NamespaceName.ModuleName

/// Function documentation
let test x = 2 * x
```

However, now that I'm thinking about it, you can extend an F# module in a separate file. Not sure what happens if you try to document the module extension. I'm not sure off the top of my head about documenting namespaces.

Elixir has:

```elixir
defmodule ModuleName do
@moduledoc """
Module documentation
"""

@doc """
Function documentation
"""
def test(x) do
  2 * x
end
```

Both situations in F# and Elixir will show the module and function documentation in their respective VS Code extensions.

share/wake/lib/core/tree.wake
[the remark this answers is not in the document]
ag-eitilt: With us currently leaning toward "line break after every `=`", that comment isn't particularly relevant any more; it was looking at the two `= match` and weighing the relevant benefits of leaving each on the same line.

To go back to my comment somewhere else about other deeply-nested `match` trees, it might wind up cleaner to refactor these with more `require` (this isn't the best example since the second `match` still has three cases and `t` doesn't look great on a line by itself, but I hope you get the idea):
```suggestion
    def helper t =
        require Bin l x r =
            t
        else
            Bin 1 Tip y Tip
        match (cmp x y)
            GT =
                balanceL (helper l) x r
            EQ =
                t
            LT =
                balanceR l x (helper r)
```

extensions/vscode/vscode.wake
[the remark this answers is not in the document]
ag-eitilt: The Haskell trick which likely works the same in F# is that it's not actually annotating the type of `y`, it's annotating the type of `3` -- `let (y :: Int) = 3` is *probably* legal, but it's not how I see people writing it in Haskell.

PR #1055, opened 2022-10-26
fix small issues formatting the internal repo

V-FEXrt:
Fix small formatting errors for the following cases
 - missing a whitespace before/after `=` in `def` and `tuple`
 - missing a whitespace before/after `:` in topic
 - comments before a "fits first" `match`
 - lambda subtree
 - The memo cleanup vector had a **bunch** of duplicate pointers, a set is sufficient and efficient


ag-eitilt (approved):
As far as I know the context, looks good to me!

PR #1056, opened 2022-10-26
try alt formatting rules

V-FEXrt:
Tried some of the easier suggested rules from the other PR. Putting it up for comparison

### DIFF
--- a/share/wake/lib/core/json.wake
+++ b/share/wake/lib/core/json.wake
@@ -13,161 +13,328 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 package wake
 
 # The JSON data type
 export data JValue =
-  JString  String
-  JInteger Integer
-  JDouble  Double
-  JBoolean Boolean
-  JNull
-  JObject  (List (Pair String JValue))
-  JArray   (List JValue)
+    JString String
+    JInteger Integer
+    JDouble Double
+    JBoolean Boolean
+    JNull
+    JObject (List (Pair String JValue))
+    JArray (List JValue)
 
-export def getJString: JValue => Option String = match _
-  JString x               = Some x
-  JArray (JString x, Nil) = Some x
-  _                       = None
-export def getJInteger: JValue => Option Integer = match _
-  JInteger x               = Some x
-  JArray (JInteger x, Nil) = Some x
-  _                        = None
-export def getJDouble: JValue => Option Double = match _
-  JDouble x               = Some x
-  JInteger x              = Some (dint x)
-  JArray (JDouble x, Nil) = Some x
-  JArray (JInteger x, Nil)= Some (dint x)
-  _                       = None
-export def getJBoolean: JValue => Option Boolean = match _
-  JBoolean x               = Some x
-  JArray (JBoolean x, Nil) = Some x
-  _                        = None
-export def getJObject: JValue => Option (List (Pair String JValue)) = match _
-  JObject x               = Some x
-  JArray (JObject x, Nil) = Some x
-  _                       = None
-export def getJArray: JValue => Option (List JValue) = match _
-  JArray x = Some x
-  _        = None
+export def getJString: JValue => Option String =
+    match _
+        JString x =
+            Some x
+        JArray (JString x, Nil) =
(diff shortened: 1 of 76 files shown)


share/wake/lib/system/environment.wake
```diff
@@ -13,9 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 package wake
 
-def uname = prim "uname"
+def uname =
+    prim "uname"
```
mmjconolly, line 20: We have some polarizing choices for `=` and `\n`, being:
* Zero chars after `=`, then newline
* Up to line-length after `=`, then newline

I can try to come up with some softer in-betweens:

We could have a: 
* Max of (say) 20 chars after `=` (up to regular line length), then newline

or a even a 
* Only allow `[a-zA-Z0-9_.]` after `=`, then newline (ie, only 'simple' looking expressions expression allowed)

share/wake/lib/system/job.wake
```diff
@@ -440,108 +736,189 @@ export def defaultRunner: Runner = fuseRunner
...
+                def pwrite m p d =
+                    prim "write"
+                def json =
+                    JObject
+                    (
+                        "label" :-> JString label,
+                        "command"
+                        :-> command
+                        | map JString
+                        | JArray,
```
ag-eitilt, line 790: I think the human perspective categorizes symbolic operators into distinct categories, which the AST doesn't necessarily capture.

ag-eitilt (commented):
I think the previous PR is better *as is* and this one winds up requiring the short-medium-long tristate logic rather than being workable in its current state, but overall I do greatly prefer this.  (And, yes, I know these aren't necessarily all-one or all-the-other.)

share/wake/lib/core/list.wake
```diff
@@ -158,16 +181,22 @@ export def mapFlat (mapFn: a => List b): (list: List a) => List b =
...
+                Nil
+            h, t =
+                # don't wait on f to process tail:
+                def sub =
+                    loop t
+                match (f h)
+                    Some x =
+                        x, sub
+                    None =
+                        sub
```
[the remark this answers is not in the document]
terpstra: I mean, this code is fine. The formatting is on crack, though.

share/wake/lib/system/job.wake
terpstra: What's super confusing here is that the operator precedence is not respected.
The expression here is `"command" :-> (command | map JSrting | JArray)` and the formatting totally hides that.

share/wake/lib/core/json.wake
```diff
@@ -413,7 +595,12 @@ def mergeValueList (describeCompat: Boolean) (values: List JValue): Result JValu
...
     def compatibilityMessage upper lower error =
         if describeCompat then
-            ("For compatibility, JSON objects should not have duplicate keys. While we try to recover from this where possible, ", lower, error, Nil)
+            (
+                "For compatibility, JSON objects should not have duplicate keys. While we try to recover from this where possible, ",
+                lower,
+                error,
+                Nil
+            )
             | cat
```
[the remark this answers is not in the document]
bmitc: I am a strong believer that the style of

```
(
    <stuff>
)
```

should be discouraged in a language with ML heritage. It feels very C/C++, C#, Java, and Dart like. I know it's a broken record at this point, haha, but having a literal list syntax would help this a bit, as (as far as I can tell) the usual way to format a multiline list in other MLs is:

```fsharp
[ "For compatibility, ...",
  lower,
  error ]
```

(We should probably verify that, but that's what I feel I've seen, and it's certainly how I currently prefer do it.)

I know I am touching upon a oft visited point, but we are already needing to surround lists with `()` quite often. In wait of the literal list syntax, I'd prefer a list like this to be formatted as

```fsharp
( "For compatibility, ...",
  lower,
  error, )
```

When `[]` comes into play, then the formatter could seemingly "easily" (🙈) reformat this to the above example using `[]`.

V-FEXrt:
The main comments in this PR have been applied to the formatter code. Closing this now for a re-upload with the current state of the world

PR #1058, opened 2022-11-01
format unary subtree

V-FEXrt:
Formats UNARY subtrees. Any wsnl between a unary and the parameter is removed

### DIFF
--- a/tests/wake-format/basic/basic.wake
+++ b/tests/wake-format/basic/basic.wake
@@ -696,6 +696,8 @@ def buildRE2WASM Unit =
 
     Pass (SysLib "1.0" headers objects cflags lflags)
 
+def x = (x ,) ++ xs
+
 def a =
   # comment
   def b = 5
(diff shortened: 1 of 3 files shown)


ag-eitilt (approved):
I find this to be one of the more artistic elements of coding, where you might want to avoid a space to visually bind a unary operator to its argument in a compound expression (`x && !y`) but include the space when you want to emphasize it (the `! matches` in the changed output), but with the formatter we're going to have to pick one way or the other across the board, and this this looks good enough to me.

I assume this works correctly for unary postfix operators?  I've not gone through the test input to see if we have a unary comma or anything (if not, maybe something like `x , ++ xs` -> `x, ++ xs` could provide a good test case; I don't know what the precedence is there).  Or are they a different CST node type?

V-FEXrt:
Yeah, it handles postfix unary as well. I added the suggested test case. I did need to play around with the precedence though

V-FEXrt:
> but include the space when you want to emphasize it
imo this is something where we should encourage parens to emphasize instead of relying on the space. `!(complicated thing)` vs `! complicated thing`

PR #1060, opened 2022-11-08
enact def + = newline rules

V-FEXrt:
Enacts the rules for newlining defs as discussed in chat.

Specially
- Same rules as current formatter (the first bikeshed PR) plus
- Top level = always newlines
- except no newlines for single top level literals/identifiers

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -92,6 +92,30 @@ static inline bool is_floating_comment(wcl::doc_builder& builder, ctx_t ctx, CST
   return it->second.bound_to.id() == TOKEN_COMMENT;
 }
 
+// determines if the rest of the node
+// - only has one sibling node
+// - and that node is a "single" thing (literal or identifier)
+static inline bool is_single_literal(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node,
+                                     const token_traits_map_t& traits) {
+  CSTElement copy = node;
+  if (!copy.isNode()) {
+    copy.nextSiblingNode();
+  }
+
+  size_t node_count = 0;
+  while (!copy.empty()) {
+    node_count++;
+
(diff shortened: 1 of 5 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -262,11 +287,16 @@ auto Emitter::rhs_fmt() {
...
       return c->has_newline() && c->last_width() == 0;
    }, full_fmt)
-   // If the RHS has a leading comment then we must use the full_fmt
+    // If the RHS has a leading comment then we must use the full_fmt
    .pred([this](const wcl::doc_builder& builder, ctx_t ctx, const CSTElement& node,
             const token_traits_map_t& traits) {
       return count_leading_newlines(token_traits, node) > 0;
    }, full_fmt)
+    // Always newline when requested unless the thing to be formatted is a "single literal".
+    // Used for top-level defs and top-level "constatn" defs
```
ag-eitilt, line 296:
```suggestion
    // Used for top-level defs and top-level "constant" defs
```

tools/wake-format/emitter.h
```diff
@@ -70,6 +72,9 @@ class Emitter {
   // Marks all node elements that have had their formatting disabled
   void mark_no_format_nodes(CSTElement node);
 
+  // Marks top level nodes. Any node on 'top level' will have top level = true
+  void mark_top_level_notes(CSTElement node);
```
ag-eitilt, line 76:
```suggestion
  void mark_top_level_nodes(CSTElement node);
```
...and its uses.

ag-eitilt (approved):
Couple spelling fixes, but otherwise looks good!

tools/wake-format/emitter.h
V-FEXrt: nice catches! Thanks!

PR #1061, opened 2022-11-09
core: use explicit Order for sequence processing

ag-eitilt:
The existing behaviour of passing *either* a less-than function or an equality function to sorting-like operations has proven repeatedly confusing to people -- notably, `groupBy` is expected to take equality but gives confusing results when passed `==` because it actually needs `<`.  This avoids that confusion entirely by requiring the functions take an `Order` rather than a `Boolean`; the algorithms have not been updated to take advantage of the tri-state type but even just the explicit return should be a user-side benefit.

I have left the `distinctRunBy` functions unchanged (expecting `Boolean` equality) as an option for any types where a total ordering is unable to be defined.


V-FEXrt (approved):
Seems like a strict improvement to me.

Will require a bump to 0.29.X with the breaking API change.

ag-eitilt:
Yeah, definitely needs a version bump.  I remember discussing something else which would also need a bump in the meeting a few weeks ago (can't remember what exactly that was, though) so it seemed like a good time to get this in.

PR #1062, opened 2022-11-09
map: use standard semantics for mexists

ag-eitilt:
In the existing types, `exists`/`vexists`/`texists` requests a predicate function and checks if it succeeds for any value, while `tcontains` requests a specific object and checks if it's equal to any value.  In the `Map` interface, I had mistakenly named the function which requests a specific key and checks if any value exists associated with that key `mexists` instead; this renames it to match the similar `tcontains` pattern, and introduces a new predicate-requesting function in its place.

This will break any code currently using "`mexists`", but since the type signatures are irreconciably different, everything will fail loudly rather than users silently getting unexpected behaviour in downstream logic.


V-FEXrt (approved):
What is our migration strategy for this change and the previous one? Seems like good candidates for a quick migration.

ag-eitilt:
The versioned package has been updated so we can release 0.29.0 whenever and just update any uses as we come across them.  On the other hand, neither of these are *critical* changes, so if anyone has anything else approaching a publishable state, we can definitely wait for that to merge and release it all at once.

PR #1063, opened 2022-11-09
only allow flat if for trivial ternaries

V-FEXrt:
Only allow flat `if - then - else -` when  `-` is a "simple" thing. Right now that is defined as a literal or identifier but that can be refined over time.

### DIFF
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -632,7 +632,11 @@ def tarball Unit =
     require Pass tarball =
         def cmd =
             def gnutar = which "gnutar"
-            def tar = if gnutar ==* "gnutar" then which "tar" else gnutar
+            def tar =
+                if gnutar ==* "gnutar" then
+                    which "tar"
+                else
+                    gnutar
             def files = map getPathName (manifest, testManifest, spec, srcs) | sortBy (_ <* _)
             tar,
             "--mtime={time}",
@@ -747,7 +751,12 @@ export def vfoldl (combiningFn: accum => element => accum): accum => Vector elem
 
 
 def loop a s e =
-    if s == e then e else if stopFn (vget a s) then s else loop a (s + 1) e
+    if s == e then
+        e
+    else if stopFn (vget a s) then
+        s
+    else
+        loop a (s + 1) e
 
 
 def loop aaaaa sssss eeeee =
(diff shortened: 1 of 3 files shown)


tests/wake-format/basic/stdout
```diff
@@ -747,7 +751,12 @@ export def vfoldl (combiningFn: accum => element => accum): accum => Vector elem
 
 
 def loop a s e =
-    if s == e then e else if stopFn (vget a s) then s else loop a (s + 1) e
+    if s == e then
+        e
+    else if stopFn (vget a s) then
+        s
+    else
+        loop a (s + 1) e
```
ag-eitilt, line 759: What would the format be if this were `...else if shouldStop then s else t`?  (Basically, does `else if` respect the explosion status of the initial `if`?)

ag-eitilt (approved):
So long as `else if` is handled as expected, looks good to me!  Will `require` be moved to this in a followup PR?

tests/wake-format/basic/stdout
V-FEXrt: yep! `if` explosion propagates to all `if`s in the tree. Previously had to add a case specifically for that. Went ahead and added an explicit test case for it

V-FEXrt:
>  Will require be moved to this in a followup PR?
I don't think require ever meets the bar of being simple right?

in something like
```
require Pass a = b
```
`Pass a` isn't "simple"

V-FEXrt:
Also, slightly adjacent, how do we will about `else match`?

i.e.
```
def loop a s e =
    if s == e then
        e
    else match e
        0 = 0
        n = 5
```

vs
```
def loop a s e =
    if s == e then
        e
    else 
        match e
            0 = 0
            n = 5
```

ag-eitilt:
> > Will require be moved to this in a followup PR?
>
> I don't think require ever meets the bar of being simple right?

Ah, good point.  I was imagining simple `else` cases but not thinking about the other half of the expression.  There might still be valid discussion around *if* the `require` can go on a single line according to the rules of `def` *and if* the `else` case is simple we *might* be able to condense it onto two lines rather than four, but I don't think it should ever be one.

> Also, slightly adjacent, how do we will about `else match`?

I like your first example, but that might be a point where we need to poll the others.

PR #1064, opened 2022-11-10
format simple constructor applies on the same line

V-FEXrt:
An apply that "looks" like a constructor, something like 

```
Identifier (
  some,
  expression
)
```

is now formatted without a newline between the lhs and rhs

Bonus: the assign operator (`:->` )  is always emitted "flat" as adding newlines makes the code confusing.

Bonus #2: if-else-match has else and match on the same line

### DIFF
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -447,8 +447,7 @@ publish compileC =
     emccFn (makeCompileC "a" "b") # comment
 
 publish compileC =
-    emccFn
-    (
+    emccFn (
         makeCompileC # comment
         "wasm-c11-release"
         _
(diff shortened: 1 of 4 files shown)


V-FEXrt:

tests/wake-format/basic/stdout
```diff
@@ -982,3 +981,61 @@ export def !(x: Boolean): Boolean =
...
+        def imp m p c = prim "write"
+        def pre input = Pair input Unit
+        def post = match _
+            Pair (Fail f) _ = Fail f
+            Pair (Pass output) Unit =
+                if mode < 0 || mode > 0x1ff then
+                    Fail (makeError "write {path}: Invalid mode ({strOctal mode})")
+                else match (imp mode path content)
+                    Fail f = Fail (makeError f)
+                    Pass path = Pass (editRunnerOutputOutputs (path, _) output)
```
ag-eitilt, line 1007: Are these the constructors the main PR was targeting?  I don't see much else other than the initial `emccFn (` which seems to be exercised, though I've also not looked at the C++ to know *what* to look for here.  Nevertheless, looks good!

ag-eitilt (approved):

JakeSiFive (approved):
C++ code looks good to me.

tests/wake-format/basic/stdout
V-FEXrt: No, I think the test here was the "else match" part. Though these constructors not getting exploded does show a bit of that functionality

PR #1065, opened 2022-11-10
Newline top level pipes/dollars

V-FEXrt:
PTAL a look at the formatting @ag-eitilt but I wouldn't focus on the C++ implementation for the moment. I want to tweak how we add things to ctx_t so it'll probably change a bit.

This is the implementation of the rules
 -  always newline `|` and `$` when they are the "root" of a binary op chain
 - when `|` and `$` are not the root, apply the normal binop formatting rules (don't newline if you fit)

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -1131,22 +1131,30 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
     parts = collect_right_binary(op_token, node);
   }
 
-  std::vector<wcl::optional<wcl::doc>> choices = {
-      // 1
-      combine_flat(op_token, ctx, parts),
-      // 2
-      combine_explode_first(op_token, ctx, parts),
-      // 3
-      combine_explode_last(op_token, ctx, parts),
-      // 4
-      combine_explode_all(op_token, ctx, parts),
-      // 5
-      combine_explode_first_compress(op_token, ctx, parts),
-      // 6
-      combine_explode_last_compress(op_token, ctx, parts),
-  };
+  if (!ctx.nested_binop &&
+      (op_token.id() == TOKEN_OP_DOLLAR ||
+       (op_token.id() == TOKEN_OP_OR && op_token.fragment().segment().str() == "|"))) {
+    MEMO_RET(select_best_choice({
+        combine_explode_first(op_token, ctx.binop(), parts),
+        combine_explode_last(op_token, ctx.binop(), parts),
+        combine_explode_all(op_token, ctx.binop(), parts),
+    }));
+  }
 
-  MEMO_RET(select_best_choice(choices));
+  MEMO_RET(select_best_choice({
+      // 1
+      combine_flat(op_token, ctx.binop(), parts),
+      // 2
+      combine_explode_first(op_token, ctx.binop(), parts),
+      // 3
+      combine_explode_last(op_token, ctx.binop(), parts),
+      // 4
+      combine_explode_all(op_token, ctx.binop(), parts),
+      // 5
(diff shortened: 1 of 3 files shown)


V-FEXrt:

ag-eitilt (approved):
Not looked at the C++, but the output looks good to me.

tools/wake-format/emitter.cpp
```diff
@@ -887,6 +887,32 @@ wcl::optional<wcl::doc> Emitter::combine_apply_flat(ctx_t ctx,
   return {wcl::in_place_t{}, std::move(doc)};
 }
 
+wcl::optional<wcl::doc> Emitter::combine_apply_constructor(ctx_t ctx,
```
JakeSiFive, line 890: rebase?

tools/wake-format/emitter.cpp
```diff
@@ -1106,22 +1131,30 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
     parts = collect_right_binary(op_token, node);
   }
 
-  std::vector<wcl::optional<wcl::doc>> choices = {
+  if (!ctx.nested_binop &&
+      (op_token.id() == TOKEN_OP_DOLLAR ||
```
JakeSiFive, line 1135: It might be good to add the check for a leading "$"

tools/wake-format/emitter.cpp
```diff
@@ -1106,22 +1131,30 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
...
+       (op_token.id() == TOKEN_OP_OR && op_token.fragment().segment().str() == "|"))) {
```
JakeSiFive, line 1136: I think you should also check for a leading pipe instead of the whole string being a pipe. This gives us room to add operators in the future like `|>` and stuff.

tools/wake-format/types.h
```diff
@@ -44,6 +44,7 @@ struct ctx_t {
   // 1: By default, don't explode unless you have to
   // 2: if prefer_explode, explode if you can
   bool prefer_explode = false;
+  bool nested_binop = false;
```
JakeSiFive, line 47: Do we need both of these? I forgot the discussion around all this but I'm wondering if nested_binop can replace prefer_explode?

V-FEXrt: Yeah, prefer_explode is more general and influences things more broadly. nested_binop just says "every child node from here on out is inside of at least one other binop" which turns out to be useful signal for many different formatting decisions

tools/wake-format/emitter.cpp
V-FEXrt: Confirming now, but I think the reason I had to do this was because `||` is also TOKEN_OP_OR which we don't want to hit this case

V-FEXrt: Yep, confirmed. This matches `||` which looks not good

```
    def post = match _
        Pair (Pass output) Unit =
            if mode < 0
            || mode > 0x1ff then
                Fail (makeError "write {path}: Invalid mode ({strOctal mode})")
            else ....... cut ......
```

JakeSiFive: ok yeah then just be conservative for now

tools/wake-format/types.h
JakeSiFive: What are those formatting decisions?

mmjconolly:
Didn't read the c++, but this looks nicer to me than having the lines extend rightwards

JakeSiFive (approved):
I gotta run but I think I'm good with this. I want to take a look at this more later but I think I can refactor it later if I'm not happy with it after I think about it more.

V-FEXrt:
Sounds good!

PR #1066, opened 2022-11-11
Tweak the defination of simple parts to be more permissive

V-FEXrt:
- Simple 2 item applies like `Some ""` where each item is a literal or id is now considered a "simple literal"
  - this also means a top level `def x = Some ""` doesn't get newlined/is considered a constant
- Simple 2 part binop like `x < 0` where the lhs/rhs is a literal or id is now considered a "simple binop", and the shape `if simple_binop then simple_literal else simple_literal` is eligible to be on a single line
- `|` or `$` nested inside of parenthesis are now eligible to be on a single line


V-FEXrt:

PR #1067, opened 2022-11-11
Run current formatter over repo

V-FEXrt:
Most recent set of rules for folks to take a look at.

Following things are still being worked on
- trailing newline at the end of the file
- single line, top level defs should only get 1 newline, not 2
- if a def/require has a CST_BLOCK it should have a freshline before the last item
- require should follow similar singleline / multiline rules that `if-then-else` has
- some top level defs aren't newlining correctly
- several comments from the original bikeshed also need to be reviewed/decided on

### DIFF
--- a/build.wake
+++ b/build.wake
@@ -13,32 +13,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 package build_wake
 from wake import _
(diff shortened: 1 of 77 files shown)


V-FEXrt:

PR #1068, opened 2022-11-16
tree: fix bug in tinsertWith/minsertWith

ag-eitilt:
Using `tinsertWith`/`minsertWith` on larger structures would sometimes duplicate some elements and drop others.  I tracked it down to a one-character copy-paste error, and added a regression test (likely not the minimum for reproduction but it should still work) to avoid running into the same issue in the future.


V-FEXrt (approved):
LGTM, thanks for pinning this down!

PR #1069, opened 2022-11-17
list: distinctBy did originally take an Order function

ag-eitilt:
#1061 switched a few list functions to take `Order` rather than `Boolean` predicates, and updated the versioned package to maintain the compatibility API.  In doing the latter, though, I overlooked that `distinctBy` was the *one* function which was already taking an `Order`.  This reverts that one function in `v0_28_wake` as it's still compatible with the 0.29 version.


terpstra:
Wait. Why did we change to taking an order function instead of a comparison? You can't express a partial order with a function which returns `Order`, so it's a loss of compatability to unnecessarily expect `Order` where `a => a => Boolean` would have sufficed.

terpstra:
I mean, I guess it doesn't matter so much since most of these functions used the `<` operator in a way that assumed a total order.

ag-eitilt:
Exactly.  `sortBy` and `groupBy` required less-than, and `distinctBy` wanted an `Order` anyway.  There were several people who had tried passing `==` to `groupBy` and it was not at all clear why the returned list was not as expected.  Maybe the better solution *is* to make equality versions, but even if we can do that without too slow of an algorithm, we'd want a different name/type to make the version bump fail loudly.

ag-eitilt:
Ah, I see what you're getting at with allowing a partial ordering, especially for `sortBy`.  I would still argue for Order on the basis of both A., consistency with `groupBy` (which needs to be Order/`==` for UX) and `distinctBy`, and B., a lack of partial-ordered types in typical Wake, but I do accept that we might want to add something new in under a different name for people who are really pushing the boundaries of this as a build language.

ag-eitilt:
Even if `Order`'s not optimal, do you mind if I merge this?  I'm hoping to release 0.29 by the end of the day.

ag-eitilt:
Going ahead with merging based on previous discussion and to unblock bugfixing via 0.29 in a private repo.

PR #1070, opened 2022-11-17
Support nested targets in CST_BLOCK

V-FEXrt:
I noticed that 

```
def wrappedFib n =
  target fib n = if n < 2 then 1 else fib (n-1) + fib (n-2)
  fib n
```

from some older docs wasn't supported and caused a parsing error.

I updated the parser, and then also had to update the cst->dst conversion. 

Now the original code works correctly


V-FEXrt:
Per DM with Wes, this was intentionally removed so I'm closing this

PR #1073, opened 2022-11-28
Fix wake clean dir bugs

V-FEXrt:
Fixes the following `wake --clean` bugs
 - Can't use wake clean if something creates `.`
 - Wake clean doesn't work on non-linux posix systems

### DIFF
--- a/tests/command-line/clean/pass.sh
+++ b/tests/command-line/clean/pass.sh
@@ -1,0 +1,5 @@
+#! /bin/sh
+WAKE="${1:+$1/wake}"
+"${WAKE:-wake}" -q --init .
+"${WAKE:-wake}" -q -x 'write "bug" ""'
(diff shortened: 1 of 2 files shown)


tests/command-line/clean/pass.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
+WAKE="${1:+$1/wake}"
+"${WAKE:-wake}" -q --init .
+"${WAKE:-wake}" -q -x 'write "bug" ""'
```
ag-eitilt, line 4: Specifically, this doesn't *directly* create a `Path "."`, but it calls `mkdir` to create that (current) directory so it covers what we're looking for.

ag-eitilt (approved):
Looks good as far as I can tell!

PR #1074, opened 2022-11-28
system: make incremental artifact creation atomic

ag-eitilt:
Incremental jobs could leave the workspace in a broken state if the user aborted Wake partway through the tarball being generated, as the resulting (truncated) file would have an incompatible hash with the database version.  This makes that archival atomic to ensure only the complete artifacts are hashed..


V-FEXrt (approved):

ag-eitilt:
The merge into master failed due to a 503 contacting Docker; I think that's just transitory, but I'll be keeping a close eye on the next few PRs just in case.

PR #1075, opened 2022-11-28
profile steps before executing wake source

V-FEXrt:
Prints out a progress message if startup takes longer than a second.

### DIFF
--- a/src/parser/wakefiles.cpp
+++ b/src/parser/wakefiles.cpp
@@ -191,8 +191,8 @@ static bool push_files_should_recurse(int dirfd, const std::string &path, struct
 }
 
 static bool push_files(std::vector<std::string> &out, const std::string &path, int dirfd,
-                       const RE2 &re, size_t skip, std::chrono::_V2::steady_clock::time_point start,
-                       size_t *explored) {
+                       const RE2 &re, size_t skip,
+                       std::chrono::time_point<std::chrono::steady_clock> start, size_t *explored) {
   auto dir = fdopendir(dirfd);
   if (!dir) {
     close(dirfd);
--- a/src/parser/wakefiles.cpp
+++ b/src/parser/wakefiles.cpp
@@ -171,6 +171,12 @@ bool push_packaged_stdlib_files(std::vector<std::string> &out, const re2::RE2 &r
 
 #else
 
+struct profile_data {
+  std::chrono::time_point<std::chrono::steady_clock> start{std::chrono::steady_clock::now()};
+  size_t explored{0};
+  bool alerted_user{false};
+};
+
 static bool push_files_should_recurse(int dirfd, const std::string &path, struct dirent *f,
                                       bool *failed) {
   struct stat sbuf;
@@ -191,8 +197,7 @@ static bool push_files_should_recurse(int dirfd, const std::string &path, struct
 }
 
 static bool push_files(std::vector<std::string> &out, const std::string &path, int dirfd,
-                       const RE2 &re, size_t skip,
-                       std::chrono::time_point<std::chrono::steady_clock> start, size_t *explored) {
+                       const RE2 &re, size_t skip, struct profile_data *profile) {
   auto dir = fdopendir(dirfd);
   if (!dir) {
     close(dirfd);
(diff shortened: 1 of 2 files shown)


JakeEhrlich:
oh this is fantastic, we kept talking about doing this but I never got around to it

src/parser/wakefiles.cpp
```diff
@@ -170,8 +171,33 @@ bool push_packaged_stdlib_files(std::vector<std::string> &out, const re2::RE2 &r
 
 #else
 
+struct profile_data {
+  std::chrono::time_point<std::chrono::steady_clock> start{std::chrono::steady_clock::now()};
+  size_t explored{0};
+  bool alerted_user{false};
+};
+
+static bool push_files_should_recurse(int dirfd, const std::string &path, struct dirent *f,
```
JakeSiFive, line 180: This needs a comment to explain what it does

tools/wake/main.cpp
```diff
@@ -578,17 +581,29 @@ int main(int argc, char **argv) {
...
 
-    wakefiles.emplace_back(terminalReporter, i.c_str());
+  bool alerted_slow_cache = false;
+
+  for (size_t i = 0; i < wakefilenames.size(); i++) {
+    auto &wakefile = wakefilenames[i];
+    auto now = std::chrono::steady_clock::now();
+    if (std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() > 1000) {
+      std::cout << "Scanning " << i << "/" << wakefilenames.size()
+                << " wake files. Cache may be cold.\r" << std::flush;
```
JakeSiFive, line 592: Please say that its the "kernel file cache" to shift blame away from wake. People may assume that "cache" is a wake thing and will gripe at us if they do.

tools/wake/main.cpp
```diff
@@ -608,6 +623,11 @@ int main(int argc, char **argv) {
     }
   }
 
+  if (alerted_slow_cache) {
```
JakeSiFive, line 626: This isn't correct, you need the `i + 1` above instead

src/parser/wakefiles.cpp
```diff
@@ -170,8 +171,33 @@ bool push_packaged_stdlib_files(std::vector<std::string> &out, const re2::RE2 &r
...
+  if (fstatat(dirfd, f->d_name, &sbuf, AT_SYMLINK_NOFOLLOW) != 0) {
+    fprintf(stderr, "Failed to fstatat %s/%s: %s\n", path.c_str(), f->d_name, strerror(errno));
+    *failed = true;
+    return false;
+  }
+
+  return S_ISDIR(sbuf.st_mode);
+}
+
 static bool push_files(std::vector<std::string> &out, const std::string &path, int dirfd,
```
JakeSiFive, line 199: I know you didn't write this but if you could explain what it does and what "push" means it would be helpful.

tools/wake/main.cpp
V-FEXrt: the `i+1` isn't why this is needed.

The print above only occurs every 1 second, which means that we will almost never print the last 100 or so without this

JakeSiFive: Ah I see gotcha.

Another idea that we could do if you want: We could employ `\b` here to make this look really fancy btw. Then The message would just disappear instead of needing to be printed out.

src/parser/wakefiles.cpp
```diff
@@ -199,8 +199,8 @@ static bool push_files_is_directory(int dirfd, const std::string &path, struct d
   return S_ISDIR(sbuf.st_mode);
 }
 
-#Recursively fills *out *with all paths under *dirfd *matching *re *.Automatically
-#skips directories that cannot have source files in them(.git, .build, .fuse).
+// Recursively fills *out *with all paths under *dirfd *matching *re *.Automatically
```
JakeSiFive, line 202: What's with the asterisks in this and the above comments? They're confusing me.

JakeSiFive (approved):

src/parser/wakefiles.cpp
V-FEXrt: omg. They were right but the # "comment" plus the formatting moved them. 🤦 .

tools/wake/main.cpp
V-FEXrt: Personal preference is to leave the message instead of clearing it

PR #1079, opened 2022-12-08
Use npm ci to build in CI per the docs

V-FEXrt:
Per https://docs.npmjs.com/cli/v9/commands/npm-ci, use `npm ci` instead of `npm install`.

In the long term we may want a `make vscode-ci` and `make vscode-dev` but this unblocks CI for now and enables landing other important changes.

### DIFF
--- a/extensions/vscode/vscode.wake
+++ b/extensions/vscode/vscode.wake
@@ -63,7 +63,7 @@ export def vscode _: Result Path Error =
         setVersion (replace `~` "-" release) "{here}/package.json"
 
     require Pass nodeModules =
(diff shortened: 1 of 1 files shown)


richardxia (approved):

V-FEXrt:
Looks like the problem was initially triggered by the latest bump of the emscripten docker image. I've pinned the version here https://github.com/sifive/wake/pull/1080 to the most recent working one to enable landing some high prior PRs but its worth setting it back to pulling the latest as soon as I can fix the dependency issues with different versions. In the meantime, I'll still land the `npm ci` change raised here after the other change lands

PR #1080, opened 2022-12-08
check older emscripten image

V-FEXrt:


ag-eitilt (approved):
I'd hope we pay down this debt before we release the next version to be sure we catch any changes it might have on the extension, but this seems a reasonable enough short-term fix to get the other PRs in.  (Although, I do also feel that allowing the CI to always move forward even when it might break unrelated PRs might not be the best model to begin with; I think my ideal setup would be a pinned version in the required checks and an additional "latest" advisory check -- maybe even as a post-merge -- to let us know when we need to update the pins.)

richardxia (approved):
> I'd hope we pay down this debt before we release the next version to be sure we catch any changes it might have on the extension, but this seems a reasonable enough short-term fix to get the other PRs in. (Although, I do also feel that allowing the CI to always move forward even when it might break unrelated PRs might not be the best model to begin with; I think my ideal setup would be a pinned version in the required checks and an additional "latest" advisory check -- maybe even as a post-merge -- to let us know when we need to update the pins.)

100% second this. GitHub itself sometimes sends those advisory checks. I'm not sure if they scan Dockerfiles, especially when we've named them funny, but they do dependency scans a lot of other types of dependencies.

V-FEXrt:
I think the pinned + latest advisory is a good plan for us moving forward. Thanks for the suggestion! :)

PR #1082, opened 2022-12-14
Use latest emsdk

V-FEXrt:
Use latest emsdk and fix several bugs along the way

- directories shouldn't ever match as executables
- ENVIRONMENT should be `web` for wasm builds
- wasm build should explicitly use the emsdk's vendored node, not the system node
- future fix: the vsix file seems to have the wrong intermediate name


.circleci/dockerfiles/wasm
```diff
@@ -1,6 +1,6 @@
-FROM emscripten/emsdk:3.1.26
+FROM emscripten/emsdk:latest
```
richardxia, line 1: Since emscripten [doesn't follow semver or have any notion of stability for changes to even the third version component in the version string](https://github.com/emscripten-core/emscripten/blob/42a2de77143836ecbd17485aaf23fb3f182e7901/ChangeLog.md), should we just pin this to the latest fixed version rather than to `latest`? Otherwise, the next time they push a new version, it has the potential to break our build again.

extensions/vscode/vscode.wake
```diff
@@ -62,9 +62,16 @@ export def vscode _: Result Path Error =
     require Pass packageJSON =
         setVersion (replace `~` "-" release) "{here}/package.json"
 
+    require Pass npm =
+        whichInEnvPath "npm"
+        | getOrFail "Unable to find npm in $PATH".makeError
+
+    def emsdk = dirname npm
```
richardxia, line 69: Maybe name this `emsdkDir` given that `emsdk` is also the name of the executable provided by emscripten, just to avoid potential confunsion?

richardxia (approved):
A few minor comments, but this otherwise LGTM!

.circleci/dockerfiles/wasm
V-FEXrt: I was thinking about that, and I'm not sure. If we just pin to the lasted verified build then there is a high chance it'll go a long time without being updated then eventually it'll bitrot without us noticing. Then when we update next time it'll be a huge pain and may require several fixes for several versions

At least this way we get a loud alert when it breaks then we can land a quick fix to pin to the previous version while we fix the build.

I can see both arguments though

richardxia: Yeah, I think that makes sense, and at the end of the day, it's up to the team managing the project (i.e. not me) to decide which side they want to lean on, so this sounds good to me.

JakeSiFive: For the most part we use "latest" for things because we use the system version of libraries. I generally advocate for pinning/vendoring but since we already use latest we should be consistent. I also find Ashley's argument persuasive for things likely to fall by the wayside.

JakeSiFive (approved):

.circleci/dockerfiles/wasm
richardxia: I fully support your decision here, but I just had a thought pop into my head arguing in favor of pinning, and I figure I should mention it here in case if it affects your decision-making process. One downside to allowing the version to float to the latest version is that it can make hotfixes to older release branches more annoying to deal with. For continuously delivered software such as web apps, you don't usually manage release branches, but for many tools, you often do.

As a hypothetical scenario, suppose next month emscripten makes another change that causes wake's `master` branch builds to fail, which you fix on the `master` branch, but then six months from now, we find a critical bug in wake that we want to backport to the 0.29.x release series. When we make the backport, the CI builds will still fail because they will pull the latest emscripten, not the version that was used for the previous wake 0.29.x release, which may cause you to scramble to actually identify which version of emscripten was used in the last successful build of wake 0.29.x and then pin it on that branch.

Anyway, I think it's still reasonable to use `latest` as long as you're willing to deal with the scenario above or if you have other processes in place to deal with it (e.g. maybe part of the release branch ceremony would involve pinning the emscripten Docker image to a specific version, or even just removing the entire VS Code emscripten part of the wake CI build after a certain point).

JakeSiFive: backporting is a very strong argument for pinning

JakeSiFive: Maybe the rule should be that we only use "latest" if there's a stable interface and here there just isn't one? I certainly lean pinning in the vast majority of cases.

On Fuchsia we used pinning but had "auto rollers" that would automatically try to update to latest and then we could see if the auto rollers were ever failing. Best of both worlds. I wonder how hard it would be to set up an autoroller?

ag-eitilt: I think I mentioned in passing a month or two ago, but it might be nice to have the required-to-merge CI check be pinned to a known-working version, and have a secondary check (optional and/or lower-frequency) which runs the latest version.  It's missing the "auto" part of your autoroller, but does a similar job of alerting us when something breaks in future environments, with presumably less setup.

JakeSiFive: That's actually implementable with what we have today though! I like it! I vote we do that.

PR #1084, opened 2023-01-04
Add --single-fork option to wakebox to ensure fuse-waked exits before wakebox

JakeSiFive:
We've observed some unusual cases where the fuse mounts stay around and fuse-waked seems to have been killed already (I think this is as yet unverified in our case). Rather than searching for the bug in our signal handling and exit logic or digging into how these processes are being killed, Richard suggested that we might try `auto_unmount` so that if fuse-waked dies without unmounting for some reason, it should unmount anyway.

### DIFF
--- a/tools/fuse-waked/fuse-waked.cpp
+++ b/tools/fuse-waked/fuse-waked.cpp
@@ -1537,7 +1537,7 @@ int main(int argc, char *argv[]) {
    * This anti-feature was added in 2.4.0 and removed in 3.0.0.
    */
   if (fuse_opt_add_arg(&args, "wake") != 0 || fuse_opt_add_arg(&args, "-o") != 0 ||
-      fuse_opt_add_arg(&args, "nonempty") != 0) {
+      fuse_opt_add_arg(&args, "nonempty,auto_unmount") != 0) {
 #else
   if (fuse_opt_add_arg(&args, "wake") != 0) {
 #endif
--- a/tools/fuse-waked/fuse-waked.cpp
+++ b/tools/fuse-waked/fuse-waked.cpp
@@ -1401,14 +1401,27 @@ int main(int argc, char *argv[]) {
   int log, null;
   bool madedir;
   struct rlimit rlim;
+  bool daemonize = true;
 
-  if (argc != 3) {
-    fprintf(stderr, "Syntax: fuse-waked <mount-point> <min-timeout-seconds>\n");
+  if (argc != 4) {
+    fprintf(stderr,
+            "Syntax: fuse-waked <daemonize|no-daemonize> <mount-point> <min-timeout-seconds>\n");
     goto term;
   }
-  path = argv[1];
 
-  linger_timeout = atol(argv[2]);
+  if (strcmp(argv[1], "daemonize") == 0) {
+    daemonize = true;
+  } else if (strcmp(argv[1], "no-daemonize") == 0) {
+    daemonize = false;
+  } else {
+    fprintf(stderr,
+            "Syntax: fuse-waked <daemonize|no-daemonize> <mount-point> <min-timeout-seconds>\n");
+    goto term;
+  }
+
+  path = argv[2];
+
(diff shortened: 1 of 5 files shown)


tools/fuse-waked/fuse-waked.cpp
```diff
@@ -1537,7 +1537,7 @@ int main(int argc, char *argv[]) {
    * This anti-feature was added in 2.4.0 and removed in 3.0.0.
    */
   if (fuse_opt_add_arg(&args, "wake") != 0 || fuse_opt_add_arg(&args, "-o") != 0 ||
-      fuse_opt_add_arg(&args, "nonempty") != 0) {
+      fuse_opt_add_arg(&args, "nonempty,auto_unmount") != 0) {
 #else
   if (fuse_opt_add_arg(&args, "wake") != 0) {
```
JakeSiFive, line 1542: unclear if we should add auto_unmount here as well. This is sort of the catch all.

richardxia, line 1542: Should we also add it to this code path? Otherwise things might regress when we eventually upgrade the FUSE version, since the `#if` block above only targets a very narrow range of versions.

richardxia: Whoops, I my GitHub hadn't loaded your comment when I first opened the diff.

JakeSiFive: heh jynx again, I'm uncertain as my own comment on this states. I'd have to dig in the libfuse repo to see what the first version to support this is? I'm also down to just add it. I would imagine 2.9.7 is the oldest that we'll be dealing with.

richardxia: Not sure what the version number was, but it was first released in 2011, so I think we're probably fine: https://github.com/libfuse/libfuse/blob/d44bf3a4ac1ce1b56672ac9ae4a5caa3a8ee6844/ChangeLog#L223-L225

Although while digging through the changelog, I did find something that we should maybe investigate before merging this in: https://github.com/libfuse/libfuse/blob/d372d3f80b5437e652ea501d8a4a917f7001b9d7/ChangeLog.rst#libfuse-301-2017-04-10

> Document that -o auto_unmount implies -o nodev,nosuid.

I don't know off the top of my head if we rely on either of those things.

JakeSiFive:
```
Most of the generic mount options described in mount are
       supported (ro, rw, suid, nosuid, dev, nodev, exec, noexec, atime,
       noatime, sync, async, dirsync). Filesystems are mounted with
       nodev,nosuid by default, which can only be overridden by a
       privileged user.
   ```
   
Oh good catch but luckily it looks like that was already happening

richardxia: Oh, nice find. Yeah, this LGTM, then.

richardxia (approved):

V-FEXrt (approved):

JakeSiFive:
ok so the latest chnage adds the `--single-fork` option to wakebox and adds an option to deamonize or not-deamonize fuse-waked. This required some subtle tweaks here and there to make work. Additionally when `--single-fork` is used we call `waitpid` on the child process to ensure that it exits before wakebox does. In order to make that happen we have to close the live file before we waitpid or else we'll deadlock. At current the live file is only closed when wakebox closes which doesn't seem entirely right to me but its perfectly fine under our current uses. However with --single-fork we have to close it as the last operation on that mount.

My main complaint with this option is that its highly specific to a use case where a single wakebox invocation is run, and there's a race for fuse-waked to exit cleanly before everything in the waekbox cgroup is SIGKILLed when wakebox exits. It's a use case that we seem to need to support but its VERY specific.

JakeSiFive:
People affected by this issue can use `unshare -mr` with wakebox in order to work around it. This is a much simpler fix. Closing this issue.

PR #1087, opened 2023-01-11
Create stable and fyi builders for vscode target

V-FEXrt:
Creates a wasm-stable and wasm-latest dockerfile. Stable variant blocks CI with a known working version while latest uses the most up to date version potentially failing, but doesn't block CI.


.circleci/config.yml
```diff
@@ -25,13 +25,25 @@ jobs:
...
 
-  vscode:
+  vscode_latest:
     machine:
       image: ubuntu-2004:202201-02
     steps:
       - attach_workspace:
           at: /tmp/workspace
-      - run: docker build -f /tmp/workspace/.circleci/dockerfiles/wasm -t wake-wasm .
+      - run: docker build -f /tmp/workspace/.circleci/dockerfiles/wasm-stable -t wake-wasm .
```
ag-eitilt, line 34: Should this be `docker build -f .../wasm-latest`?

Also, this may just be me not knowing CircleCI, but where do we set which of these are required?  Is that a separate UI?

V-FEXrt: > Should this be docker build -f .../wasm-latest?
yep, bad copy+passte

> Also, this may just be me not knowing CircleCI, but where do we set which of these are required? Is that a separate UI?
working on that bit of code atm

.circleci/config.yml
```diff
@@ -220,7 +232,10 @@ workflows:
       - ubuntu_22_04:
           requires:
             - tarball
-      - vscode:
+      - vscode_latest:
+          requires:
+            - tarball
+      - vscode_stable:
```
JakeSiFive, line 238: We need someway to not block CI on these

JakeSiFive: ok nevermind, conversation offline indicates that for some reason `release` and `macos` are the only two CI jobs that block so by becoming a dep of `release` this blocks but since latest isn't a dep of `release` or `macos` it doesn't block...I'm not sure where the blocking nature of `release` and `macos` comes from but I'm happy enough with empirical evidence here (which Ashley tested)

.circleci/dockerfiles/wasm-stable
```diff
@@ -0,0 +1,7 @@
+FROM emscripten/emsdk:3.1.29
```
richardxia, line 1: In case you didtn' know, Dockerfiles do support an [`ARG`](https://docs.docker.com/engine/reference/builder/#arg) directive for supplying build-time arguments via [`docker build --build-arg`](https://docs.docker.com/engine/reference/commandline/build/#-set-build-time-variables---build-arg). It's up to you whether you want to have separate Dockerfiles anyway, since I could see some utility in that, especially if you needed to change other stuff in the Dockerfile (maybe the list of apt packages?), but I don't think it's a huge deal either way because this Dockerfile is small and we already have a plethora of Dockerfiles for the different Linux distros.

richardxia (approved):

.circleci/dockerfiles/wasm-stable
V-FEXrt: Done!

PR #1088, opened 2023-01-13
Add ability to emit compile commands for wake code

JakeSiFive:
This change adds a "simple-enough" means to create a compile_commands.json file for the wake repo. The process is as follows

```
rm -rf compile_commands # Make sure we remove old compile commands
EMIT_COMPILE_COMMANDS=1 make # It shouldn't matter if the build is clean, this also works with `wake build <variant>`
python3 ./scripts/compile_commands.py
```

...wait I just realized that that always works so I think I'm going to add a bash script too

### DIFF
--- a/share/wake/lib/gcc_wake/gcc.wake
+++ b/share/wake/lib/gcc_wake/gcc.wake
@@ -63,7 +63,6 @@ def doCompileC variant gcc flags headers cfile =
 
     def get f = prim "get_hash"
     def path_hash = get (cfile.getPathName)
-    #require Pass _ = mkdir "compile_commands"
     def cmd_file_path = "compile_commands/{path_hash}.json"
 
     require Pass _ = write cmd_file_path (prettyJSON outJson)
--- a/share/wake/lib/gcc_wake/gcc.wake
+++ b/share/wake/lib/gcc_wake/gcc.wake
@@ -25,15 +25,50 @@ export def releaseLFlags = (Nil)
 export def staticCFlags =  ("-Wall", "-Wno-format-security", "-O2", "-flto", Nil)
 export def staticLFlags  = ("-flto", "-static", "-s", Nil)
 
+target pwd _ =
+    makeExecPlan ("pwd",) Nil
+    | setPlanLocalOnly True
+    | runJob
+    | getJobStdout
+
 def doCompileC variant gcc flags headers cfile =
     def obj = replace `\.c(pp)?$` ".{variant}.o" cfile.getPathName
     def cmdline =
         gcc, flags ++
         ("-c", cfile.getPathName, "-frandom-seed={obj}", "-o", obj, Nil)
-    makeExecPlan cmdline (cfile, headers)
-    | setPlanLabel "compile {replace `^.*/` '' gcc} {obj}"
-    | runJob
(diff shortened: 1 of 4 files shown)


JakeSiFive:
ok now you just run `./scripts/compile_commands` and it builds it for you, hopefully without fail or mishap. Using the hashes of the files should greatly reduce any possible errors that could occur since only deleting or moving a file would cause a slightly wrong compile_commands file to be emitted (it would just have an extra entry for a file that no longer existed which most editors should tolerate)

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -25,15 +25,50 @@ export def releaseLFlags = (Nil)
 export def staticCFlags =  ("-Wall", "-Wno-format-security", "-O2", "-flto", Nil)
 export def staticLFlags  = ("-flto", "-static", "-s", Nil)
 
+target pwd _ =
+    makeExecPlan ("pwd",) Nil
+    | setPlanLocalOnly True
```
V-FEXrt, line 30: Is there a specific reason this needs LocalOnly?

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -25,15 +25,50 @@ export def releaseLFlags = (Nil)
...
+
+    def outJson = JObject (
+        "directory" :-> JString dir,
+        "arguments" :-> JArray $ map JString cmdline,
+        "file" :-> JString "cfile.getPathName",
+    )
+
+    def get f = prim "get_hash"
+    def path_hash = get (cfile.getPathName)
+    #require Pass _ = mkdir "compile_commands"
```
V-FEXrt, line 66: nit: commented code

V-FEXrt (approved):
I'm testing locally now but LGTM with a question and a nit

V-FEXrt:
Actually, if I can be a bit more annoying, Is it possible to put compile_commands and maybe compile_commands.json under `.build`?

share/wake/lib/gcc_wake/gcc.wake
JakeSiFive: If it isn't the pwd will be from inside the workspace so it will always emit `/workspace` which isn't what we want here. I was hopeful that compile_commands.json would allow relative paths but the spec doesn't say one way or the other and it only gives absolute paths as an example so I needed an absolute path, otherwise `"."` would have worked

JakeSiFive:
> Actually, if I can be a bit more annoying, Is it possible to put `compile_commands/` and maybe `compile_commands.json` under `.build`?

I can make the directory hidden if you want but I don't think they should go in .build, that's only used by wakebox at this time. The compile commands spec recomends that `compile_commands.json` be at the top level of the build so its where the spec recommends it go now.

V-FEXrt:
okay, everything works out of the box with `compile_commands.json` at the root, so maybe shouldn't mess with that but would be good to hide `compile_commands/`

richardxia:
Thank you for doing this thing that I asked for in a random comment of a PR that, to be fair, had a slightly different scope: https://github.com/sifive/wake/pull/815#issuecomment-971871432

PR #1089, opened 2023-01-17
Warn on unused import

V-FEXrt:
Emit an LSP warning on unused import. Right now it only works for qualified imports but should be possible to support import all imports.

One thing I'm not sure about is how we roll out to existing users without causing a bunch of warnings? Maybe we are okay with doing that? Not sure if many folks use "warn as fail" or not.

### DIFF
--- a/src/dst/bind.cpp
+++ b/src/dst/bind.cpp
@@ -1183,17 +1183,18 @@ static std::unique_ptr<Expr> fracture(std::unique_ptr<Top> top) {
 
       for (auto &import : file.content->imports.topics) {
         const std::string &qualified = import.second.qualified;
+        const std::string &topic_qualified = "topic " + import.second.qualified;
         const Location &location = import.second.fragment.location();
 
         filename = location.filename;
-        imports.insert({qualified, {0, location}});
+        imports.insert({topic_qualified, {0, location}});
 
         std::size_t at_pos = qualified.find("@");
         if (at_pos == std::string::npos) {
           continue;
         }
 
-        unqualified_to_qualified.insert({qualified.substr(0, at_pos), qualified});
+        unqualified_to_qualified.insert({qualified.substr(0, at_pos), topic_qualified});
       }
 
       // File doesn't have any import for us to verify
@@ -1220,7 +1221,7 @@ static std::unique_ptr<Expr> fracture(std::unique_ptr<Top> top) {
       }
 
       for (auto &publish : file.pubs) {
-        // If the publish was an imported publish, mark it as used
+        // Publishing to an imported topic should mark the topic as used
         auto qual_it = unqualified_to_qualified.find(publish.first);
         if (qual_it == unqualified_to_qualified.end()) {
           continue;
--- a/src/dst/bind.cpp
+++ b/src/dst/bind.cpp
@@ -1263,9 +1263,8 @@ static std::unique_ptr<Expr> fracture(std::unique_ptr<Top> top) {
           continue;
         }
 
-        WARNING(import.second.second, "unused import of '" << import.first
-                                                           << "'; consider removing. ("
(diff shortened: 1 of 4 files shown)


V-FEXrt:
Capture22

src/dst/bind.cpp
```diff
@@ -882,18 +882,24 @@ static std::unique_ptr<Expr> fracture(Top &top, bool anon, const std::string &na
     // don't fail if unbound; leave that for the second pass
     rebind_ref(binding, ref->name, ref->fragment);
     return expr;
-  } else if (expr->type == &Subscribe::type) {
+  }
```
JakeSiFive, line 885: Why change this at all?

src/dst/bind.cpp
```diff
@@ -1148,6 +1194,26 @@ static std::unique_ptr<Expr> fracture(std::unique_ptr<Top> top) {
...
+    }
+
+    auto &i = gbinding.defs[it->second];
+
+    // Generated defs shouldn't be included
+    if (i.expr->flags & FLAG_SYNTHETIC) {
+      continue;
+    }
+
+    if (i.uses == 1) {
```
JakeSiFive, line 1212: Is the reason `uses` has 1 and not zero because the import "uses" it? If so I think a comment should be added here explaining that.

src/dst/bind.cpp
V-FEXrt: ah, it was an unrelated cleanup. Was trying to figure out what the code did and this was de-cluttering so I could think better/understand what was going on

src/dst/bind.cpp
V-FEXrt: Hmm, I though I had left a comment already, it is `1` here because anything _exported_ (which is everything imported) gets it usage incremented by 1 by virtue of being exported.

 Ignoring this PR,  if we don't increment all exports by 1 then any unused export triggers a warning

V-FEXrt:
Hmm, poking around this may not be sufficient actually. I just tried something that should have triggered a warning but didn't. I think `uses` may be a global count not a file local one. If so, it only actually warns when an import is unused _anywhere_

JakeSiFive:
I wonder if setting "uses" to 1 on exported things was just a hack that Wes added to make things not warn on exported functions

V-FEXrt:
> I wonder if setting "uses" to 1 on exported things was just a hack that Wes added to make things not warn on exported functions

Mentioned it above, but if you don't set 1, then unused exported functions will warn with "function is unused" which is not desirable for exported functions since you can't know if/when and exported lib function may be used

JakeSiFive:
> Mentioned it above, but if you don't set 1, then unused exported functions will warn with "function is unused" which is not desirable for exported functions since you can't know if/when and exported lib function may be used

Right, I read what you wrote but I was pointing out that that is a hack and not really a "use".

V-FEXrt:
Ah, I see. Yeah you could certainly argue that it is a hack. I don't know of any other reasons it is done that way.

V-FEXrt:
Okay! This is ready for a review! PTAL

src/dst/bind.cpp
```diff
@@ -1148,6 +1150,137 @@ static std::unique_ptr<Expr> fracture(std::unique_ptr<Top> top) {
     }
   }
 
+  for (auto &package : top->packages) {
+    for (auto &file : package.second->files) {
+      std::map<std::string, std::pair<int, Location>> imports = {};
```
JakeSiFive, line 1155: You should document what the pair does or replace it with a small struct. You can define a struct inline inside this function I think. It wasn't obvious to me that the integer was a use count.

JakeSiFive (approved):
This seems much better! It's quite clean. I just requested a couple comments but I think it makes sense from there.

PR #1092, opened 2023-01-18
Mirror CircleCI workflow as a GHActions workflow

V-FEXrt:
This PR moves all of our CircleCI jobs over to GH Actions. It doesn't delete any of the CircleCI jobs for now. I want to punt that until later in case something goes wrong and we need to revert. So for an amount of time, we will have CI jobs in both.

I have turned off the doc deploy step in CircleCI however as both GHActions and CircleCI would be fighting over the single resource.

Main changes
 - All builds were ported to GH Actions 
 - Doc deploy was turned off in CircleCI and recreated in GHActions
 - A *new* job was created that will automatically build and create a draft release whenever a tag is created. This should make it easier to create new wake releases.


JakeSiFive:
I'll prioritize reviewing this today

.github/workflows/build.yaml
```diff
@@ -1,26 +1,648 @@
...
+          uses: actions/download-artifact@v3
+          with:
+            name: dockerfiles
+            path: dockerfiles
+
+        - name: Uncompress tarball
+          run: tar xvJf build/wake_*.tar.xz -C build
+
+        - name: Set tests as executable
+          run: chmod -R +x build/tests
```
richardxia, line 230: Just trying to think of how to DRY things up, and for this step, why do we need to do this after extracting the artifact? I don't know how GitHub's artifacts work under the hood, but normally archive file formats (e.g. tar) should preserve the executability bit.

richardxia (approved):
I didn't review this in great detail, since I assume that the tricky bits with the exact `docker` commands were copies verbatim from the CircleCI jobs, but this overall LGTM!

Upon first glance, it feels like there's a lot of repeated boilerplate, but I don't think there's really a good way to avoid it because there are just enough little differences between the different distros that would defeat any benefit from refactoring.

One last nit-picky comment: The title of the PR is misleading, since "Actions" and "Workflows" are different kinds of things in GH Actions. What you've changed here is a Workflow, and a Custom Action would be more like a reusable step that you call from the workflow.

V-FEXrt:
>  The title of the PR is misleading, since "Actions" and "Workflows" are different kinds of things in GH Actions

yeah, my bad. This PR started out very different and I forgot to update the title

.github/workflows/build.yaml
V-FEXrt: > I don't know how GitHub's artifacts work under the hood, but normally archive file formats (e.g. tar) should preserve the executability bit.

It gets lost in in the upload flow :(, its a common enough question that it has a top level note in the documentation. https://github.com/actions/upload-artifact#permission-loss

richardxia: I had suspected and was afraid you were going to say that :(

V-FEXrt: I do want to DRY here if possible.

The best option is via a `matrix` which basically repeats a job but with different parameters. And its _almost_ here with the following params.

- docker image
 - build command
 - test command
 - artificats glob
 - install destination
 - upload destination
 
 We could certainly just do extra steps like adding the tests/debian dirs even when they aren't needed to cover the differences, but with 6 parameters its hard for me to decided if its worth it. I'm personally leaning towards yes

richardxia: If you're willing to spend time to DRY it up, I'd support that effort, since with the 7 or so different environments, I think it'd be worth it. But if it ends up being too difficult, I don't think it would be the worst thing to leave things as they are right now.

V-FEXrt: This should be done now!

.circleci/config.yml
```diff
@@ -251,9 +251,4 @@ workflows:
             - ubuntu_22_04   # LTS
             - vscode_stable
             - wasm
-      - docs_deploy:
-          requires:
-            - release
-          filters:
-            branches:
-              only: master
+
```
ag-eitilt, line 254: If we weren't about to delete this file, I'd be more concerned about the blank line at the end. 😉

.github/workflows/build.yaml
```diff
@@ -1,26 +1,363 @@
...
+          run: touch .nojekyll
+
+        - name: Upload
+          uses: actions/upload-pages-artifact@v1
+          with:
+            path: .
+
+    deploy_docs:
+      # Only run the 'deploy_docs' job if this workflow was triggered by a push to master
+      if: github.event_name == 'push' && github.ref_type == 'branch' && github.ref_name == 'master'
```
ag-eitilt, line 266: I assume this catches PRs being merged as well?

.github/workflows/build.yaml
```diff
@@ -1,26 +1,363 @@
...
+      needs: docs
+      name: Deploy Docs to GH Pages
+      steps:
+        - name: Deploy
+          id: deployment
+          uses: actions/deploy-pages@v1
+
+    release:
+      # Only run the 'release' job if this workflow was triggered by the creation of a tag
+      if: github.event_name == 'push' && github.ref_type == 'tag'
```
ag-eitilt, line 286: I foresee a need to update the "Make a release" workflow docs.  That's probably a good thing; this *feels* like it should allow a bit easier releases.

.github/workflows/build.yaml
```diff
@@ -1,26 +1,363 @@
...
+          uses: actions/download-artifact@v3
+          with:
+            name: release_vscode
+            path: vscode
+
+        - name: Download Alpine
+          uses: actions/download-artifact@v3
+          with:
+            name: release_alpine
+            path: apline
```
ag-eitilt, line 308: I don't see any typos beyond this one.
```suggestion
            path: alpine
```

.github/workflows/build.yaml
```diff
@@ -1,26 +1,363 @@
...
+      runs-on: ubuntu-latest
+      needs: tarball
+      name: Docker Build (${{ matrix.target }})
+      strategy:
+        matrix:
+          include:
+            - target: alpine
+              dockerfile: alpine
+              extra_docker_build_args:
+              extra_docker_run_args:
```
ag-eitilt, line 90: The syntax highlighter is complaining about these being empty; I imagine the keys are in here because you want them here, but would explicit empty strings break anything?  (It fixes the highlighter, but I'm not sure about the Workflows parser.)
```suggestion
              extra_docker_build_args: ""
              extra_docker_run_args: ""
```

ag-eitilt (commented):
Don't want to block this in case I don't see the fix for a bit, but there are a couple, very small fixes which you might want to make.  On the whole, though, this looks promising!  (I've not seen what it looks like in the CI interface yet.)

.github/workflows/build.yaml
V-FEXrt: Right so the whole workflow triggers on any of 

1) PR opened
2) commit pushed to master
3) Tag starting with a `v` created and pushed

The filter for this specific job says only run when a commit is pushed to master, and since we only let commits land on master via a PR this runs right after a PR is merged

.github/workflows/build.yaml
V-FEXrt: Yep, I'm planning on making the update after this lands. I need to land this before update the docs to make sure I get the details right since I can only test the full flow after it lands,

V-FEXrt:
I'm going to remove `MacOS build` as required since it was renamed to `MacOS 10.15 Build`. Once this lands on master, I'll re-add it as required with the correct name. That was I can land this without blocking other PRs in the meantime

.github/workflows/build.yaml
```diff
@@ -1,26 +1,366 @@
...
+          with:
+            name: tarball
+            path: wake_*.tar.xz
+
+        - name: Upload Dockerfiles
+          uses: actions/upload-artifact@v3
+          with:
+            name: dockerfiles
+            path: .circleci/dockerfiles/
+            retention-days: 1
```
JakeSiFive, line 50: What are retention-days?

.github/workflows/build.yaml
```diff
@@ -1,26 +1,366 @@
...
+          run: make test && make unittest
+
+        - name: Generate Docs
+          run: mkdir www && ./bin/wake --no-workspace --html > www/index.html
+
+        - name: wake to rst
+          run: PATH=$(pwd)/bin:$PATH PYTHONPATH=$(pwd)/scripts python3 scripts/wake2rst.py
+
+        - name: Upload tarball
+          uses: actions/upload-artifact@v3
```
JakeSiFive, line 40: What does the upload-artifact action do?

.github/workflows/build.yaml
```diff
@@ -1,26 +1,366 @@
...
+    release:
+      # Only run the 'release' job if this workflow was triggered by the creation of a tag
+      if: github.event_name == 'push' && github.ref_type == 'tag'
+
+      runs-on: ubuntu-latest
+      needs: [tarball, docker_builds]
+      name: Create Release
+      steps:
+        - name: Download Tarball
+          uses: actions/download-artifact@v3
```
JakeSiFive, line 293: what does download-artifact do?

.github/workflows/build.yaml
```diff
@@ -1,26 +1,366 @@
...
+            TAG="${{ github.ref_name }}"
+            VERSION="${TAG:1}"
+            cp centos_7_6/wake-*-1.x86_64.rpm centos-7-6-wake-${VERSION}-1.x86_64.rpm
+            cp rocky_8/wake-*-1.x86_64.rpm rocky_8-wake-${VERSION}-1.x86_64.rpm
+            cp debian_bullseye/wake_*-1_amd64.deb debian-bullseye-wake_${VERSION}-1_amd64.deb
+            cp ubuntu_18_04/wake_*-1_amd64.deb ubuntu-18-04-wake_${VERSION}-1_amd64.deb
+            cp ubuntu_22_04/wake_*-1_amd64.deb ubuntu-22-04-wake_${VERSION}-1_amd64.deb
+
+        - name: Create Release
+          uses: marvinpinto/action-automatic-releases@latest
```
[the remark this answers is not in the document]
JakeSiFive: MIT license and yeah this is someone else's action. I reviewed the action code and it looks fine and seems to have a lot of users but I think we should consider keeping our own fork of this

.github/workflows/build.yaml
V-FEXrt: number of days the artifact is kept before being deleted. Default (and max) is 90. I set it to 1 for artifacts that are only used as intermediates for other steps to be a good steward

.github/workflows/build.yaml
V-FEXrt: uploads the contents of a folder or path to a key-value store that is a new store for each workflow run

.github/workflows/build.yaml
```diff
@@ -1,26 +1,366 @@
...
+              extra_docker_build_args: ''
+              extra_docker_run_args: -u build
+              build_cmd: make -C wake-* wasm
+              test_cmd: ls # no-op
+              # TODO: This doesn't fully capture all the wake contents
+              # but that is okay since wake-wasm in still in its infancy
+              # and is not deployed/depended on
+              install_src_glob: build/wake-*/bin/*.wasm*
+
+      steps:
```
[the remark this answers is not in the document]
V-FEXrt: yeah, old commit where each of this were one a separate job

.github/workflows/build.yaml
```diff
@@ -1,26 +1,366 @@
...
+        - name: Install
+          run: install -D -t release/${{ matrix.target }} ${{ matrix.install_src_glob }}
+
+        - name: Upload
+          uses: actions/upload-artifact@v3
+          with:
+            name: release_${{ matrix.target }}
+            path: release/${{ matrix.target }}
+
+    macos_10_15:
```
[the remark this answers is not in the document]
V-FEXrt: Unclear. This is just the rule we had before. I don't really want to hit on that with this PR and we should review this separately

.github/workflows/build.yaml
V-FEXrt: reverse of upload-artifact

.github/workflows/build.yaml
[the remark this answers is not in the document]
JakeSiFive: Sure sounds good. Worst case can't be any worse than what we had before

.github/workflows/build.yaml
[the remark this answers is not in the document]
JakeSiFive: Ashley and I spoke offline about this. I'm still in favor of forking but if enough others or if Matthew feels that a commit hash is an acceptable means of pinning the action code then I'm down. If we're not down to fork or if we can't reach consensus on this soon enough we'll just remove this code and fall back to manually uploading like we were before. This would be a nice plus though.

.github/workflows/build.yaml
ag-eitilt: > > I don't know how GitHub's artifacts work under the hood, but normally archive file formats (e.g. tar) should preserve the executability bit.
> 
> It gets lost in in the upload flow :(, its a common enough question that it has a top level note in the documentation. https://github.com/actions/upload-artifact#permission-loss

[A few sections down](https://github.com/actions/upload-artifact#maintaining-file-permissions-and-case-sensitive-files), it looks like that might only apply to the raw uploads, and making a tarball is actually given as a solution for maintaining the permissions.  I'm sure you've included this line as a reaction to trying and failing to run the tests without it, so there's probably something else going on which is *also* stripping those permissions before the files get packaged up, but that just underscores how complex this whole remote-execution-via-ostensibly-data-files system is.  (Very handy though, no matter its foibles!)

.github/workflows/build.yaml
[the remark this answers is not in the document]
JakeSiFive: Further research reveals that git is using SHA-1DC which is somehow hardened against the SHA1 attacks but is otherwise compatible and at present has no publicly known attacks against it. So a commit hash is probably secure enough for our purposes.

JakeSiFive (approved):
I don't know what the latest error is about but this all looks good to me now once its working

.github/workflows/build.yaml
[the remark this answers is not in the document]
V-FEXrt: Done! PTAL!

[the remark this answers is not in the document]
V-FEXrt: nvm, just saw the most recent comment. Thanks!

.github/workflows/build.yaml
V-FEXrt: yeah adding it was in response to a failed test. We can explore improving it later, but it seems like a small issue and I'd like to get this landed to minimize the time with the CI in a weird state.

PR #1094, opened 2023-01-25
Fix GH Action Release Job

V-FEXrt:
I was using bad path matching which caused the artifacts to be missed. This fixes that.


ag-eitilt (approved):
Rubber stamp.

PR #1095, opened 2023-01-25
Release 0.29.2

V-FEXrt:
Prepare for the 0.29.2 release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,19 @@ jobs:
             - tests
             - scripts/sphinx/build/html
 
-  vscode:
+  vscode_latest:
+    machine:
+      image: ubuntu-2004:202201-02
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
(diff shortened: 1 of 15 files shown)


ag-eitilt (approved):
My "Merge from v0.29.1-prep" or whatever branch it is at the top of the commit list was my own experiments with a new release flow: by making the Debian changelog release and similar administrative changes in a separate `*-prep` branch which then gets *merged* into the `release-*`, we can also merge that same branch back into `master` and not have situations like we had a while back where the changelog doesn't seem to have been updated for the past four releases simply because the updates never made it out of the `v*` branches.  I wasn't able to get the process for doing so quite smooth enough to write up, but it might be something to consider as you update the release instructions for the new CI.

V-FEXrt:
I'm actually going to propose something more slightly radical in the meeting today :)

V-FEXrt:
Okay! decision approved, we are getting rid of release branches. Going to axe the PR and start the process over from the beginning with the new flow

PR #1099, opened 2023-01-27
Make the version tag safe to use in node

V-FEXrt:
Landing the non `alpha` tag on `master` broke how versions were generating for non-release CI builds. This fixes it so that node/npm doesn't fail to build in those cases by inserting `safe` where `alpha` normally goes if it is missing

### DIFF
--- a/vendor/git.wake
+++ b/vendor/git.wake
@@ -62,20 +62,20 @@ def buildOn Unit = match (subscribe releaseOn)
 #       ^^^
 # which breaks node's version requiement and is considered invalid
 #
-# Capture the parts and insert '-post-' between the version and the commit. 
+# Capture the parts and insert '-plus-' between the version and the commit. 
 # Do nothing if the version is already safe.
 def mk_node_semver_safe v =
-    def unsafe_version_regex = `(v?[0-9]+\.[0-9]+\.[0-9]+)-([0-9]+)-(.*)`
-    #                           ^^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^  ^^^^
-    #                           |                          |        |
-    #                           |                          |        | The "rest", everything after the '-' after the commit number
-    #                           |                          | The number of commits since the tagged version, surrounded by '-'
-    #                           | Matches the version tag, which should be 3 numbers separated by '.' with an optional leading 'v'
+    def unsafe_version_regex = `^(v?[0-9]+\.[0-9]+\.[0-9]+)-([0-9]+)-(.*)`
+    #                            ^^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^  ^^^^
+    #                            |                          |        |
+    #                            |                          |        | The "rest", everything after the '-' after the commit number
+    #                            |                          | The number of commits since the tagged version, surrounded by '-'
+    #                            | Matches the version tag, which should be 3 numbers separated by '.' with an optional leading 'v'
     require True = matches unsafe_version_regex  v
     else v
 
     match (extract unsafe_version_regex v)
         version, commit, rest, Nil =
-            (version, "-post-", commit, "-", rest,)
+            (version, "-plus-", commit, "-", rest,)
             | cat
         _ = v
--- a/vendor/git.wake
+++ b/vendor/git.wake
@@ -20,26 +20,6 @@ topic releaseAs: String
 def buildAs Unit = match (subscribe releaseAs)
     version, Nil = Pass version
     _ =
-        # Process the version when the version number and commit count are adjacent.
-        # v0.15.0-4-g6efe8e9
-        #       ^^^
-        # Without intervention this becomes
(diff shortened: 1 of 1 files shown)


V-FEXrt:
` sed "s/@VERSION@/0.29.2-safe.2+g3f4c102b/g" "extensions/vscode/package.json.in" > "extensions/vscode/package.json.tmp"`
CI job reflect the expected change as well

vendor/git.wake
```diff
@@ -20,7 +20,28 @@ topic releaseAs: String
...
+        #       ^^^
+        # which breaks node's version requiement and is considered invalid
+        #
+        # Capture the two halves 'v0.15.0', '4-g6efe8e9' and insert '-safe-' in the middle
+        # Do nothing if the version is already safe.
+        def mk_node_semver_safe v =
+            require True = matches `.+[0-9]-[0-9].+` v
+            else v
+
+            match (extract `(.+[0-9])-([0-9].+)` v)
```
JakeSiFive, line 37: 1) You can reuse the same regex above
2) This regex is allowed to return `v0.15.0-4` for the first group and `g6efe8e9` for the second group which as I understand it is not what you want. I think you can fix this by anchoring the start of the regex.

JakeSiFive (changes requested):

vendor/git.wake
ag-eitilt: Even then, I think the `.+` could still swallow the dash.  It probably needs `[^-]+`, or be explicit about NPM's requirements and do `(v?[0-9]+\.[0-9]+\.[0-9]+)` (is the `v` required as well?)

ag-eitilt (commented):
I'll let Jake's "Request changes" stand for the regex fixes, rather than doubling up.

vendor/git.wake
JakeSiFive: Yeah you're right since `.+` could eat the `-` as long as there are two hyphens and a number between them (which is the case it seems). Going fully explicit (your way) is one way to solve it, or the lazy way would be to replace `.` with `[^\-]` to avoid the first hyphen being consumed. Which way we go depends on the variance we expect and that is demanded by node.

Ideally the first part should be fully explicit like Sam's regex to what node expects and assuming node allows anything after `-safe` or `-post` or whatever to be any other string then that part should be explicit to what we expect to receive from git. If we don't have a clear guideline on what git will give us then that part should be generic (e.g. `.+`)

vendor/git.wake
```diff
@@ -20,7 +20,28 @@ topic releaseAs: String
...
     _ =
+        # Process the version when the version number and commit count are adjacent.
+        # v0.15.0-4-g6efe8e9
+        #       ^^^
+        # Without intervention this becomes
+        # v0.15.0.4-g6efe8e9
+        #       ^^^
+        # which breaks node's version requiement and is considered invalid
+        #
+        # Capture the two halves 'v0.15.0', '4-g6efe8e9' and insert '-safe-' in the middle
```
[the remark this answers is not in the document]
V-FEXrt: `post` is fine. I was just trying to make it obvious that something was inserted and this wasn't the normal flow

V-FEXrt:
> (is the v required as well?)
Not for node, but we enforce it by convention

vendor/git.wake
```diff
@@ -51,3 +53,29 @@ def buildOn Unit = match (subscribe releaseOn)
...
+#       ^^^
+# Without intervention this becomes
+# v0.15.0.4-g6efe8e9
+#       ^^^
+# which breaks node's version requiement and is considered invalid
+#
+# Capture the parts and insert '-post-' between the version and the commit. 
+# Do nothing if the version is already safe.
+def mk_node_semver_safe v =
+    def unsafe_version_regex = `(v?[0-9]+\.[0-9]+\.[0-9]+)-([0-9]+)-(.*)`
```
JakeSiFive, line 68: nit: I'd still anchor this to make possible bugs less confusing in the future.

JakeSiFive (approved):

vendor/git.wake
```diff
@@ -51,3 +53,29 @@ def buildOn Unit = match (subscribe releaseOn)
...
+    #                            |                          |        |
+    #                            |                          |        | The "rest", everything after the '-' after the commit number
+    #                            |                          | The number of commits since the tagged version, surrounded by '-'
+    #                            | Matches the version tag, which should be 3 numbers separated by '.' with an optional leading 'v'
+    require True = matches unsafe_version_regex  v
+    else v
+
+    match (extract unsafe_version_regex v)
+        version, commit, rest, Nil =
+            (version, "-plus-", commit, "-", rest,)
```
V-FEXrt, line 79: Renamed this to `plus` so the version name "reads" correctly
`v0.15.0-plus-4-g6efe8e9`

JakeSiFive (approved):

PR #1100, opened 2023-01-28
Suppress slow disk warning when --quiet is set

V-FEXrt:
I'm not sure that I'm happy with this implementation, but I don't know how else to do it.

The MacOS builder is running slow enough that the warning is emitted to stdout and causing the tests to fail (however only on release build for some reason). I wanted to nip this in the bud before it became a larger issue.

Since the tests all set the -q flag, this will suppress the warning from triggering during the tests.

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -570,7 +570,7 @@ int main(int argc, char **argv) {
   bool enumok = true;
   std::string libdir = make_canonical(find_execpath() + "/../share/wake/lib");
   auto wakefilenames =
-      find_all_wakefiles(enumok, workspace, verbose, libdir, ".", quiet ? stdout : stderr);
+      find_all_wakefiles(enumok, workspace, verbose, libdir, ".", user_warn);
   if (!enumok) {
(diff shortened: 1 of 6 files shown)


tools/wake/main.cpp
```diff
@@ -553,15 +553,25 @@ int main(int argc, char **argv) {
...
+  auto wakefilenames =
+      find_all_wakefiles(enumok, workspace, verbose, libdir, ".", user_warn);
   if (!enumok) {
     if (verbose) std::cerr << "Workspace wake file enumeration failed" << std::endl;
     // Try to run the build anyway; if wake files are missing, it will fail later
     // The unreadable location might be irrelevant to the build
   }
 
+  if (quiet) {
+    fclose(user_warn);
```
JakeSiFive, line 572: We really badly need `defer`, I think I'll go implement it real quick

JakeSiFive (approved):

JakeSiFive (approved):
Thanks for fixing my mistakes 😬

PR #1101, opened 2023-01-30
add wcl::defer

JakeSiFive:
I think I'm going to modify this so that the function inside is a wcl::optional. This will give us the desired move semantics by default AND will allow default construction....still I'm trying think of how to modify this so that you don't have to use std::function<void()> and yet you can still declare it. Ideally I want to capture this pattern

```
wcl::defer opt_resource_defer;
ResourceType optional_resource;
if (my_condition_here) {
  optional_resource = make_the_resource;
  opt_resource_defer = wcl::make_defer([&]() {
    clean_up_resource(optional_resource);
  });
}
```

### DIFF
--- a/src/wcl/defer.h
+++ b/src/wcl/defer.h
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2022 SiFive, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You should have received a copy of LICENSE.Apache2 along with
+ * this software. If not, you may obtain a copy at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


JakeSiFive:
Ok so this adds two ways to use defer

simple case:
```
Resource my_resource = make_resource();
auto my_resource_defer = wcl::make_defer([&]() {
  clean_up_resource(my_resource);
});
```

slow but easy conditional case
```
Resource my_resource;
wcl::opt_defer my_resource_defer;
if (my_condition) {
  my_resource = make_resource();
  my_resource_defer = [&]() {
    clean_up_resource(my_resource);
  };
}
```

fast but hard conditional route
```
Resource my_resource = make_resource();
auto my_resource_defer = wcl::make_defer([&]() {
  clean_up_resource(my_resource);
});
if (my_condition) {
  my_resource = make_resource();
} else {
  my_resource_defer.nullify();
}
```

V-FEXrt (approved):
Big QOL addition! Thanks!

PR #1102, opened 2023-01-31
Explicitly call into add_bool instead of relying on implict choice by compiler

V-FEXrt:
Some of the json functions in the LSP code were incorrectly using the new bool json function do to implict operator overloading.

This broke the LSP; the change below makes bool `add`s explicit


JakeSiFive (approved):
Glad we found this subtle bug! That was awesome teamwork!

JakeSiFive:
I'm gonna squash and merge for you because I'm working on shared caching right now

PR #1105, opened 2023-02-01
Create logging.md

mwachs5:
I was having trouble understanding how wake logging worked. While the docs on the functions are pretty good, having a high level overview helped me understand a lot better. Hoping others might benefit.

### DIFF
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -37,11 +37,8 @@ It also controls which *actual* stream it is going to coming out of `wake`
 
 There are a number of built in loggers in wake which all start with `log...`.
 You can also make your own logger using `mkLogLevel`.
-You can use this to customize the color you want it to be shown as,
-or (more commonly) to grab output from a `Job` for consuming within your wake program,
-or redirecting outside of wake using shell redirect.
-
-
+You can use this to customize the color you want it to be shown as (if at all),
+and giving more fine-grained control over which wake output streamit should go to.
 
 For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
 If you match a line in the table, stop!
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -9,7 +9,7 @@ for commented code.
 Every `Job` will emit a stderr and a stdout.
 
 When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
-We call this which `stream` each goes to (we often refer to each `stream` as a `logLevel`).
+We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
 
 Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
 `wake` will store all the text in its database,
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -11,8 +11,8 @@ Wake has a concept of loggers,
 to which text can be written during wake execution by various sources.
 A logger combines a string log level (e.g. "error", "info", "my_custom_level")
 and a display format (color and intensity).
-There can be multiple loggers with the same log level, though this is
-not a common use case.
+There can be multiple loggers with the same log level,
+though colors may not display as expected.
 
 |Term Used In this Document | What wake code/libraries call it |
 ----------------------------|----------------------------------|
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -1,44 +1,30 @@
-# Logging
+# Logging & Console Output
 
-This article gives a high level overview of how wake logging works.
+This article gives a high level overview of how wake logging works,
+and other things that impact what you see on the console when wake is executing or running database commands.
 Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
 for commented code.
 
-## Things that Might Appear on Output Streams
+## Loggers
 
-* the stdout and stderr of executed Jobs (default to logInfo and logError, respectively, control with setPlanStdout, setPlanStderr)
-* The command line ('echo') of executed Jobs (default to logEcho, control with setPlanEcho)
-* The results of calling print/println (default to logInfo, control with printLevel, printlnLevel)
-* The pass/fail results of ____. (default to ...?)
+Wake has a concept of loggers,
+to which bytes can be written during wake execution by various sources.
+Based on command line flags, each logger's output will be directed to
+zero or more of wake's outputs, possibly with additional formatting applied.
+The wake codebase and standard libraries often refer to a _logger_ as a `logLevel`.
 
-## How Logging works on Wake Execution commands
-
-Every `Job` run by wake tracks its own stdout and stderr,
-and when the underlying process prints to stdout or stderr,
-it will go to the `Job`-specific stdout and stderr,
-not the stdout and stderr of the `wake` process itself.
-
-When writing your wake code,
-you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout,
-stderr, or other arbitrary file descriptors.
-These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
-We call this which _logger_ each goes to (wake code often refers to a _logger_ as a `logLevel`).
-
-Regardless of to which logger you direct the `Job`'s `Stderr` and `Stdout`,
-`wake` will store all the text in its database,
-discarding the information about the logger it was destined for at the time it was stored.
-It will however know whether it came from the `Job`'s stderr/stdout.
-
-So what's the difference between the different loggers?
-What you see on your console when you run wake.
-It controls *what* you see (depending on the command line flags) and what it looks like (what color it is).
-It also controls which *actual* stream it is going to coming out of `wake`
+What's the difference between the different loggers?
+Which of wake's own output streams the results are sent to.
+This can influence what you see on your console when you run wake.
+The logger, in combination with the command line flags you invoke wake with,
+controls *what* you see. The logger also controls what it looks like (what color it is).
+It also controls which *actual* stream it is going to coming out of the executing `wake` process
 (stderr or stdout, or other file descriptors 3-5).
 
-There are a number of built in loggers in wake which all start with `log...`.
+`wake` has a number of built in loggers in wake which all start with `log...`.
 You can also make your own logger using `mkLogLevel`.
 You can use this to customize the color you want it to be shown as (if at all),
-and giving more fine-grained control over which wake output streamit should go to.
+and giving more fine-grained control over which wake output stream it should go to.
 
 For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
 If you match a line in the table, stop!
@@ -46,31 +32,80 @@ This is what will appear on the console as you watch wake execute, and what is s
 
 |Flag               | logDebug | logInfo | logEcho | logReport | logWarning | logError | logNever | mkLogLevel "foo" Green|
 |-------------------|----------|---------|---------|-----------|------------|----------|----------|----------------------
-|Color on Console   |  Blue    |  Dim    |  default| Magenta   |  Yellow    |   Red    |          |   Green         |          
-|--fd:1="foo"       |          |         |         |           |            |          |          |         x       |
-|--fd:1="foo,info"  |          |    x    |         |           |            |          |          |         x       |
+|Color on Console   |  Blue    |  Dim    |  default| Magenta   |  Yellow    |   Red    |          |      Green      |          
+|--stdout="foo"     |          |         |         |           |            |          |          |         x       |
+|--stdout="foo,info" |         |    x    |         |           |            |          |          |         x       |
 |--debug            |      x   |    x    |    x    |     x     |      x     |    x     |          |                 |
 |--verbose          |          |    x    |    x    |     x     |      x     |    x     |          |                 |
 |--quiet            |          |         |         |           |            |    x     |          |                 |
-|--no-tty           |          |         |    x    |     x     |      x     |    x     |          |                 |
+|--no-tty<sup>*</sup>|         |         |    x    |     x     |      x     |    x     |          |                 |
 |(default)          |          |         |         |     x     |      x     |    x     |          |                 |
 
-Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
 and it always appears there.
 
-`println` and `print` go to `logReport`. You can make print statements that go to other levels with `printlnLevel` or `printLevel`.
-These statements are *not* saved to the wake database, no matter which `stream` they are sent to.
+The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
+These must be opened. One example shell command would be, for a wake code that had `mkLogLevel "foo"`:
 
-Of course one can always manually redirect or tee stdin/stdout to files within the command executed by the `Job`, outside
-of these mechanisms.
+```
+wake --fd3="foo" 3&>foo.txt
+```
 
-## How Logging works on Wake Database Commands
+These can be concatenated with commas and include the normal log levels, so to capture both `mkLogLevel "foo"` and `logInfo` to a file:
 
-The `--verbose` command line argument has a different functionalilty when used with data base commands (`--last`, `-o`, `--failed`, etc).
+```
+wake --fd3="foo,info" 3&>foo.txt
+```
+
+## What Gets Sent To Loggers
+
+### Job Execution
+
+Every `Job` run by wake tracks its own stdout and stderr,
+and when the underlying process prints to stdout or stderr,
+it will go to the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
+Every `Job` also has an "echo", which is the command that the job is executing.
+
+When writing your wake code,
+you can control which logger these `Job`-specific stdout, stderr, and echo are sent to.
+This will influence whether they appear on the `wake` process's own stdout,
+stderr, or other arbitrary file descriptors.
+These can be controlled with `setPlanStderr`, `setPlanStdout`, and `setPlanEcho` before you `runJob`.
+These default to `logInfo`, `logError`, and `logEcho`, respectively.
+
+Regardless of to which logger you direct the `Job`'s `Stderr`, `Stdout`, and `Echo`,
+`wake` will store all the text in its database,
+discarding the information about the logger it was destined for at the time it was stored.
+It will however know whether it came from the `Job`'s stderr/stdout/echo.
+
+
+### Other Things that Appear on Wake's Output Streams During Execution
+
+We have already covered the stdout, stderr, and echo of executed `Job`s.
+
+The results of calling `print`/`println` within `wake` code are sent to `logReport` by default.
+To send them somewhere else, use `printLevel`, `printlnLevel` instead.
+
+The final output of whatever wake function or expression is called is always directed to
+wake's `stdout` unless `--quiet` is set.
+This is outside of the logger mechanism, so it is not possible to redirect this within wake
+code or with command line flags.
+
+Similarly, the "spinner" and progress bar is always sent to wake's `stdout` unless `--no-tty`
+or `--quiet` is set.
+
+## What is Displayed to user on Wake Database Commands
+
+The `--verbose` command line argument has a different functionality when used with data base commands (`--last`, `-o`, `--failed`, etc).
 
 Running `wake <database command> --verbose` will show ALL of stdout and stdin from all `Job`s,
 regardless of what logger they were directed to.
 Running a database command without `--verbose` will show NO stdout or stdin from any `Job`.
+The command executed will always be displayed regardless of command line arguments.
 
-Recall that results of `print` or `println` are not stored in the database, so will never be revealed with any combination
-of flags.
+Other things that are observed on stdout from executing wake (the results of `print` or `println`, etc)
+are not stored in the database, so will never be revealed with any combination of flags on database commands.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -9,24 +9,24 @@ for commented code.
 Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
 
 When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
-We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
+We call this which _logger_ each goes to (we often refer to a _logger_ as a `logLevel`).
 
-Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
+Regardless of to which logger you direct the `Job`'s `Stderr` and `Stdout`,
 `wake` will store all the text in its database,
-discarding the information about the `stream` it was destined for at the time it was stored.
+discarding the information about the logger it was destined for at the time it was stored.
 It will however know whether it came from the `Job`'s stderr/stdout.
 
-So what's the difference between the different streams?
+So what's the difference between the different loggers?
 What you see on your console when you run wake.
 It controls *what* you see (depending on the command line flags) and what it looks like (what color it is).
 It also controls which *actual* stream it is going to coming out of `wake`
 (stderr or stdout, or other file descriptors 3-5).
 
-There are a number of built in streams in wake which all start with `log...`.
-You can also make your own stream using `mkLogLevel`.
+There are a number of built in loggers in wake which all start with `log...`.
+You can also make your own logger using `mkLogLevel`.
 You can use this to customize the color you want it to be shown as,
-or (more commonly) to grab output from a `Job` for passing to another `Job`,
-consuming within your wake program, or redirecting outside of wake using shell redirect.
+or (more commonly) to grab output from a `Job` for consuming within your wake program,
+or redirecting outside of wake using shell redirect.
 
 Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
 and it always appears there.
@@ -55,10 +55,11 @@ of these mechanisms.
 
 ## How Logging works on Wake Database Commands
 
-Confusingly, the command line arguments in the database alias with those when executing, but they have different behavior.
+The `--verbose` command line argument has a different functionalilty when used with data base commands (`--last`, `-o`, `--failed`, etc).
 
-Running `wake <database command> --verbose` will show ALL of stdout and stdin from all `Job`s, regardless of what wake
-stream they were directed to. Running a database command without `--verbose` will show NO stdout or stdin from any `Job`.
+Running `wake <database command> --verbose` will show ALL of stdout and stdin from all `Job`s,
+regardless of what logger they were directed to.
+Running a database command without `--verbose` will show NO stdout or stdin from any `Job`.
 
 Recall that results of `print` or `println` are not stored in the database, so will never be revealed with any combination
 of flags.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -11,7 +11,10 @@ and when the underlying process prints to stdout or stderr,
 it will go the `Job`-specific stdout and stderr,
 not the stdout and stderr of the `wake` process itself.
 
-When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+When writing your wake code,
+you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout,
+stderr, or other arbitrary file descriptors.
+These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
 We call this which _logger_ each goes to (wake code often refers to a _logger_ as a `logLevel`).
 
 Regardless of to which logger you direct the `Job`'s `Stderr` and `Stdout`,
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -6,7 +6,7 @@ for commented code.
 
 ## How Logging works on Wake Execution commands
 
-Every `Job` run by wake will have its own stdout and stderr,
+Every `Job` run by wake tracks its own stdout and stderr,
 and when the underlying process prints to stdout or stderr,
 it will go to the `Job`-specific stdout and stderr,
 not the stdout and stderr of the `wake` process itself.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -36,8 +36,8 @@ and give more fine-grained control over which wake output stream it should go to
 | logWarning        |    Yellow       | "warning" |
 | logError          |     Red         | "error"   |
 | logNever          |     N/A         | "null"    |
-| mkLogLevel "foo" Green |  Green     | "foo"     |
-| mkLogLevel "foo"  Bright Green | Bright Green   | "foo"      |
+| mkLogLevel "foo" (Some Green) |  Green     | "foo"     |
+| mkLogLevel2 "foo" (Some Green) (Some Bright) | Bright Green   | "foo"      |
 
 ## Directing Logger Output
 
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -33,7 +33,7 @@ and it always appears there.
 
 For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
 If you match a line in the table, stop!
-This is what will appear on the console as you watch wake execute:
+This is what will appear on the console as you watch wake execute, and what is sent to wake's stdout:
 
 Flag               | logDebug | logInfo | logEcho | logReport | logWarning | logError | logNever | mkLogLevel "foo" Green|
 --------------------------------------------------------------------------------------------------------------------------
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -8,7 +8,7 @@ for commented code.
 
 Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
 
-When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
 We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
 
 Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -14,7 +14,7 @@ We call this which `stream` each goes to (we often refer to a `stream` as a `log
 Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
 `wake` will store all the text in its database,
 discarding the information about the `stream` it was destined for at the time it was stored.
-It will however know whether it came from the `Job`'s stdin/stdout.
+It will however know whether it came from the `Job`'s stderr/stdout.
 
 So what's the difference between the different streams?
 What you see on your console when you run wake.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -66,7 +66,7 @@ i.e. what is sent to ``wake`'s stdout:
 <sup>*</sup> `--no-tty` also prevents any coloring from being applied.
 
 Unless overridden with `--stderr=...`,
-the only logger output that ever appears on wake's stderr is outputs from loggers with log level "error"
+the only logger output that ever appears on wake's stderr is from those with log level "error"
 (e.g. `logError`).
 
 The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -1,0 +1,65 @@
+# Logging
+
+This article gives a high level overview of how wake logging works.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## How Logging works on Wake Execution commands
+
+Every `Job` will emit a stdin and a stdout.
+
+When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+We call this which `stream` each goes to (we often refer to each `stream` as a `logLevel`).
+
+Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
+`wake` will store all the text in its database,
+discarding the information about the `stream` it was destined for at the time it was stored.
+It will however know whether it came from the `Job`'s stdin/stdout.
+
+So what's the difference between the different streams?
+What you see on your console when you run wake.
+It controls *what* you see (depending on the command line flags) and what it looks like (what color it is).
+It also controls which *actual* stream it is going to coming out of `wake`
+(stderr or stdout, or other file descriptors 3-5).
+
+There are a number of built in streams in wake which all start with `log...`.
+You can also make your own stream using `mkLogLevel`.
+You can use this to customize the color you want it to be shown as,
+or (more commonly) to grab output from a `Job` for passing to another `Job`,
+consuming within your wake program, or redirecting outside of wake using shell redirect.
+
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
+If you match a line in the table, stop!
+This is what will appear on the console as you watch wake execute:
+
+Flag               | logDebug | logInfo | logEcho | logReport | logWarning | logError | logNever | mkLogLevel "foo" Green|
+--------------------------------------------------------------------------------------------------------------------------
+Color on Console   |  Blue    |  Dim    |  default| Magenta   |  Yellow    |   Red    |          |   Green        |          
+--------------------------------------------------------------------------------------------------------------------
+--fd:1="foo"       |          |         |         |           |            |          |          |         x       |
+--fd:1="foo,info"  |          |    x    |         |           |            |          |          |         x       |
+--no-execute       |          |         |         |           |            |    x     |          |                 |
+--debug            |      x   |    x    |    x    |     x     |      x     |    x     |          |                 |
+--verbose          |          |    x    |    x    |     x     |      x     |    x     |          |                 |
+--quiet            |          |         |         |           |            |    x     |          |                 |
+--no-tty           |          |         |    x    |     x     |      x     |    x     |          |                 |
+(default)          |          |         |         |     x     |      x     |    x     |          |                 |
+
+`println` and `print` go to `logReport`. You can make print statements that go to other levels with `printlnLevel` or `printLevel`.
+These statements are *not* saved to the wake database, no matter which `stream` they are sent to.
+
+Of course one can always manually redirect or tee stdin/stdout to files within the command executed by the `Job`, outside
+of these mechanisms.
+
+## How Logging works on Wake Database Commands
+
+Confusingly, the command line arguments in the database alias with those when executing, but they have different behavior.
+
+Running `wake <database command> --verbose` will show ALL of stdout and stdin from all `Job`s, regardless of what wake
+stream they were directed to. Running a database command without `--verbose` will show NO stdout or stdin from any `Job`.
+
+Recall that results of `print` or `println` are not stored in the database, so will never be revealed with any combination
+of flags.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -6,7 +6,7 @@ for commented code.
 
 ## How Logging works on Wake Execution commands
 
-Every `Job` will emit a stderr and a stdout.
+Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
 
 When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
 We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -6,7 +6,7 @@ for commented code.
 
 ## How Logging works on Wake Execution commands
 
-Every `Job` will emit a stdin and a stdout.
+Every `Job` will emit a stderr and a stdout.
 
 When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
 We call this which `stream` each goes to (we often refer to each `stream` as a `logLevel`).
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -6,7 +6,10 @@ for commented code.
 
 ## How Logging works on Wake Execution commands
 
-Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
+Every `Job` run by wake will have its own stdout and stderr,
+and when the underlying process prints to stdout or stderr,
+it will go the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
 
 When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
 We call this which _logger_ each goes to (wake code often refers to a _logger_ as a `logLevel`).
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -65,7 +65,7 @@ i.e. what is sent to ``wake`'s stdout:
 
 <sup>*</sup> `--no-tty` also prevents any coloring from being applied.
 
-Unless overridden with `--stderr=...` the *only* thing that ever appears on wake's stderr is outputs from loggers
+Unless overridden with `--stderr=...` the only logger output that ever appears on wake's stderr is outputs from loggers
 with log level "error" (e.g. `logError`).
 
 The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -8,7 +8,7 @@ for commented code.
 
 Every `Job` run by wake will have its own stdout and stderr,
 and when the underlying process prints to stdout or stderr,
-it will go the `Job`-specific stdout and stderr,
+it will go to the `Job`-specific stdout and stderr,
 not the stdout and stderr of the `wake` process itself.
 
 When writing your wake code,
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -97,7 +97,7 @@ stderr, or other arbitrary file descriptors.
 These can be controlled with `setPlanStdout`, `setPlanStderr` before you `runJob`.
 These default to `logInfo` and `logError` respectively.
 
-Regardless of to which logger you direct the `Job`'s `Stderr` and `Stdout`,
+Regardless of the logger to which you direct the `Job`'s `Stderr` and `Stdout`,
 `wake` will store all the text in its database,
 discarding the information about the logger it was destined for at the time it was stored.
 It will however know whether it came from the `Job`'s stderr/stdout.
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -6,9 +6,9 @@ for commented code.
 
 ## Things that Might Appear on Output Streams
 
-* the stdout and stderr of executed Jobs (default to logInfo and logError, respectively)
-* The command line ('echo') of executed Jobs (default to logEcho)
-* The results of calling print/println (default to logInfo)
+* the stdout and stderr of executed Jobs (default to logInfo and logError, respectively, control with setPlanStdout, setPlanStderr)
+* The command line ('echo') of executed Jobs (default to logEcho, control with setPlanEcho)
+* The results of calling print/println (default to logInfo, control with printLevel, printlnLevel)
 * The pass/fail results of ____. (default to ...?)
 
 ## How Logging works on Wake Execution commands
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -117,8 +117,7 @@ wake's `stdout` unless `--quiet` is set.
 This is outside of the logger mechanism,
 so it is not possible to redirect this within wake code or with command line flags.
 
-Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty`
-or `--quiet` is set.
+Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty` is set.
 
 ## What is Displayed to user on Wake Database Commands
 
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -43,8 +43,6 @@ and give more fine-grained control over which wake output stream it should go to
 
 Based on command line flags and its log level,
 each logger's output will be directed to zero or more of wake's outputs streams.
-Depending on the logger's color control and command line flags,
-additional formatting can be applied.
 
 For `wake`'s own stdout, the logger's color & intensity are used to apply
 additional formatting to the output before displaying it.
@@ -55,20 +53,20 @@ If you match a line in the table, stop!
 This is what will appear on the console as you watch wake execute,
 i.e. what is sent to ``wake`'s stdout:
 
-| log level         | "debug"  | "info"  | "echo"  | "report"  | "warning"  | "error"  | "null"  |  "foo" |
-|-------------------|----------|---------|---------|-----------|------------|----------|---------|--------|
-|--stdout="foo"     |          |         |         |           |            |          |         |    x   |
-|--stdout="foo,info" |         |    x    |         |           |            |          |         |    x   |
-|--debug            |      x   |    x    |    x    |     x     |      x     |    x     |         |        |
-|--verbose          |          |    x    |    x    |     x     |      x     |    x     |         |        |
-|--quiet            |          |         |         |           |            |    x     |         |        |
-|--no-tty<sup>*</sup>|         |         |    x    |     x     |      x     |    x     |         |        |
-|(default)          |          |         |         |     x     |      x     |    x     |         |        |
+| log level           | "debug"  | "info"  | "echo"  | "report"  | "warning"  | "error"  | "null"  |  "foo" |
+|---------------------|----------|---------|---------|-----------|------------|----------|---------|--------|
+|--stdout="foo"       |          |         |         |           |            |          |         |    x   |
+|--stdout="foo,info"  |          |    x    |         |           |            |          |         |    x   |
+|--debug              |      x   |    x    |    x    |     x     |      x     |    x     |         |        |
+|--verbose            |          |    x    |    x    |     x     |      x     |    x     |         |        |
+|--quiet              |          |         |         |           |            |    x     |         |        |
+|--no-tty<sup>*</sup> |          |         |    x    |     x     |      x     |    x     |         |        |
+|(default)            |          |         |         |     x     |      x     |    x     |         |        |
 
 <sup>*</sup> `--no-tty` also prevents any coloring from being applied.
 
-Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is logger
-outputs with log level "error" (e.g. `logError`).
+Unless overridden with `--stderr=...` the *only* thing that ever appears on wake's stderr is outputs from loggers
+with log level "error" (e.g. `logError`).
 
 The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
 One example shell command would be, for some wake code that had `mkLogLevel "foo"`:
--- a/share/doc/wake/logging.md
+++ b/share/doc/wake/logging.md
@@ -4,6 +4,13 @@ This article gives a high level overview of how wake logging works.
 Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
 for commented code.
 
+## Things that Might Appear on Output Streams
+
+* the stdout and stderr of executed Jobs (default to logInfo and logError, respectively)
+* The command line ('echo') of executed Jobs (default to logEcho)
+* The results of calling print/println (default to logInfo)
+* The pass/fail results of ____. (default to ...?)
+
 ## How Logging works on Wake Execution commands
 
 Every `Job` run by wake tracks its own stdout and stderr,
@@ -34,8 +41,7 @@ You can use this to customize the color you want it to be shown as,
 or (more commonly) to grab output from a `Job` for consuming within your wake program,
 or redirecting outside of wake using shell redirect.
 
-Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
-and it always appears there.
+
 
 For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
 If you match a line in the table, stop!
@@ -51,6 +57,9 @@ This is what will appear on the console as you watch wake execute, and what is s
 |--quiet            |          |         |         |           |            |    x     |          |                 |
 |--no-tty           |          |         |    x    |     x     |      x     |    x     |          |                 |
 |(default)          |          |         |         |     x     |      x     |    x     |          |                 |
+
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
 
 `println` and `print` go to `logReport`. You can make print statements that go to other levels with `printlnLevel` or `printLevel`.
 These statements are *not* saved to the wake database, no matter which `stream` they are sent to.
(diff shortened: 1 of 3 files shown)


share/doc/wake/logging.md
```diff
@@ -0,0 +1,65 @@
...
+This article gives a high level overview of how wake logging works.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## How Logging works on Wake Execution commands
+
+Every `Job` will emit a stderr and a stdout.
+
+When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+We call this which `stream` each goes to (we often refer to each `stream` as a `logLevel`).
```
mwachs5, line 12:
```suggestion
We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,65 @@
...
+You can use this to customize the color you want it to be shown as,
+or (more commonly) to grab output from a `Job` for passing to another `Job`,
+consuming within your wake program, or redirecting outside of wake using shell redirect.
+
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
+If you match a line in the table, stop!
+This is what will appear on the console as you watch wake execute:
```
mwachs5, line 36:
```suggestion
This is what will appear on the console as you watch wake execute, and what is sent to wake's stdout:
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,65 @@
...
+
+Every `Job` will emit a stderr and a stdout.
+
+When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
+
+Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
+`wake` will store all the text in its database,
+discarding the information about the `stream` it was destined for at the time it was stored.
+It will however know whether it came from the `Job`'s stdin/stdout.
```
mwachs5, line 17:
```suggestion
It will however know whether it came from the `Job`'s stderr/stdout.
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,65 @@
+# Logging
+
+This article gives a high level overview of how wake logging works.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## How Logging works on Wake Execution commands
+
+Every `Job` will emit a stderr and a stdout.
```
JakeSiFive, line 9: I think "a" vs "an" depends how you say this. If you say "stood error" then "a" is correct but if you're like me and say "es-tee-dee error" then you want "an" here.

Maybe to be more widely clear if you said "ever `Job` will emit some (possibly empty) string to its stderr and stdout" it might be more explict and avoid the pronunciation ambiguity.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
```
richardxia, line 9: Nit 1: stderr and stdout are file descriptors which just exist for pretty much all Unix processes. You probably mean "emit to" rather than "emit", but it might be clearer to just say "print to".

Nit 2: Not all Jobs will necessarily write to either stderr or stdout; it depends on whether the job itself tries to print to either of those file descriptors.

Here's my version of the sentence I think you wanted:

```suggestion
Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
```

JakeSiFive (approved):
This looks awesome! Thank you!

Please give Sam a chance to review this tomorrow as well.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+
+When writing your wake code, you can control where each goes with `setPlanStderr` and `setPlanStdout` before you `runJob`.
```
richardxia, line 11: I think a more precise way of wording this that is still accessible to users who don't have a deep understanding of Unix processes, file descriptors, and whatnot would maybe be the following:

```suggestion
When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+We call this which `stream` each goes to (we often refer to a `stream` as a `logLevel`).
```
richardxia, line 12: I would suggest not formatting "stream" in monospace because it doesn't actually correspond to anything in code.

richardxia (commented):
I have to run now, and I didn't read the whole thing, but 💯 on doing this, and I'll leave my feedback on the first few sentences.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+Regardless of which `stream` you direct the `Job`'s `Stderr` and `Stdout`,
+`wake` will store all the text in its database,
+discarding the information about the `stream` it was destined for at the time it was stored.
+It will however know whether it came from the `Job`'s stderr/stdout.
+
+So what's the difference between the different streams?
+What you see on your console when you run wake.
+It controls *what* you see (depending on the command line flags) and what it looks like (what color it is).
+It also controls which *actual* stream it is going to coming out of `wake`
+(stderr or stdout, or other file descriptors 3-5).
```
richardxia, line 23: This sentence is a bit confusing because you are using the word "stream" differently than in the first sentence of the paragraph. It's almost tautological because you are effectively saying that the difference between the different "streams" is which "stream" they use.

Perhaps we just need to avoid using the term "stream" to refer to a wake `LogLevel`, since it makes it difficult for us to refer to the [Unix concept of the same name](https://en.wikipedia.org/wiki/Standard_streams).

Notably, a Unix stream doesn't dictate the color because color codes can be present within the stream to change the colors of specific characters displayed by a terminal, rather than the entire stream.

If we're going to bikeshed about naming here, I'd like to pitch renaming `LogLevel` to `Logger`, since I think that more closely matches how many other languages ([Java](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html), [Python](https://docs.python.org/3/howto/logging.html)) call it, where the logging level is just a property of a logger.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+
+There are a number of built in streams in wake which all start with `log...`.
+You can also make your own stream using `mkLogLevel`.
+You can use this to customize the color you want it to be shown as,
+or (more commonly) to grab output from a `Job` for passing to another `Job`,
```
richardxia, line 28: My understanding is that LogLevels don't affect this at all.

share/doc/wake/logging.md
mwachs5: right, i found it really awkward to call it a "log level" as the sentences don't make sense.

But what i am trying to distinguish here is exactly what you helped rewrite in the first sentence -- each running job has its own output streams. Where you tell wake to send it via the `logLevel` mechanis, to one of *wake*'s output streams is different than those.

share/doc/wake/logging.md
mwachs5: the "this" you are referring to is "grabbing output from a `Job` for passing to another `Job`'?

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+consuming within your wake program, or redirecting outside of wake using shell redirect.
```
mwachs5, line 29: what about the "consuming within your wake program" use case?

share/doc/wake/logging.md
richardxia: Yes. Sorry, the GitHub diff view for leaving comments really obfuscates which lines of code will actually be shown even when I specifically select just a single line to comment on. I guess there's no way to indicate that you really are commenting on one line.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+
+`println` and `print` go to `logReport`. You can make print statements that go to other levels with `printlnLevel` or `printLevel`.
+These statements are *not* saved to the wake database, no matter which `stream` they are sent to.
+
+Of course one can always manually redirect or tee stdin/stdout to files within the command executed by the `Job`, outside
+of these mechanisms.
+
+## How Logging works on Wake Database Commands
+
+Confusingly, the command line arguments in the database alias with those when executing, but they have different behavior.
```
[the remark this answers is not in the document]
mwachs5: I mean, it's very confusing for me as a user, especially when I have not really absorbed what is a "database command" or otherwise.

But I also don't think it's useful for me to say "Something is confusing" in a document, so happy to reword this.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+
+Running `wake <database command> --verbose` will show ALL of stdout and stdin from all `Job`s, regardless of what wake
+stream they were directed to. Running a database command without `--verbose` will show NO stdout or stdin from any `Job`.
+
+Recall that results of `print` or `println` are not stored in the database, so will never be revealed with any combination
+of flags.
```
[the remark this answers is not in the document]
mwachs5: It doesn't matter whether you can/can't do this, the point is that people should not expect to see it. It's not super intuitive what the database is *storing*. If I run `wake -v -o <some output>` it is reasonable user expectation to see the same results as I would have seen by running `wake -v <some command that generated some output>`.

I think all this is under development and changing, and the plan would be to print it out with the database commands "more like the user would have seen it" by running wake -v, and then I think it *will* become even more confusing that the printlns don't show up in that view.

share/wake/lib/core/print.wake
```diff
@@ -80,27 +80,27 @@ export def makeLogLevel2 (name: String) (colour: Option Colour) (intensity: Opti
...
 export def logReport: LogLevel =
     makeLogLevel "report" (Some Magenta)
 
-# logEcho: logged to stdout when run with -v
+# logEcho: logged to stdout when run with -v (default console color)
 export def logEcho: LogLevel =
     makeLogLevel "echo" None
 
-# logInfo: logged to stdout when run with -v
+# logInfo: logged to stdout when run with -v (Dim)
```
mwachs5, line 99:
```suggestion
# logInfo: logged to stdout when run with -v (Dim version of default console color)
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
+If you match a line in the table, stop!
+This is what will appear on the console as you watch wake execute, and what is sent to wake's stdout:
+
+|Flag               | logDebug | logInfo | logEcho | logReport | logWarning | logError | logNever | mkLogLevel "foo" Green|
+|-------------------|----------|---------|---------|-----------|------------|----------|----------|----------------------
+|Color on Console   |  Blue    |  Dim    |  default| Magenta   |  Yellow    |   Red    |          |   Green         |
```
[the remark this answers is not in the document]
richardxia: Your response is contradictory, since you say the "console default" and then you say whatever scalatest overrides the console default with. Your console default is determined by your terminal emulator's settings, specifically the foreground color (normally you set both the foreground and the background, since most of the time you either want dark text on a light background or light text on a dark background). Programs can change the foreground and background colors by setting control codes that affect either the foreground or background colors, and the colors used by wake appear to be the [ANSI 3-bit/4-bit colors](https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit) ((black + white + 6 other colors) &times; bright modifier).

Most well-behaved programs will not emit color codes to stdout/stderr when not connected to a tty (e.g. stdout is piped to a file or another program), but I've noticed that most Scala programs really do not seem to follow those conventions at all. I have no idea what happens in wake when a program prints color codes and whether wake overrides them or whether the program's original color codes come out.

share/doc/wake/logging.md
mwachs5: used logger instead of stream. It reads much better now, thank you.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,65 @@
...
+This article gives a high level overview of how wake logging works.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## How Logging works on Wake Execution commands
+
+Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
+
+When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+We call this which _logger_ each goes to (we often refer to a _logger_ as a `logLevel`).
```
mwachs5, line 12:
```suggestion
We call this which _logger_ each goes to (wake code often refers to a _logger_ as a `logLevel`).
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,65 @@
+# Logging
+
+This article gives a high level overview of how wake logging works.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## How Logging works on Wake Execution commands
+
+Every `Job` run by wake will have its own stdout and stderr, and when the underlying process prints to stdout or stderr, it will go the `Job`-specific stdout and stderr, not the stdout and stderr of the `wake` process itself.
```
mwachs5, line 9:
```suggestion
Every `Job` run by wake will have its own stdout and stderr,
and when the underlying process prints to stdout or stderr,
it will go the `Job`-specific stdout and stderr,
not the stdout and stderr of the `wake` process itself.
```

trying to be good about breaking lines on punctuation

share/doc/wake/logging.md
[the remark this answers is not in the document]
richardxia: I can see that it may be unintuitive what the database is storing, but I think that's completely orthogonal to logging, which is what this document is about.

> If I run wake -v -o <some output> it is reasonable user expectation to see the same results as I would have seen by running wake -v <some command that generated some output>.

I'm not entirely sure that I agree with that, but I think it depends on what someone thinks `-o` is supposed to do. Once again, I'm going to claim that this should be explained in a separate document about the various commands that query job information from the database, not the document about logging, since logging doesn't affect the database commands.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,68 @@
...
+for commented code.
+
+## How Logging works on Wake Execution commands
+
+Every `Job` run by wake will have its own stdout and stderr,
+and when the underlying process prints to stdout or stderr,
+it will go the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
+
+When writing your wake code, you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout, stderr, or other arbitrary file descriptors. These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
```
mwachs5, line 14:
```suggestion
When writing your wake code,
you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout,
stderr, or other arbitrary file descriptors.
These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
```

share/doc/wake/logging.md
richardxia: I don't actually understand what you mean by that.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,64 @@
...
+(stderr or stdout, or other file descriptors 3-5).
+
+There are a number of built in streams in wake which all start with `log...`.
+You can also make your own stream using `mkLogLevel`.
+You can use this to customize the color you want it to be shown as,
+or (more commonly) to grab output from a `Job` for passing to another `Job`,
+consuming within your wake program, or redirecting outside of wake using shell redirect.
+
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
```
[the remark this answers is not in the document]
mwachs5: agree, will reorder tomorrow

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: the program's original color codes come out. I mean, there are 2 arguments to the color (in wake) -- the color override and the intensity override. "console default" isn't the right term, but I want to say something like "whatever color you would have seen if you were just running it, not using wake". do you have a better suggestion for that term?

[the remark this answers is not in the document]
richardxia: > (but have not re-read the code in question since you wrote this comment)

Having read the code now, here's an explanation of how it works:

In wake-land, [`makeLogLevel2` eventually makes a call](https://github.com/sifive/wake/blob/e596242b4a009550e2bd06ea21206f9aedeeea18/share/wake/lib/core/print.wake#L61) to the [primitive function `colour`](https://github.com/sifive/wake/blob/a84668a15cdb076e9796631aa32f6cb7f4941a08/src/runtime/string.cpp#L391-L398), which accepts two arguments: a string name (the logger name), and an integer colo<s>u</s>r. Note that this color encodes both the hue and the the intensity using the following bit mask:

```
[MSB]   5    4          3        2  1  0  [LSB]
      [intensity]      ???      [  hue  ]

intensity = 0 => unset
intensity = 1 => dim
intensity = 2 => bright

??? = apparently set to 1 (via adding 8) when you want to set a color, including black,
      but otherwise ignored by the rest of the wake code? Maybe just a sentinel
      so that when you explicitly want to set a color of black, the rest of the wake
      code can see that you explicitly wanted black.

hue = standard 3-bit ANSI encoding
```

The primitive function `colour` just calls out to [`set_status_colour()`](https://github.com/sifive/wake/blob/e596242b4a009550e2bd06ea21206f9aedeeea18/src/runtime/status.cpp#L280), which is a C++ function that just stores the name and integer color (still encoded) into a hash table. Later on, when wake wants to actually write something to a logger/stream/whatever we're calling it, it calls [`status_write()`](https://github.com/sifive/wake/blob/e596242b4a009550e2bd06ea21206f9aedeeea18/src/runtime/status.cpp#L300-L314), which decodes the color and intensity and calls a pair of helper functions, [`term_colour()` and `term_intensity()`](https://github.com/sifive/wake/blob/e596242b4a009550e2bd06ea21206f9aedeeea18/src/runtime/term.cpp#L42-L57).

`term_colour()` is basically just a wrapper around the [`term` `setaf`](https://man7.org/linux/man-pages/man5/terminfo.5.html) capability, which gets the control code for setting the foreground color to an ANSI color.

`term_intensity()` looks up the control code for setting dim or bold (bright). The actual behavior of dim or bold depends on your terminal emulator (some like mine _actually_ bold the text, but some will just change the text to a brighter color).

Anyway, based on my reading of this code, I think you are right. Dim will send the Dim control code to the user's terminal. Most non-Scala programs will not attempt to set color codes when run under wake, but Scala programs will, and setting the color intensity to dim will actually dim the text if the program set color codes (unless the program overrides the dimness code).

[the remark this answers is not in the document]
richardxia: Sorry, I need to remember to refresh before hitting submit, especially when I've been typing comments for a long time. Didn't see your previous comment.

> the program's original color codes come out. I mean, there are 2 arguments to the color (in wake) -- the color override and the intensity override. the Dim color used by logInfo only overrides the second one without touching the first one. "console default" isn't the right term, but I want to say something like "whatever color you would have seen if you were just running it, not using wake". do you have a better suggestion for that term?

I think we can just leave it at dim. I wasn't sure if wake was itself shifting into 4-bit ANSI, which has a "bright" set of the original 8 colors in 3-bit ANSI, but it looks like what is labeled as "bright" in wake actually corresponds to "bold" in ANSI, not "bright" in ANSI, so I was confused.

I would really suggest renaming the Wake enum value from `Bright` to `Bold`, since not everyone's terminal emulators will actually brighten the text (mine will in fact bold it).

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: have reworded

share/doc/wake/logging.md
mwachs5: like, can i do something in wake like send some Job's stdout to "megalog" (`def logMega = mkLogLevel "megalog"`, then send a job's stdout to `logMega`), then do some sort of in-wake decision making on the text found in `logMega`.

mwachs5: In other words, is the only way to actually use the custom case shell redirect? If so I should just say that and then maybe give a little cookbook at the end. If there are other use cases like consuming the output within wake, then maybe that's another cookbook or I'd leave out all the cookbooks

richardxia: > like, can (should?) I do something in wake like send some Job's stdout to "megalog" (def logMega = mkLogLevel "megalog", then send a job's stdout to logMega), then do some sort of in-wake decision making on the text found in logMega.

That's not possible within wake itself.

> In other words, is the only way to actually use the custom case shell redirect? If so I should just say that and then maybe give a little cookbook at the end. If there are other use cases like consuming the output within wake, then maybe that's another cookbook or I'd leave out all the cookbooks

No and yes. Yes if you want to make sure nothing else is written to stdout and stderr, which is important when you want to consume the content within the custom stream programmatically. No if you don't care, in which case you can just redirect one of the builtin streams (loggers?) if you wanted to.

The shell redirection thing is a shell-specific thing. You can consume the custom streams without redirection if you directly read the file descriptors, such as if you invoked wake via another programming language like Python.

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: Maybe. I guess this is a question of "is this a document for someone writing wake code, or someone runnign wake code, or both"?

If it's for the latter category, the experience is "I run a wake command. A bunch of stuff spews to my screen. I see if I run `wake --last` I can see some of that stuff." I'd like to give more intuition on what that stuff is, and why some of it doesnt' show up.

Let me ask a different related question. Are there other things in wake aside from this whole logging mechanism and print/println that make things show up on the user's console? I'd like to make that in the scope of this document (the first part), then I'd like to consider what to say about it, if anything, in the second part.

Like, does returning a Fail result automatically put something on the console (not sure this sentence even makes sense)? Is that reflected in the database?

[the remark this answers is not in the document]
richardxia: > Maybe. I guess this is a question of "is this a document for someone writing wake code, or someone runnign wake code, or both"?

I assumed the former, since none of the log level stuff is really relevant to someone only running `wake`, since as someone only running `wake` rules built by others, you're really just at the whims of whatever LogLevels were chosen by the person who authored the rules.

> Let me ask a different related question. Are there other things in wake aside from this whole logging mechanism and print/println that make things show up on the user's console? I'd like to make that in the scope of this document (the first part), then I'd like to consider what to say about it, if anything, in the second part.
>
> Like, does returning a Fail result automatically put something on the console (not sure this sentence even makes sense)? Is that reflected in the database?

You brought up the only other thing that I can think of, which is the final returned value is printed to screen. This is true regardless of whatever the return value is, i.e. `Pass` or `Unit` or anything just get printed at the end. I believe they go to stdout, which is one of the reasons why you _have_ to map your logging streams to fd:3 or some other file descriptor or else the final return value gets included into stdout.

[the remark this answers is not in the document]
richardxia: Sorry, didn't answer the final question, but no, this is not reflected in the database because it's not a Job. The wake database holds Job information, and Jobs are only created with `runJob`.

Nothing else is a Job.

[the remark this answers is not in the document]
mwachs5: >  since none of the log level stuff is really relevant to someone only running wake

I mean, they see things getting printed out in different colors, some of which show up when they do or do not specify different flags, so I do think that is very relevant to the users.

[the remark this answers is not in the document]
mwachs5: (I guess I am saying I think I am mixing the two users/reader types indiscriminately in this document, which is sloppy, I can think about more clearly dividing it in more of a 'so you are writing wake code' vs 'so you're running wake commands').

share/doc/wake/logging.md
```diff
@@ -0,0 +1,71 @@
...
+it will go the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
+
+When writing your wake code,
+you can control whether these `Job`-specific stdout and stderr are also printed to `wake`'s own stdout,
+stderr, or other arbitrary file descriptors.
+These can be controlled with `setPlanStderr` and `setPlanStdout` before you `runJob`.
+We call this which _logger_ each goes to (wake code often refers to a _logger_ as a `logLevel`).
+
+Regardless of to which logger you direct the `Job`'s `Stderr` and `Stdout`,
```
ag-eitilt, line 20: One more thing which may or may not be worth mentioning here, you can also call `setPlanEcho` with a LogLevel to change where and how noisily the command line/label is printed.  That's generally of much less utility, but it's broadly the same mechanisms involved.

ag-eitilt (changes requested):
This is a mix of optional suggestions and changes which are worth blocking for, so sorry for not making it obvious which comment is which.  I'm sure you'll be able to figure out what's worth changing, and don't worry about any comments you disagree with.

On the whole, though, big thanks for writing this up!

share/doc/wake/logging.md
```diff
@@ -0,0 +1,71 @@
...
+`wake` will store all the text in its database,
+discarding the information about the logger it was destined for at the time it was stored.
+It will however know whether it came from the `Job`'s stderr/stdout.
+
+So what's the difference between the different loggers?
+What you see on your console when you run wake.
+It controls *what* you see (depending on the command line flags) and what it looks like (what color it is).
+It also controls which *actual* stream it is going to coming out of `wake`
+(stderr or stdout, or other file descriptors 3-5).
```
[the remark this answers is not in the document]
mwachs5: right i was debating adding those to the --help output in this PR too

share/doc/wake/logging.md
```diff
@@ -0,0 +1,71 @@
...
+or redirecting outside of wake using shell redirect.
+
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+For stdout, the following table should be read top to bottom considering the command line arguments you've passed to wake.
+If you match a line in the table, stop!
+This is what will appear on the console as you watch wake execute, and what is sent to wake's stdout:
+
+|Flag               | logDebug | logInfo | logEcho | logReport | logWarning | logError | logNever | mkLogLevel "foo" Green|
```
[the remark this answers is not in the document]
mwachs5: >  but it's probably a good idea to clarify that logEcho is the command run

I dont know what you mean, `logEcho` and `logInfo` are both "log level"s of the same sort of shape? One has name "info" and one has name "echo" and they have different colors.

I think what you are saying is that there are OTHER things, not just running-job-stderr/stdout and println/print that show up on the console, and these are:

* The command run (will always get sent to the `logEcho` logger)

Not sure what you mean by 

> logInfo is the internal Job's stdout.

What is "the internal Job"?

And then I want to flesh out what @richardxia said about the results getting sent to some particular logger.

> f I do a printlnLevel logError does that get printed to Wake's stdout just formatted as if it were an error?

That is my understanding

share/doc/wake/logging.md
mwachs5: > how noisily the command line/label 

Good to know! I will take a look at that too.

share/doc/wake/logging.md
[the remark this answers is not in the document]
ag-eitilt: > > logInfo is the internal Job's stdout.
> 
> What is "the internal Job"?

Sorry, the command that is being shelled out for -- so in `makePlan "" Nil "echo Hello world!"` the "Hello world!" is output on `logInfo` by default.  That mapping is not quite obvious from the "Info" name, so it might be worth an explicit mention.

> The command run (will always get sent to the logEcho logger)

I said this more clearly below with the "how noisily the command line/label" comment.  The full `"echo \\"Hello world!\\""` command gets printed to `logEcho` which may or may not be what people expect from *that* name, depending on how familiar they are with things like Batch's `@echo off`/Bash's `set +x`.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,71 @@
...
+
+So what's the difference between the different loggers?
+What you see on your console when you run wake.
+It controls *what* you see (depending on the command line flags) and what it looks like (what color it is).
+It also controls which *actual* stream it is going to coming out of `wake`
+(stderr or stdout, or other file descriptors 3-5).
+
+There are a number of built in loggers in wake which all start with `log...`.
+You can also make your own logger using `mkLogLevel`.
+You can use this to customize the color you want it to be shown as,
```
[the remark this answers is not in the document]
mwachs5: yeah, that's what I meant by the "or more commonly..." right after this, but I need to rewrite that bit after the feedback from @richardxia

share/doc/wake/logging.md
```diff
@@ -0,0 +1,71 @@
...
+|-------------------|----------|---------|---------|-----------|------------|----------|----------|----------------------
+|Color on Console   |  Blue    |  Dim    |  default| Magenta   |  Yellow    |   Red    |          |   Green         |          
+|--fd:1="foo"       |          |         |         |           |            |          |          |         x       |
+|--fd:1="foo,info"  |          |    x    |         |           |            |          |          |         x       |
+|--no-execute       |          |         |         |           |            |    x     |          |                 |
+|--debug            |      x   |    x    |    x    |     x     |      x     |    x     |          |                 |
+|--verbose          |          |    x    |    x    |     x     |      x     |    x     |          |                 |
+|--quiet            |          |         |         |           |            |    x     |          |                 |
+|--no-tty           |          |         |    x    |     x     |      x     |    x     |          |                 |
```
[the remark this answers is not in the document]
mwachs5: I'm just calling it like I see it :)

Does -no-tty strip the colors entirely?

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: > the "Hello world!" is output on logInfo by default.

I believe that is equivalent to saying "if you don't override setPlanStdout for a Job, it goes to `logInfo`". (And probably I should say "if you don't override setPlanStderr for a Job, it goes to `logError`" (assuming that is true?)

share/doc/wake/logging.md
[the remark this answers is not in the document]
ag-eitilt: It does, yeah.  Gets rid of all the fancy stuff to be able to output to the most basic interfaces/logs.

share/doc/wake/logging.md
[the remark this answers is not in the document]
ag-eitilt: It is true, and if you're already saying that, then great!  I probably just didn't see it, or saw it after writing the comment and forgot to go back to this.

share/doc/wake/logging.md
```diff
@@ -4,6 +4,13 @@ This article gives a high level overview of how wake logging works.
 Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
 for commented code.
 
+## Things that Might Appear on Output Streams
+
+* the stdout and stderr of executed Jobs (default to logInfo and logError, respectively, control with setPlanStdout, setPlanStderr)
+* The command line ('echo') of executed Jobs (default to logEcho, control with setPlanEcho)
+* The results of calling print/println (default to logInfo, control with printLevel, printlnLevel)
```
ag-eitilt, line 11: I didn't actually know we had `print`/`printLevel`; you learn something new every day!  Apparently, though, the comments above the definitions are wrong since these don't actually go to `logInfo`:
```suggestion
* The results of calling print/println (default to logReport, control with printLevel, printlnLevel)
```

share/doc/wake/logging.md
```diff
@@ -4,6 +4,13 @@ This article gives a high level overview of how wake logging works.
...
+* The pass/fail results of ____. (default to ...?)
```
ag-eitilt, line 12: This is outside the logger handling.  It's the output of whatever Wake function or expression is called, which is *normally* a pass/fail but does not have to be:
```
$ wake -x '1 + 2'
3
```
This gets printed to Wake's stdout unless `--quiet` is set, but can't be affected by any of the log-level-based flags.

share/doc/wake/logging.md
```diff
@@ -52,6 +58,9 @@ This is what will appear on the console as you watch wake execute, and what is s
 |--no-tty           |          |         |    x    |     x     |      x     |    x     |          |                 |
 |(default)          |          |         |         |     x     |      x     |    x     |          |                 |
 
+Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
```
ag-eitilt, line 58: Moving the existing suggestion to the new diff location.
```suggestion
Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
```

share/doc/wake/logging.md
```diff
@@ -30,12 +37,8 @@ It also controls which *actual* stream it is going to coming out of `wake`
...
 There are a number of built in loggers in wake which all start with `log...`.
 You can also make your own logger using `mkLogLevel`.
-You can use this to customize the color you want it to be shown as,
-or (more commonly) to grab output from a `Job` for consuming within your wake program,
-or redirecting outside of wake using shell redirect.
-
-Unless overridden with `--fd:2=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
-and it always appears there.
+You can use this to customize the color you want it to be shown as (if at all),
+and giving more fine-grained control over which wake output streamit should go to.
```
ag-eitilt, line 41:
```suggestion
and giving more fine-grained control over which wake output stream it should go to.
```

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: ok, then let me break out into another table of "what color does it show as" and mention that --no-tty also strips the colors

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: no I am not saying that yet, I am going to break this out into a more explicit section based on this feedback

share/doc/wake/logging.md
mwachs5: excellent thanks

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: ok, i think renaming `Bright` to `Bold` is outside the scope of this doc, and I think this thread is mostly resolved...? I decided to keep the colors in the existing table but made it focus on the fact that these are the loggers, because the colors are intrinsic to the loggers

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: ok, I tried to break this out more clearly into things that are about jobs running and "other stuff". let me know what you think.

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: I added these to the help output and added a bit more to this doc. Also realized that --fd:1 and --fd:2 are not real things :)

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: I decided to just leave the colors in place and added an asterisk for --no-tty

share/doc/wake/logging.md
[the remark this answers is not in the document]
mwachs5: ok i think i broke this out more clearly into the different "kinds" of things we are talking about

share/doc/wake/logging.md
mwachs5: added this information in parallel with stdout and stderr

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+and other things that impact what you see on the console when wake is executing or running database commands.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## Loggers
+
+Wake has a concept of loggers,
+to which bytes can be written during wake execution by various sources.
+Based on command line flags, each logger's output will be directed to
+zero or more of wake's outputs, possibly with additional formatting applied.
```
mwachs5, line 13:
```suggestion
zero or more of `wake`'s own output streams, possibly with additional formatting applied.
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
+These must be opened. One example shell command would be, for a wake code that had `mkLogLevel "foo"`:
+
+```
+wake --fd3="foo" 3&>foo.txt
```
mwachs5, line 54:
```suggestion
wake --fd3="foo" mywakecode 3&>foo.txt
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+|(default)          |          |         |         |     x     |      x     |    x     |          |                 |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
+These must be opened. One example shell command would be, for a wake code that had `mkLogLevel "foo"`:
```
mwachs5, line 51:
```suggestion
These must be opened. One example shell command would be, for some wake code that had `mkLogLevel "foo"`:
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+|(default)          |          |         |         |     x     |      x     |    x     |          |                 |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
+and it always appears there.
+
+The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
+These must be opened. One example shell command would be, for some wake code that had `mkLogLevel "foo"`:
```
mwachs5, line 51:
```suggestion
One example shell command would be, for some wake code that had `mkLogLevel "foo"`:
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+### Job Execution
+
+Every `Job` run by wake tracks its own stdout and stderr,
+and when the underlying process prints to stdout or stderr,
+it will go to the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
+Every `Job` also has an "echo", which is the command that the job is executing.
+
+When writing your wake code,
+you can control which logger these `Job`-specific stdout, stderr, and echo are sent to.
```
mwachs5, line 74:
```suggestion
you can control which loggers these `Job`-specific stdout, stderr, and echo are sent to.
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+and when the underlying process prints to stdout or stderr,
+it will go to the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
+Every `Job` also has an "echo", which is the command that the job is executing.
+
+When writing your wake code,
+you can control which loggers these `Job`-specific stdout, stderr, and echo are sent to.
+This will influence whether they appear on the `wake` process's own stdout,
+stderr, or other arbitrary file descriptors.
+These can be controlled with `setPlanStderr`, `setPlanStdout`, and `setPlanEcho` before you `runJob`.
```
mwachs5, line 77:
```suggestion
These can be controlled with `setPlanStdout`, `setPlanStderr`, and `setPlanEcho` before you `runJob`.
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+One example shell command would be, for some wake code that had `mkLogLevel "foo"`:
+
+```
+wake --fd3="foo" mywakecode 3&>foo.txt
+```
+
+These can be concatenated with commas and include the normal log levels, so to capture both `mkLogLevel "foo"` and `logInfo` to a file:
+
+```
+wake --fd3="foo,info" 3&>foo.txt
```
mwachs5, line 60:
```suggestion
wake --fd3="foo,info" mywakecode 3&>foo.txt
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+This can influence what you see on your console when you run wake.
+The logger, in combination with the command line flags you invoke wake with,
+controls *what* you see. The logger also controls what it looks like (what color it is).
+It also controls which *actual* stream it is going to coming out of the executing `wake` process
+(stderr or stdout, or other file descriptors 3-5).
+
+`wake` has a number of built in loggers in wake which all start with `log...`.
+You can also make your own logger using `mkLogLevel`.
+You can use this to customize the color you want it to be shown as (if at all),
+and giving more fine-grained control over which wake output stream it should go to.
```
richardxia, line 27:
```suggestion
You can use this to customize the color you want it to be shown as (if at all)
and give more fine-grained control over which wake output stream it should go to.
```

Grammatically, "customize" and "give" should be the same tense. Also, the comma should be removed because there's no subject in the second line.

share/doc/wake/logging.md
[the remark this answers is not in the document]
richardxia: Sorry, what was the rationale for not mentioning the colors separately from the command-line options? I still feel like it's a bit confusing to see orthogonal data in one table like this.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+|--debug            |      x   |    x    |    x    |     x     |      x     |    x     |          |                 |
+|--verbose          |          |    x    |    x    |     x     |      x     |    x     |          |                 |
+|--quiet            |          |         |         |           |            |    x     |          |                 |
+|--no-tty<sup>*</sup>|         |         |    x    |     x     |      x     |    x     |          |                 |
+|(default)          |          |         |         |     x     |      x     |    x     |          |                 |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on stderr coming out of wake is `logError`,
```
richardxia, line 47: Once again, you are conflating the name of the variable with the name of the logger. You should call it "error" (or probably the "error logger").

You can have multiple variables point to the same object in memory, so the only canonical name will be the name of the actual logger.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+and it always appears there.
+
+The additional command line arguments `--fd:3`, `--fd:4`, `--fd:5` can be used to output to file descriptors 3, 4, 5.
+One example shell command would be, for some wake code that had `mkLogLevel "foo"`:
+
+```
+wake --fd3="foo" mywakecode 3&>foo.txt
```
richardxia, line 54:
```suggestion
wake --fd:3="foo" mywakecode 3>foo.txt
```

The CLI option has a colon in its name. The way you put the `&` in the bash command will cause it to parse in a very different way than you would expect. `foocmd 3&>bar` parses the same as `foocmd 3 &>bar`, and the latter means call `foocmd` with a single argument of `3` and redirect both stdout and stderr to a file named `foo.txt`. (`&>` is shorthand for redirect both stdout and stderr.)

`&` unfortunately has lots of very different meanings in bash depending on exactly where it is. Without any redirections, it means background the process. On the left-hand side of a `>` redirection, `&` means redirect both stdout and stderr to the thing on the right. On the right-hand side of the `>` redirection, if it is followed by a number, then it means redirect to the file descriptor with that number. `>&3` means redirect stdout to file descriptor 3; `>3` means redirect stdout to file named `3`.

In your example here, you don't want any `&` at all because you're only trying to redirect fd 3 to a normally named file.

share/doc/wake/logging.md
richardxia, line 60:
```suggestion
wake --fd:3="foo,info" 3>foo.txt
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+
+## What Gets Sent To Loggers
+
+### Job Execution
+
+Every `Job` run by wake tracks its own stdout and stderr,
+and when the underlying process prints to stdout or stderr,
+it will go to the `Job`-specific stdout and stderr,
+not the stdout and stderr of the `wake` process itself.
+Every `Job` also has an "echo", which is the command that the job is executing.
```
richardxia, line 71: I don't think I actually understand what this last part means. Is this some wake feature that I don't know about? What exactly is the "echo"? Is it a property on the `Job` tuple? Is it a column in the jobs table of the job database? What type of thing is an "echo"? stdout and stderr are file-like streams on Unix processes that most programmers are very familiar with. "echo" clearly is not one of those, but you don't identify the type thing that "echo" is, so this sentence is difficult to understand.

The text in the following paragraphs don't really help me much either, so I think something needs to be clarified, since seeing "echo" mentioned alongside "stdout" and "stderr" is very unexpected to me.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+
+The results of calling `print`/`println` within `wake` code are sent to `logReport` by default.
+To send them somewhere else, use `printLevel`, `printlnLevel` instead.
+
+The final output of whatever wake function or expression is called is always directed to
+wake's `stdout` unless `--quiet` is set.
+This is outside of the logger mechanism, so it is not possible to redirect this within wake
+code or with command line flags.
+
+Similarly, the "spinner" and progress bar is always sent to wake's `stdout` unless `--no-tty`
```
richardxia, line 98: Are the "spinner" and "progress bar" two different things? I can only think of one animated thing I see, which I guess must be the progress bar, but sometimes people use "spinner" as a generic term that can also mean a "progress bar" (that might be my web development influence).

richardxia (commented):
The "What Gets Sent To Loggers" section is excellent, but I still have some concerns about the correctness or the presentation of some of the information of the "Loggers" section.

share/doc/wake/logging.md
mwachs5: yeah now I am curious what actually happens if I `mkLogLevel "mega"` in two places... I assume that both created loggers would still be impacted by e.g. `--stderr=mega`

share/doc/wake/logging.md
mwachs5: I learned this too in the process of writing the doc, The 'echo' is referring to/identical to the command you're actually going to execute with the job (if I understand correctly). I could just call it "the command", but setPlanEcho is what you use to set where it is logged, so 🤷 . In other words, in the database you are storing a bunch of things, like the environment variables, the command you are going to execute, the stderr and stdout. The "echo" is identical to "the command you are going to execute" (I think? maybe there is other stuff in the echo like the label of the job or something?)

mwachs5: https://github.com/sifive/wake/blob/63151ddf4a75ed597a9b92588569be6fe28fa677/share/wake/lib/system/sources.wake#L45

share/doc/wake/logging.md
mwachs5: I'm not 100% sure if this sentence is true about going to stdout...

share/doc/wake/logging.md
mwachs5: I'm not quite sure what you are suggesting. I *think* you saying that the source of truth of the logger is the string itself, so "error" or "mega", and `mkLogLevel` just makes something that will direct to that source-of-truth logger? But I think that's not quite true, because I can `mkLogLevel "mega" Green` and `mkLogLevel "mega" Blue` and those are different loggers, coupling the level and the color displayed...

mwachs5: let me re-read what I've written and your comments and try to be more consistent. I want to use these definitions:

* log level is something like "error", "info", "foo"
* a logger couples a log level and a color
* command line arguments influence what happens to things sent to loggers of a given log level
* outputs are directed to a logger

WDYT?

share/doc/wake/logging.md
richardxia: But what "type" of thing is an "echo"? There's no equivalent concept in Unix, so the explanation above this sentence that talks about the "underlying process" doesn't make sense when applied to "echo".

It sounds like this thing is basically just equivalent to a wake `printLogLevel` call that wake does just before running the Job with the command that the Job is about to run. If this is the case, then I think it would be best for you to explain it separately from stdout and stderr, because those actually correspond to things in the Unix process model (streams that are accessible via file descriptors within the process), whereas "echo" appears to only exist in wake-land.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+
+This article gives a high level overview of how wake logging works,
+and other things that impact what you see on the console when wake is executing or running database commands.
+Check out [print.wake](https://github.com/sifive/wake/blob/master/share/wake/lib/core/print.wake)
+for commented code.
+
+## Loggers
+
+Wake has a concept of loggers,
+to which bytes can be written during wake execution by various sources.
```
mwachs5, line 11:
```suggestion
to which text can be written during wake execution by various sources.
```

share/doc/wake/logging.md
richardxia: > I'm not quite sure what you are suggesting. I think you saying that the source of truth of the logger is the string itself, so "error" or "mega", and mkLogLevel just makes something that will direct to that source-of-truth logger? But I think that's not quite true, because I can mkLogLevel "mega" Green and mkLogLevel "mega" Blue and those are different loggers, coupling the level and the color displayed...

No, I'm saying in

```wake
def foo = mkLogLevel "mega" Green
def bar = foo
```

`foo` and `bar` are wake variable names that point to the same object in memory, which is a logger by the name of "mega". I'm saying it's incorrect or at least a little bit confusing for you to refer to the loggers by their wake variable names rather than by the value of the `name` field on these objects.

share/doc/wake/logging.md
mwachs5: > It sounds like this thing is basically just equivalent to a wake printLogLevel call that wake does just before running the Job with the command that the Job is about to run.

Sort of, but it is controlled in analagous way as SetPlanStderr and SetPlanStdout. I could break this out separately into the "other stuff" section (i had it there originally, but then moved it back here, because this DOES appear in the database as it goes with an executing job). And, you set it like you set setPlanStderr, with setPlanEcho. 

Maybe I can just not introduce a noun of 'echo' and just call it 'the command you are going to execute' except for saying you override that with `setPlanEcho`.

mwachs5: or "command echo" or "echoed command"?

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+
+The results of calling `print`/`println` within `wake` code are sent to `logReport` by default.
+To send them somewhere else, use `printLevel`, `printlnLevel` instead.
+
+The final output of whatever wake function or expression is called is always directed to
+wake's `stdout` unless `--quiet` is set.
+This is outside of the logger mechanism, so it is not possible to redirect this within wake
+code or with command line flags.
+
+Similarly, the "spinner" and progress bar is always sent to wake's `stdout` unless `--no-tty`
```
mwachs5, line 98:
```suggestion
Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty`
```

share/doc/wake/logging.md
mwachs5: right, i think this is in aligment with what I am saying... the strings are really their log level and what the command lines are dealing with. There may be many loggers out there (by whatever wake variable name) with the same stringy log level.

richardxia: I think some of my misunderstanding is that you're trying to use terminology that, although makes more sense, is inconsistent with the wake code code.

> log level is something like "error", "info", "foo"
> a logger couples a log level and a color

The confusing thing is that the wake code calls the second thing a `LogLevel` and the first thing a `name`.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+## Loggers
+
+Wake has a concept of loggers,
+to which bytes can be written during wake execution by various sources.
+Based on command line flags, each logger's output will be directed to
+zero or more of `wake`'s own output streams, possibly with additional formatting applied.
+The wake codebase and standard libraries often refer to a _logger_ as a `logLevel`.
+
+What's the difference between the different loggers?
+Which of wake's own output streams the results are sent to.
```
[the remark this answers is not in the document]
mwachs5: I completely rewrote all this, so marking this as resolved

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+This can influence what you see on your console when you run wake.
+The logger, in combination with the command line flags you invoke wake with,
+controls *what* you see. The logger also controls what it looks like (what color it is).
```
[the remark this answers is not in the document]
mwachs5: I rewrote all this, so please take a look

share/doc/wake/logging.md
```diff
@@ -0,0 +1,111 @@
...
+It also controls which *actual* stream it is going to coming out of the executing `wake` process
+(stderr or stdout, or other file descriptors 3-5).
+
+`wake` has a number of built in loggers in wake which all start with `log...`.
```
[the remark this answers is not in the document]
mwachs5: I explicitly added a table to distinguish what this doc is calling things vs what wake calls them in the code, and then tried to use the terms consistently/correctly -- hopefully this is more clear now.

share/doc/wake/logging.md
mwachs5: I added a table to distinguish terms used in this doc vs what wake calls them internally, then tried to use the former correctly and consistently throughout the doc.

share/doc/wake/logging.md
mwachs5: (I made it just say progress bar, still not sure about the --quiet)

mwachs5:
Not sure why this is actually failing CI, perhaps there are some tests for the docstrings / `--help` output?

ag-eitilt:
> Not sure why this is actually failing CI, perhaps there are some tests for the docstrings / `--help` output?

We've had recurring (upstream) issues with the MacOS image; looks like they've popped up again.  Nothing to do with the PR.

share/doc/wake/logging.md
mwachs5: Reworded all this, I think it is more clear now (I brought the 'echo' part back down to the "other stuff" area)

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+|-------------------|-----------------|-----------|
+| logDebug          |     Blue        | "debug"   |
+| logInfo           |     Dim         | "info"    |
+| logEcho           |      N/A        | "echo"    |
+| logReport         |    Magenta      | "report"  |
+| logWarning        |    Yellow       | "warning" |
+| logError          |     Red         | "error"   |
+| logNever          |     N/A         | "null"    |
+| mkLogLevel "foo" Green |  Green     | "foo"     |
+| mkLogLevel "foo"  Bright Green | Bright Green   | "foo"      |
```
ag-eitilt, line 40:
```suggestion
| mkLogLevel "foo" (Some Green) |  Green     | "foo"     |
| mkLogLevel2 "foo" (Some Green) (Some Bright) | Bright Green   | "foo"      |
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+|--debug              |      x   |    x    |    x    |     x     |      x     |    x     |         |        |
+|--verbose            |          |    x    |    x    |     x     |      x     |    x     |         |        |
+|--quiet              |          |         |         |           |            |    x     |         |        |
+|--no-tty<sup>*</sup> |          |         |    x    |     x     |      x     |    x     |         |        |
+|(default)            |          |         |         |     x     |      x     |    x     |         |        |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on wake's stderr is outputs from loggers
+with log level "error" (e.g. `logError`).
```
ag-eitilt, line 69: @JakeSiFive, @V-FEXrt: Do we make any tighter guarantees about the stderr than we do the stdout?  It's the only thing that appears once you hit the wake code, definitely, but there might be more things output by the tool itself (slow disc warnings?) that appear here.

Also, I'm still not quite clear on how both stdout and stderr naming `"error"` doesn't result in duplicated output as Wake prints it to both file descriptors; we definitely don't have to go into the details in this document, but it might be a nice thing to briefly mention, especially if the mechanism is less-than-intuitive (e.g. something like *only* Job stderr is put on Wake's stderr).

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+not the stdout and stderr of the `wake` process itself.
+
+When writing your wake code,
+you can control which loggers these `Job`-specific stdout and stderr are sent to.
+This will influence whether they appear on the `wake` process's own stdout,
+stderr, or other arbitrary file descriptors.
+These can be controlled with `setPlanStdout`, `setPlanStderr` before you `runJob`.
+These default to `logInfo` and `logError` respectively.
+
+Regardless of to which logger you direct the `Job`'s `Stderr` and `Stdout`,
```
ag-eitilt, line 100: This is perfectly valid grammar, but the double preposition feels like it stumbles a bit on the first reading.  Maybe rearrange the "to"?
```suggestion
Regardless of the logger to which you direct the `Job`'s `Stderr` and `Stdout`,
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+## Consideration for `printLevel` vs `setPlanStdout`
+
+When writing wake code where one wants to see a `Job`'s stderr/stdout
+regardless of whether that job executes in a given `wake` execution
+(e.g. a job which collects and reports passing test results),
+it is better to  `setPlanStdout logNever`, then explicitly `getJobStdout` and
+send the result to a `printLevel`.
+This is because the `Job`'s stdout is only sent to the logger if it actually executes
+(vs its outputs being retrieved from cache),
+whereas the `printLevel` is dependent on just the wake execution.
```
ag-eitilt, line 152: This is definitely a good note to include, and nicely illustrates a non-obvious point about Wake output printing.  However, it might be good to explicitly counterbalance that example with the fact that `getJobStdout` -> `println`/`printlnLevel` (and I do think you should use the "ln" version(s) both times in this paragraph) will only print the entire output *after* the Job has finished, while `setPlanStdout` allows it to be streamed line-by-line as it's printed by the Job.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+wake's `stdout` unless `--quiet` is set.
+This is outside of the logger mechanism,
+so it is not possible to redirect this within wake code or with command line flags.
+
+Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty`
+or `--quiet` is set.
+
+## What is Displayed to user on Wake Database Commands
+
+The `--verbse` and `--metadata` flags impact what is shown per-job by the database
```
ag-eitilt, line 124:
```suggestion
The `--verbose` and `--metadata` flags impact what is shown per-job by the database
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+(`--last, --failed, --job`, etc) commands.
+They do not impact which `Job`s information is shown for.
```
ag-eitilt, line 126: Specifically, this is talking about `--verbose` when it's given alongside a database command.  `--verbose` *without* a database command is described in the table above.  Maybe put the predicate first?
```suggestion
When used with a database command (`--last, --failed, --job`, etc),
the `--verbose` and `--metadata` flags impact what is shown per-job by the database.
They do not impact the selection of the `Job`s for which that information will be shown.
```

ag-eitilt (approved):
Other than the `--verbose` typo and the `mkLogLevel` example fixes, I've made a number of comments which are subjective editing, or which might be worth a deeper look in follow-up PRs but which aren't critical to solve before we merge the first version of this.  Thanks for all the work!

V-FEXrt (approved):

share/doc/wake/logging.md
V-FEXrt: > Do we make any tighter guarantees about the stderr than we do the stdout? 
No, stderr and stdout are not safe to depend on in any general context. The only guaranteed way to have a clean output/capture is to create a new stream (because anyone can output to the default streams) and then to redirect said stream to an fd (which may be stdout/stderr if desired)

mwachs5: >  I'm still not quite clear on how both stdout and stderr naming "error" doesn't result in duplicated output as Wake prints it to both file descriptors; 

I am also not clear on this, this came out of just me reading the code so I could be wrong.

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+for commented code.
+
+## Loggers & Log Levels
+
+Wake has a concept of loggers,
+to which text can be written during wake execution by various sources.
+A logger combines a string log level (e.g. "error", "info", "my_custom_level")
+and a display format (color and intensity).
+There can be multiple loggers with the same log level, though this is
+not a common use case.
```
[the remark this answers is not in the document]
mwachs5: ah, ok, maybe I should change this to say "is discouraged / not well-supported"?

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+|-------------------|-----------------|-----------|
+| logDebug          |     Blue        | "debug"   |
+| logInfo           |     Dim         | "info"    |
+| logEcho           |      N/A        | "echo"    |
+| logReport         |    Magenta      | "report"  |
+| logWarning        |    Yellow       | "warning" |
+| logError          |     Red         | "error"   |
+| logNever          |     N/A         | "null"    |
+| mkLogLevel "foo" (Some Green) |  Green     | "foo"     |
+| mkLogLevel2 "foo" (Some Green) (Some Bright) | Bright Green   | "foo"      |
```
mwachs5, line 40:
```suggestion
| mkLogLevel2 "bar" (Some Green) (Some Bright) | Bright Green   | "bar"      |
```

if the colors aren't well supported with multiple loggers with the same "log level" aka name string using different colors, let's not put it here

share/doc/wake/logging.md
mwachs5: Ok yeah I should tighten up there may be other cruft on stderr

share/doc/wake/logging.md
mwachs5, line 15:
```suggestion
There can be multiple loggers with the same log level,
though colors may not display as expected.
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+so it is not possible to redirect this within wake code or with command line flags.
+
+Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty`
+or `--quiet` is set.
+
+## What is Displayed to user on Wake Database Commands
+
+The `--verbose` and `--metadata` flags impact what is shown per-job by the database
+(`--last, --failed, --job`, etc) commands.
+They do not impact which `Job`s information is shown for.
```
mwachs5, line 126:
```suggestion
When used with a database command (`--last, --failed, --job`, etc),
the `--verbose` and `--metadata` flags impact what is shown per-job by the database.
They do not impact the selection of the `Job`s for which that information will be shown.
```

share/doc/wake/logging.md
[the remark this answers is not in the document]
ag-eitilt: Probably "not well-supported" for now.  I can definitely see use cases for it and I don't want to say "is discouraged" now if we implement everything required for them in the future, but at the moment doing so is definitely not ideal.

share/doc/wake/logging.md
ag-eitilt: Oh, I didn't even consider that having the same name could be read as defining these in the same codebase; I just figured it was "insert your name here subject to the other described restrictions".  Good fix!

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+|--stdout="foo,info"  |          |    x    |         |           |            |          |         |    x   |
+|--debug              |      x   |    x    |    x    |     x     |      x     |    x     |         |        |
+|--verbose            |          |    x    |    x    |     x     |      x     |    x     |         |        |
+|--quiet              |          |         |         |           |            |    x     |         |        |
+|--no-tty<sup>*</sup> |          |         |    x    |     x     |      x     |    x     |         |        |
+|(default)            |          |         |         |     x     |      x     |    x     |         |        |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+Unless overridden with `--stderr=...` the *only* thing that ever appears on wake's stderr is outputs from loggers
```
mwachs5, line 68:
```suggestion
Unless overridden with `--stderr=...` the only logger output that ever appears on wake's stderr is outputs from loggers
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,152 @@
...
+|--debug              |      x   |    x    |    x    |     x     |      x     |    x     |         |        |
+|--verbose            |          |    x    |    x    |     x     |      x     |    x     |         |        |
+|--quiet              |          |         |         |           |            |    x     |         |        |
+|--no-tty<sup>*</sup> |          |         |    x    |     x     |      x     |    x     |         |        |
+|(default)            |          |         |         |     x     |      x     |    x     |         |        |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+Unless overridden with `--stderr=...` the only logger output that ever appears on wake's stderr is outputs from loggers
+with log level "error" (e.g. `logError`).
```
mwachs5, line 69:
```suggestion
Unless overridden with `--stderr=...`,
the only logger output that ever appears on wake's stderr is outputs from loggers with log level "error"
(e.g. `logError`).
```

share/doc/wake/logging.md
```diff
@@ -0,0 +1,153 @@
...
+|--verbose            |          |    x    |    x    |     x     |      x     |    x     |         |        |
+|--quiet              |          |         |         |           |            |    x     |         |        |
+|--no-tty<sup>*</sup> |          |         |    x    |     x     |      x     |    x     |         |        |
+|(default)            |          |         |         |     x     |      x     |    x     |         |        |
+
+<sup>*</sup> `--no-tty` also prevents any coloring from being applied.
+
+Unless overridden with `--stderr=...`,
+the only logger output that ever appears on wake's stderr is outputs from loggers with log level "error"
+(e.g. `logError`).
```
mwachs5, line 70:
```suggestion
Unless overridden with `--stderr=...`,
the only logger output that ever appears on wake's stderr is from those with log level "error"
(e.g. `logError`).
```

share/doc/wake/logging.md
mwachs5: have added to this paragraph, thanks!

mwachs5:
I am not able to merge this myself, so feel free to merge if you are happy with it!
The only open question (from me) is whether the progress bar is impacted by `--quiet` or not

share/doc/wake/logging.md
```diff
@@ -0,0 +1,157 @@
...
+The results of calling `print`/`println` within `wake` code are sent to `logReport` by default.
+To send them somewhere else, use `printLevel`, `printlnLevel` instead.
+
+The final output of whatever wake function or expression is called is always directed to
+wake's `stdout` unless `--quiet` is set.
+This is outside of the logger mechanism,
+so it is not possible to redirect this within wake code or with command line flags.
+
+Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty`
+or `--quiet` is set.
```
mwachs5, line 121:
```suggestion
Similarly, the progress bar is always sent to wake's `stdout` unless `--no-tty` is set.
```

PR #1107, opened 2023-02-03
Create TermInfoBuf and FdBuf

V-FEXrt:
This should be the final prep PR and the next PR will include actually emitting color in `--failed` and the like.

This PR introduces TermInfoBuf, FdBuf, and the associated tests but does not directly use any of them.

### DIFF
--- a/src/runtime/term.cpp
+++ b/src/runtime/term.cpp
@@ -23,8 +23,333 @@
 #ifndef __EMSCRIPTEN__
 
 #include <curses.h>
+#include <sys/ioctl.h>
 #include <term.h>
+#include <termios.h>
 #include <unistd.h>
+
+#include <cstring>
+#include <set>
+#include <sstream>
+
+#define XTERM256_FOREGROUND_ESCAPE 38
+#define XTERM256_BACKGROUND_ESCAPE 48
+#define XTERM256_8BIT_ESCAPE 5
+
+void TermInfoBuf::clear_codes() {
+  codes.clear();
+  cur_code = -1;
+}
+
+// Puts a character to `buf`
+void TermInfoBuf::put(char c) { buf.sputc(c); }
+
+// Puts a string to `buf`
+void TermInfoBuf::putstr(const char *str) { buf.sputn(str, strlen(str)); }
+
+// Updates `cur_code` with the next read digit
+void TermInfoBuf::update_code(char digit) {
+  if (cur_code < 0) cur_code = 0;
+  cur_code = cur_code * 10 + digit;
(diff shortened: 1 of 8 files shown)


src/runtime/term.cpp
```diff
@@ -23,9 +23,325 @@
...
+        put(c);
+      }
+      break;
+    case State::esc_state:
+      // TODO: Add other things to this state,
+      // specifically all the escape sequences we
+      // we want to ignore. There should be two
+      // cases, one character cases and two character
+      // cases. For the two character case we'll need
+      // an extra case.
```
JakeSiFive, line 275: I think this todo is todone

src/runtime/term.cpp
```diff
@@ -23,9 +23,325 @@
...
+
+int FdBuf::overflow(int c) {
+  if (c == EOF) {
+    return EOF;
+  }
+
+  char ch = c;
+  ssize_t r = write(fd, &ch, sizeof(ch));
+  if (r == -1) {
+    // TODO(jake): what should be done here?
```
JakeSiFive, line 337: We should just roll over and die. The only error that isn't catastrophic here is `EINTR` so what you could do if you wanted to be smart is this

```
do {
  size_t r = write(fd, &ch, sizeof(ch));
  if (r == -1 && errno != EINTR) { /* roll over and die in a hole */ }
} while (errno != EINTR);
```

That's a maximally safe way to do the write I suppose.

src/runtime/term.cpp
```diff
@@ -58,42 +402,47 @@ const char *term_intensity(int code) {
 
 const char *term_normal() { return sgr0 ? sgr0 : ""; }
 
-bool term_init(bool tty_) {
+bool term_init(bool tty_, bool skip_atty) {
```
JakeSiFive, line 405: please put a todo to reconsider skip_atty after full color support is added.

JakeSiFive (approved):

src/runtime/term.cpp
```diff
@@ -23,9 +23,323 @@
...
+
+static void write_num(std::streambuf &buf, int x) {
+  std::string s = std::to_string(x);
+  buf.sputn(s.c_str(), s.size());
+}
+
+// In case we have to back up from the num_state to default_state,
+// we put all previous characters. This happens if at some point
+// we have to flush all thus-far collected numbers because the
+// escape code is invalid. There is some special edge case handeling
```
mwachs5, line 233:
```suggestion
// escape code is invalid. There is some special edge case handling
```

PR #1109, opened 2023-02-07
Run jobs in pseudoterminal and re-emit with color

V-FEXrt:
This is the last change in the stack of changes to improve the output format of previously ran jobs. It runs the job in a pseudo terminal to capture the colored output then replays the colors when requested to be printed back out.


share/wake/lib/gcc_wake/pkgconfig.wake
```diff
@@ -51,6 +51,7 @@ target pkgConfigImp args =
     require Some output =
         result.getJobStdout
         | getWhenFail ""
+        | replace `[\r]` ""
```
ag-eitilt, line 54: Is the character group for clarity here?  (Which is likely enough of a purpose to leave it in.)
```suggestion
        | replace `\r` ""
```

ag-eitilt (commented):
Not reviewing since the Wake code is really not the focus of the PR, but it's exciting watching these make their way through!

mwachs5:
can you share an in-color screenshot?

share/wake/lib/gcc_wake/pkgconfig.wake
JakeSiFive: yep, I asked Ashley to put it in there like that.

JakeSiFive:
Screen Shot 2023-02-06 at 6 31 21 PM

JakeSiFive:
Notice that when wake prints the error while running, it prints out in all yellow, when it prints it out in --failed however it prints it out in color as you would have seen it if you had ran it from the terminal yourself.

src/runtime/status.cpp
```diff
@@ -299,16 +302,26 @@ void status_set_bulk_fd(int fd, const char *streams) {
...
       int colour = s.colour % 8;
       int intensity = s.colour / 16;
-      if (colour != TERM_DEFAULT) write_all_str(s.fd, term_colour(colour));
-      if (intensity != TERM_DEFAULT) write_all_str(s.fd, term_intensity(intensity));
+      if (colour != TERM_DEFAULT) {
+        raw << term_colour(colour);
+      }
+      if (intensity != TERM_DEFAULT) {
+        raw << term_intensity(intensity);
+      }
```
JakeSiFive, line 319: nit: can be written without brackets

....I'm kidding lol, don't change this

JakeSiFive (approved):
Very happy with this change. After the last change went in this change was so nice and small!

mwachs5:
> Notice that when wake prints the error while running, it prints out in all yellow, when it prints it out in --failed however it prints it out in color as you would have seen it if you had ran it from the terminal yourself.

So I can reflect this change properly in my document, are you saying that the `logLevel` that the stdout/stderr is not saved to the database, so it does not impact the color displayed in this view? 

(A little confused how this is different than what is currently on master -- do you normally strip all the color encodings before saving to the database, now we save and reflect them)?

JakeSiFive:
> > Notice that when wake prints the error while running, it prints out in all yellow, when it prints it out in --failed however it prints it out in color as you would have seen it if you had ran it from the terminal yourself.
> 
> So I can reflect this change properly in my document, are you saying that the `logLevel` that the stdout/stderr is not saved to the database, so it does not impact the color displayed in this view?
> 
> (A little confused how this is different than what is currently on master -- do you normally strip all the color encodings before saving to the database, now we save and reflect them)?

So to answer this question I need to make sure you're aware of a few things about how color in terminals works (which may be review, sorry if it is).

Color is output by special escape sequences. These escape sequences do play well files and reading those files later. For instance if you opened a file with color escape sequences in it in vim it might be unreadable. To combat this, most tools will check to see if they're piping to a terminal, and they will also inspect the `TERM` environment variable to determine what escape sequences to emit (although there is a solid subset defined by ANSI that works basically everywhere). Previously wake was creating pipes and starting up subprocesses with their stdout and stderr set as pipes. When the subprocess ran it would call `isatty` on the stdout/stderr to see if it could print in color or not. Previously all applications would see that this was false and would fail to print in color. In wake then we would color coat the output based on what stream it came from. It was decided way back when I wrote this code that that stream coloring had to stay. So I wrote a filter to filter out the escape sequences (and also translate between terminal colors). So when you see the stream output while wake is running you used to be seeing raw output from the program that wasn't filtered, now you're seeing filtered output that was output with color but the filtered output was then recolored in monochrome by wake so that the old behavior was matched. But!! We still store the raw output exactly as the program output it in the database which means that we can output stdout/stderr in color as the program intended with --failed.

So to sumerize, the big change here is that we're running programs in psuedo terminals now so that they actully use color. To maintain the old behavoir of wake itself we had to do some crazy complicated color filtering/translation shenanigans. Because of this --failed will now be in color with bolds and such, where as before this change there would be no color/bolding etc...

V-FEXrt:
Closing this. #1120 has all the changes that matter from this.

PR #1118, opened 2023-02-09
Massively improved filtering and testing for TermInfoBuf

JakeSiFive:
This change drastically improves TermInfoBuf and its testing. I now have tests that near comprehensively generate content over a range of inputs that should be a super set of the vt100 mode escape sequences seen here: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Single-character-functions and the latest code passes all of these tests. Additionally to ensure Unicode characters are not unduly mutated (they shouldn't be because escape sequences tend to be 7-bit) we've also added explicit Unicode support so that all Unicode is entirely unfettered.

### DIFF
--- a/src/runtime/term.cpp
+++ b/src/runtime/term.cpp
@@ -254,6 +254,11 @@ void TermInfoBuf::flush_nums() {
 
 int TermInfoBuf::sync() { return buf.pubsync(); }
 
+static inline bool isIgnoredSingleByteCommand(char c) {
+  // CR, BEL, BS, ENQ, SI, SO
+  return c == '\r' || c == 0x7 || c == 0x8 || c == 0x5 || c == 0xF || c == 0xE;
+}
+
 int TermInfoBuf::overflow(int c) {
   static const char tcis[] = " #%()*+-./";
   static std::set<char> two_character_ignore_starters(tcis, tcis + sizeof(tcis));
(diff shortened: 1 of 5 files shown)


tools/wake-unit/term.cpp
```diff
@@ -384,3 +387,250 @@ TEST(term_dumb_gcc_smoke_test) {
...
+
+TEST(term_git_smoke_test) {
+  CONVERTS_TO("\033[?1h\033=\r2023-02-08 12:31:48 -0600\033[m\r\n\r\033[K\033[?1l\033>",
+              "2023-02-08 12:31:48 -0600\033(B\033[m\n");
+}
+
+TEST(term_dumb_git_smoke_test) {
+  CONVERTS_TO_DUMB("\033[?1h\033=\r2023-02-08 12:31:48 -0600\033[m\r\n\r\033[K\033[?1l\033>",
+                   "2023-02-08 12:31:48 -0600\n");
+}
```
JakeSiFive, line 399: These tests ensure that the immediate problem is fixed

tools/wake-unit/term.cpp
```diff
@@ -384,3 +387,250 @@ TEST(term_dumb_gcc_smoke_test) {
...
+  CONVERTS_TO_DUMB("\xE0\x80\r", "\xE0\x80\r");
+  CONVERTS_TO_DUMB("\xE0\x80\033", "\xE0\x80\033");
+
+  CONVERTS_TO_DUMB("\xF0\x80\x80\r", "\xF0\x80\x80\r");
+  CONVERTS_TO_DUMB("\xF0\x80\x80\033", "\xF0\x80\x80\033");
+
+  // However if we go just one more we'll fall out of the unicode states
+  // and special characters should be removed
+  CONVERTS_TO_DUMB("\xF0\xFF\xFF\xFF\r", "\xF0\xFF\xFF\xFF");
+}
```
JakeSiFive, line 433: These are basically testing that unicode works the way that we expect it to. Namely that it interprets the leading UTF-8 character as a marker to get the length of the character.

tools/wake-unit/term.cpp
```diff
@@ -384,3 +387,250 @@ TEST(term_dumb_gcc_smoke_test) {
...
+      auto p = func(rng);
+      input += p.first;
+      expected += p.second;
+    }
+
+    // Now we just hope they're equal!
+    CONVERTS_TO_DUMB(input, expected)
+        << "\ninput = " << json_escape(input) << "\nfunctions = " << functions << "\n";
+  }
+}
```
JakeSiFive, line 636: This beast of a test attempts to murder TermBufInfo. Unfortunately its kind of hard to generate expected strings for the non-dumb case but given that all the changes in this are fully exercised by the dumb filtering (and that's also what we're most concerned about) we just leave it as is.

src/runtime/term.cpp
```diff
@@ -263,33 +268,99 @@ int TermInfoBuf::overflow(int c) {
...
+        put('\n');
+        break;
+      }
+
+      // We want to ignore characters like BEL, CR, etc...
+      if (isIgnoredSingleByteCommand(c)) {
+        break;
+      }
+      // Check for unicode continuations. It's worth
+      // noting that this makes TermInfoBuf incompaitable
```
V-FEXrt, line 284: typo: incompatible

V-FEXrt (approved):
Happy to see the detailed fuzz tests in there!

I'll try this change out locally and report back but LGTM

V-FEXrt:
Okay! everything looks good locally!

V-FEXrt:
Well maybe not. looks like CI is still falling over hard

JakeSiFive:
ok Ashley got this working on his side, his changes are coming!

PR #1119, opened 2023-02-09
Add helper document for F# developers coming to Wake

bmitc:
* Renamed `from-scala.adoc` to `wake-for-scala-developers.md`
* Added `wake-for-fsharp-developers.md`

The latter document is one that I started up when I first joined SiFive and started developing in Wake. I've added to it a bit, but for the most part it has become a static document. Since there is an existing and similar document for Scala already in the Wake repository, I thought I would add in my F# document to aid in any transition for any future developer coming to Wake from either F# or even OCaml (given F# and OCaml's several similarities).

The rename was to make the document names consistent and a bit more explicit for what the files contain. I could not find any reference to the `from-scala.adoc` document.


share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
+# Wake for F# developers
+
+F# and Wake are both ML-dialects, so they share very similar heritages as far as programming languages go. The major difference is that F# is intended to be a general-purpose programming language and is multi-paradigm since it also supports imperative and OOP programming, whereas Wake is intended to be a programming language to support a build system and does not support the imperative or OOP programming paradigms. However, while simple, the Wake programming language is rather powerful.
```
ag-eitilt, line 3: Slight tweak to avoid suggesting Wake's not sufficiently powerful.
```suggestion
F# and Wake are both ML-dialects, so they share very similar heritages as far as programming languages go. The major difference is that F# is intended to be a general-purpose programming language and is multi-paradigm since it also supports imperative and OOP programming, whereas Wake is intended to be a programming language to support a build system and does not support the imperative or OOP programming paradigms. However, while more simple, the Wake programming language is rather powerful.
```

share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
...
+
+> [1; 2; 3];;
+val it: int list = [1; 2; 3]
+```
+
+In Wake, cons is [`,`](https://github.com/sifive/wake/blob/master/share/wake/lib/core/list.wake#L43-L50) and the empty list is [`Nil`](https://github.com/sifive/wake/blob/master/share/wake/lib/core/list.wake#L41-L42). There is currently no literal list syntax in Wake. You often need to surround a list value with parentheses, such as `(1, 2, 3, Nil)`.
+
+```wake
+> wake -vx '1, 2, 3, Nil'
+1, 2, 3, Nil: List Integer = 1, 2, 3, Nil
```
ag-eitilt, line 48: Wake does indeed duplicate the list, but it seems like that might be a bug.  Are there cases of more complex data where that's helpful/renders differently?

share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
...
+    Rectangle (length: Double) (width: Double)
+```
+
+## Pattern Matching
+
+F#
+```fsharp
+let rec area shape =
+    match shape with
+    | Circle radius             -> radius * radius
```
ag-eitilt, line 75:
```suggestion
    | Circle radius             -> pi * radius * radius
```

share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
...
+```wake
+export def area (shape: Shape) =
+    match shape
+        Circle radius          = pi *. (radius ^. 2.0)
+        Square side            = area (Rectangle side side)
+        Rectangle length width = length *. width
+```
+
+Note that in F#, recursive functions require the `rec` keyword, whereas in Wake, recursive functions require no special keyword.
+
```
ag-eitilt, line 90: I'm guessing this got left behind after reorganizing the file at some point?
```suggestion
```

share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
...
+// This is a record with a generic type parameter
+type Person<'T> = {Name: Name; JobFunction: 'T}
+```
+
+Wake
+```wake
+# This is a tuple
+export tuple Name =
+    First: String
+    Last: String
```
ag-eitilt, line 128: It's not as clean, but I'm guessing the F# dev would want all the accessors visible?
```suggestion
export tuple Name =
    export First: String
    export Last: String
```

share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
...
+
+# This a value of type `Name`
+export def name = Name "John" "Smith"
```
ag-eitilt, line 131: Here, on the other hand, that's a bit of extra noise.
```suggestion
def name = Name "John" "Smith"
```

share/doc/wake/wake-for-fsharp-developers.md
```diff
@@ -0,0 +1,150 @@
...
+
+# This is a tuple with a type parameter
+export tuple Person a =
+    Name: Name
+    JobFunction: a
+
+# This is a value of type `Person Integer`
+export def person = Person (Name "John" "Smith") 4
```
ag-eitilt, line 139: These two may or may not want the same changes (it would be nice to provide the sleaker not-export-every-line tuple format).

ag-eitilt (changes requested):
This definitely wants more work at some point, but the Scala version isn't all that much more extensive and this isn't bad as-is.  A few small fixes, a few general comments:

> Renamed `from-scala.adoc` to `wake-for-scala-developers.md`

.adoc and .md are different filetypes with incompatible syntax; you're going to want to fix the extension.

share/doc/wake/wake-for-fsharp-developers.md
JakeSiFive: Nope, this is correct. It prints out `<what you asked for>: <type> = <normal form>` in verbose mode. If what you ask for is in normal form it will be duplicated

share/doc/wake/wake-for-fsharp-developers.md
bmitc: That's a good point, but I already added it from above. Haha. I think I'll leave it for now with the `export`s added it for consistency. It might make sense to later update this section to point to some more general Wake documentation.

share/doc/wake/wake-for-fsharp-developers.md
bmitc: It is, but I think I'll leave it to match the F# code since I don't mark the binding as private there (bindings are public in the module by default). The binding was actually missing, but I added it [here](https://github.com/sifive/wake/pull/1119/commits/661f3f971e4c9acae7e6d4d0a8592de338190cc2).

share/doc/wake/wake-for-fsharp-developers.md
bmitc: Good call. [Updated here.](https://github.com/sifive/wake/pull/1119/commits/661f3f971e4c9acae7e6d4d0a8592de338190cc2)

share/doc/wake/wake-for-fsharp-developers.md
bmitc: Haha. Whoops. It should actually be `System.Math.PI` (assuming there's no declarations). Thanks! [Fixed here.](https://github.com/sifive/wake/pull/1119/commits/661f3f971e4c9acae7e6d4d0a8592de338190cc2)

share/doc/wake/wake-for-fsharp-developers.md
bmitc: I hadn't reviewed that since I wrote it I think, and so I removed/reworded a large portion of the middle, as it's pretty much irrelevant to a doc like this I think. Wake should already be sold by this point. :) [Updated here.](https://github.com/sifive/wake/pull/1119/commits/661f3f971e4c9acae7e6d4d0a8592de338190cc2)

share/doc/wake/wake-for-fsharp-developers.md
bmitc: That's exactly what happened. [Removed here.](https://github.com/sifive/wake/pull/1119/commits/fdbe702140ce35f66cd39a064b2f5ae439c1a33f) (Needed to remove an extra newline.)

bmitc:
Totally blanked on the `adoc` to `md` rename. Will fix that.

bmitc:
"Fixed" the rename issue that included a format change by simply changing the extension back to `adoc`. It would be nice to standardize on `md` files, in my opinion, but I will defer the conversion for a later standalone PR (if I ever get to it).

share/doc/wake/wake-for-fsharp-developers.md
ag-eitilt: I actually had the newline removed in the suggestion as well, but apparently GitHub doesn't like line+blankline -> (remove both) changes in comments.

ag-eitilt (approved):

share/doc/wake/wake-for-fsharp-developers.md
bmitc: Ah, interesting. No worries!

PR #1121, opened 2023-02-13
implement last-used and move old functionality to last-executed

V-FEXrt:
Running `--last` or `--last-used` will include all jobs where `used_id == the highest run_id`. This creates a stable output that ignore the cache status when capturing jobs.

`--last-executed` functions as `--last` used to where only jobs that *executed* in the last run are captured.


JakeSiFive (approved):
ezpz

PR #1122, opened 2023-02-14
Bugsquashing

JakeSiFive:
This change adds some significant improvements to the shared cache and identifies a few more areas to work on.

1) It adds hash_str for all the value hashing stuff we do. This is in place of the original `hash` which is non-deterministic because it hashes addresses which was a big surprise to me
2) Files containing the same content (and thus that have the same hash) are now handled correctly where as before it would cause errors or cache misses
3) These changes have been successfully tested on a large repository.

We do know of a few issues however:
1) output directories and symlinks are not handled correctly by AddJobRequest or Cache::add
2) I realized that I left the threaded file copies in the code even now that its in the wake binary. This is a problem because one of those threads could catch important signals which is bad. I think we should be able to do async IO here instead. Unfortunately copy_file_range cannot be made async a la https://man7.org/linux/man-pages/man7/aio.7.html but since we check for the ioctal reflink first anyway I think that's fine. So copying a file async would essentially be first checking if you can reflink, otherwise append a write job to an async io queue. That's an optimization for later.

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -67,6 +67,7 @@ export def installAs (dest: String) (file: Path): Result Path Error =
     def foutputs _ =
         sdest, Nil
     makeExecPlan cmd inputs
+    | setPlanLabel "installAs: {dest} {file.getPathName}"
     | setPlanEnvironment Nil
     | setPlanLocalOnly True
     | setPlanFnOutputs foutputs
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -154,10 +154,9 @@ export def editPlanShare (f: Boolean => Boolean): Plan => Plan =
 
 # Get a unique hash-code for the job
 export def getPlanHash (plan: Plan): Integer =
-  def hash_str str = prim "hash_str"
   def Plan _ cmd _ env dir stdin _ _ _ _ _ _ _ _ _ _ _ = plan
   def sig = cat (implode cmd, "\0\0", implode env, "\0\0", dir, "\0\0", stdin,)
-  require Some out = intbase 16 (hash_str sig)
+  require Some out = intbase 16 (hashString sig)
   else unreachable "hash_str returned a non-hex string!!!"
   out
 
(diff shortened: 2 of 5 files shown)


share/wake/lib/system/io.wake
```diff
@@ -67,6 +67,7 @@ export def installAs (dest: String) (file: Path): Result Path Error =
     def foutputs _ =
         sdest, Nil
     makeExecPlan cmd inputs
+    | setPlanLabel "installAs: {dest} {file.getPathName}"
```
V-FEXrt, line 70: good cleanup, thanks!

share/wake/lib/system/job.wake
```diff
@@ -268,6 +272,8 @@ export def mkJobCacheRunner ((Runner name score baseDoIt): Runner): Runner =
 
         require False = cacheHit
         else
+          def _ = write ".cache-hit/read.{prefix}.json" "//{label}\n{jobCacheJsonIn}"
+          def _ = write ".cache-hit/out.{prefix}.json" "//{label}\n{cacheResult}"
```
V-FEXrt, line 276: Is this for debugged or does it serve another purpose?

src/runtime/string.cpp
```diff
@@ -633,6 +635,28 @@ static PRIMFN(prim_filter_term_codes) {
...
+}
+
+static void blake2b(const std::string &str, uint64_t (*out)[4]) {
+  blake2b_state S;
+  blake2b_init(&S, sizeof(*out));
+  blake2b_update(&S, reinterpret_cast<const uint8_t *>(str.data()), str.size());
+  blake2b_final(&S, reinterpret_cast<uint8_t *>(*out), sizeof(*out));
+}
+
+static PRIMFN(prim_hash_str) {
```
V-FEXrt, line 649: Can we make this a stdlib function instead of having consumers rely on the `prim` syntax?

share/wake/lib/system/job.wake
JakeSiFive: Its for debugging. I should probably wrap it in an environment variable.

src/runtime/string.cpp
JakeSiFive: Yeah I can add that

PR #1124, opened 2023-02-15
Shared Caching Eviction Tool

V-FEXrt:


tools/evict-shared-cache/eviction-policy.h
```diff
@@ -0,0 +1,25 @@
+#include <iostream>
+#include <string>
+
+struct EvictionPolicy {
```
V-FEXrt, line 4: Need to flesh out the actual parameters here

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+    print_help(argv[0]);
+    exit(EXIT_FAILURE);
+  }
+
+  std::unique_ptr<EvictionPolicy> policy = make_policy(argv[0], policy_name);
+  policy->init();
+
+  // TODO: Implement the following structure
+
+  // while stdin and stdout open
```
V-FEXrt, line 97: @JakeSiFive  Seem reasonable?

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+  //   read byte from stdin
```
JakeSiFive, line 98: Let's read in larger chunks at a time and then scan for the null terminator instead.

tools/evict-shared-cache/eviction-policy.h
JakeSiFive: I think initially just having the job id is sufficient.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <iostream>
+#include <memory>
+
+#include "eviction-policy.h"
+#include "gopt/gopt-arg.h"
+#include "gopt/gopt.h"
+
+void print_help(const char* argv0) {
+  // clang-format off
+  std::cout << std::endl
```
JakeSiFive, line 33: I'd make this `cerr` instead and I'd exit unconditionally.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+     << "Usage: " << argv0 << " [OPTIONS]" << std::endl
+     << "  --cache  DIR     Evict from shared cache DIR"  << std::endl
+     << "  --policy POLICY  Evict using POLICY"           << std::endl
+     << "  --help   -h      Print this message and exit"  << std::endl
+     << "Commands (read from stdin):" << std::endl
+     << "  init             Initialize the shared cache database with eviction status tables" << std::endl
+     << "  write JOB_ID     JOB_ID was written into the shared cache" << std::endl
+     << "  read JOB_ID      JOB_ID was read from the shared cache" << std::endl
```
JakeSiFive, line 41: I think `init` will happen unconditionally on startup like you've already done it. Also note that init is per-policy

tools/evict-shared-cache/evict-shared-cache.cpp
JakeSiFive: I think we should add a step that looks over all the jobs in the database and deletes the files/folders associated with any jobs not in the database. That makes us far more tolerant to failing to delete something from the file system and keeps the database as a good source of truth.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+  // while stdin and stdout open
+  //   read byte from stdin
+  //   if byte not null
+  //     append to command buffer
+  //     continue
+  //   else
+  //     parse command buffer into json
+  //     convert json command into relevant Policy funtion call
+  //     convert return from policy into json
+  //     serialize json into stdout terminated with null
```
JakeSiFive, line 106: I'm not sure we need to return anything at this time. Perhaps an `ack` is sufficient. Eviction doesn't have to be perfect.

tools/evict-shared-cache/eviction-policy.h
JakeSiFive: What do you think these would be returning?

V-FEXrt: I didn't have anything specific in mind. Maybe a simple success/fail is sufficient. Though maybe that isn't even all that valuable. Not sure how wake recovers from a eviction failure

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: SGTM

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: So the policy call just removes thing from the database while the main loop actually deletes the files?

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: I didn't put the exit in here because sometimes I want `exit(EXIT_SUCCESS)` and sometimes I want `exit(EXIT_FAILURE)`

tools/evict-shared-cache/eviction-policy.h
JakeSiFive: Yeah I think on eviction failure we should just not care. Unless the eviction processes closes the pipe we should assume its still responding even if it can't delete that one thing.

tools/evict-shared-cache/evict-shared-cache.cpp
JakeSiFive: I just mean on startup we should do a more robust cleanup to remove anything that might have been messed up

JakeSiFive: Deleting files should probably be done in a different thread though for sure.

tools/evict-shared-cache/evict-shared-cache.cpp
JakeSiFive: I never thought it was bad to exit successfully when asking for help personally but I can accept that argument. The main issue with using `cout` is that it will be pipped back to the process where as we can choose to log cerr instead which would be a better place to be notified that we messed up the call.

V-FEXrt: Sure, I  `EXIT_SUCCESS`  when the user does `--help` but when they do something wrong like

`./bin/evict --policy`

I print that the policy needs a value, call `print_help` then `exit(FAIL)`

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -49,11 +50,50 @@ std::unique_ptr<EvictionPolicy> make_policy(const char* argv0, const char* polic
...
+  std::cerr << "Unknown policy: " << policy << std::endl;
   print_help(argv0);
   exit(EXIT_FAILURE);
 }
 
+std::string read_command() {
+  std::string command = "";
+
+  // TODO: no idea if this is the right condition
+  while (std::cin.good() && std::cout.good()) {
```
JakeSiFive, line 62: I think this is fine but if you really wanted to be paranoid you should use `fcntl(F_GETFD)` and see if it errors out with `EBADF` in which case the file handle is closed for sure.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -49,11 +50,50 @@ std::unique_ptr<EvictionPolicy> make_policy(const char* argv0, const char* polic
...
+    uint8_t buffer[4096] = {};
+    ssize_t count = read(0, static_cast<void*>(buffer), 4096);
```
JakeSiFive, line 64: I think you can use STDIN_FILENO instead of 0 for the file descriptor.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -49,11 +50,50 @@ std::unique_ptr<EvictionPolicy> make_policy(const char* argv0, const char* polic
...
+
+    // Nothing new to process
+    if (count == 0) {
+      continue;
+    }
+
+    // An error occured during read
+    if (count < 0) {
+      std::cout << "Failed to read from stdin: " << strerror(errno) << std::endl;
```
JakeSiFive, line 73: nit: cerr

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -49,11 +50,50 @@ std::unique_ptr<EvictionPolicy> make_policy(const char* argv0, const char* polic
...
+        continue;
+      }
+
+      if (c != i - 1) {
+        std::cerr << "Multiple commands in one read. Trailing command will be dropped."
+                  << std::endl;
+      }
+
+      return command;
+    }
```
JakeSiFive, line 90:
```
std::vector<std::string> commands;
....
uint8_t* iter = buffer;
const uint8_t* buffer_end = buffer + count;
while (iter < buffer_end) {
    auto end = std::find(iter, buffer_end, '\0');
    command.append(iter, end);
    if (end != buffer_end) {
      commands.emplace_back(std::move(command));
    }
    iter = end;
}
```

to avoid loops and to handle batched commands we can do something like this instead.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -77,31 +117,60 @@ int main(int argc, char** argv) {
...
-  //     append to command buffer
-  //     continue
-  //   else
-  //     parse command buffer into json
-  //     convert json command into relevant Policy funtion call
-  //     maybe send an ack back
+  std::string command = "";
+
+  // TODO: no idea if this is the right condition
+  while (std::cin.good() && std::cout.good()) {
```
JakeSiFive, line 137: deja-vu?

V-FEXrt: ope yeah, bad refactor

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: oops typo. Thanks!

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: What is the point of only `emplace_back`ing inside of  `if (end != buffer_end)` here?

`end` always signifies the end of a command so shouldn't it always emplace iter...end as long as '\0' was found?

JakeSiFive: end != buffer_end is exactly the condition you need to check for to see if you found a null terminator or not. If end == buffer_end that means there was no null terminator so the next read will have to fill the buffer.

As another thought. This is going to have to be a class method so that command can persist across invocations.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+      // last read consumed the buffer, yield control
+      if (count < 4096) {
+        return commands;
+      }
+
+      // yield if 100 commands have been buffered up without yielding
+      if (commands.size() > 100) {
+        return commands;
+      }
```
JakeSiFive, line 119: These are neat optimizations. The first one prevents doing an extra `read` to find out that nothing is left and the second one keeps that latency of commands lower.

JakeSiFive: We might want to use system polling to avoid busy-reading as well....hmmm....I'll have to think about that.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,175 @@
...
+    std::cerr << "Eviction policy not specified" << std::endl;
+    print_help(argv[0]);
+    exit(EXIT_FAILURE);
+  }
+
+  std::unique_ptr<EvictionPolicy> policy = make_policy(argv[0], policy_name);
+  policy->init();
+
+  CommandParser cmd_parser;
+  while (true) {
```
JakeSiFive, line 164: this should be while stdin/stdout are open

JakeSiFive (commented):
Looking preety good!

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: That is checked inside of read commands atm. Didn't  want to double check

tools/evict-shared-cache/evict-shared-cache.cpp
V-FEXrt: we could also get away with an exponential read backoff. 

basically anytime a read returns nothing we double the wait time between reads (maybe capping at like 5 seconds). then anytime a read returns something we reset the timer. Basically we adaptively read less once we stop seeing messages

V-FEXrt:
Okay! This is ready for review. I'm going to put integration with shared caching into a separate PR

tools/evict-shared-cache/evict-shared-cache.cpp
JakeSiFive: since we already have pooling I'm inclined to use it instead I think. The logic should look like

1) Perform all database operations in the commands buffer
2) use system poll to detect IO events, this blocks if there is no IO event.
3) keep processing IO events to accumulate more commands in the command buffer.
4) if a close IO event is encountered, exit cleanly
5) otherwise go back to step 1

That's better in all ways than exponential back off I think. Look at the polling library that Wes already wrote. It's fairly easy to use.

JakeSiFive: Yeah we should do that now as part of the general framework I think.

tools/evict-shared-cache/command.h
```diff
@@ -0,0 +1,73 @@
...
+
+    const JAST& command = json.get("command");
+    if (command.kind != JSON_STR) {
+      std::cerr << "Expected string for 'command' key" << std::endl;
+      return false;
+    }
+
+    const std::string& command_str = command.value;
+
+    CommandType type = CommandType::Read;
```
JakeSiFive, line 52: nit: initing this seems strange to me.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,188 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+struct CommandParser {
+  std::string command_buff = "";
+
+  std::vector<std::string> read_commands() {
```
JakeSiFive, line 81: Yeah I think I've concluded that we should use system polling here actually. Look at poll.h in src/runtime. You'll probably want to put that in src/util or src/wcl instead of in runtime so that its more widely usable.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,188 @@
...
+        return commands;
+      }
+
+      // yield if 100 commands have been buffered up without yielding
+      if (commands.size() > 100) {
+        return commands;
+      }
+    }
+    // stdin or stdout were closed. Exit cleanly
+    exit(EXIT_SUCCESS);
```
JakeSiFive, line 123: We should return something instead of just exiting so that we can choose to exit cleanly in the future.

tools/evict-shared-cache/evict-shared-cache.cpp
JakeSiFive: Left a comment above. I think we should actually use poll instead of checking to see if the file is open (a file being closed is an IO event that Poll will notify you of, if `read` return `EBADF` then the handle is closed). We should also return something like maybe wcl::optional<std::vector<std::string>> so that the empty optional case indicates that stdin was closed.

src/util/poll.cpp
```diff
@@ -44,7 +45,7 @@ struct Poll::detail {
   int pfd;
 };
 
-Poll::Poll() : imp(new Poll::detail) {
+Poll::Poll() : imp(std::make_unique<Poll::detail>()) {
```
JakeSiFive, line 48: thancc

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,199 @@
...
+    struct timespec timeout;
+    timeout.tv_sec = 1;
+
+    // Sleep until timeout or a signal arrives
+    std::vector<int> ready_fds = poll.wait(&timeout, &saved);
+
+    // Nothing is ready within timeout, yield control
+    if (ready_fds.size() == 0) {
+      return CommandParserState::Continue;
+    }
```
JakeSiFive, line 84: We should just wait indefinitely not yield control. Process all database commands before entering this function I think. We don't really need timely deletes.

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,199 @@
...
+  Poll poll;
+
+  CommandParser() { poll.add(STDIN_FILENO); }
+
+  CommandParserState read_commands(std::vector<std::string>& commands) {
+    commands = {};
+
+    sigset_t saved;
+    struct timespec timeout;
+    timeout.tv_sec = 1;
```
JakeSiFive, line 76: I think you can wait -1 seconds to wait forever

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+      // last read consumed the buffer, yield control
+      if (count < 4096) {
+        return CommandParserState::Continue;
+      }
+
+      // yield if 100 commands have been buffered up without yielding
+      if (commands.size() > 100) {
+        return CommandParserState::Continue;
+      }
```
JakeSiFive, line 128: Also here, no need to yield in this situation, just keep parsing. Since we can sleep indefinitely until a message comes along we don't need to yield ever. Just do as much work as we can in each place.

JakeSiFive: This also enables bigger batch transactions which will be nice I think.

JakeSiFive (approved):

PR #1125, opened 2023-02-16
Fix several small issues with shared caching

JakeSiFive:
These were all known issues that we haven't hit testing against real code yet but which I believe will come up. At this time this represents all known bugs with shared caching. The next step is to test it more widely on more code and to fuzz test the cache type in C++.

Issues fixed:
* Output directories would be ignored so if a tool created a directory with a non-standard mode or if it that created a directory that contained no files that directory would not be created. We now store these in the cache database and recreate them faithfully. If we did have to hash them they would hash to all zeros but we don't need to do that.
* Symlinks would have been hashed to the content of the file they pointed to and the content of the file they pointed to would have been copied instead. We store the readlink value in the cache database and recreate these faithfully now. We don't have to worry about hashing these because on input they're handled like anything else and on output shim-wake will later hash them instead of us.
* We had some code to speed up hashing of files by doing them in parallel but that introduced new threads which would mess with signal handling. I undid this. We took a big performance hit (on one real example we went from taking 22 seconds to 45 seconds). This is unfortunate but not a deal breaker since its still saving time over what would be a 10 minute job otherwise. We can optimize things later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,8 +39,3 @@ compile_commands.json
 *.swp
 *.swo
 *.swn
-/tests/job-cache/symlink-dir/bar/
-tests/job-cache/symlink-dir/test.txt
-tests/job-cache/basic-fetch/test.txt
-tests/job-cache/dup-output/test1.txt
-tests/job-cache/dup-output/test2.txt
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,9 @@ manifest.wake
 *.o
 .build
 .fuse*
-/.job-cache
+.cache-hit
+.cache-misses
+.job-cache
 compile_commands.json
 /.compile_commands
 /.vscode/*.log
@@ -37,3 +39,8 @@ compile_commands.json
 *.swp
 *.swo
 *.swn
+/tests/job-cache/symlink-dir/bar/
+tests/job-cache/symlink-dir/test.txt
+tests/job-cache/basic-fetch/test.txt
+tests/job-cache/dup-output/test1.txt
+tests/job-cache/dup-output/test2.txt
--- a/src/job-cache/job-cache.cpp
+++ b/src/job-cache/job-cache.cpp
@@ -1163,48 +1163,32 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
   // The MatchingJob is currently using sandbox paths.
   // We need to redirect those sandbox paths to non-sandbox paths
   // here we redirect the output files.
+  auto redirect_path = [&find_request](std::string& path) {
+    std::vector<std::string> output_path_vec = split_path(path);
+    auto pair = find_request.dir_redirects.find_max(output_path_vec.begin(), output_path_vec.end());
+    if (!pair.first) return;
+    std::string rel_path = join('/', pair.second, output_path_vec.end());
+    path = wcl::join_paths(*pair.first, rel_path);
+  };
+
   for (auto &output_file : result->output_files) {
-    std::vector<std::string> output_path_vec = split_path(output_file.path);
-    auto pair = find_request.dir_redirects.find_max(output_path_vec.begin(), output_path_vec.end());
-    if (!pair.first) continue;
-    std::string rel_path = join('/', pair.second, output_path_vec.end());
-    output_file.path = wcl::join_paths(*pair.first, rel_path);
+    redirect_path(output_file.path);
   }
 
-  // Similarly, we need to redirect the output directories.
   for (auto &output_dir : result->output_dirs) {
-    std::vector<std::string> output_path_vec = split_path(output_dir.path);
-    auto pair = find_request.dir_redirects.find_max(output_path_vec.begin(), output_path_vec.end());
-    if (!pair.first) continue;
-    std::string rel_path = join('/', pair.second, output_path_vec.end());
-    output_dir.path = wcl::join_paths(*pair.first, rel_path);
+    redirect_path(output_dir.path);
   }
 
-  // Ditto but for output symlinks
   for (auto &output_symlink : result->output_symlinks) {
-    std::vector<std::string> output_path_vec = split_path(output_symlink.path);
-    auto pair = find_request.dir_redirects.find_max(output_path_vec.begin(), output_path_vec.end());
-    if (!pair.first) continue;
-    std::string rel_path = join('/', pair.second, output_path_vec.end());
-    output_symlink.path = wcl::join_paths(*pair.first, rel_path);
+    redirect_path(output_symlink.path);
   }
 
-  // Ditto but for the input files
   for (auto &input_file : result->input_files) {
-    std::vector<std::string> path_vec = split_path(input_file);
-    auto pair = find_request.dir_redirects.find_max(path_vec.begin(), path_vec.end());
-    if (!pair.first) continue;
-    std::string rel_path = join('/', pair.second, path_vec.end());
-    input_file = wcl::join_paths(*pair.first, rel_path);
+    redirect_path(input_file);
   }
 
-  // Finally, same thing but for the output directories.
   for (auto &input_dir : result->input_dirs) {
-    std::vector<std::string> path_vec = split_path(input_dir);
-    auto pair = find_request.dir_redirects.find_max(path_vec.begin(), path_vec.end());
-    if (!pair.first) continue;
-    std::string rel_path = join('/', pair.second, path_vec.end());
-    input_dir = wcl::join_paths(*pair.first, rel_path);
+    redirect_path(input_dir);
   }
 
   // TODO: We should really return a different thing here
--- a/src/job-cache/job-cache.cpp
+++ b/src/job-cache/job-cache.cpp
@@ -21,6 +21,34 @@
 
 #include "job-cache.h"
 
+#include <errno.h>
+#include <fcntl.h>
+#include <json/json5.h>
+#include <sqlite3.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/file.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <util/execpath.h>
+#include <util/mkdir_parents.h>
+#include <util/term.h>
+#include <wcl/defer.h>
+#include <wcl/filepath.h>
+#include <wcl/trie.h>
+#include <wcl/xoshiro_256.h>
+
+#include <algorithm>
+#include <future>
+#include <iostream>
+#include <map>
+#include <random>
+#include <thread>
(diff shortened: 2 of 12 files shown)


src/job-cache/job-cache.cpp
```diff
@@ -86,35 +87,30 @@ static void rmdir_no_fail(const char *dir) {
 // supported by the filesystem it copies instead.
 #ifdef __APPLE__
```
mmjconolly, line 89: I'd have assumed that the shared-cache were linux-only, as the macos implementation of wakebox can't hide the .fuse directory when a payload command resolves an absolute path

JakeSiFive: The shared cache itself should be able to work on mac but that may be an issue that I was about to spend a lot time debugging. The tests failed when they ran on mac despite succeeding on the tarball step. I guess I need to update these tests to pass without testing on mac.

JakeSiFive: Thanks for the heads up!

JakeSiFive:
BTW, on the slow down. The machine I was testing on was actually loaded. Reads should be just as fast. I was thinking about it but the hashing only happens on write not on read so there should be no slow down on read and I'm not sure we care too much about the slow down on write since the write is non-blocking.

tests/job-cache/basic-fetch/pass.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/sh
+
+if [ $(uname) = Darwin ] ; then
```
mmjconolly, line 3:
```suggestion
if [ $(uname) != Linux ] ; then
```

JakeSiFive: Good call, thanks!

src/job-cache/job-cache.h
```diff
@@ -128,6 +138,9 @@ struct AddJobRequest {
   std::vector<InputFile> inputs;
   std::vector<InputDir> directories;
   std::vector<OutputFile> outputs;
+  // TODO: Add mode to avoid having to fstat again
```
V-FEXrt, line 141: this is done no?

src/job-cache/job-cache.cpp
```diff
@@ -90,25 +138,44 @@ static void copy(int src_fd, int dst_fd) {
...
+  size_t size = buf.st_size;
+  do {
+    intptr_t written = sendfile(dst_fd, src_fd, &idx, size);
+    if (written < 0) {
+      log_fatal("sendfile(src_fd = %d, NULL, dst_fd = %d, size = %d, 0): %s", src_fd, dst_fd,
+                buf.st_size, strerror(errno));
+    }
+    idx = written;
+    size -= written;
+  } while (size != 0);
```
V-FEXrt, line 171: this looks eerily similar to writing a raw NAND device driver lol

src/job-cache/job-cache.cpp
```diff
@@ -759,33 +901,51 @@ AddJobRequest::AddJobRequest(const JAST &job_result_json) {
   //       need this loop. Since this job was just run, wake
   //       will eventually hash all these files so the fact
   //       that we have to re-hash them here is a shame.
-  // TODO: It is incorrect to do this all async without signal
-  //       handling being done correctly.
   // TODO: This code does not handle directories or symlinks correctly
-  std::vector<std::future<OutputFile>> future_outputs;
-
+  // TODO: Use aio_read and do these hashes online and interleaveed
+  //       so that the IO can be in parallel despite the hasing being
```
V-FEXrt, line 906: typo: `hasing` -> `hashing`

src/job-cache/job-cache.cpp
```diff
@@ -759,33 +901,51 @@ AddJobRequest::AddJobRequest(const JAST &job_result_json) {
...
+        log_fatal("readlink(%s): %s", src.c_str(), strerror(errno));
+      }
+      sym.path = output_file.second.get("path").value;
+      sym.value = std::string(link, link + size);
+      output_symlinks.emplace_back(std::move(sym));
+      continue;
+    }
+
+    // Handle regular files but ignore everything else.
+    if (!S_ISREG(buf.st_mode)) continue;
```
V-FEXrt, line 941: Do we expect a bunch of things that aren't dirs, links, or regular files? 

if not, might be good to add a debug warning before the continue

src/job-cache/job-cache.h
JakeSiFive, line 141: Ah crap I did this one already, I'll remove that TODO before this lands.

src/job-cache/job-cache.cpp
```diff
@@ -915,35 +1085,68 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
...
+    std::string output_path = wcl::join_paths(output_dir, rel_path);
+    std::vector<std::string> output_path_vec = split_path(output_path);
+    return std::make_pair<std::string, std::vector<std::string>>(std::move(output_path),
+                                                                 std::move(output_path_vec));
+  };
+
   if (success) {
-    // Now copy/reflink all files into their final place
+    // First output all the directories (assumed to be sorted by length).
+    // This ensures that all directories are already made witht the
```
V-FEXrt, line 1109: typo: `witht` -> `with`

src/job-cache/job-cache.cpp
```diff
@@ -915,35 +1085,68 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
...
+    // expected mode.
+    for (const auto &output_dir : result->output_dirs) {
+      // Rewrite the path based on the avliable rewrites
```
V-FEXrt, line 1112: typo: `avliable` -> `available`

src/job-cache/job-cache.cpp
```diff
@@ -915,35 +1085,68 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
...
-      } else {
-        const auto &output_dir = *pair.first;
-        const auto &rel_path = join('/', pair.second, path_vec.end());
-        std::string output_path = wcl::join_paths(output_dir, rel_path);
-        std::vector<std::string> output_path_vec = split_path(output_path);
-        mkdir_all(output_path_vec.begin(), output_path_vec.end());
-        copy_or_reflink(tmp_file.c_str(), output_path.c_str(), mode);
-      }
+
+      // Rewrite the path based on the avliable rewrites
```
V-FEXrt, line 1125: typo: `avliable` -> `available`

src/job-cache/job-cache.cpp
```diff
@@ -968,6 +1171,22 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
...
+  }
+
+  for (auto &output_symlink : result->output_symlinks) {
+    std::vector<std::string> output_path_vec = split_path(output_symlink.path);
+    auto pair = find_request.dir_redirects.find_max(output_path_vec.begin(), output_path_vec.end());
+    if (!pair.first) continue;
+    std::string rel_path = join('/', pair.second, output_path_vec.end());
+    output_symlink.path = wcl::join_paths(*pair.first, rel_path);
+  }
+
```
V-FEXrt, line 1189: what is this code doing?

JakeSiFive: This is one of those times you're baiting me to add a comment huh ;)

These rename sandbox paths into non-sandbox paths. The paths that I expect in the json are sandbox absolute paths and thats also what I store in the database. The paths outside the sandbox are generally relative. You can add redirects to redirect any sandbox path to a non-sandbox location. For instance you can redirect "/workspace" to "./"

src/job-cache/job-cache.cpp
JakeSiFive: No I don't think so. The only other options are things like devices, character devices (I think that's a terminal?), block devices, and sockets...stuff that you can't cache even if you wanted to and stuff that fuse isn't going to tell us about.

src/job-cache/job-cache.cpp
JakeSiFive: heh. Hopefully we don't have the issue where the first 4 physical blocks were being used for critical stuff and they were getting overloaded because the designers were like "it will probably never have that many cycles on it" and then all of our infra started to fail one by one at the exact same time...yeah that wasn't fun.

src/job-cache/job-cache.h
JakeSiFive: yeah I'll remove it

src/job-cache/job-cache.h
```diff
@@ -116,6 +114,18 @@ struct OutputFile {
...
+
+struct OutputDirectory {
+  std::string path;
+  mode_t mode;
+};
+
+// TODO: Add mode to avoid having to fstat again
+struct OutputSymlink {
+  std::string value;
+  std::string path;
```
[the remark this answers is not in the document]
JakeSiFive: OutputSymlink has the path to the symlink and the value you'd get from calling readlink on that symlink. OutputDirectory just has the path and the permission bits for the directory.

.gitignore
```diff
@@ -37,3 +39,8 @@ compile_commands.json
 *.swp
 *.swo
 *.swn
+/tests/job-cache/symlink-dir/bar/
+tests/job-cache/symlink-dir/test.txt
+tests/job-cache/basic-fetch/test.txt
+tests/job-cache/dup-output/test1.txt
+tests/job-cache/dup-output/test2.txt
```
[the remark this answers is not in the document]
JakeSiFive: That's probably a better idea huh?

src/job-cache/job-cache.cpp
JakeSiFive: I added comments, let me know if that helps.

V-FEXrt: eh goal wasn't directly to bait you into writing a comment but I did have trouble following what it was doing so a comment or function extraction felt reasonable

src/job-cache/job-cache.cpp
JakeSiFive: I can and probably should extract this out into a function.

src/job-cache/job-cache.cpp
V-FEXrt: 👀

src/job-cache/job-cache.cpp
V-FEXrt: I'll not block on it, but I'll strongly urge you to consider it :)

V-FEXrt (approved):

src/job-cache/job-cache.cpp
JakeSiFive: Yeah I already did it and its upload now :) I just had stuff to do this morning so it took me a while to finish.

PR #1127, opened 2023-02-17
PID namespace for wakebox isolation

mmjconolly:
Introduces a PID namespace for wakebox (linux-only).

Wakebox was leaking the existence of a file via `/proc/self/fd`. 
The file was used to let the `fuse-waked` daemon know when all users of it's fuse mountpoint had closed, by wakebox leaving the file open as child processes forked.
The file was not actually accessible by wakebox's payload command however, as it is hidden by the mount table changes where we bind-mount `workspace/.fuse/$UID.$GID` onto `workspace/`.

The motivation of this change, is that some payload commands actually try to read the files symlinked via `/proc/self/fd`, so those payloads would encounter the unexpected situation of a file that _is_ open, but where the path is not accessible.

Changes (linux only): 
* the live file for fuse-waked is closed on exec
* instead of exec-ing the payload command directly, a `clone()` is called with a flag to create a new PID namespace
* the function started by clone() is PID1, mounts a new procfs, fork/execs the payload command, and then waits for _all_ children to terminate. Any orphaned children are re-parented to this new PID1.

The process tree view from inside wakebox (on linux)
```
$ wakebox ps
    PID TTY          TIME CMD
      1 pts/96   00:00:00 wb-pid-ns
      2 pts/96   00:00:00 ps

$ wakebox ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0  15008   216 pts/96   S+   15:52   0:00 wakebox ps aux
root           2  0.0  0.0  44704  3448 pts/96   R+   15:52   0:00 ps aux
```

Open files according to the payload command:
```
# Before this PR:
$ old-wakebox bash -c "ls -l /proc/self/fd"
total 0
lrwx------. 1 root root 64 Feb 16 16:00 0 -> /dev/pts/96
lrwx------. 1 root root 64 Feb 16 16:00 1 -> /dev/pts/96
lrwx------. 1 root root 64 Feb 16 16:00 2 -> /dev/pts/96
lr-x------. 1 root root 64 Feb 16 16:00 3 -> /proc/1234566/fd
lrwx------. 1 root root 64 Feb 16 16:00 4 -> /home/user/wake/.fuse/1000.1000/.l.1234567

# This PR applied:
$ wakebox bash -c "ls -l /proc/self/fd"
total 0
lrwx------. 1 root root 64 Feb 16 16:01 0 -> /dev/pts/96
lrwx------. 1 root root 64 Feb 16 16:01 1 -> /dev/pts/96
lrwx------. 1 root root 64 Feb 16 16:01 2 -> /dev/pts/96
lr-x------. 1 root root 64 Feb 16 16:01 3 -> /proc/2/fd
```

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -805,17 +805,11 @@ export def getJobStatus (job: Job): Status = match (getJobReport job)
         else
             Signaled (-status)
 
-# Implement FUSE-based Runner
-# The FUSE runner on linux supports a few isolation options via resources:
-# - "isolate/user": the Job will appear to run as root
-# - "isolate/host": make the hostname appear to be "build.local"
-# - "isolate/net": disables network access
-# - "isolate/workspace": makes the build appear run in /var/cache/wake
-#    ... if /var/cache/wake does not exist, a directory 'build/wake' is
-#        used relative to where wake has been installed
+# Location of the wake executable
 export def wakePath: String =
-    prim "execpath" # location of the wake executable
+    prim "execpath"
 
+# Implement FUSE-based Runner
 export def fuseRunner: Runner =
     def fuse = "{wakePath}/wakebox"
     def score _ = Pass 2.0
@@ -895,6 +889,8 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
                     "stdin" :-> JString stdin,
                     "resources" :-> res | map JString | JArray,
                     "version" :-> JString version,
+                    "isolate-network" :-> JBoolean False,
+                    "isolate-pids" :-> JBoolean False,
                     "mount-ops" :-> JArray (JObject ("type" :-> JString "workspace", "destination" :-> JString ".", Nil), Nil),
                     "usage" :-> JObject (
                         "status" :-> JInteger status,
--- a/src/wakefs/fuse.cpp
+++ b/src/wakefs/fuse.cpp
@@ -37,6 +37,10 @@
 #include <iostream>
 #include <sstream>
 
+#ifdef __linux__
+#include <sys/prctl.h>
+#endif
+
 #include "compat/rusage.h"
 #include "json/json5.h"
 #include "namespace.h"
@@ -79,8 +83,8 @@ bool json_as_struct(const std::string &json, json_args &result) {
   return true;
 }
 
-static int execve_wrapper(const std::vector<std::string> &command,
-                          const std::vector<std::string> &environment) {
+int execve_wrapper(const std::vector<std::string> &command,
+                   const std::vector<std::string> &environment) {
   std::vector<const char *> cmd_args;
   for (auto &s : command) cmd_args.push_back(s.c_str());
   cmd_args.push_back(0);
@@ -148,6 +152,8 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
       exit(1);
 
     if (!do_mounts(args.mount_ops, args.daemon.mount_subdir, envs_from_mounts)) exit(1);
+
+    prctl(PR_SET_NAME, "wb-mount-ns", 0, 0, 0);
 #endif
 
     if (chdir(args.command_running_dir.c_str()) != 0) {
@@ -185,8 +191,13 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
       }
     }
 
+#ifdef __linux__
+    pidns_args nsargs = {command, args.environment};
+    exec_in_pidns(&nsargs);
+#else
     int err = execve_wrapper(command, args.environment);
     std::cerr << "execve " << command[0] << ": " << strerror(err) << std::endl;
+#endif
     exit(1);
   }
 
--- a/src/wakefs/fuse.cpp
+++ b/src/wakefs/fuse.cpp
@@ -195,7 +195,7 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
 #ifdef __linux__
     if (args.isolate_pids) {
       pidns_args nsargs = {command, args.environment};
-      exec_in_pidns(&nsargs);
+      exec_in_pidns(nsargs);
     } else {
 #endif
       int err = execve_wrapper(command, args.environment);
--- a/src/wakefs/fuse.cpp
+++ b/src/wakefs/fuse.cpp
@@ -65,6 +65,7 @@ bool json_as_struct(const std::string &json, json_args &result) {
   result.stdin_file = jast.get("stdin").value;
 
   result.isolate_network = jast.get("isolate-network").kind == JSON_TRUE;
+  result.isolate_pids = jast.get("isolate-pids").kind == JSON_TRUE;
 
   result.hostname = jast.get("hostname").value;
   result.domainname = jast.get("domainname").value;
@@ -137,7 +138,7 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
     return false;
   }
 
-  if (!args.daemon.connect(args.visible)) return false;
+  if (!args.daemon.connect(args.visible, args.isolate_pids)) return false;
 
   struct timeval start;
   gettimeofday(&start, 0);
@@ -192,11 +193,15 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
     }
 
 #ifdef __linux__
-    pidns_args nsargs = {command, args.environment};
-    exec_in_pidns(&nsargs);
-#else
-    int err = execve_wrapper(command, args.environment);
-    std::cerr << "execve " << command[0] << ": " << strerror(err) << std::endl;
+    if (args.isolate_pids) {
+      pidns_args nsargs = {command, args.environment};
+      exec_in_pidns(&nsargs);
+    } else {
+#endif
+      int err = execve_wrapper(command, args.environment);
+      std::cerr << "execve " << command[0] << ": " << strerror(err) << std::endl;
+#ifdef __linux__
+    }
 #endif
     exit(1);
   }
--- a/src/wakefs/namespace.cpp
+++ b/src/wakefs/namespace.cpp
@@ -433,11 +433,13 @@ bool get_workspace_dir(const std::vector<mount_op> &mount_ops,
 
 bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
                            const std::string &hostname, const std::string &domainname) {
-  // Later we will create a PID namespace.
-  // TODO explain this, if it works in CI.
+  // If we have permissions in the current mount namespace, create a fresh /proc mount.
+  // This can avoid issues where the mount propagation setting for /proc in the
+  // current namespace is set to MS_UNBINDABLE.
   int err = mount("proc", "/proc", "proc", 0, nullptr);
-  if (errno != EPERM && err == -1) {
-    std::cerr << "mount('proc', '/proc', 'proc', 0, nullptr): " << strerror(errno) << std::endl;
+  if (err == -1 && errno != EPERM && errno != EBUSY) {
+    std::cerr << "pre namespace setup: mount('proc', '/proc', 'proc', 0, nullptr): "
+              << strerror(errno) << std::endl;
     exit(1);
   }
 
@@ -484,7 +486,8 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
   // than leaking process ids from the exterior namespace.
   int err = mount("proc", "/proc", "proc", 0, nullptr);
   if (err == -1) {
-    std::cerr << "mount('proc', '/proc', 'proc', 0, nullptr): " << strerror(errno) << std::endl;
+    std::cerr << "new pid namespace: mount('proc', '/proc', 'proc', 0, nullptr): "
+              << strerror(errno) << std::endl;
     exit(1);
   }
 
--- a/src/wakefs/namespace.cpp
+++ b/src/wakefs/namespace.cpp
@@ -518,7 +518,7 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
   exit(-WTERMSIG(status));
 }
 
-[[noreturn]] void exec_in_pidns(pidns_args& nsargs) {
+[[noreturn]] void exec_in_pidns(pidns_args &nsargs) {
   alignas(16) uint8_t stack_buf[4096];
   void *child_stack = reinterpret_cast<void *>(stack_buf + sizeof(child_stack));
 
--- a/src/wakefs/namespace.cpp
+++ b/src/wakefs/namespace.cpp
@@ -518,21 +518,21 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
   exit(-WTERMSIG(status));
 }
 
-[[noreturn]] void exec_in_pidns(pidns_args *nsargs) {
+[[noreturn]] void exec_in_pidns(pidns_args& nsargs) {
   alignas(16) uint8_t stack_buf[4096];
   void *child_stack = reinterpret_cast<void *>(stack_buf + sizeof(child_stack));
 
   // Create a new thread in a PID namespace, calling pidns_init.
-  pid_t child_pid = clone(pidns_init, child_stack, CLONE_NEWPID | SIGCHLD, nsargs);
+  pid_t child_pid = clone(pidns_init, child_stack, CLONE_NEWPID | SIGCHLD, &nsargs);
   if (child_pid == -1) {
-    std::cerr << "clone " << nsargs->command[0] << ": " << strerror(errno) << std::endl;
+    std::cerr << "clone " << nsargs.command[0] << ": " << strerror(errno) << std::endl;
     exit(1);
   }
 
   // While we wait, the subdir_live_file for the fuse daemon will be held open.
   int status = 1;
   if (waitpid(child_pid, &status, 0) == -1) {
-    std::cerr << "waitpid (clone) " << nsargs->command[0] << std::endl;
+    std::cerr << "waitpid (clone) " << nsargs.command[0] << std::endl;
     exit(1);
   }
 
--- a/src/wakefs/namespace.cpp
+++ b/src/wakefs/namespace.cpp
@@ -433,6 +433,14 @@ bool get_workspace_dir(const std::vector<mount_op> &mount_ops,
 
 bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
                            const std::string &hostname, const std::string &domainname) {
+  // Later we will create a PID namespace.
+  // TODO explain this, if it works in CI.
+  int err = mount("proc", "/proc", "proc", 0, nullptr);
+  if (errno != EPERM && err == -1) {
+    std::cerr << "mount('proc', '/proc', 'proc', 0, nullptr): " << strerror(errno) << std::endl;
+    exit(1);
+  }
+
   uid_t real_euid = geteuid();
   gid_t real_egid = getegid();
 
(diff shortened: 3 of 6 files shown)


mmjconolly:
I haven't finished testing this manually (let alone writing tests), but I thought I'd get some eyes on this as I've not been able to give it a lot of time lately

JakeSiFive:
> the function started by clone() is PID1, mounts a new procfs, fork/execs the payload command, and then waits for all children to terminate. Any orphaned children are re-parented to this new PID1.

Do you wait on all children or do you wait on all processes in the namespace? Waiting on all processes in the namespace would let us avoid using unshare as a means to avoid fuse-waked getting SIGKILLed

src/wakefs/fuse.cpp
```diff
@@ -127,6 +129,32 @@ static bool collect_result_metadata(const std::string daemon_output, const struc
   return !result_ss.fail();
 }
 
+static void exec_in_pidns(pidns_args *nsargs) {
+  size_t stack_size = 1024 * 1024;
+  uint8_t *child_stack = (uint8_t *)malloc(stack_size) + stack_size;
```
JakeSiFive, line 134: malloc assures 8-byte aligned pointers but the x86_64 abi requires 16-byte aligned pointers. We should bump the pointer to alignment.

src/wakefs/namespace.cpp
```diff
@@ -459,4 +459,48 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
+  int err = mount("proc", "/proc", "proc", 0, NULL);
+  if (err == -1) {
+    std::cerr << "mount pidns proc" << strerror(err) << std::endl;
+    exit(1);
+  }
+
+  pid_t pid = fork();
+  if (pid == 0) {
+    // Execute the user-specified command.
+    pidns_args *args = (pidns_args *)arg;
```
JakeSiFive, line 480: reinterpret_cast instead of using a C-cast

```
const auto *args = reinterpret_cast<const pidns_args*>(arg);
```

src/wakefs/fuse.cpp
```diff
@@ -127,6 +129,32 @@ static bool collect_result_metadata(const std::string daemon_output, const struc
...
+
+  // Create a new thread in a PID namespace, calling pidns_init.
+  pid_t child_pid = clone(pidns_init, child_stack, CLONE_NEWPID | SIGCHLD, nsargs);
```
JakeSiFive, line 137: Ok I had to go read the clone docs. I knew it was how pthread_create and fork (and exec etc...) were all implemented but I didn't know the details of how to actually do that. Since neither CLONE_VM nor CLONE_THREAD is called, this is basically a fancy fork call. That also means that we don't have to worry about avoiding using TLS which was one of my first comments on this before I read more. Does that all sound correct to you?

src/wakefs/fuse.cpp
```diff
@@ -127,6 +129,32 @@ static bool collect_result_metadata(const std::string daemon_output, const struc
   return !result_ss.fail();
 }
 
+static void exec_in_pidns(pidns_args *nsargs) {
```
JakeSiFive, line 132: I think this can be a const reference instead of a pointer.

src/wakefs/fuse.cpp
```diff
@@ -79,8 +81,8 @@ bool json_as_struct(const std::string &json, json_args &result) {
   return true;
 }
 
-static int execve_wrapper(const std::vector<std::string> &command,
-                          const std::vector<std::string> &environment) {
+int execve_wrapper(const std::vector<std::string> &command,
```
JakeSiFive, line 84: This function now also needs a "NO TLS" warning on it as well.

src/wakefs/namespace.cpp
```diff
@@ -459,4 +459,48 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
+  prctl(PR_SET_NAME, "wb-pid-ns", 0, 0, 0);
+
+  // A fresh mount of procfs over the previous namespace's /proc.
+  // which we can only do as we're already in a mount namespace.
+  // Note that this is not a bind mount of an existing /proc.
+  // The new procfs will have the correct data for the new PID namespace, rather
+  // than leaking process ids from the exterior namespace.
+  int err = mount("proc", "/proc", "proc", 0, NULL);
+  if (err == -1) {
+    std::cerr << "mount pidns proc" << strerror(err) << std::endl;
```
JakeSiFive, line 473: Needs a space between strerror and the message
```
std::cerr << "mount('proc', '/proc', 'proc', 0, nullptr): " << strerror(err) << std::endl;
```

src/wakefs/namespace.cpp
```diff
@@ -459,4 +459,48 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
+int pidns_init(void *arg) {
+  // We should be in a new PID namespace now, label the process in 'ps'.
+  prctl(PR_SET_NAME, "wb-pid-ns", 0, 0, 0);
+
+  // A fresh mount of procfs over the previous namespace's /proc.
+  // which we can only do as we're already in a mount namespace.
+  // Note that this is not a bind mount of an existing /proc.
+  // The new procfs will have the correct data for the new PID namespace, rather
+  // than leaking process ids from the exterior namespace.
+  int err = mount("proc", "/proc", "proc", 0, NULL);
```
JakeSiFive, line 471: use nullptr instead of NULL please

src/wakefs/namespace.cpp
```diff
@@ -459,4 +459,48 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
+  // Wait for child processes, we are init(1) in this PID namespace.
+  int status;
+  while ((pid = waitpid(-1, &status, 0)) != 0) {
+    if (pid == -1) {
+      if (errno == ECHILD) {
+        break;  // all children have terminated
+      } else {
+        std::cerr << "waitpid: " << strerror(errno) << std::endl;
+        exit(errno);
+      }
```
JakeSiFive, line 495: You don't need the else branch here since you break before

src/wakefs/fuse.cpp
```diff
@@ -127,6 +129,32 @@ static bool collect_result_metadata(const std::string daemon_output, const struc
...
+  // Create a new thread in a PID namespace, calling pidns_init.
+  pid_t child_pid = clone(pidns_init, child_stack, CLONE_NEWPID | SIGCHLD, nsargs);
+  if (child_pid == -1) {
+    std::cerr << "clone " << nsargs->command[0] << ": " << strerror(errno) << std::endl;
+    exit(1);
+  }
+
+  // While we wait, the subdir_live_file for the fuse daemon will be held open.
+  int status = 1;
+  if (waitpid(child_pid, &status, 0) == -1) {
```
JakeSiFive, line 145: Ah so it looks like we only wait on the child to finish and not everything in the namespace. So we'll keep using unshare I guess.

JakeSiFive (commented):
This looks awesome! I'm excited to see this work.

src/wakefs/fuse.cpp
JakeSiFive: Ah, sadly that's not possible because glibc's clone makes modifications to the stack in the current address space before jumping into the child (which makes sense, it has to setup the arguments and return address etc...) so you have to give it some extra space. So I recommend `alignas(16) uint8_t child_stack[4096]` as that will protect you from any issues but it will still be on the stack so the child processes will get a plenty big stack (much bigger than 4096 bytes).

Another thing I got wrong but I see you got right is that you have to give the *last* byte because the stack grows towards lower addresses. You have an off by 1 error in yours though since the pointer you pass to clone has to be a valid byte, not just past the end of it. So add in a `+ sizeof(child_stack) - 1` but then you'll need to re-align it which means you have to do some pointer arithmetic...or you can just increase the size of child_stack by one so that its aligned after the subtraction.

```
alignas(16) uint8_t child_stack[4097];
void* stk = reinterpret_cast<void*>(child_stack + sizeof(child_stack) - 1);
pid_t child_pid = clone(pidns_init, stk, CLONE_NEWPID | SIGCHLD, nsargs);
```

I think that's all correct now.

src/wakefs/fuse.cpp
mmjconolly: I think I need to wait on the child regardless, so that we can capture it's exit code

src/wakefs/fuse.cpp
richardxia: I don't think the `- 1` is necessary. The official [man page](https://man7.org/linux/man-pages/man2/clone.2.html) has an example at the very bottom that points to just past the allocation. I think the reason it's fine is normally the stack pointer points to the _most recent_ value pushed to the stack, and the contract for a function is for it to change the stack pointer and to only use the address up to and including the new stack pointer value.

Stack-allocating the stack for the child process might be a bit dangerous, since if you overflow the child process's stack, it could clobber the parent process's stack.

I was a bit concerned about the `malloc` as well at first, but after reading through the rest of the PR more carefully, I see that the process that calls `exec_in_pidns()` will never return since it always exits.

richardxia:
Just to double check my understanding, with this change, there are now three processes involved?

- Grandparent
  - Calls `run_in_fuse()`, which will call the new `exec_in_pidns()` function
  - Never returns from `exec_in_pdns()`
  - Will hold on to the FUSE file descriptor
- Parent
  - The result of the `clone()` call in `exec_in_pidns()`, which starts executing `pidns_init()`
  - Is created in a new PID namespace
  - Is responsible for setting up the new PID namespace / `/proc` file system
  - Is PID 0 and is the "`init`" process of this new namespace, reaping child processes
- Child
  - The result of the `fork()` call in `pidns_init()`
  - All it does it call `execve_wrapper()` to actually run the command that the user specified

Both the grandparent and the parent `wait()` for all child processes to terminate, and the parent is guaranteed(?) to catch all descendant processes because it is the root of a new PID namespace, so even if user command leaves orphaned processes, they will get reparented to this PID root process.

At a high level, this makes sense to me. One question I have, though, is whether Grandparent and Parent need to be separate processes, or could we leak the FUSE file descriptor into the PID root process?

src/wakefs/fuse.cpp
JakeSiFive: Nothing will clobber because the memory is copy on write in the child process like it is with fork(). You have to specify CLONE_VM in order to get it to *not* copy on write. But you seem to be right about not needing the -1.

richardxia: I meant in the glibc `clone()` function, which does a bit of work in the current address space before actually calling the `clone` syscall.

JakeSiFive:
> so even if user command leaves orphaned processes, they will get reparented to this PID root process.
Also my earilier question
> Do you wait on all children or do you wait on all processes in the namespace?
I need some help understanding this. I'm not clear on if the Parent (per Richard's description) or the Child is the root of the namespace. I would have assumed the Parent was the root of the namespace. But the Child is the process that waits on all children. It would be nice if the root of the namespace did wait on all children because then we wouldn't need the unshare hack.

src/wakefs/fuse.cpp
richardxia: Right, yeah, I think I agree that that should be enough.

src/wakefs/fuse.cpp
```diff
@@ -127,6 +129,32 @@ static bool collect_result_metadata(const std::string daemon_output, const struc
...
+    std::cerr << "waitpid (clone) " << nsargs->command[0] << std::endl;
+    exit(1);
+  }
+
+  if (WIFEXITED(status)) {
+    exit(WEXITSTATUS(status));
+  } else {
+    exit(-WTERMSIG(status));
+  }
```
V-FEXrt, line 154: nit:

```
  if (WIFEXITED(status)) {
    exit(WEXITSTATUS(status));
  }
  exit(-WTERMSIG(status));
```

src/wakefs/namespace.cpp
```diff
@@ -459,4 +459,48 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
+    }
+  }
+
+  if (WIFEXITED(status)) {
+    status = WEXITSTATUS(status);
+  } else {
+    status = -WTERMSIG(status);
+  }
+  exit(status);
```
V-FEXrt, line 504: nit:

```
  if (WIFEXITED(status)) {
    exit(WEXITSTATUS(status));
  }
  exit(-WTERMSIG(status));
```

src/wakefs/fuse.cpp
mmjconolly: Is the consensus: stay with malloc() but use 1 page as size?

JakeSiFive: My understanding of the consensus was this. Having the stack allocated array be a page in size ensures that glibc won't clobber anything you care about and since you're using part of the stack the child processes stack will be close to the same size as the parent process's stack and there's no allocation.

```
alignas(16) uint8_t child_stack[4096];
void* stk = reinterpret_cast<void*>(child_stack + sizeof(child_stack));
pid_t child_pid = clone(pidns_init, stk, CLONE_NEWPID | SIGCHLD, nsargs);
```

src/wakefs/fuse.cpp
JakeSiFive: Also can you mark this function as `[[ noreturn ]]`

https://en.cppreference.com/w/cpp/language/attributes/noreturn

That will mainly make it clear that this function doesn't return. It also lets the compiler do some optimizations but that doesn't really matter here.

src/wakefs/namespace.cpp
JakeSiFive: I'm fine with either way but personally I prefer what Matthew has now. Just saying its a preference thing.

src/wakefs/namespace.cpp
```diff
@@ -459,4 +459,48 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
   return true;
 }
 
+int pidns_init(void *arg) {
```
JakeSiFive, line 462: You should mark this as `[[ noreturn ]]` as well: https://en.cppreference.com/w/cpp/language/attributes/noreturn

mmjconolly:
> Just to double check my understanding, with this change, there are now three processes involved?
>
>     * Grandparent
>       * Calls `run_in_fuse()`, which will call the new `exec_in_pidns()` function
>       * Never returns from `exec_in_pdns()`
>       * Will hold on to the FUSE file descriptor
>     * Parent
>       * The result of the `clone()` call in `exec_in_pidns()`, which starts executing `pidns_init()`
>       * Is created in a new PID namespace
>       * Is responsible for setting up the new PID namespace / `/proc` file system
>       * Is PID 0 and is the "`init`" process of this new namespace, reaping child processes
>     * Child
>       * The result of the `fork()` call in `pidns_init()`
>       * All it does it call `execve_wrapper()` to actually run the command that the user specified

In my head it's 4, I call them 'wakebox', who creates 'ws-mount-ns', who creates 'ws-pid-ns', and then 'payload'

My mental model of this is:

* wakebox
    * Reads the input json
    * Connects to fuse, will hold onto the FUSE file descriptor 
    * Calls `run_in_fuse()` which fork()s, creating ws-mount-ns
    * Waits for result of fork()
    * Writes the output json
    * Is not itself in any user namespace (unless someone does something externally)
* ws-mount-ns
    * unshare()s into a user/mount namespace
    * Will hold onto the FUSE file descriptor
    * Makes changes to mount table (such as replacing rootfs, or loading a squashfs)
    * Reads any environment modifications that are requested by mounts
    * calls `exec_in_pidns`, which clone()s to create ws-pid-ns in a PID namespace
    * waits for ws-pid-ns to terminate and captures its exitcode
    * ws-mount-ns never returns, exits supplying ws-pid-ns's exit code to wakebox
* ws-pid-ns
    * is in a PID namespace at all times, and inherets ws-mount-ns's user/mount namespace
    * is process id 1 in the new namespace, any decedents will be re-parented to ws-pid-ns if needed
    * Still has the FUSE file descriptor (haven't verified)
    * mounts /proc
    * fork()s out to the real payload command
    * waits for all (potentially re-parented) children
    * exits supplying exit code to ws-mount-ns (who supplies it to wakebox, who puts it in output json)
    * if ws-pid-ns somehow terminates before its decedents, they are all destroyed
 * payload command
    * is in a user/mount/pid namespace
    * is PID 2, and may spawn more children
    * does not have FUSE file descriptor
    * any children who try to escape will be re-parented to ws-pid-ns

> Both the grandparent and the parent `wait()` for all child processes to terminate, and the parent is guaranteed(?) to catch all descendant processes because it is the root of a new PID namespace, so even if user command leaves orphaned processes, they will get reparented to this PID root process.

Yeah this matches my understanding

> One question I have, though, is whether Grandparent and Parent need to be separate processes,

I think it may be possible to change the fork() that creates ws-mount-ns to a clone(), so and have it create the pid namespace at the same time as the other namespaces. Clone is linux-only so it would be a larger refactor to do so.
ws-mount-ns and ws-pid-ns would become one, rather than involving the grandparent wakebox as asked in the question.

or could we leak the FUSE file descriptor into the PID root process?

the PID root process would obtain the fuse file descriptor (I haven't verified though), as we've not yet exec(). However I feel that's ok as the payload command wouldn't obtain it, and has the benefit of strongly connecting the fuse daemon's liveliness with the destruction of the pid namespace

mmjconolly:
> I need some help understanding this. I'm not clear on if the Parent (per Richard's description) or the Child is the root of the namespace. I would have assumed the Parent was the root of the namespace. But the Child is the process that waits on all children. It would be nice if the root of the namespace did wait on all children because then we wouldn't need the unshare hack.

Is this refering to the use of unshare() within wakebox, or externally wrapping wakebox?

If you're referring to the external wrapping, addressing that is not part of the PR and waiting on more children wouldn't help, as the connection to fuse-waked happens outside of any user namespace - from the perspective of a not-wrapped wakebox process

For usage of unshare() within wakebox, in my reply to Richard I address this by describing how its possible (though I'm reluctant) to flatten the wb-mount-ns and wb-pid-ns into one

src/wakefs/namespace.cpp
V-FEXrt: Yep def preference thing. That's why its a nitpick :)

JakeSiFive:
> Is this refering to the use of unshare() within wakebox, or externally wrapping wakebox?
> 
> If you're referring to the external wrapping, addressing that is not part of the PR and waiting on more children wouldn't help, as the connection to fuse-waked happens outside of any user namespace - from the perspective of a not-wrapped wakebox process
> 
> For usage of unshare() within wakebox, in my reply to Richard I address this by describing how its possible (though I'm reluctant) to flatten the wb-mount-ns and wb-pid-ns into one

Yeah I meant the external unshare. I wasn't thinking about the order here correctly apparently!

richardxia:
> For usage of unshare() within wakebox, in my reply to Richard I address this by describing how its possible (though I'm reluctant) to flatten the wb-mount-ns and wb-pid-ns into one

Yeah, makes sense to not do this, at least if we're still planning to try to support macOS.

src/wakefs/fuse.cpp
mmjconolly: > I think this can be a const reference instead of a pointer.

I need to pass `nsargs` into the clone function as a void pointer, is there a way to transform a reference into a pointer or is that not advisable?

src/wakefs/namespace.cpp
mmjconolly: errno is defined as an int, but I think its understood that it's really char-sized
```
$  errno -l | wc -l
134
```

JakeSiFive:
> Given that MacOS support is being dropped for wake generally, I feel that this PR could be refactored to consolidate the two intermediate wb-mount-ns/wb-pid-ns processes into the one clone() call

I'm good with that.

V-FEXrt:
Same

mmjconolly:
The previously reviewed code is bb38624eee6c172ad905e46b1a567e65f3c122da
the new changes are in 8c3086666136b09c6d8814a3d6c5fde9d057f092

Changes since last review:
* the PID namespace is optional, and off by default
* some linux-only ifdefs that were near code I was modifying were removed

I'd like to roll this out as default-off as a risk mitigation strategy.
If we're later confident that this works on heavier workloads, we can set default-on.

mmjconolly (commented):
I'm going to change this to use a wakebox spec file option, rather than CLI option

share/wake/lib/system/job.wake
```diff
@@ -805,17 +805,11 @@ export def getJobStatus (job: Job): Status = match (getJobReport job)
...
-# - "isolate/net": disables network access
-# - "isolate/workspace": makes the build appear run in /var/cache/wake
-#    ... if /var/cache/wake does not exist, a directory 'build/wake' is
-#        used relative to where wake has been installed
+# Location of the wake executable
 export def wakePath: String =
-    prim "execpath" # location of the wake executable
+    prim "execpath"
 
+# Implement FUSE-based Runner
```
mmjconolly, line 812: This comment was edited as it listed options that haven't existed for a long time

mmjconolly:
I think this is ready for re-review

src/wakefs/fuse.cpp
```diff
@@ -37,6 +37,10 @@
 #include <iostream>
 #include <sstream>
 
+#ifdef __linux__
```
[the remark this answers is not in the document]
mmjconolly: The wasm target doesn't seem to have this header

src/wakefs/fuse.cpp
```diff
@@ -185,8 +192,17 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
       }
     }
 
-    int err = execve_wrapper(command, args.environment);
-    std::cerr << "execve " << command[0] << ": " << strerror(err) << std::endl;
+#ifdef __linux__
```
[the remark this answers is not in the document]
mmjconolly: As above but with some indirection, 
exec_in_pidns is defined in a linux-only context, as it relies on functions that are not in wasm

src/wakefs/fuse.cpp
JakeSiFive: Sorry I didn't get back on this, A reference can have `&` called on it to make it a pointer which can then be cast to a void pointer.

src/wakefs/fuse.cpp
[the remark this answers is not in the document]
JakeSiFive: ah gross. I want to re-configure some things at some point so that the wasm target doesn't build most things like this that don't make sense for it.

JakeSiFive:
Ok I spoke with Matthew and after I spoke with Matthew here's my (thus far approved) understanding of whats going on here in my own words:

ok so here's my current understanding

Without isolate-pid flow:
1. wakebox eventually calls run_in_fuse
2. run_in_fuse creates a child process called wb-mount-ns and this child process setups its namespaces and mounts (this also involves starting up the fuse daemon)
3. wb-mount-ns calls execve_wrapper, which just execs and because the live file created in step 2 was not marked as close-on-exec, the file handle is leaked
4. the wakebox parent process waits on wb-mount-ns 

With isolate-pid flow:
1. wakebox eventually calls run_in_fuse
2. run_in_fuse creates a child process called wb-mount-ns and this child process setups its namespaces and mounts (this also involves starting up the fuse daemon)
3. wb-mount-ns calls exec_in_pidns 
4. exec_in_pidns creates a child process called wb-pid-ns 
5. wb-pid-ns mounts a new /proc fs
6. wb-pid-ns forks to create a child process
7. That child process calls execve_wrapper, which just execs the payload but this time the live file is closed because close-on-exec was set in step 2
8. wb-pid-ns  waits on all child processes (which includes the payload, and anything that it might create that gets reparented to wb-pid-ns )
9. wb-mount-ns  waits on wb-pid-ns
10. the wakebox parent process waits on wb-mount-ns

src/wakefs/namespace.cpp
```diff
@@ -465,4 +475,71 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
+    }
+  }
+
+  if (WIFEXITED(status)) {
+    exit(WEXITSTATUS(status));
+  }
+  exit(-WTERMSIG(status));
+}
+
+[[noreturn]] void exec_in_pidns(pidns_args *nsargs) {
```
JakeSiFive, line 521: very minor nit: This could still be a reference, I don't think it can be `const` anymore because you'll have to const_cast it to get to a `void*` and I'd rather not do that.

JakeSiFive (approved):
Ok, I think I've managed to approximently grok what this is actually doing and I can't personally think of any step that does not appear to be required to me (except maybe in the isolate pid case we could shave 1 fork but I think it would clutter the code since that fork is required in the no-isolate-pid case). Still maybe Richard or Wes can think of some way to refactor this that would be beneficial. I still don't have sufficient background to confidently say there is no other better way to structure this.

I don't see any major remaining pure-code issues (just adding one minor nit). So I'm going to mark this as approved with a request that Richard also take a look.

mmjconolly:
Using reference instead of pointer added in ca96d60dfb28094125465736b084610a873ffd67

mmjconolly:
@richardxia do you seem ok with this? It's disabled by default

richardxia:
> @richardxia do you seem ok with this? It's disabled by default

This PR has completely paged out of my memory, but just taking @mmjconolly's [recent summary of changes](https://github.com/sifive/wake/pull/1127#issuecomment-1605044194) he made at face value, yes, I think this sounds good to me.

mmjconolly:
From manual testing I believe this does the right thing

The payload command does not have `fuse-waked`'s live file, and when a payload command tries to double-fork away it is trapped as a child of `wb-pid-ns`, which does hold the live file open, so if the `wb-pid-ns` is killed the payload is killed and the fuse-daemon is notified to exit.

PR #1133, opened 2023-02-24
Update `sortBy` examples to use `Order` and to be executable

bmitc:
`sortBy` was recently updated in 0.30 to use `Order` comparison rather than `Boolean` comparison, but the examples and documentation weren't updated. I have updated the examples to use comparisons returning an `Order` in addition to making the examples executable. Personally, I think explicit examples should be executable, as that's typically the first thing a user is going to do by copying them, executing them, and then modifying them to learn a function's behavior. The more generic examples are useful, but I think those probably belong as part of the documentation string.

### DIFF
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -637,7 +637,7 @@ export def transpose: List (List a) => List (List a) =
 #
 # Guarantees:
 #  - The output is a permutation of `list`
-#  - If `0 <= x < y < len list` then `cmpFn (at list x) (at list y)` (ignoring None)
+#  - If `0 <= x < y < len list` then `cmpFn (at list x) (at list y) | isLT` (ignoring None)
 #
 # Example:
 # ```
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -646,10 +646,10 @@ export def transpose: List (List a) => List (List a) =
 #
 # Example:
 # ```
-#   sortBy (_<_) (6, 1, 4, 2, Nil) = 1, 2, 4, 6, Nil
-#   sortBy p Nil = Nil
-#   sortBy p (x, Nil) = (x, Nil)
-#   sortBy (\x\y if x == y then False else ! p x y) xs = reverse (sortBy p xs)
+#   sortBy (_<=>_) (6, 1, 4, 2, Nil) = 1, 2, 4, 6, Nil
+#   sortBy (_<=>_) Nil = Nil
+#   sortBy (_<=>_) (1, Nil) = (1, Nil)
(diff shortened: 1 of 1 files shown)


bmitc (commented):
The lines
```wake
# ```
#   if lessThanFn x y then ! lessThanFn y x
#   if lessThanFn x z then lessThanFn x y || lessThanFn y z
# ```
```
are still incorrect, as they still assume that `lessThanFn` returns a boolean. I wasn't for sure what to do with them, but I'm thinking they should just be removed. `!` has an analog for `Order`, although Wake doesn't currently have such a function, but I don't know what the analog of `||` is for `Order`.

I think it might be useful to have a negation function for `Order`. Barring the name, it would just be:
```wake
export def notOrder (order: Order) =
    match order
        LT = GT
        EQ = EQ
        GT = LT
```

JakeSiFive (approved):
Thanks!

bmitc:
@JakeSiFive Should I just remove the lines I called out [here](https://github.com/sifive/wake/pull/1133#pullrequestreview-1313823204)? I don't think they're applicable anymore given that an `Order`ing function is expected now.

JakeSiFive:
> The lines
> 
> ```
> # ```
> #   if lessThanFn x y then ! lessThanFn y x
> #   if lessThanFn x z then lessThanFn x y || lessThanFn y z
> # ```
> ```
> 
> are still incorrect, as they still assume that `lessThanFn` returns a boolean. I wasn't for sure what to do with them, but I'm thinking they should just be removed. `!` has an analog for `Order`, although Wake doesn't currently have such a function, but I don't know what the analog of `||` is for `Order`.
> 
> I think it might be useful to have a negation function for `Order`. Barring the name, it would just be:
> 
> ```
> export def notOrder (order: Order) =
>     match order
>         LT = GT
>         EQ = EQ
>         GT = LT
> ```

Yeah you can just remove them...why did we change this anyway? It seems kind of strange to me (lol watch me be the one that changed it to an order lol)

JakeSiFive:
looks like people were confused by it: https://github.com/sifive/wake/pull/1061

Let's just delete it. I think a function that returns `Order` is implied to be an ordering and I think if people are getting confused by orderings that return booleans then they aren't going to understand those conditions either.

bmitc:
I'm not entirely sure about the update. I personally don't really see the need for it because I didn't find the previous behavior confusing, and I don't particularly like dealing with `Order`, especially since we're missing a helper function or two.

But yea, removed those lines now.

I also adjusted one of the guarantees that I missed the first time [here](https://github.com/sifive/wake/pull/1133/commits/d8d69514962aee8565017aa4289f1624d7f923c9). Can you cross-check me on that change?

JakeSiFive:
Yeah I'm also not a fan but if Sam has evidence that people found it confusing and has a hunch that this is less confusing I'm willing to trust that anecdotal evidence. For what its worth I just went through and updated a large code base using wake and all of the sortBy/groupBy functions were very easy to fix so I think our helpers are mostly sufficient.

PR #1134, opened 2023-02-24
Use wake-specific staging mountpoint for squashfs mounts

mmjconolly:
When using wakebox's squashfs support, if a user has particular versions of libfuse then they can not mount over a non-empty directory (and this was happing in a namespace, and we moved the mount directly aftewards).

We were using `/mnt`, in this PR we create a dedicated directory to put these squashfs mounts before moving them.

(We stage/move them because within the squashfs it contains a file which must be read to say where to really mount it)

### DIFF
--- a/src/wakefs/namespace.cpp
+++ b/src/wakefs/namespace.cpp
@@ -191,7 +191,7 @@ static bool do_squashfuse_mount(const std::string &source, const std::string &mo
     return false;
   }
 
-  int err = mkdir_with_parents(mountpoint, 0777);
+  int err = mkdir_with_parents(mountpoint, 0555);
   if (0 != err) {
     std::cerr << "mkdir_with_parents ('" << mountpoint << "'):" << strerror(err) << std::endl;
     return false;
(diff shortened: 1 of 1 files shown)


mmjconolly:
This now has an extra change
The mountpoint staging areas have the write-bit removed when created, as we don't want anyone putting files in them if some versions of libfuse are going to refuse mounting over non-empty directories

mmjconolly:
(perhaps fixed circleci auth for my user..)

JakeSiFive (commented):
ok so if I understand you're making a read only directory in `tmp` and mounting over it and you're doing this so that you can read the mount file out of the squashfs to figure out where you need to mount it but you have to use a temporary directory to do that that is empty.

Do you have a reason to think that .wakebox-mount-squashfs will always be empty? If not, I think you should add a large random number to that to ensure its empty. xoshiro_rng in src/wcl is made to do this exactly.

mmjconolly:
> Do you have a reason to think that .wakebox-mount-squashfs will always be empty?

I guess it's the same as for `/tmp/.wakebox-mount`, in that only wakebox should be creating that, so if it is not-empty then that user is just breaking themselves.

If we really want to protect them from that, adding the random numbers could work, but could lead to a build-up of directories in the host's /tmp dir (as this isn't in the mount namespace) in cases where wakebox is terminated abnormally

JakeSiFive:
eh ok fair, if wakebox doesn't put anything there (at least not in this namespace) then this seems fine to me.

JakeSiFive (approved):

V-FEXrt (approved):

JakeSiFive:
Oh yeah. I want us to get to the system that the Chrome team has where every 6 weeks they release chrome and they never ask what features people want in and they never delay a release. There's just a release team that maintains a release branch every 6 weeks and does QA and what makes it in makes it in and what doesn't doesn't. I'd like for our releases to be even more frequent than that actully.

PR #1135, opened 2023-02-24
Remove version libs and delete `print`

JakeSiFive:
This change removes `print` from wake and deletes all versioned libraries. Going forward the new policy should be to make versioned libraries only if needed. We have good reason to believe that it should be relatively easy to release wake and that the versioned libraries are causing more pain than they're worth.


mmjconolly (approved):

ag-eitilt:
I know the ship has sailed by this point, but I'm not a fan of this change, and certainly not as it's been implemented.  I do agree we were gathering quite a number of seemingly-unnecessary versions when we really only needed a small handful and that the general policy over what goes into those packages/what purpose they served needed clarification and recording, but we moved away from a plain `from wake import` for a reason not too long ago, and I don't think that the reasons behind that have changed any or that the requirements for supporting the structure we had were too onerous.  (See also #1050 where we explicitly decided *not* to remove *some* of the packages to maintain support for even older codebases than this change will affect.)

I suppose we might be able to get by with enforcing a dual `from wake import _ \n from wake_next import explicitName` redundancy as needed (though I severely doubt it will actually be used in practice, not to mention that it means we have to search for every deprecated symbol individually rather than just the obsolete package name), but as it stands and unless I've missed the discussion in the meeting notes, we currently have no structure in place which will allow us to migrate over the large change to the JSON API, for example: the old `//` is still used throughout the large private codebase, and any migration strategy we use will *have* to support both APIs for at least one Wake version -- there's simply too many uses to update everything all at once.  Even manually toggling a `from wake_old import` only on the relevant files will be very prone to merge conflicts.  By explicitly saying "this file was written for the API of this version", we got the ability to defer updates as piecemeal work for free.

JakeSiFive:
We're still allowing versioned APIs as needed for migrations. The rule is just "don't use them if you don't have to"

PR #1138, opened 2023-02-27
move make_canonical to wcl

V-FEXrt:
Moves `make_canonical` to `wcl`, adds tests, and per DM removes windows support


JakeSiFive (approved):
lgtm

PR #1141, opened 2023-03-02
Integrate eviction tool into job-cache flow

V-FEXrt:


build.wake
```diff
@@ -36,9 +36,41 @@ export def install: List String => Result String Error = match _
     dest, Nil = doInstall (in cwd dest) "default" | rmap (\_ "INSTALL")
     _ = Fail "no directory specified (try: install /opt/local)".makeError
 
+def bootstrapTarget targetFn variant =
```
V-FEXrt, line 39: I have no idea how you'll feel about this but it seemed like a neat solution @JakeSiFive 

I actually had the thought that it would be cool to gut the Makefile as much as possible. Maybe we could have the Makefile just build wake and shim-wake, then run a bootstrap build flow where everything is build it wake (would require `setLocalPlanOnly True` since we wouldn't have wakebox yet)

src/job-cache/job-cache.cpp
```diff
@@ -31,6 +31,7 @@
 #include <sys/file.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
```
V-FEXrt, line 34: I think I can delete this

src/job-cache/job-cache.cpp
```diff
@@ -1191,6 +1194,16 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
     redirect_path(input_dir);
   }
 
+  // TODO: write a constructor that takes type and job_id
```
V-FEXrt, line 1197: still need to do this TODO

tools/evict-shared-cache/evict-shared-cache.cpp
```diff
@@ -108,6 +96,10 @@ struct CommandParser {
         }
         iter = end + 1;
       }
+
+      if (count < 4096) {
```
V-FEXrt, line 100: Something like this is required or we will not process any messages until just before exiting

tools/wake/main.cpp
```diff
@@ -283,6 +283,7 @@ int main(int argc, char **argv) {
 
   if (lsp) {
     std::string lsp = wcl::make_canonical(find_execpath() + "/../lib/wake/lsp-wake");
+    std::cerr << lsp << std::endl;
```
V-FEXrt, line 286: nit: delete this

tools/evict-shared-cache/evict-shared-cache.cpp
JakeSiFive: Yeah because I misunderstood the read semantics I agree now.

src/job-cache/job-cache.cpp
V-FEXrt: nope, this is needed for waitpid

JakeSiFive (approved):

mmjconolly:
Please consider having descriptions for PRs, it makes it much easier to others to understand when viewing later

PR #1142, opened 2023-03-02
Small code cleanup

V-FEXrt:
- Renames the main files for each tool to `main.cpp`
- Clears an "ignored result" warning
- use ` _` in file names instead of `-` for consistency


mmjconolly:
one of the modified files `src/job-cache/job-cache.cpp` has a dash in the name, in case that's something you're looking to keep consistent

V-FEXrt:
You're too good/too quick @mmjconolly 

I also saw that but I didn't really like having `job-cache/job_cache.h` as an import path so I didn't change that one. 

Probably a good idea to just rename the folder as well

mmjconolly (approved):
rubber stamp

src/job_cache/job_cache.cpp
```diff
@@ -1145,7 +1145,9 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
       mkdir_all(pair.second.begin(), pair.second.end());
 
       // Lastly make the symlink
-      symlink(pair.first.c_str(), output_symlink.value.c_str());
+      if (symlink(pair.first.c_str(), output_symlink.value.c_str()) == -1) {
+        std::cerr << "[WARN] failed to create symlink" << std::endl;
```
JakeSiFive, line 1149: This should be `log_fatal` I just missed checking the error.

JakeSiFive (approved):
please make the log_fatal change

V-FEXrt:
@JakeSiFive I'm not super familiar with the tests for shared caching but it looks like log_fatal broke the `symlink-dir`  test.

```
./pass.sh ../../../bin
Pass ("bar", "bar/baz.txt", "test.txt", Nil)
[pid=13869, 2023-03-06 12:35:56] symlink(./bar/baz.txt, test.txt): File exists
```

Any pointers on what I should do to fix it? This is unexpected behavior right?

JakeSiFive:
looking at this failure now

src/job_cache/job_cache.cpp
```diff
@@ -1145,10 +1145,11 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
       mkdir_all(pair.second.begin(), pair.second.end());
 
       // Lastly make the symlink
-      if (symlink(pair.first.c_str(), output_symlink.value.c_str()) == -1) {
-        log_fatal("symlink(%s, %s): %s", pair.first.c_str(), output_symlink.value.c_str(),
-                  strerror(errno));
+      std::string tmp_link = rng.unique_name();
```
JakeSiFive, line 1148: It needs to be adjacent to the output_symlink location so this should be `output_symlink.value + "." + rng.unique_name()` that also makes debugging easier if we ever find one of those files lying around.

src/job_cache/job_cache.cpp
```diff
@@ -1145,10 +1145,11 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
...
+      if (symlink(pair.first.c_str(), tmp_link.c_str()) == -1) {
+        log_fatal("symlink(%s, %s): %s", pair.first.c_str(), tmp_link.c_str(), strerror(errno));
       }
+      rename_no_fail(tmp_link.c_str(), output_symlink.value.c_str());
```
JakeSiFive, line 1152: The man page for rename is very cryptic about how it works with symlinks. I'm not sure I understand what's going on. I think you need to do a link-unlink instead to get the right semantics...normally rename is just a link-unlink but maybe its different for symlinks?

PR #1143, opened 2023-03-02
Fix parsing wake version

### LINKED ISSUE #1139: .wakeroot version check fails to parse wake's own version number
```
$ echo 0.40.0 > .wakeroot
$ wake -x Unit
.wakeroot: major version is not a number (0.30.0)
$ wake --version
wake 0.30.0
```

V-FEXrt:
Fixes #1139 

For some reason we were skipping the first character of the wake version meaning `0.30.0` would be seen as `.30.0` during semver parsing. My thought was that previously it started with `v` and we needed to skip that but I went back to 0.18.0 and even that didn't have a leading `v` so this must be a fairly longstanding bug.

I still don't know why we were skipping so if anyone can give context that would be great, this does fix the issue though.


mmjconolly:
I think it's just that we've never used this version-check functionality before

mmjconolly (approved):

PR #1144, opened 2023-03-03
Move term.h into util and fix "silent" util deps

V-FEXrt:
term.h was split across runtime and util and needs to exist properly in one library. This change makes that happen. In doing this, it was discovered that many tools depended on util without declaring so. This was also fixed.


tools/lsp-wake/lsp.cpp
```diff
@@ -64,9 +64,6 @@ static const char *ServerNotInitialized = "-32002";
 
 DiagnosticReporter *reporter;
 
-const char *term_colour(int _) { return ""; }
-const char *term_normal() { return ""; }
-
```
V-FEXrt, line 69: I feel somewhat worried about this change. No idea why these were stubbed out in lsp.cpp but now that it is probably depending on util/term.h these were defined twice and needed to be deleted

Makefile
```diff
@@ -6,7 +6,7 @@ CC	:= cc -std=c11
 CXX	:= c++
 CXX_VERSION := -std=c++14
 CFLAGS	:= -Wall -Wno-format-security -O2 -DVERSION=$(VERSION)
-LDFLAGS	:= -lpthread
+LDFLAGS	:= -lpthread -lncurses -ltinfo
```
V-FEXrt, line 9: this is needed since the bootstrapped tools all depend on util/ and util/ now needed them. We could maybe make "term.h" its own library if we wanted to not have this here

V-FEXrt:
I'm pretty sure this change is needed to fix CI for #1141. Will confirm in a bit

mmjconolly:
Seems to have `bin/evict-shared-cache` commited as binary file

mmjconolly:
[c53e5c9](https://github.com/sifive/wake/pull/1144/commits/c53e5c95158c24b5277773d38cb24a0d968895f8) wont remove the file from history (and thus storage), unless this branch is either force pushed, or if the PR is squashed-as-merge with it's initial branch deleted

V-FEXrt:
Went ahead and squashed then forced pushed but the general policy is to squash-and-merge for sifive/wake

mmjconolly:
> but the general policy is to squash-and-merge for sifive/wake

ah ok

JakeSiFive (approved):
Yeah this sounds like a good idea to me! I forgot why we waited to move term.cpp over but glad we did it!

PR #1146, opened 2023-03-07
Set labels for all plans

V-FEXrt:
There were several Plans within wake that didn't have labels. This change adds labels everywhere they were missing.

### DIFF
--- a/extensions/vscode/vscode.wake
+++ b/extensions/vscode/vscode.wake
@@ -70,6 +70,7 @@ export def vscode _: Result Path Error =
 
     require Pass nodeModules =
         makeExecPlan (npm, "ci", Nil) (packageJSON, packageLockJSON, Nil)
+        | setPlanLabel "Install npm dependencies (npm ci)"
         | editPlanEnvironment (addEnvironmentPath emsdkBinDir)
         | setPlanDirectory @here
         | runJob
@@ -115,6 +116,7 @@ export def vscode _: Result Path Error =
 
     require Pass outputs =
         makeExecPlan (npx, "@vscode/vsce@2.15.0", "package", Nil) vsceFiles
+        | setPlanLabel "Package vsix (vsce package)"
         | editPlanEnvironment (addEnvironmentPath emsdkBinDir)
         | setPlanDirectory here
         | setPlanStdin (relative @here vsceStdin.getPathName)
--- a/extensions/vscode/vscode.wake
+++ b/extensions/vscode/vscode.wake
@@ -70,7 +70,7 @@ export def vscode _: Result Path Error =
 
     require Pass nodeModules =
         makeExecPlan (npm, "ci", Nil) (packageJSON, packageLockJSON, Nil)
-        | setPlanLabel "Install npm dependencies (npm ci)"
+        | setPlanLabel "npm ci: Install npm dependencies"
         | editPlanEnvironment (addEnvironmentPath emsdkBinDir)
         | setPlanDirectory @here
         | runJob
--- a/extensions/vscode/vscode.wake
+++ b/extensions/vscode/vscode.wake
@@ -116,7 +116,7 @@ export def vscode _: Result Path Error =
 
     require Pass outputs =
         makeExecPlan (npx, "@vscode/vsce@2.15.0", "package", Nil) vsceFiles
-        | setPlanLabel "Package vsix (vsce package)"
+        | setPlanLabel "vsce: Package vsix extension"
         | editPlanEnvironment (addEnvironmentPath emsdkBinDir)
         | setPlanDirectory here
         | setPlanStdin (relative @here vsceStdin.getPathName)
--- a/share/wake/lib/gcc_wake/gcc.wake
+++ b/share/wake/lib/gcc_wake/gcc.wake
@@ -27,7 +27,7 @@ export def staticLFlags  = ("-lpthread", "-flto", "-static", "-s", Nil)
 
 target pwd _ =
     makeExecPlan ("pwd",) Nil
-    | setPlanLabel "pwd: Get pwd"
+    | setPlanLabel "pwd: Get workspace directory"
     | setPlanLocalOnly True
     | runJob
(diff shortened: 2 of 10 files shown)


mmjconolly (approved):
rubber stamp

extensions/vscode/vscode.wake
```diff
@@ -70,6 +70,7 @@ export def vscode _: Result Path Error =
 
     require Pass nodeModules =
         makeExecPlan (npm, "ci", Nil) (packageJSON, packageLockJSON, Nil)
+        | setPlanLabel "Install npm dependencies (npm ci)"
```
ag-eitilt, line 73: This additional context is certainly better than we had previously, but (bikeshedding) I wonder if it would be better to put the command/tool first on the line for easy at-a-glance scanning?
```suggestion
        | setPlanLabel "npm ci: Install npm dependencies"
```

share/wake/lib/gcc_wake/pkgconfig.wake
```diff
@@ -33,16 +33,17 @@ export def flattenSysLibs (packages: List SysLib): SysLib =
...
     else None
 
-    def cmdline = pkgConfig, args
+    def cmdline = pkgConfig, flags ++ pkgs
     def addenv list = match (getenv "PKG_CONFIG_PATH")
         Some p = "PKG_CONFIG_PATH={p}", list
         None   = list
     def result =
         makeExecPlan cmdline Nil
+        | setPlanLabel "pkg-config: {pkgs | catWith " "}"
```
ag-eitilt, line 46: This is the format I mentioned above.  "What is pkg-config doing?  This list of packages."

share/wake/lib/system/io.wake
```diff
@@ -36,6 +36,7 @@ target writeImp inputs mode path content =
                     Pass path = Pass (editRunnerOutputOutputs (path, _) output)
         makeRunner "write" (\_ Pass 0.0) pre post virtualRunner
     makeExecPlan ("<write>", str mode, path, Nil) inputs
+    | setPlanLabel "<write> {str mode} {path}"
```
ag-eitilt, line 39: I was going to say something about Plans without labels potentially falling back to the command line, but I realized that even if that's what we *show*, that's not what's *stored*.  This line may or may not change the live output, but it is definitely important for making the database queries look nice.  (`wake --last --metadata | grep Job` is only of limited use without it.)

share/wake/lib/system/sources.wake
```diff
@@ -90,6 +91,7 @@ export def claim (file: String): Result Path Error =
             # Its important that the command here is distinct from what `source` would use
             # so that claim and source do not overlap.
             makeExecPlan ("<claim>", str time, file, Nil) Nil
+            | setPlanLabel "<claim> {file}"
```
ag-eitilt, line 94: Now that we're separating the label from the command, do we want to change the format?  The `<name>` angle brackets were only required to ensure it wouldn't clash with some tool/script which also happened to be called `claim`.
```suggestion
            | setPlanLabel "claim: {file}"
```

share/wake/lib/system/sources.wake
```diff
@@ -152,6 +154,7 @@ export def claimFileAsPathIn (outputDirectory: Path) (existingFile: String) (des
         def visible = outputDirectory, Nil
         def cmdline = which "cp", existingFile, desiredWorkspacePath, Nil
         makeExecPlan cmdline visible
+        | setPlanLabel "claimFileAsPathIn: cp {existingFile} {desiredWorkspacePath}"
```
ag-eitilt, line 157: Same here, this seems like it could be simplified quite a bit (I'm not a huge fan of using Wake function names as label prefixes, but definitely not strongly opposed even so).
```suggestion
        | setPlanLabel "claim: {existingFile} -> {desiredWorkspacePath}"
```
```suggestion
        | setPlanLabel "cp: {existingFile} -> {desiredWorkspacePath}"
```

tests/runtime/symlinks/test.wake
```diff
@@ -5,11 +5,13 @@ export def goodLink Unit =
     require Pass dat =
         datFile Unit
     makeExecPlan ("ln", "-s", "datFile", "goodLink", Nil) (dat, Nil)
+    | setPlanLabel "goodLink"
```
ag-eitilt, line 8: In lieu of a tool name here, it might be better to label with the role?  Though I definitely don't want to introduce too strong a precedent for that; we definitely do *not* want anything which leads people toward `addLabelContext`.
```suggestion
    | setPlanLabel "test (symlinks): goodLink"
```

vendor/re2c.wake
```diff
@@ -23,6 +23,7 @@ def re2cOk Unit: Result String Error =
 
     require Some x, Some y, _ =
         makeExecPlan (re2c, "--version", Nil) Nil
+        | setPlanLabel "re2c --version"
```
ag-eitilt, line 26:
```suggestion
        | setPlanLabel "re2c: Get version number"
```

ag-eitilt (approved):
Looks good as is, thanks!  I just had a couple thoughts of ways which we may (or may not) want to polish it further.

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -27,7 +27,7 @@ export def staticLFlags  = ("-lpthread", "-flto", "-static", "-s", Nil)
 
 target pwd _ =
     makeExecPlan ("pwd",) Nil
-    | setPlanLabel "pwd"
+    | setPlanLabel "pwd: Get pwd"
```
ag-eitilt, line 30: Rather tautological, eh? 😉 
```suggestion
    | setPlanLabel "pwd: Get workspace directory"
```

ag-eitilt (approved):

share/wake/lib/gcc_wake/gcc.wake
V-FEXrt: We love a good tautology

PR #1147, opened 2023-03-07
Support capturing jobs via label glob

V-FEXrt:
Running wake with `--label` allows you to provide a glob  to capture jobs based on the job label. Commandline globbing isn't supported since commandline is stored as a blob not a string. This may be added in the future.

```
> wake --label '*123*'
> wake--label '*re2c*'


# re2c: Get version number(299)
$ /usr/bin/re2c --version


re2c 3.0


# re2c: src/parser/lexer.cpp(320)
$ /usr/bin/re2c -8 --no-generation-date --input-encoding utf8 src/parser/lexer.re -o src/parser/lexer.cpp




# re2c: src/json/jlexer.cpp(332)
$ /usr/bin/re2c -8 --no-generation-date --input-encoding utf8 src/json/jlexer.re -o src/json/jlexer.cpp

> wake --label 'lemon*'


# lemon: m4 src/parser/parser.y.m4(33)
$ /usr/bin/m4 src/parser/parser.y.m4
...
...
```

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -107,6 +107,7 @@ void print_help(const char *argv0) {
     << "    --clean          Delete all job outputs"                                     << std::endl
     << "    --input  -i FILE Capture jobs which read FILES"                              << std::endl
     << "    --output -o FILE Capture jobs which wrote FILES"                             << std::endl
+    << "    --label     GLOB Capture jobs where label matches GLOB"                      << std::endl
     << "    --job       JOB  Captre the job with the specified job id"                   << std::endl
     << "    --last     -l    See --last-used"                                            << std::endl
     << "    --last-used      Capture all jobs used by last build. Regardless of cache"   << std::endl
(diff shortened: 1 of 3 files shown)


mmjconolly:
It's hard to tell what's happening in the example log due to the whitespace lines
What does this segment mean?
```
> wake --label '*123*'
> wake--label '*re2c*'
```

ag-eitilt:
I like the syntax (barring the #1154 issue), but haven't gone through the code yet; depending on how busy tomorrow turns out to be, I'll try to take a look then.

> It's hard to tell what's happening in the example log due to the whitespace lines What does this segment mean?
> 
> ```
> > wake --label '*123*'
> > wake--label '*re2c*'
> ```

My guess is that there's no label which contains the text "123", and so the first line is showing the (non-existent) result when nothing matches, while the second line (which is just missing a space in transcription) starts the example of searching for "re2c".

V-FEXrt:
Yup @ag-eitilt  is correct!

tools/wake/main.cpp
```diff
@@ -582,6 +601,12 @@ int main(int argc, char **argv) {
     }
   }
 
+  if (label) {
+    std::string glob = argv[1];
+    std::replace(glob.begin(), glob.end(), '*', '%');
```
JakeSiFive, line 606: I'm slightly skeptical of this line TBH but its probably fine. I saw land it and we can just say "oh darn" if we see a bug in the future here. 

We should add `?` being replaced with `_` as well btw since that's the single character setup for SQL

JakeSiFive (approved):
Code looks good! I say land it

tools/wake/main.cpp
V-FEXrt: > We should add ? being replaced with _ as well btw since that's the single character setup for SQL
want me to add that before landing?

JakeSiFive: Yeah I don't see why not.

PR #1148, opened 2023-03-08
wakebox: Allow --bind with --params file.

richardxia:
It is sometimes useful to take a params file created by a wake run and run it interactively with additional bind mounts. Previously, --bind was ignored when --params is specified, but now additional binds are appended to the ones specified by the params file.

### DIFF
--- a/tools/wakebox/wakebox.cpp
+++ b/tools/wakebox/wakebox.cpp
@@ -60,6 +60,12 @@ void print_help() {
       "                             Use 'eval $WAKEBOX_CMD' to run the command from params file. \n"
       "    -i --allow-interactive   Use default stdin, ignoring the params json file's stdin     \n"
       "                             value.                                                       \n"
+#ifdef __linux__
+      "    -b --bind DIR1:DIR2      Place the directory (or file) at DIR1 within the command's   \n"
+      "                             view of the filesystem at location DIR2.                     \n"
+      "                             May be specified multiple times.                             \n"
+      "                             These are applied after the params file mount ops.           \n"
+#endif
       "    -I --isolate-retcode     Don't allow COMMAND's return code to impact wakebox's return \n"
       "                             code.                                                        \n"
       "                                                                                          \n"
@@ -141,7 +147,7 @@ int run_interactive(const std::string &rootfs, const std::vector<std::string> &t
 }
 
 int run_batch(const char *params_path, bool has_output, bool use_stdin_file, bool use_shell,
-              bool isolate_retcode, const char *result_path) {
+              bool isolate_retcode, const char *result_path, const std::vector<mount_op> &binds) {
   // Read the params file
   std::ifstream ifs(params_path);
   const std::string json((std::istreambuf_iterator<char>(ifs)), (std::istreambuf_iterator<char>()));
@@ -159,6 +165,9 @@ int run_batch(const char *params_path, bool has_output, bool use_stdin_file, boo
     std::cerr << "No command was provided." << std::endl;
     return 1;
   }
+
+  // Append CLI-specified binds after params file binds.
+  args.mount_ops.insert(args.mount_ops.end(), binds.begin(), binds.end());
 
   if (use_shell) {
     std::stringstream ss;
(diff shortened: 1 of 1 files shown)


tools/wakebox/wakebox.cpp
```diff
@@ -259,25 +270,25 @@ int main(int argc, char *argv[]) {
...
+    const std::string s = binds[i];
+    std::string source = s.substr(0, s.find(":"));
+    std::string destination = s.substr(s.find(":") + 1);
+    if (source.empty() || destination.empty() || s.find(":") == std::string::npos) {
+      std::cerr << "Invalid bind: " << s << std::endl;
+      return 1;
+    }
+    bind_ops.push_back({"bind", source, destination});
+  }
+
```
richardxia, line 284: Looks like the macOS builds are failing. When I was reading the code, I was actually kind of curious why this wasn't broken on macOS, given that the `--bind` option is declared under an `#ifdef __linux__`, but I guess the answer is that it _is_ broken on macOS but CI isn't exercising that case because we only try to read the `--bind` option under wakebox's interactive mode.

Should I put everything in this `for` loop in an `#ifdef __linux__` but leave the `bind_ops` outside so that on macOS it just becomes an empty vector?

mmjconolly: As we don't support any mount options on MacOS other than placement of the fuse-waked fuse mount, I think it's reasonable to have `bind_ops` as linux-only (or empty for non-linux)

tools/wakebox/wakebox.cpp
```diff
@@ -72,6 +72,14 @@ void print_help() {
...
       "    -i --allow-interactive   Use default stdin, ignoring the params json file's stdin "
       "value.  \n"
+      "    -b --bind DIR1:DIR2      Place the directory (or file) at DIR1 within the command's "
+      "view  \n"
+      "                             of the filesystem at location DIR2.                            "
+      "  \n"
+      "                             May be specified multiple times.                               "
+      "  \n"
+      "                             These are are applied after the params file mount ops.         "
+      "  \n"
```
mmjconolly, line 82: Does this get printed on MacOS?

tools/wakebox/wakebox.cpp
mmjconolly: Mounting is typically an ordered activity, so I'm wondering if users will care about the ordering here, say if you write
```
wakebox --bind $dir1:/dest1 -p params.json --bind $dir2:/dest2
```
Perhaps the user may want to use `--bind $dir1:/dest1` to modify what's happening before `params.json`.
Though if they're replacing the rootfs I guess it may be less of a use-case to move things around beforehand

richardxia: > Mounting is typically an ordered activity, so I'm wondering if users will care about the ordering here, say if you write

I was wondering about this myself, but I wasn't sure if that was something that we could easily handle with the current option parsing library, since I think, by default, all the `--bind` values get grouped into a single array.

Is this something I should look into? I think we may need to change our option parsing approach to be more of an imperative builder pattern rather than the current declarative approach if we want to keep track of the relative order of different options.

mmjconolly: We could just put something in the `--help` saying that the bind options always happen after(?) the params file is read

mmjconolly: Oh you already have that in the --help

tools/wakebox/wakebox.cpp
richardxia: I haven't built this branch on a Mac yet, but I believe it would. I can add an ifdef around this section, though maybe now's the time to mention that the strings here are really difficult to read because it looks like the string literals have been wrapped at a short line length than the strings they contain, causing the newline characters and sometimes even the last word of the printed line to appear on the following source line.

Should I fix up the string literals in this file first?

tools/wakebox/wakebox.cpp
```diff
@@ -72,6 +72,14 @@ void print_help() {
       "  \n"
       "    -i --allow-interactive   Use default stdin, ignoring the params json file's stdin "
       "value.  \n"
+      "    -b --bind DIR1:DIR2      Place the directory (or file) at DIR1 within the command's "
+      "view  \n"
+      "                             of the filesystem at location DIR2.                            "
+      "  \n"
+      "                             May be specified multiple times.                               "
+      "  \n"
+      "                             These are are applied after the params file mount ops.         "
```
mmjconolly, line 81:
```suggestion
      "                             These are applied after the params file mount ops.         "
```

tools/wakebox/wakebox.cpp
mmjconolly: Yeah an #ifdef is ok
I'm guessing the formatting was overlooked when the clang-format was applied to the whole repo

tools/wakebox/wakebox.cpp
richardxia: I ended up rebasing this PR on top of https://github.com/sifive/wake/pull/1150, where I fixed up the line wrapping first, but I just wanted to mention that I did fix this typo as well.

tools/wakebox/wakebox.cpp
richardxia: This new text is now guarded by an `#ifdef`.

mmjconolly (approved):

PR #1150, opened 2023-03-08
wakebox: Make help source text wrapping legible.

richardxia:
When the codebase was reformatted, it forced line wrapping at the source code level a bit earlier than what the printed text would have allowed. This made it much less readable in the source code, since half the lines would only contain the newline escape character and whitespace.

This hard wraps each line of the help text such that each printed line corresponds to one source line at the expense of losing a few characters of space per line.

I wanted to fix this before making actual changes to the help text in https://github.com/sifive/wake/pull/1148, especially because it looks like I'll be adding an `#ifdef __linux__` in the middle of the help text strings.

If you care to see the difference in wrapping, here's the before and after (on Linux):

### Before
```
Usage: wakebox [OPTIONS] [COMMAND...]

Interactive options
    -r --rootfs FILE         Use a squashfs file as the command's view of the root filesystem.
    -t --toolchain FILE      Make a toolchain visible on the command's view of the filesystem.
                             May be specified multiple times.
    -b --bind DIR1:DIR2      Place the directory (or file) at DIR1 within the command's view
                             of the filesystem at location DIR2.
                             May be specified multiple times.
    -x                       Shorthand for '--bind $PWD:$PWD'
    COMMAND                  The command to run.

Batch options
    -p --params FILE         Json file specifying input parameters.
    -o --output-stats FILE   Json file written containing output results and return code.
    -s --force-shell         Run shell instead of command from params file.
                             Implies --allow-interactive.
                             Use 'eval $WAKEBOX_CMD' to run the command from params file.
    -i --allow-interactive   Use default stdin, ignoring the params json file's stdin value.
    -I --isolate-retcode     Don't allow COMMAND's return code to impact wakebox's return code.

Other options
    -h --help                Print usage
```

### After
```
Usage: wakebox [OPTIONS] [COMMAND...]

Interactive options
    -r --rootfs FILE         Use a squashfs file as the command's view of the root
                             filesystem.
    -t --toolchain FILE      Make a toolchain visible on the command's view of the
                             filesystem.
                             May be specified multiple times.
    -b --bind DIR1:DIR2      Place the directory (or file) at DIR1 within the command's
                             view of the filesystem at location DIR2.
                             May be specified multiple times.
    -x                       Shorthand for '--bind $PWD:$PWD'
    COMMAND                  The command to run.

Batch options
    -p --params FILE         Json file specifying input parameters.
    -o --output-stats FILE   Json file written containing output results and return code.
    -s --force-shell         Run shell instead of command from params file.
                             Implies --allow-interactive.
                             Use 'eval $WAKEBOX_CMD' to run the command from params file.
    -i --allow-interactive   Use default stdin, ignoring the params json file's stdin
                             value.
    -I --isolate-retcode     Don't allow COMMAND's return code to impact wakebox's return
                             code.

Other options
    -h --help                Print usage
```


mmjconolly (approved):

V-FEXrt:
> When the codebase was reformatted, it forced line wrapping at the source code level a bit earlier 

@richardxia if the old format is preferable then below can be used to temporarily disable formatting

```
// clang-format off
"some string where"
     "I know better"
"than the formatter"
// clang-format on
```

richardxia:
> @richardxia if the old format is preferable then below can be used to temporarily disable formatting

Thanks, I had considered doing that, but didn't, for no reason in particular. I think it's ultimately your decision which you prefer, since I'm not a core contributor.

By the way, it doesn't look like I have merge access, so one of you will need to merge this and my other PR from today.

V-FEXrt (approved):

V-FEXrt:
@richardxia  I think you may have to re-auth CircleCI. Ever since the GH 2FA rollout all user-level CircleCI tokens were invalidated.

 https://support.circleci.com/hc/en-us/articles/360043002793-Troubleshooting-CircleCI-Access-After-Enabling-Github-SSO

richardxia:
Thanks, I have no idea if I properly re-auth'd into CircleCI, but I rebased and force-pushed back to this branch.

But still, I don't have merge access, since GitHub is telling me that I can't merge because I am not authorized to, not because CI hasn't passed, so even after the CircleCI jobs succeed, one of you will need to hit the merge button.

mmjconolly:
I've changed some repo settings, did it change for you?

richardxia:
> I've changed some repo settings just now, is it now available to you?

Ah, yes, looks like it's now only saying that I can't merge until the CI tests pass. Thanks!

PR #1155, opened 2023-03-10
Fix output for verbose format

V-FEXrt:


terpstra:
This does not fix the issue. It just removes the color.

terpstra:
The problem is that `indent` injects "\n" in a way that does not work with an enclosing `for` loop.
This has been broken since 26e237d590c624b1c2f07f80e083d87db7d3b26a.

terpstra:
I pushed a proper fix.

V-FEXrt:
I see, yeah there were two bug there because the color previously being output wasn't going through `TermInfoBuf` which means it wasn't getting replayed correctly either

V-FEXrt:
@terpstra as part of the org wide GitHub SSO you need to reauthorize CircleCI. All the old tokens were invalidated.

https://support.circleci.com/hc/en-us/articles/360043002793-Troubleshooting-CircleCI-Access-After-Enabling-Github-SSO

JakeSiFive (approved):
LGTM

PR #1158, opened 2023-03-13
Don't send response to unsupported $/ notifications

V-FEXrt:
Per the spec, certain notifications (a message send without and `id` field) may be ignored. This change ignores those messages when they are unrecognized instead of generating an error.

### DIFF
--- a/tools/lsp-wake/lsp.cpp
+++ b/tools/lsp-wake/lsp.cpp
@@ -64,6 +64,11 @@ static const char *ServerNotInitialized = "-32002";
 
 DiagnosticReporter *reporter;
 
+static inline bool message_is_notification(JAST request) {
+  const JAST &entry = request.get("id");
+  return entry.kind == JSON_NULLVAL;
+}
+
 class LSPServer {
  public:
   LSPServer() : astree() {}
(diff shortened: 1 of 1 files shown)


richardxia (approved):
Thanks for submitting this fix!

One thing I noticed while reading this code recently, though, is that we've already used the word "notification" to mean something different than the JSON-RPC concept of a notification. For example, the `MethodResult` struct has a `notification` field, which is unrelated to the notion of a JSON-RPC notification being a message without an `id`.

It probably doesn't block this PR, but we might want to rename all the existing usages of the word "notification" to not collide with the LSP and JSON-RPC spec. I haven't read our code deeply enough to 100% understand what we're using it for, but from a cursory glance I think it's something similar to a warning or an alert, so those are potential words we could use instead.

V-FEXrt:
Yeah, I'm actually just generally not happy with how we handle the various JSON models. I started mocking up a refactor but it quickly became very large due to how deeply coupled the current code is. I'm trying to see if I can justify putting the work in for that major refactor. If I can then this should be fixed as part of that (MethodResult would honestly just go away)

https://github.com/sifive/wake/tree/lsp-typed-json is the branch if you are interested

richardxia:
Yeah, I'm almost always in favor of using stronger types when possible and just keeping the untyped JSON at the boundaries.

V-FEXrt:
Just confirmed with Colin that this fixes the issue. Will circle around on the larger refactor after chatting with some folks. Landing this now

PR #1160, opened 2023-03-16
modify kernel cache message

mmjconolly:
Removes the `Kernel file cache may be cold.` string as it's speculation, they might just have a slow/loaded disk/computer.

Adds an extra `/r` on the remaining message, as I suspect we want all of them to have it so the message disappears at next write?

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -676,8 +676,8 @@ int main(int argc, char **argv) {
     auto now = std::chrono::steady_clock::now();
     if (!quiet && is_stdout_tty &&
         std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() > 1000) {
-      std::cout << "Scanning " << i + 1 << "/" << wakefilenames.size()
-                << " wake files. Kernel file cache may be cold.\r" << std::flush;
+      std::cout << "Scanning " << i + 1 << "/" << wakefilenames.size() << " wake files.\r"
+                << std::flush;
       start = now;
       alerted_slow_cache = true;
     }
@@ -713,7 +713,7 @@ int main(int argc, char **argv) {
 
   if (!quiet && alerted_slow_cache && is_stdout_tty) {
     std::cout << "Scanning " << wakefilenames.size() << "/" << wakefilenames.size()
-              << " wake files. Kernel file cache may be cold." << std::endl;
+              << " wake files.\r" << std::endl;
   }
 
   if (in) {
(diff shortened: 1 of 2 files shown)


JakeSiFive (approved):

tools/wake/main.cpp
```diff
@@ -713,7 +713,7 @@ int main(int argc, char **argv) {
 
   if (!quiet && alerted_slow_cache && is_stdout_tty) {
     std::cout << "Scanning " << wakefilenames.size() << "/" << wakefilenames.size()
-              << " wake files. Kernel file cache may be cold." << std::endl;
+              << " wake files.\r" << std::endl;
```
ag-eitilt, line 716: Ashley can definitely give a better answer, but I think this deliberately *didn't* have the `\r` because it was meant to stay on the screen as a record of "yeah, we took a long time to scan everything, but we did find them eventually!"  (Note the `wakefilenames.size() << "/" << wakefilenames.size()` having the total number on both sides.)

JakeSiFive: Now that you say it, I remember recommending this or backspaces and I remember him not wanting to do it for some reason.

mmjconolly: If I run `wake -x Unit` on a slow machine, I should be left with nothing but `Unit`, but I see
```
$ wake -x Unit                                                                                                                                                   
Finding wake files is taking longer than expected. Kernel file cache may be cold. (45753 explored).                                                                                            
Scanning 267/267 wake files. Kernel file cache may be cold.                                                                                                                                    
Unit
```

ag-eitilt: I'm not sure we've guaranteed that (unused symbol warnings will still be printed, for example, though that might be to stderr?), but I do like it in principle.

JakeSiFive: That only happens when you output to a terminal. If we used your solution without a terminal you'd still have bad bytes so I don't think this is an issue.

JakeSiFive: (plying devil's advocate of course, I'm happy with using \r unconditionally here).

mmjconolly: @V-FEXrt do you have any thoughts on these changes?

V-FEXrt: Hi sorry, was heads down in code today. Looking at the review now but my personal preference was to not erase. (but IIRC it was a preference and nothing else)

V-FEXrt: Okay, yep. I'm happy with clearing. But can you make this line just a newline then?

`std::cout << "Scanning " << wakefilenames.size() << "/" << wakefilenames.size() << " wake files.\r" << std::endl;`
              
`\r` following by `std::endl` is just the same as only emitting the `std::endl`


Also FWIW, we only print this when `is atty = true` so scripts _shouldn't _ see the output only humans

V-FEXrt (approved):

JakeSiFive:
Let's land this

PR #1161, opened 2023-03-17
Don't consume positional args for input/output flags

V-FEXrt:
Running `wake -o file1 file2 --verbose` didn't capture `--verbose` since `-o` consumed positional args leading `--verbose` to be treated as a file. Fix this by changing -i/-o use a required arg repeated for each instance. `wake -o file1 -o file2 --verbose`

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -406,7 +406,7 @@ int main(int argc, char **argv) {
 
   // Arguments are forbidden with these options
   bool noargs = init || job || last_use || last_exe || failed || tagdag || html || global ||
-                exports || api || exec;
+                exports || api || exec || label || input || output;
   bool targets = argc == 1 && !noargs;
 
   bool nodb = init;
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -105,8 +105,8 @@ void print_help(const char *argv0) {
     << "    --timeline       Print the timeline of wake jobs as HTML"                    << std::endl
     << "    --list-outputs   List all job outputs"                                       << std::endl
     << "    --clean          Delete all job outputs"                                     << std::endl
-    << "    --input  -i FILE Capture jobs which read FILES"                              << std::endl
-    << "    --output -o FILE Capture jobs which wrote FILES"                             << std::endl
+    << "    --input  -i FILE Capture jobs which read FILE. (repeat for multiple files)"  << std::endl
+    << "    --output -o FILE Capture jobs which wrote FILE. (repeat for multiple files)" << std::endl
     << "    --label     GLOB Capture jobs where label matches GLOB"                      << std::endl
     << "    --job       JOB  Captre the job with the specified job id"                   << std::endl
     << "    --last     -l    See --last-used"                                            << std::endl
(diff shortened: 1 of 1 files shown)


tools/wake/main.cpp
```diff
@@ -178,9 +182,9 @@ int main(int argc, char **argv) {
     {0, "in", GOPT_ARGUMENT_REQUIRED},
     {'x', "exec", GOPT_ARGUMENT_REQUIRED},
     {0, "job", GOPT_ARGUMENT_REQUIRED},
-    {'i', "input", GOPT_ARGUMENT_FORBIDDEN},
-    {'o', "output", GOPT_ARGUMENT_FORBIDDEN},
-    {0, "label", GOPT_ARGUMENT_FORBIDDEN},
+    {'i', "input", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, input_files.data(), max_pairs},
+    {'o', "output", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, output_files.data(), max_pairs},
+    {0, "label", GOPT_ARGUMENT_REQUIRED},
```
ag-eitilt, line 187: It may not be a straightforward implementation (and is probably a separate PR anyway), but do we want to enable repeating `--label` and `--job` to be able to run multiple queries at once?

mmjconolly (approved):
lgtm

ag-eitilt (approved):
So far as I can tell, without knowledge of gopt, this looks good!

tools/wake/main.cpp
V-FEXrt: I thought about that but it has some weird semantics. Maybe still work adding though.

Setting different flags should be intersection `--label --failed` -> show me failed jobs that match label while setting the _same_ flag has to be union `--job 1 --job 2` -> "show me jobs 1 and 2"

It gets weird when you mix them `--job 1 --job 2 --job 3 --failed` -> show me only the failed jobs in the set 1,2,3

JakeSiFive (approved):
Thanks for fixing this!

tools/wake/main.cpp
ag-eitilt: Ah, right, that does get a bit strange.  And multiple `--label`s could actually go both ways -- though, actually, that's the same question we have for `--input` (is it any job which reads *any* of these files, or any job which reads them *all*) so it seems like we're already setting a precedent for union within a single tag and intersection between different ones.

JakeSiFive: Yeah I think we decided that intersection is more useful than union for some of the fields but currently everything is union based...but also for --job and --input/--output union is clearly more useful lol

For instance there was a myth going around that I also fell for that `--failed --last` would intersect the two...but --failed already implied --last and doing both would just repeat the failed jobs...its not even really union.

V-FEXrt: > though, actually, that's the same question we have for --input 

Yeah, right now its union. We can explore changing that later but I think the policy in the short term will be repeated flags are union (even if only input/output can be repeated right now) and different flags are intersection (I swear I created a PR for this but idk where it went)

PR #1165, opened 2023-03-17
Intersection based db filters

V-FEXrt:
This a quick and dirty implementation. Probably not what we want to land but using it to start the conversation.

ATM It doesn't correctly handle the case where no jobs intersect (it falls back to a weird union since all counts are 1)


V-FEXrt:
https://github.com/sifive/wake/pull/1166 supersedes this

PR #1166, opened 2023-03-17
Intersect capture flags when multiple set

V-FEXrt:
When setting two different job capture flags (like `--last` and `--output`) intersect the results from each flag

(@ag-eitilt this was the PR I mentioned in that one comment)

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -73,6 +73,40 @@
 
 static CPPFile cppFile(__FILE__);
 
+namespace {
+
+template <class T>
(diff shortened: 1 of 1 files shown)


JakeSiFive (commented):
Sick ok so multiple --inputs union with each other, same for --output and in the future we might add --job that would union but then things intersect. So --input together with --output find the jobs that both read and output specific files. So you could for instance say list a series of output files but then specify one input file to see which of several jobs actully read a given input file. That seems pretty useful to me!

JakeSiFive (approved):
LGTM!

PR #1167, opened 2023-03-17
Even cleaner logging

JakeSiFive:
This change will add log line headers and generally make the status system a little bit better for wake internal development. It's still a work in progress and landing this feature depends on having the discussed .wakerc feature so I think this is going to sit here for a bit and I'll come back to it later when that's done. For now I'll go back to shared caching.


mmjconolly:
Please provide some before/after comparisons of logs

JakeSiFive:
> Please provide some before/after comparisons of logs

I will once we have a way to configure log headers. We don't have that right now thoguh so I'm not going to work on this until we do. I think @V-FEXrt is working on it however.

JakeSiFive:
Now that we have configs, I'll pick this back up once eviction is working. I'm about half way done getting up the individual PRs for eviction so we're fairly close.

JakeSiFive:
Here are some examples of what the new logging will look like. The log header will always be colored according to the default stream color for that stream. The log header itself Can be whatever you want, even the empty string if you so choose. There are currently 3 supported variables called "$stream" ,"$time" and "$extra". The "$extra" variable is just some extra bit of information to explain where the content is coming from. For jobs this is always the label. For other things this will either be empty or something else, I haven't decided just yet. This is just an MVP. The variables are subject to change.

A config looks like this (which is also the current default)
```
{
    log_header: "[$stream] $extra: ",
    log_header_extra_width: 25
}
```

Here are some example screen shots showing different configs, note the log header and that the output from the program is displayed in full color.

Screen Shot 2023-04-21 at 12 11 13 PM
Screen Shot 2023-04-21 at 12 11 31 PM
Screen Shot 2023-04-21 at 12 11 48 PM

JakeSiFive:
Here's a sample of what things look like now without any log header

ag-eitilt:
I like it, though I wonder if `$source` might be a bit more clear than `$extra` (I do definitely get why you aren't using `$label`).  Are you looking at any options to give these their own lines?  Might be helpful for particularly long labels/info or thin terminals, though I love your `...` inclusion for that in the standard case.
```
[warning][c++ src/util/config.cpp]
src/runtime/status.cpp: In member function 'void StatusBuf::emit_header()':
src/runtime/status.cpp:213:23: error: expected ';' before 'continue'
        out << tmp_extra
                        ^
                        ;
        continue;
        ~~~~~~~~
```
Does `log_header_extra_width` *only* affect the `$extra` field?  The stream names are different lengths as well, and I personally would be looking for some format which starts all the output in the same column.

V-FEXrt:
Can you update README.md to document the new config options?

.wakeroot
```diff
@@ -1,2 +1,4 @@
 {
+    "log_header": "",
+    "log_header_extra_width": 0
```
V-FEXrt, line 3: any specific reason we aren't using the default here?

share/wake/lib/core/print.wake
```diff
@@ -38,75 +38,35 @@ from wake export type LogLevel
...
-            Yellow  = 3
-            Blue    = 4
-            Magenta = 5
-            Cyan    = 6
-            White   = 7
-    def intensityCode = match intensity
-        None   = 0
-        Some i = 16 * match i
-            Dim    = 1
-            Bright = 2
```
V-FEXrt, line 77: lol don't have to worry about formatting these anymore (do you need to rebase?)

src/runtime/database.cpp
```diff
@@ -1111,18 +1111,18 @@ void Database::replay_output(long job, const char *stdout, const char *stderr) {
...
-      status_write(fd == 2 ? stderr : stdout, str, len);
-      needlf[fd - 1] = str[len - 1] != '\n';
-    }
+    if (len) status_get_generic_stream(fd == 2 ? stderr : stdout) << std::string(str, len);
   }
   finish_stmt(why, imp->replay_log, imp->debugdb);
-  if (needlf[0]) status_write(stdout, "\n", 1);
-  if (needlf[1]) status_write(stderr, "\n", 1);
+  // TODO: Add job_id to these
+  if (stdout_acc.size()) status_get_generic_stream(stdout) << std::endl;
```
V-FEXrt, line 1124: You never actually write to `stdout_acc`/`stderr_acc` right?

.wakeroot
JakeSiFive: Because then it would clutter the stdout with info

JakeSiFive:
> Does log_header_extra_width only affect the $extra field? The stream names are different lengths as well, and I personally would be looking for some format which starts all the output in the same column.

Yes it only effects that. Stream names are padded to the length of "warning" right now so that they remain aligned.

share/wake/lib/core/print.wake
JakeSiFive: Yeah lol true haha. I probably need to rebase. You've been writing code like a mad man.

src/runtime/job.cpp
```diff
@@ -744,8 +755,57 @@ static void launch(JobTable *jobtable) {
...
+      }
+      jobtable->imp->term_bufs[fd_err] =
+          std::make_unique<TermInfoBuf>(fd_buf.get(), !isatty(fd_err));
+      jobtable->imp->fd_bufs[fd_err] = std::move(fd_buf);
+    }
+
+    // Now we need to figure out if we're outputting to a null buffer to avoid rendering
+    // or if we're going to output to a nicely rendered StatusBuf.
+    auto job_id_str = wcl::some(task.job->label->as_str());
+    wcl::optional<std::string>{wcl::in_place_t{}, "job_id = " + std::to_string(task.job->job)};
```
V-FEXrt, line 793: nit: `job_id = 123` still feels a  bit too verbose. maybe `job(123)`?

ag-eitilt:
> > Does log_header_extra_width only affect the $extra field? The stream names are different lengths as well, and I personally would be looking for some format which starts all the output in the same column.
> 
> Yes it only effects that. Stream names are padded to the length of "warning" right now so that they remain aligned.

Is that hardcoded, or is it the length of the longest stream?  Because if I wanted to be particularly tightly-filtered I could run a `makeLogLevel "only_show_this" None` followed by a `--stdout=only_show_this`.

src/runtime/job.cpp
```diff
@@ -941,21 +979,10 @@ bool JobTable::wait(Runtime &runtime) {
           runtime.heap.guarantee(WJob::reserve());
           runtime.schedule(WJob::claim(runtime.heap, entry->job.get()));
           ++done;
-          if (!imp->batch && !entry->stderr_buf.empty()) {
-            if (entry->stderr_buf.back() != '\n') entry->stderr_buf.push_back('\n');
-            status_write(entry->job->stream_err.c_str(), entry->stderr_buf.data(),
-                         entry->stderr_buf.size());
-            entry->stderr_buf.clear();
-          }
```
V-FEXrt, line 949: I'm not sure what this is doing, and it looks like it doesn't have a translation? Are we dropping/loosing a feature here?

src/runtime/status.cpp
```diff
@@ -82,6 +85,163 @@ static void status_clear() {
...
+  // if the last thing was a variable, this might be the empty string
+  // and that's ok.
+  out.emplace_back(std::move(cur));
+  return out;
+}
+
+void StatusBuf::emit_header() {
+  auto fmt_vec = split_by_var(WakeConfig::get()->log_header);
+
+  // Push the current state sot hat we can overwrite
```
V-FEXrt, line 152: nit: typo `sot hat` -> `so that`

JakeSiFive:
> Are you looking at any options to give these their own lines? Might be helpful for particularly long labels/info or thin terminals, though I love your ... inclusion for that in the standard case.

Nope, not looking into that at all right now. That would require sending everything (both stdout and stderr) through a single choke point and then multiplexing back out to stdout/stderr. This was already a quite complex change to our very broken status system, that would be a whole extra thing.

src/runtime/job.cpp
JakeSiFive: Nope :) I just cleaned it up a fuck load

This was line buffering. Line buffering is now handled by the StatusBuf directly

JakeSiFive:
> Is that hardcoded, or is it the length of the longest stream? Because if I wanted to be particularly tightly-filtered I could run a makeLogLevel "only_show_this" None followed by a --stdout=only_show_this

Currently hard coded to the length of "warning". It is no longer possible to make new log levels with this change.

src/runtime/job.cpp
JakeSiFive: It's actually a typo that that was left in. We're using the label now instead.

src/util/config.cpp
```diff
@@ -201,21 +199,25 @@ static std::vector<std::string> find_disallowed_keys(const JAST& json,
   return disallowed;
 }
 
-bool init(const std::string& wakeroot_path) {
+bool WakeConfig::init(const std::string& wakeroot_path) {
   if (_config != nullptr) {
     std::cerr << "Cannot initialize config twice" << std::endl;
-    assert(false);
+    exit(1);
```
V-FEXrt, line 205: nit: I prefer assert here since this is a "code failure" not a "user failure". It'll point to the exact line instead of needed to grep for "Cannot initialize config twice". And if we ever see this error then its our fail so it should blow up

src/util/config.cpp
```diff
@@ -230,43 +232,54 @@ bool init(const std::string& wakeroot_path) {
...
+      std::cerr << wakeroot_path << ": ";
+      std::cerr << "Key '" << key << "' may not be set in .wakeroot";
       if (user_config_allowed_keys.count(key) > 0) {
-        s << " but it may be set in user config";
+        std::cerr << " but it may be set in user config";
       }
-      s << ".";
-      reporter->reportWarning(Location(wakeroot_path.c_str()), s.str());
+      std::cerr << "." << std::endl;
+      ;
```
V-FEXrt, line 241: using `reportWarning` was an explicit choice so that our warnings were more consistent (asked about this in a DM and that's what we landed on). Don't mind that its changing, but we should consider that decision/ensure the warning is consistent with others

tests/.wakeroot
```diff
@@ -1 +1 @@
-{}
+{"log_header":"", "log_header_extra_width":0}
```
V-FEXrt, line 1: Ah, if these aren't optional then we need `wake --init` or similar to generate a default `.wakeroot`. We don't want to increase the barrier to entry for a new project

tests/config/nominal/.wakeroot
```diff
@@ -1,3 +1,5 @@
 {
-  "user_config": "./wake_user.json"
+  "user_config": "./wake_user.json",
+  "log_header":"",
+  "log_header_extra_width":0
```
V-FEXrt, line 4: nominal test should set custom values and ensure they are pulled through

V-FEXrt (changes requested):

.wakeroot
V-FEXrt: 🤷‍♂️ fair, I disagree but I suppose that is why it's configurable

src/runtime/job.cpp
```diff
@@ -899,6 +946,7 @@ bool JobTable::wait(Runtime &runtime) {
       std::shared_ptr<JobEntry> entry = it->second;
       assert(entry);
 
+      std::string job_id_str = "job_id = " + std::to_string(entry->job->job);
```
V-FEXrt, line 949: Is this one also a typo?

tests/.wakeroot
ag-eitilt: I'd much prefer having these be optional for backwards compatibility.  Besides, these feel like something which is more of a user preference than a repo preference, so forcing every repo to include them anyway seems off.

ag-eitilt:
> It is no longer possible to make new log levels with this change.

I don't like this.  We've suggested custom log levels as a means of pulling out specific job outputs for piping into external scripts (`--fd:3=custom`).  This breaks a few use cases and reduces power/repo-workflow-customizability in general, for what I'm sure were ease-of-implementation reasons, but even if that were the case during development that feature should be restored/rearchitected before this merges.

src/util/config.cpp
JakeSiFive: It doesn't always fail though because a lot of times asserts don't trigger.

src/util/config.cpp
JakeSiFive: Right but I can't do that inside of the runtime because I don't have a reporter defined.

tests/.wakeroot
JakeSiFive: They are optional, but the defaults break the tests.

tests/config/nominal/.wakeroot
JakeSiFive: I guess I'm not sure what you mean by that?

src/runtime/job.cpp
JakeSiFive: yep!

JakeSiFive:
> > It is no longer possible to make new log levels with this change.
> 
> I don't like this. We've suggested custom log levels as a means of pulling out specific job outputs for piping into external scripts (`--fd:3=custom`). This breaks a few use cases and reduces power/repo-workflow-customizability in general, for what I'm sure were ease-of-implementation reasons, but even if that were the case during development that feature should be restored/rearchitected before this merges.

I'm happy to replace use cases but the status system is currently very broken. This is a change towards fixing it any making it more flexible. The logging changes coming will make repo-workflow-customizability strictly better in the future.

tests/.wakeroot
ag-eitilt: Ah, got it.  Overrides for tests is fine!

src/util/config.cpp
V-FEXrt: strange. I use this pattern frequently. This probably deserves a grep across the repo then. (Also what good are asserts if we can't  trust them?)

tests/config/nominal/.wakeroot
V-FEXrt: I just want `log_header` and `log_header_extra_width` to have some value a random repo would set. Right now all of our tests for config only test the "empty" values or the default values

src/util/config.cpp
JakeSiFive: They do trigger if you set the DEBUG define is set correctly but you have to make sure its set correctly in every translation unit (and it wasn't set in this one)

ag-eitilt:
> > I don't like this. We've suggested custom log levels as a means of pulling out specific job outputs for piping into external scripts (`--fd:3=custom`). This breaks a few use cases and reduces power/repo-workflow-customizability in general, for what I'm sure were ease-of-implementation reasons, but even if that were the case during development that feature should be restored/rearchitected before this merges.
> 
> I'm happy to replace use cases but the status system is currently very broken. This is a change towards fixing it any making it more flexible. The logging changes coming will make repo-workflow-customizability strictly better in the future.

It's very definitely broken, and I'm definitely not against seeing it all fixed.  If you can offer an interface where Wake can be told to print only the output from a specific job (not its dependencies) or `println`, without other potential clutter from system info, and without having to run once and then query the database, I'll be happy.  I've never particularly liked the `1>/dev/null 3&>1` contortions or whatever syntax it is that the current interface requires.

JakeSiFive:
> It's very definitely broken, and I'm definitely not against seeing it all fixed. If you can offer an interface where Wake can be told to print only the output from a specific job (not its dependencies) or println, without other potential clutter from system info, and without having to run once and then query the database, I'll be happy. I've never particularly liked the 1>/dev/null 3&>1 contortions or whatever syntax it is that the current interface requires.

The plan is to offer a `--label` flag that would filter all output that didn't match a label. Would that suffice?

ag-eitilt:
> The plan is to offer a `--label` flag that would filter all output that didn't match a label glob. Would that suffice?

Yeah, that seems like it could be a good bit better than the current interface in nearly every way, so long as we're sure we also aren't adding the final return value echo to the stdout.  Any chance of getting that added before whatever release includes this code, so we don't have a version where we can't pipe around the target output, even if it's not part of this specific PR?

JakeSiFive:
I can just add it to this PR I think, it shouldn't be too bad. That way we don't lose functionality even between releases.

tests/.wakeroot
```diff
@@ -1 +1 @@
-{}
+{"log_header":"", "log_header_source_width":0}
```
V-FEXrt, line 1: I can't remember what we said  before but if these can just be `{}` I'd prefer that. If they can't then feel free to leave it

tests/config/nominal/.wakeroot
```diff
@@ -1,3 +1,5 @@
 {
-  "user_config": "./wake_user.json"
+  "user_config": "./wake_user.json",
+  "log_header":"",
```
V-FEXrt, line 3: I still want the `nominal` case to set these values to something arbitrary.

ie 
```json
{
  "user_config": "./wake_user.json",
  "log_header": "$label",
  "log_header_source_width": 13
}
```

V-FEXrt (approved):

src/runtime/job.cpp
```diff
@@ -467,11 +481,9 @@ JobTable::JobTable(Database *db, ResourceBudget memory, ResourceBudget cpu, bool
   // Double-check that ::parse() did not do something crazy.
   assert(imp->limit > 0);
 
-  std::stringstream s;
-  s << "wake: targeting utilization for " << imp->limit << " threads and "
-    << ResourceBudget::format(imp->phys_limit) << " of memory." << std::endl;
-  std::string out = s.str();
-  status_write("echo", out.data(), out.size());
+  status_get_generic_stream("echo")
```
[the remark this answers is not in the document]
JakeSiFive: echo is output no matter what right now. We'll need extra knobs to adjust that separately in the future. For the use case that Sam is currently requesting I replace that feature is not needed. In future work we can add more knobs to control things like that.

tests/.wakeroot
JakeSiFive: It adds log headers if I don't add those things and consequently I'd have to go change all of the stdout/stderr parts if I didn't add this.

PR #1173, opened 2023-03-23
Add note on Wake arguments vs. List String targets

ag-eitilt:
The way we pass arguments to `List String` CLI targets (no `-x` when running) is mostly intuitive if you've got a "subcommand" you're already expecting to pass arguments to, but even one-off targets which don't do anything with their arguments still gobble anything following them, rather than the potentially-expected behaviour of flags being passed to `wake` itself.  I've added a paragraph in the tutorial to clarify that.


JakeSiFive (approved):
Good note! Thank you!

PR #1178, opened 2023-03-29
Add directory_range to wcl

JakeSiFive:
This change adds a wrapper around posix readdir in order to make iterating directories simpler, more C++-ish, and improve error handling for it.

### DIFF
--- a/src/wcl/filepath.h
+++ b/src/wcl/filepath.h
@@ -17,11 +17,100 @@
 
 #pragma once
 
+#include <dirent.h>
+#include <sys/types.h>
+
 #include <sstream>
 #include <string>
 #include <vector>
 
+#include "result.h"
+
 namespace wcl {
+
+enum class file_type { block, character, directory, fifo, symlink, regular, socket, unknown };
+
+struct directory_entry {
+  std::string entry_name;
+  file_type type;
+};
+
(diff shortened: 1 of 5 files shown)


src/wcl/filepath.h
```diff
@@ -17,12 +17,101 @@
...
+  }
+  directory_range(const directory_range&) = delete;
+  directory_range(directory_range&& dir) : dir(dir.dir), dir_path(std::move(dir_path)) {
+    dir.dir = nullptr;
+  }
+
+  static result<directory_range, posix_error_t> open(const std::string& path) {
+    directory_range out{path};
+    if (out.dir == nullptr) {
+      return make_error<directory_range, posix_error_t>(errno);
```
V-FEXrt, line 104: nit: doesn't result have a helper function for this exact case?

V-FEXrt (approved):
If possible can you write a test for this?

The code looks fine but hard to actually verify it in isolation

src/wcl/filepath.h
JakeSiFive: oh yeah it does, but I wrote these out of order and then noticed the pattern later. I'll go back and fix this.

V-FEXrt (approved):

PR #1179, opened 2023-03-29
html: Refactor and modernize HTML/CSS/JS

richardxia:
This makes a number of non-functional changes to the `wake --html` output to modernize the code and make it easier to work with (IMO, at least). See the individual commits for more information about each change, but at a high level, this:

- explicitly declares the document to be HTML5, since it's no longer 2010
- relies less on DOM APIs that require parsing strings and instead uses the APIs that let you directly modify elements in a structured way. e.g. `Element.cssText` -> `Element.style`, `Element.setAttribute()` -> `Element.<specific attribute name>`.
- moves almost all styling to CSS, and JS is only used to add or remove CSS class names on elements
- uses the official [`data-`/`dataset` API](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) for setting custom attributes and not making up our own root-namespace attributes
- use [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) (a.k.a. variables), a somewhat recent CSS feature. This will make it a bit easier to support both a light and dark theme, which I will do in a followup PR that I am about to submit.

This should have no functional (nor aesthetic) impact on the generated HTML files.

### DIFF
--- a/share/wake/html/main.js
+++ b/share/wake/html/main.js
@@ -19,11 +19,7 @@
 // tooltip
 const tooltip = document.createElement('div');
 tooltip.classList.add('tooltip');
-tooltip.style.visibility = 'hidden';
-
-// css to select uses
-const usecss = document.createElement('div');
-usecss.appendChild(document.createElement('style'));
+tooltip.classList.add('hidden');
 
 // The following variables encompass the state for expression selection mode,
 // where pressing + or - on the keyboard while the mouse is hovering over an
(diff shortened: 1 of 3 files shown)


V-FEXrt (approved):
Awesome! These  are  all great improvements. Thanks!

richardxia:
Thanks for the quick review!

PR #1180, opened 2023-03-29
html: Add new dark and light themes.

richardxia:
This builds off https://github.com/sifive/wake/pull/1179 by adding new dark and light color themes for the generated HTML pages. I added a theme toggle to the top right of the page, but I otherwise default to your OS's settings (and failing that, it defaults to the light theme).

I originally attempted to colors from popular open source themes like Solarized and Dracula, but none of them worked well for the several colored background highlights that are used in the wake generated HTML. I ended up making up my own theme using purple as the primary tint and basing the remainder of the colors off the original hues that wake used. I lightened the background color and darkened the text colors, since the original wake styles used pure black and pure white, which causes excessive contrast and can actually reduce readability. I also adjusted the foreground colors so that they have appropriate contrast with the background colors, e.g. not so dark that they are insufficiently contrasted with the background, and not so light that they are a painful to read due to excessive contrast.

I'm happy to tweak the colors if anyone has any preferences or requests.

## Dark theme
Screen Shot 2023-03-29 at 1 40 28 PM

## Light theme
Screen Shot 2023-03-29 at 1 40 42 PM

### DIFF
--- a/share/wake/html/main.js
+++ b/share/wake/html/main.js
@@ -197,11 +197,36 @@ function workspace(node) {
   return res;
 }
 
+function createThemeToggle() {
+  const prefersDark = matchMedia('prefers-color-scheme: dark');
+  document.documentElement.className = prefersDark ? 'dark' : 'light';
+
+  const themeToggle = document.createElement('div');
+  themeToggle.innerHTML = `
+    <label>Light<input type="radio" name="style" value="light"${prefersDark ? '' : ' checked'}>
+    <label>Dark<input type="radio" name="style" value="dark"${prefersDark ? ' checked' : ''}>
+  `;
+
+  const onChange = (e) => {
+    e.preventDefault();
+    document.documentElement.className = e.target.value;
+  }
+  themeToggle.querySelectorAll('input')
+    .forEach(e => e.addEventListener('change', onChange));
+
+  themeToggle.style.position = 'fixed';
+  themeToggle.style.top = '10px';
+  themeToggle.style.right = '10px';
+
+  return themeToggle;
+}
+
 document.addEventListener('DOMContentLoaded', function main () {
   document.body.appendChild(tooltip);
   const wakeData = document.getElementById('wake-data');
   const node = JSON.parse(wakeData.text);
   document.body.appendChild(workspace(node));
+  document.body.appendChild(createThemeToggle());
 });
 
 /* eslint-env browser */
--- a/share/wake/html/main.js
+++ b/share/wake/html/main.js
@@ -198,21 +198,32 @@ function workspace(node) {
 }
 
 function createThemeToggle() {
-  const prefersDark = matchMedia('prefers-color-scheme: dark');
-  document.documentElement.className = prefersDark ? 'dark' : 'light';
+  // If the user previously chose a theme, default to that theme.
+  // Otherwise, check the browser/OS preferences for dark vs. light theme.
+  let colorTheme = window.localStorage?.getItem('wakeColorTheme');
+  colorTheme ||= matchMedia('prefers-color-scheme: dark') ? 'dark' : 'light';
+
+  document.documentElement.className = colorTheme;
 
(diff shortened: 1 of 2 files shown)


share/wake/html/main.js
```diff
@@ -197,11 +197,36 @@ function workspace(node) {
   return res;
 }
 
+function createThemeToggle() {
+  const prefersDark = matchMedia("prefers-color-scheme: dark");
```
V-FEXrt, line 201: Can you add support for localStorage that way the site remembers when someone edits the toggle? It should only be a few lines of code. https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#examples

My personal preference is dark theme but (I think) I haven't bothered to setup my OS preference so it would keep defaulting to light theme

V-FEXrt (approved):

share/wake/html/main.js
richardxia: Yeah, I was debating whether or not I should use local storage for persisting the choice, so I can go add that now.

One thing to note is that if you're opening the file locally, with a `file://` protocol, then there will probably be issues with using local storage for security reasons. local storage should work fine for anything that's actually hosted off an actual HTTP server, though. I can test out a few scenarios locally to see what happens and report back on the behavior.

richardxia: Interestingly, it looks like different browsers do different things. In both Firefox and Safari, for `file://` protocols, each individual file appears to have its own, private localStorage. Chrome, however, appears to share the same localStorage across multiple files.

For anything served off HTTP(S), all pages under the same domain and port number share the same localStorage.

Anyway, I pushed up a change that now saves the user's preference to localStorage: https://github.com/sifive/wake/pull/1180/commits/2a68be780d0012c4b46d76375eb8660d5161d31a

V-FEXrt (approved):

share/wake/html/main.js
```diff
@@ -197,11 +197,49 @@ function workspace(node) {
...
+  const themeToggle = document.createElement('div');
+  themeToggle.innerHTML = `
+    <label>Light<input type="radio" name="style" value="light">
+    <label>Dark<input type="radio" name="style" value="dark">
+  `;
+
+  const onChange = (e) => {
+    e.preventDefault();
+    document.documentElement.className = e.target.value;
+    if (localStorage) {
```
V-FEXrt, line 217: nit: does nil coalescing not work here?

`localStorage?.setItem('wakeColorTheme', e.target.value);`

richardxia: Oh right, yeah, that should work. I wrote the `if` statement when I thought I was going to write something like `localStorage.wakeColorTheme = e.target.value`, and I wasn't sure if that operator worked on the LHS, but then I forgot to change it when I realized that the API was `setItem()` and not a property access.

I just tested this with local storage disabled temporarily on my browser, and it seems to work.

I also discovered a few more things:

- I should probably prefix `localStorage` with `window?.`, since apparently attempting to access an undefined variable will throw a ReferenceError, not return `undefined`. Attempting to access an undefined attribute on an object will return `undefined`, though. This doesn't matter for browsers that support `localStorage` but allow you to disable it, since `localStorage` will be defined and `null` in that case, but I'll do the access off the `window` object just in case.
- I did confirm that `foo?.bar = baz` is illegal.

I'll push up these changes in a sec.

richardxia: Slight edit: I don't need to do `window?.localStorage`, since `window` is always defined in a browser environment, so `window.localStorage?.setItem()` is sufficient.

Side rant: I am so not used to writing non-TypeScript JavaScript and having my editor just tell me when things are undefined/null.

JakeSiFive:
oh man this is loads better looking!

JakeSiFive (approved):
Not approving the code as much as the improved theme

richardxia:
Thanks for all the careful review! I'm excited to have this in for the next release of wake!

Issue #1181, opened 2023-03-29
shared cache corrupts files in workspace in some situations


JakeSiFive:
Say you run a job and it outputs a file of 14 bytes and it gets added to the shared cache. Next you tweak that job and run it again and it outputs a file to the same place in the workspace but its 16 bytes this time. So far all good...but now lets say you revert your changes and you so now you get a cache hit on that original 14 byte entry. Currently wake's shared cache will sometimes overwrite the first 14 bytes but leave the remaining 2 bytes. For a lot of file formats this doesn't matter but for many it *does* matter and must be fixed.

JakeSiFive:
Should be solved by this: https://github.com/sifive/wake/pull/1183/files

PR #1183, opened 2023-03-30
rename files into place rather than copying into them

JakeSiFive:
We saw a case where the current shared cache can corrupt the files in the workspace because it wasn't being hygenic about creating new inodes like it was supposed to. I *thought* it was working this way already but it wasn't. It's quite simple to fix and we use this trick all over the place to solve this problem. I still need to add a test that I can show previously exhibited corruption so that we don't introduce this issue again but it should be fairly easy to do that. 

This should solve https://github.com/sifive/wake/issues/1181


src/job_cache/job_cache.cpp
```diff
@@ -1140,7 +1140,9 @@ wcl::optional<MatchingJob> Cache::read(const FindJobRequest &find_request) {
 
       // Finally copy the file (as efficently as we can) to
       // the destination.
-      copy_or_reflink(tmp_file.c_str(), pair.first.c_str(), mode);
+      std::string tmp_dst = pair.first + "." + rng.unique_name();
+      copy_or_reflink(tmp_file.c_str(), tmp_dst.c_str(), mode);
+      rename_no_fail(tmp_dst.c_str(), pair.first.c_str());
```
V-FEXrt, line 1145: nit: we should write a single function for this since its just a common pattern `rename_over_no_fail()` or something

V-FEXrt (approved):

src/job_cache/job_cache.cpp
JakeSiFive: The problem is that the thing you want to rename changes, and you need access to a fast source of high entropy rng to do it so the interface is kinda strange.

V-FEXrt: Yeah, its unclear. The dep on rng makes it a mucky api. Its at least worth looking at since we expect this to continuously bite us.

I'm willing to take the API tradeoff _if_ we can avoid an entire class of bugs

mmjconolly:
Is the issue that the destination already exists? and that we don't want to do a delete as that could be a race with the new copy/create/reflink-copy?

PR #1186, opened 2023-03-30
move eviction code into library and out of tool

JakeSiFive:


V-FEXrt (approved):
All strict improvements!

I guess you were right about not being a separate tool :)

PR #1187, opened 2023-03-30
cli: Move command line parsing to dedicated class

V-FEXrt:
This PR extracts the command line variables that were polluting the main() function into a separate struct (and does nothing else)

I've had several false starts here and I think I've pinned down the core issues from before

1) I was trying to do too much refactoring all at once
2) I didn't fully understand what each argument expected

I took the time to learn what the args were expecting and I made this PR as small as possible so I think I'm happy with the result here.

There are still several changes I want to make like moving the cli validation code, adding tests, extracting more out of main, and reworking the actual cli interface but this change is the minimal precursor to all of that.

I tried out several different flags (including the more hairy ones) but please feel free to try any yourself as well

### DIFF
--- a/tools/wake/cli_options.h
+++ b/tools/wake/cli_options.h
@@ -84,9 +84,8 @@ struct CommandLineOptions {
 
   CommandLineOptions(int argc_in, char **argv_in) {
     argv = argv_in;
-    unsigned int max_pairs = argc_in / 2;
-    input_files.reserve(max_pairs);
-    output_files.reserve(max_pairs);
+    input_files = std::vector<char *>(argc_in, nullptr);
+    output_files = std::vector<char *>(argc_in, nullptr);
 
     // clang-format off
     struct option options[] {
@@ -108,8 +107,8 @@ struct CommandLineOptions {
       {0, "in", GOPT_ARGUMENT_REQUIRED},
       {'x', "exec", GOPT_ARGUMENT_REQUIRED},
       {0, "job", GOPT_ARGUMENT_REQUIRED},
-      {'i', "input", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, input_files.data(), max_pairs},
-      {'o', "output", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, output_files.data(), max_pairs},
+      {'i', "input", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, input_files.data(), (unsigned int)argc_in},
+      {'o', "output", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, output_files.data(), (unsigned int)argc_in},
       {0, "label", GOPT_ARGUMENT_REQUIRED},
       {'l', "last", GOPT_ARGUMENT_FORBIDDEN},
       {0, "last-used", GOPT_ARGUMENT_FORBIDDEN},
--- a/tools/wake/cli_options.h
+++ b/tools/wake/cli_options.h
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2022 SiFive, Inc.
+ *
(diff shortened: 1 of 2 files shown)


tools/wake/cli_options.h
```diff
@@ -0,0 +1,206 @@
...
+  const char *label;
+  char *shebang;
+  const char *tagdag;
+  const char *tag;
+  const char *api;
+  const char *fd1;
+  const char *fd2;
+  const char *fd3;
+  const char *fd4;
+  const char *fd5;
```
JakeSiFive, line 77: Can we make these std::strings? Also why is `shebang` mutable?

tools/wake/cli_options.h
```diff
@@ -0,0 +1,206 @@
...
+
+  std::vector<char *> input_files;
+  std::vector<char *> output_files;
+
+  int argc;
+  char **argv;
+
+  CommandLineOptions(int argc_in, char **argv_in) {
+    argv = argv_in;
+    unsigned int max_pairs = argc_in / 2;
```
JakeSiFive, line 87: This works for a kind of subtle reason. Because argc_in is actully 1 higher than the logical length of argv_in because the 0th argument is the tool and not the arguments you don't have to round up. In general though, calling `reserve` is not a good idea. It's generally a que to me taht someone is about to do something very wrong with a vector or someone is over-optimizing something that they shouldn't be. I would recommend removing these 3 lines.

tools/wake/cli_options.h
```diff
@@ -0,0 +1,206 @@
...
+      {0, "fatal-warnings", GOPT_ARGUMENT_FORBIDDEN},
+      {0, "heap-factor", GOPT_ARGUMENT_REQUIRED | GOPT_ARGUMENT_NO_HYPHEN},
+      {0, "profile-heap", GOPT_ARGUMENT_FORBIDDEN | GOPT_REPEATABLE},
+      {0, "profile", GOPT_ARGUMENT_REQUIRED},
+      {'C', "chdir", GOPT_ARGUMENT_REQUIRED},
+      {0, "in", GOPT_ARGUMENT_REQUIRED},
+      {'x', "exec", GOPT_ARGUMENT_REQUIRED},
+      {0, "job", GOPT_ARGUMENT_REQUIRED},
+      {'i', "input", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, input_files.data(), max_pairs},
+      {'o', "output", GOPT_ARGUMENT_REQUIRED | GOPT_REPEATABLE_VALUE, output_files.data(), max_pairs},
```
JakeSiFive, line 112: Ah yes...the "things you should not be doing with a vector" that I was expecting. If you call ".size()" on this vector it will return zero and if you use ".at()" on it it will throw an out of bounds exception

JakeSiFive: Here's how I'd fix this

1) construct an std::vector of using the constructor without reserving it, setting nullptr as the default value. Just use argc_in as the size and don't bother dividing by 2.
2) give it the data like you do here
3) copy that into an std::vector<std::string> via emplace_back but break when you hit a nullptr

tools/wake/cli_options.h
V-FEXrt: > Can we make these std::strings?
yes, but can we do that as a different PR?

> Also why is shebang mutable?
Just directly copying over from the previous implementation

JakeSiFive: sounds good to me.

tools/wake/cli_options.h
V-FEXrt: 1 and 2 are done, 3 wasn't needed as `clo.input` and `clo.output` store the actual number of args found which is later looped over

JakeSiFive: That's still super gross. The size of those vectors should be the number of arguments they contain. Let's clean that up now rather than later. It's trivial to add and is clearly the right way to do this.

PR #1188, opened 2023-03-30
Fix VS Code extension not loading

ErikParawell-SiFive:
Error received when using extension versions v0.31.0 and v0.30.0:
```
Request initialize failed with message: not compiled for this environment (did you build to HTML and try to run it not on the web, or set ENVIRONMENT to something - like node - and run it someplace else - like on the web?)
```
This PR attempts to fix this issue.


ErikParawell-SiFive:
@V-FEXrt published a new PR reverting the emsdk version

PR #1190, opened 2023-03-31
Implement an LRU eviction policy for wake's shared cache

JakeSiFive:
What it says on the tin, this implements an LRU eviction policy for wake's shared cache. There are still some TODOs however

1) This is complicated and needs at least a bit of testing as a smoke test if nothing else. Before being confident putting this live I think we need more extensive fuzz testing but at the very least I'd like to see a basic test that exercises the LRU policy.
2) Right now I have the high and low water marks set at 16GB and 8GB, that needs to be configurable somehow and I haven't yet decided how to make that configurable. It would be *super* nice to implement this for testing to solve 1)

Other than those two points, its ready for eyeballs on it.

### DIFF
--- a/src/job_cache/db_helpers.h
+++ b/src/job_cache/db_helpers.h
@@ -1,0 +1,247 @@
+/*
+ * Copyright 2023 SiFive, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You should have received a copy of LICENSE.Apache2 along with
+ * this software. If not, you may obtain a copy at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+// NOTE: PLEASE DO NOT INCLUDE THIS OUTSIDE OF .cpp FILE IN THIS DIR!!!
+//       IMPORTANTLY DO NOT INCLUDE THEM IN HEADERS IN THIS DIR!!!
+//       This file exposes sqlite3 which right now is abstracted away.
+#include <sqlite3.h>
+
+#include <random>
+#include <string>
+
+#include "job_cache_impl_common.h"
+#include "logging.h"
+#include "wcl/filepath.h"
+
+namespace job_cache {
+
+class Database {
+ private:
+  sqlite3 *db = nullptr;
+
+  // This is fairly critical to getting strong concurency.
+  // Specifically adding in exponetial back off plus randomization.
+  static inline int wait_handle(void *, int retries) {
+    // We don't ever want to wait more than ~4 seconds.
+    // If we wait more than ~4 seconds we fail.
+    constexpr int start_pow_2 = 6;
+    constexpr int end_pow_2 = 22;
+    if (retries > end_pow_2 - start_pow_2) return 0;
+
+    useconds_t base_wait = 1 << start_pow_2;
+    std::random_device rd;
+    // Wait exponetially longer the more times
+    // we've had to retry.
+    useconds_t wait = base_wait << retries;
+    // Randomize so we don't all retry at the same time.
+    wait += rd() & (wait - 1);
+    // Finally sleep, without waking up exactly when we're done.
+    // but we should never sleep longer than twice as long as we
+    // had to this way.
+    usleep(wait);
+
+    // Tell sqlite to retry
+    return 1;
+  }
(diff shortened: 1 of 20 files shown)


src/job_cache/db_helpers.h
```diff
@@ -0,0 +1,247 @@
+/*
```
JakeSiFive, line 1: These were all just moved out of job_cache.cpp so that they could be used across the job_cache folder.

src/job_cache/job_cache_impl_common.cpp
```diff
@@ -0,0 +1,251 @@
+/*
```
JakeSiFive, line 1: everything but the remove_backing_files functions is old code that was just moved here so that it would have a common location across job_cache

JakeSiFive (commented):
Leaving some notes to make review easier. Github says this is a 700 LOC change but it isn't *quite*. It's more 400-500 LOC change but you have to know which parts to ignore...still bigger than I'd like...sorry.

src/job_cache/job_cache_impl_common.cpp
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  copy(src_fd->get(), dst_fd->get());
+}
+
+#endif
+
+void remove_backing_files(const std::string &dir, int64_t job_id) {
+  uint8_t group_id = job_id & 0xFF;
+  std::string job_dir = wcl::join_paths(dir, wcl::to_hex(&group_id), std::to_string(job_id));
```
V-FEXrt, line 203: This looks like duplicated code from somewhere else right? Like is this code also used somewhere for determining where to write a file?  If so can we make a "get job_dir from job_id" function so that we don't have to keep that code aligned everywhere we do it?

src/job_cache/job_cache_impl_common.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+void remove_backing_files(std::string dir, const std::vector<int64_t> &job_ids,
+                          size_t min_removals_per_thread, size_t max_number_of_threads) {
+  // Calculate a good number of threads to use.
+  size_t actual_num_threads =
+      std::min(max_number_of_threads, std::max(1UL, job_ids.size() / min_removals_per_thread));
+
+  // Calculate how many removals will be done by each task
+  size_t actual_removals_per_thread =
+      job_ids.size() / actual_num_threads + !!(job_ids.size() % actual_num_threads);
```
V-FEXrt, line 230: > `!!(job_ids.size() % actual_num_threads)`

This looks like "add one if job_ids.size isn't an exact multiple of actual_num_threads" but it took some time to figure that out and I'm not even sure that I'm write so this probably warrants a comment

src/job_cache/schema.sql
```diff
@@ -95,5 +95,21 @@ create table if not exists output_symlinks(
   job            job_id  not null references jobs(job_id) on delete cascade);
 create index if not exists output_symlink_by_job on output_symlinks(job);
 
+-- We need to track the total size to know when to start collecting stuff
+create table if not exists total_size(
+  total_size_id integer primary key,
+  size          integer not null);
+
+insert into total_size (total_size_id, size) select 0, 0 where not exists (select * from total_size);
```
V-FEXrt, line 103: Not sure if it matters, but an id of 0 _feels_ wrong

src/job_cache/eviction_policy.cpp
```diff
@@ -80,8 +84,166 @@ struct CommandParser {
...
+  job_cache::PreparedStatement update_size;
+  job_cache::PreparedStatement get_size;
+  job_cache::PreparedStatement update_last_use;
+  job_cache::PreparedStatement find_least_recently_used;
+  job_cache::PreparedStatement remove_least_recently_used;
+  job_cache::Transaction transact;
+
+  // TODO: Right now we're using `obytes` as a proxy for the size of a job
+  //       but this is an over aproximation of the storage cost really.
+
```
V-FEXrt, line 98: mentioned in DM, but we probably want a "resync/recalc total size" query that is called on some interval

V-FEXrt (approved):
No blocking comments but a couple of potential improvements.

All in all this looks really good. Thanks for getting this out!

src/job_cache/job_cache_impl_common.cpp
JakeSiFive: Yeah sorry lol. That is in fact what it does. I always try to pair divisions and modulus when I have to use both but its kind of a silly optimization. It basically just ensures that the compiler will put the div and the mod in the same basic block and will find the optimization to emit the single divmod instruction more easily instead of doing two division operations...this is without a doubt a pre-mature and un-tested optimization lol. I've never disassembled things when I did this to see if it even matters lol.

src/job_cache/schema.sql
JakeSiFive: If you use auto-increment it starts at 1 not 0 so you might have a point. It's not wrong in any precise way that I'm aware of however. I'll change it to 1.

PR #1191, opened 2023-04-01
wasm: Revert incorrect fixes for emsdk+webpack build

V-FEXrt:
Previously I set `ENVIRONMENT=web` as a "fix" for the wasm+emsdk builds. This resulted in a passing build but broke the desktop version of the extension at runtime. (Unsurprisingly it fixed the web version)

This change reverts that and pins back to an older version of emsdk. I _think_ the revert is needed as newer versions of emsdk more readily use ES6 style exports which then in turn breaks webpack. (https://github.com/webpack/webpack/issues/16878). We can possible avoid this by setting `EXPORT_ES6=0` and that did seem to work but after spending all day deep in the weeds I just want to return to a known working state so desktop users can continue to be productive.

This comes at a cost to the web version of the LSP, for some reason its failing in a new/different way that it has before even though we should be returning to old functionality. In the short term this tradeoff seems reasonable as most users use the desktop app and syntax highlighting, the most important feature for code review, still works with the web extension.

There is some interesting discussion at https://github.com/emscripten-core/emscripten/issues/19066 as well. Hopefully we can get this resolved fairly quickly.

After landing this we should fast track a release since folks have been without a working desktop LSP for a bit.


ErikParawell-SiFive (approved):
Tested the desktop version is working. The web one is currently broken, but is a low priority item in my mind.

JakeSiFive (approved):
Yeah I'm happy to return to a known working state. I'm not sure that anyone is using the web extension right now.

richardxia (approved):

PR #1192, opened 2023-04-03
cli: Move validation and env parsing to CLO

V-FEXrt:
Moves validation logic to CommandLineOptions, also moves the env fallback for some options into CommandLineOptions


JakeSiFive (approved):
strict improvement, awesome!

PR #1196, opened 2023-04-06
decrease newlines/fix trailing newline

V-FEXrt:
Decrease newlines between top level defines to 1 instead of 2. This fixes the "too many trialing newlines at EOF" problem at the cost of slightly less separation. The separation can be regained with header/separation comments when needed.


tests/wake-format/basic/stdout
```diff
@@ -1038,4 +942,3 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
 
     json
 
-
```
ag-eitilt, line 1041: There's still an unnecessary blank line; GitHub won't show the EOF character (unless it's missing entirely and gets the red warning symbol), so any that show up in review are extra.

tests/wake-format/disallowed-input/stdout
```diff
@@ -24,7 +22,6 @@ def t12 =
     # comment
     )
 
-
 # trailing comment
```
ag-eitilt, line 27: Looks like comments trigger a different pathway for adding the newline; it's likely whatever's going on with the topic/target spacing I pointed out would also need to be addressed separately if either of those happened at the end of the file.

ag-eitilt (changes requested):
Doesn't actually solve the issue, unfortunately.

tests/wake-format/basic/stdout
V-FEXrt: Ah rip, that is what I get for trusting the diff lol

tests/wake-format/disallowed-input/stdout
V-FEXrt: Right, comments have a slightly different route, but it should just work for everything besides trailing comments.

Anything in this test case are explicitly not supported by the formatter. In this case it means that comments are not allowed to be the last thing in a file. (or at the very least, you can have it but it'll be formatted wrong)

tests/wake-format/basic/stdout
V-FEXrt: Okay, actually I'm confused. This is correct no? We want 1 empty line right?

The file above has the following content
```
...cut....
    json<newline>
<newline><eof>
```

ag-eitilt: We'd want `json\n` where where the single newline is the EOF character.  The original output had `json\n\n\n` and the change cuts it down to `json\n\n` but there's still an extra line in there.  GitHub hides the very last newline, but if it's showing something numbered then there's still an extra "empty string" being tokenized.

JakeSiFive (approved):
C++ code is good and I'm good with the formatting choice personally.

tests/wake-format/basic/stdout
V-FEXrt: I supposed I'm confused by why `json\n\n<eof>` is wrong. That feels like a pretty standard thing to do. (`json\n\n\n<eof>` is clearly wrong)

tests/wake-format/disallowed-input/stdout
V-FEXrt: yep can do

tests/wake-format/basic/stdout
ag-eitilt: Ah, I see where you're coming from then.  It's not *morally* wrong if that's your workflow, no, but it's not the style we've been using; all of the existing files in the repo I'm aware of don't do that, we've pointed it out in the past as unnecessary in code reviews as a style recommendation, and with leaving it as-is, every Wake file wherever the formatter is run will be touched to add space there where there (with potentially very few exceptions in the most obscure corners) wasn't any space previously.

tests/wake-format/disallowed-input/disallowed-input.wake
```diff
@@ -21,4 +21,4 @@ def t12 = (
   # comment
 )
 
-# trailing comment
\ No newline at end of file
+# trailing comment
```
ag-eitilt: It might be nice to keep this as no-newline to be sure the formatter adds one and only one, but this being the disallowed input, it's certainly less critical.

tests/wake-format/basic/stdout
```diff
@@ -823,10 +836,8 @@ def name Unit =
 
 # floating 6
 
-
 # floating 7
 
-
 def a b c = 5
 
 def a =
```
ag-eitilt: Yeah, why was there a double line before the middle def, but only a single one after?  Going to single-spacing hides that nicely, but if we ever want to go back to two it's going to require some digging.

ag-eitilt (approved):
Not gone through this carefully, with complete context, but looks good!

tests/wake-format/basic/stdout
V-FEXrt: That is because floating comments (`# floating 7`) also had two trailing newlines. I just forgot to update them both to be 1

tests/wake-format/disallowed-input/disallowed-input.wake
V-FEXrt: ah yeah, that wasn't intentional. I think my editor made the change

PR #1197, opened 2023-04-07
cli: list out unexpected positional args

V-FEXrt:
When failing an invocation its useful to list out the problematic arguments. When wake is invoked from a tool it might not be immediately obvious what the invocation command like was.

```
./bin/wake -o zero one two three
Unexpected positional arguments on the command-line!
        'one' 'two' 'three'
```

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -285,6 +285,11 @@ int main(int argc, char **argv) {
 
   if (noargs && clo.argc > 1) {
     std::cerr << "Unexpected positional arguments on the command-line!" << std::endl;
+    std::cerr << "\t'" << clo.argv[1] << "'";
+    for (int i = 2; i < clo.argc; i++) {
+      std::cerr << " '" << clo.argv[i] << "'";
+    }
+    std::cerr << std::endl;
     return 1;
   }
 
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -285,8 +285,8 @@ int main(int argc, char **argv) {
 
   if (noargs && clo.argc > 1) {
     std::cerr << "Unexpected positional arguments on the command-line!" << std::endl;
-    std::cerr << "\t'" << clo.argv[1] << "'";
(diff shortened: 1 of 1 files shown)


tools/wake/main.cpp
```diff
@@ -285,6 +285,11 @@ int main(int argc, char **argv) {
 
   if (noargs && clo.argc > 1) {
     std::cerr << "Unexpected positional arguments on the command-line!" << std::endl;
+    std::cerr << "\t'" << clo.argv[1] << "'";
+    for (int i = 2; i < clo.argc; i++) {
+      std::cerr << " '" << clo.argv[i] << "'";
+    }
```
ag-eitilt, line 291: Since you're not comma-separating these, you can avoid the duplication with a space-indent of one less than you want printed rather than a tab (which pushes it further right than a lot of error messages I've seen, despite being objectively the right character here).
```suggestion
    std::cerr << "   ";
    for (int i = 1; i < clo.argc; i++) {
      std::cerr << " '" << clo.argv[i] << "'";
    }
```

ag-eitilt (approved):
Definitely a good thing to add!  Always in favour of failing loudly.

tools/wake/main.cpp
V-FEXrt: Happy to change to spaces, but the duplication is still needed otherwise a trailling space gets printed

`'one' 'two' 'three' `

its not visible but its still there lol

ag-eitilt: I was keeping the extra space with the left quote, and subtracting it from the indentation (so if you're indenting four spaces, line 288 only prints *three* of them.

ag-eitilt: (I'm certainly happy keeping the tab, but with the deduplication that would mean you're indenting by `"\t "` for a total of nine columns.)

V-FEXrt: ohh, neat. That is a good way to solve that. Only works because of the proceeding spaces

PR #1199, opened 2023-04-07
Tweak the defination of simple parts to be more permissive

V-FEXrt:
- Simple 2 item applies like `Some ""` where each item is a literal or id is now considered a "simple literal"
  - this also means a top level `def x = Some ""` doesn't get newlined/is considered a constant
- Simple 2 part binop like `x < 0` where the lhs/rhs is a literal or id is now considered a "simple binop", and the shape `if simple_binop then simple_literal else simple_literal` is eligible to be on a single line
- `|` or `$` nested inside of parenthesis are now eligible to be on a single line

### DIFF
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -605,11 +605,7 @@ def tarball Unit =
     require Pass tarball =
         def cmd =
             def gnutar = which "gnutar"
-            def tar =
-                if gnutar ==* "gnutar" then
-                    which "tar"
-                else
-                    gnutar
+            def tar = if gnutar ==* "gnutar" then which "tar" else gnutar
             def files =
                 map getPathName (manifest, testManifest, spec, srcs)
                 | sortBy (_ <* _)
@@ -979,3 +975,16 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
         )
 
     json
+
+def anyArray =
+    exists (_.getJArray | isSome) value
+
+def a =
+    rec Nil (if indent > 0 then Some "" else None) lhs
+
+def a = Some ""
+
+def a =
+    def x = if a then b, else !b
+    require Pass readme = installIn "{dest}/share/doc/wake" "." readmeSource
+    x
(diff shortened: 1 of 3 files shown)


tests/wake-format/basic/stdout
```diff
@@ -605,11 +605,7 @@ def tarball Unit =
     require Pass tarball =
         def cmd =
             def gnutar = which "gnutar"
-            def tar =
-                if gnutar ==* "gnutar" then
-                    which "tar"
-                else
-                    gnutar
+            def tar = if gnutar ==* "gnutar" then which "tar" else gnutar
```
ag-eitilt, line 608: Yeah, this is about as far as I want to be pushing `if` blocks, but it's still fine for a single line.

tests/wake-format/basic/stdout
```diff
@@ -979,3 +975,16 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
...
+def anyArray =
+    exists (_.getJArray | isSome) value
+
+def a =
+    rec Nil (if indent > 0 then Some "" else None) lhs
+
+def a = Some ""
+
+def a =
+    def x = if a then b, else !b
```
ag-eitilt, line 988: I still maintain a dislike for single-line unary `,`, and this type error kind of emphasizes that.  Formatting looks good, though.

tools/wake-format/emitter.cpp
```diff
@@ -279,6 +270,147 @@ static wcl::doc binop_rhs_separator(const CSTElement& op) {
...
+
+static inline bool is_simple_apply(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node,
+                                   const token_traits_map_t& traits) {
+  if (node.id() != CST_APP) {
+    return is_simple_literal(builder, ctx, node, traits);
+  }
+
+  auto parts = collect_apply_parts(node);
+
+  if (parts.size() != 2) {
```
ag-eitilt, line 382: I don't expect there to ever be a `CST_APP` with less than two parts, but maybe `>` for better semantic readability anyway?
```suggestion
  if (parts.size() > 2) {
```

ag-eitilt (approved):
Only gave the CPP a very cursory scan, but the resulting formatting still looks good.

tests/wake-format/basic/stdout
V-FEXrt: Yeah, we can bikeshed the policy on unary `,` at some point but I'd like to push that to post MVP

JakeSiFive (approved):

PR #1200, opened 2023-04-10
format: remove trailing whitespace from bound comments

V-FEXrt:
Fixes trailing  white space that is  emitted after a  nested floating comment. Also adds more tests around floating  comments to catch other potential breakages

### DIFF
--- a/tools/wake-format/actions.h
+++ b/tools/wake-format/actions.h
@@ -87,7 +87,29 @@ struct TokenAction {
     auto it = traits.find(node);
     if (it != traits.end()) {
       for (auto n : it->second.before_bound) {
-        builder.append(n.fragment().segment().str());
+        if (n.id() == TOKEN_COMMENT) {
+          // Realign indent before writing the comment
+          freshline(builder, ctx);
+          builder.append(n.fragment().segment().str());
+          builder.append(wcl::doc::lit("\n"));
+          continue;
+        }
+
+        if (n.id() == TOKEN_NL) {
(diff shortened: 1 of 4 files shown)


ag-eitilt (commented):
Good catch!  Do we have anything removing end-of-line spacing?  (I'm pretty sure that's taken care of by default for the actual code, but we might want to strip within comments as well.)

tools/wake-format/actions.h
```diff
@@ -121,7 +143,29 @@ struct TokenReplaceAction {
     auto it = traits.find(node);
     if (it != traits.end()) {
       for (auto n : it->second.before_bound) {
-        builder.append(n.fragment().segment().str());
+        if (n.id() == TOKEN_COMMENT) {
```
JakeSiFive, line 146: I'm going to pull an Ashley and ask if we can factor out the repeated code between the repeated code here. I'm happy to not do it though as well.

JakeSiFive (approved):
C++ code looks good to me

tools/wake-format/actions.h
```diff
@@ -121,12 +91,38 @@ struct TokenReplaceAction {
...
+      }
+
+      // Realign indent in case we lost alighnment.
+      if (it->second.before_bound.size() > 0) {
         freshline(builder, ctx);
       }
     }
 
-    builder.append(str);
+    if (str == nullptr) {
```
JakeSiFive, line 121: sick, like this solution!

JakeSiFive (approved):

Issue #1106, opened 2023-02-02
Performing string interpolation inside a multiline string inserts incorrect indentation


bmitc:
Inside a Wake multiline string, if you have string interpolation, then it inserts indentation where it shouldn't be. It seems to insert an indentation of 3 spaces. Examples follow.

For the definition:
```wake
export def multiline1 _args =
    println "%
    testing %{str 123}
    %"
```
we get:
```bash
> wake multiline1
testing
Unit
```

But given
```wake
export def multiline2 _args =
    println "%
    testing %{str 123}
    %"
```
we get:
```bash
> wake multiline2
   testing 123
Unit
```

Additionally, the indentation carries over into newlines, as:
```wake
export def multiline3 _args =
    println "%
    testing %{str 123}
    asdf
    %"
```
gives:
```bash
> wake multiline3
   testing 123
   asdf
Unit
```

The only way to get rid of this is to remove some spacing. So:
```wake
export def multiline4 _args =
    println "%
 testing %{str 123}
 asdf
    %"
```
yields:
```bash
> wake multiline4
testing 123
asdf
Unit
```

However, this goes against the documentation in the [quick reference](https://github.com/sifive/wake/blob/master/share/doc/wake/quickref.md), which states that this should actually be an error.

ag-eitilt:
I'm guessing the first example shouldn't have the `%{str 123}`?

bmitc:
Updated description and the end (examples `multiline4` and `multiline5`) to better explain some indentation behavior that addresses a comment in the quick reference.

JakeSiFive:
Yeah this is a bug for sure. Adding string interpolation should not change the indentation removing feature.

bmitc:
Thanks @V-FEXrt for fixing this!

PR #1201, opened 2023-04-10
mstr: Fix extra prepended spaces on interpolation

V-FEXrt:
Fixes https://github.com/sifive/wake/issues/1106

Interpolated multiline strings would have prepended spaces in the resulting string. This is because the loop 

```
std::vector<Expr *> args;
for (CSTElement i = intp.firstChildNode(); !i.empty(); i.nextSiblingNode()) {
  if (args.size() % 2 == 0) fsm.accept(i);
}
```

intended for `fsm.accept(i)` to be called every other iteration but it was called every  iteration since `args` wasn't being appended to.

Example:

```
export def hax2 = "%
  fun day.
  happy day. %{str 123}
%"
```

Before fix:
```
> ./bin/wake -x 'hax2'
" fun day.\n happy day. 123"
```

After fix:
```
> ./bin/wake -x 'hax2'
"fun day.\nhappy day. 123"
```

### DIFF
--- a/src/dst/todst.cpp
+++ b/src/dst/todst.cpp
@@ -27,7 +27,6 @@
 #include <algorithm>
 #include <set>
 #include <sstream>
-#include <iostream>
 
 #include "expr.h"
 #include "parser/cst.h"
@@ -1153,7 +1152,7 @@ static Expr *dst_interpolate(CSTElement intp) {
 
   MultiLineStringIndentationFSM fsm;
   CSTElement i = intp.firstChildNode();
-  while(!i.empty()) {
+  while (!i.empty()) {
     fsm.accept(i);
     i.nextSiblingNode();
     i.nextSiblingNode();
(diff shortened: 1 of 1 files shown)


terpstra (approved):

src/dst/todst.cpp
```diff
@@ -1151,11 +1151,14 @@ static Expr *dst_interpolate(CSTElement intp) {
   std::string total;
 
   MultiLineStringIndentationFSM fsm;
-  for (CSTElement i = intp.firstChildNode(); !i.empty(); i.nextSiblingNode()) {
-    if (args.size() % 2 == 0) fsm.accept(i);
+  CSTElement i = intp.firstChildNode();
+  while (!i.empty()) {
+    fsm.accept(i);
+    i.nextSiblingNode();
+    i.nextSiblingNode();
```
JakeSiFive, line 1158: Here we're unconditionally skipping the next two sibling nodes. What do we expect these nodes to be? And can a comment about this be added?

PR #1202, opened 2023-04-11
Fix example for `find`

bmitc:
One of the examples for `find` had an off by 1 index error.

Can verify using `wake -x find (_==10) (85, 4, 10, 3, Nil)`.


ag-eitilt (approved):
Yep, that's definitely an error there!

V-FEXrt (approved):

PR #1204, opened 2023-04-11
format: Multiline strings now indent correct wrt delimiter

V-FEXrt:


V-FEXrt:
I still need to not copy+paste `MultiLineStringIndentationFSM` but other than that this should be ready for review of the output/format rules

tools/wake-format/emitter.cpp
```diff
@@ -1425,59 +1426,84 @@ wcl::doc Emitter::walk_literal(ctx_t ctx, CSTElement node) {
...
 
-  auto lstr_fmt = fmt()
+  // Insert the proper amount of spaces to correctly indent the line relative to base identation
+  auto inset_line = fmt().escape([prefix_size](wcl::doc_builder& builder, ctx_t ctx, CSTElement& node){
+           FMT_ASSERT(node.id() == TOKEN_WS, node, "Expected <TOKEN_WS>, Saw <" + std::string(symbolName(node.id())) + ">");
+           for (size_t i = prefix_size; i < node.fragment().segment().size(); i++) {
+             builder.append(" ");
+           }
+           node.nextSiblingElement();
+  });
```
JakeSiFive, line 1447: Spoke off line about this but I think there's a slight issue if someone types an actual tab character into a string here instead of using \t. As is it will be replaced with a space. That would be ok IMO if it were documented and matched how the runtime behaved but I don't think it does.

JakeSiFive (commented):
I think we should take a stance on the tabs issue if not out-right solve it but otherwise code looks good.

tools/wake-format/emitter.cpp
V-FEXrt: Okay it is updated to do the replace as it should but I don't think wake in general correctly handles this case

JakeSiFive (approved):

PR #1205, opened 2023-04-12
Allow -> instead of = for match case

V-FEXrt:
Allows `->` instead of `=` for match case.

Ex:

```
> cat do.wake
package build_wake
from wake import _

export def test2 Unit =
  match (Some 5)
    Some x -> x
    None   -> 0
> ./bin/wake -x 'test2 Unit'
5
```

Feel free to fight out the exact characters for `->` below. It's easy to change


JakeSiFive (approved):

mmjconolly:
It's probably worth doing a release for this one

JakeSiFive:
> It's probably worth doing a release for this one

I think this would be 32.1.0 per symver right?

V-FEXrt:
Yep!

PR #1206, opened 2023-04-12
Fix field documentation formatting for `Usage` and `Plan` tuples

bmitc:
The formatting was incorrect for use with the Wake LSP and VS Code extension (since the implementation of `Usage` and `Plan` predated the existence of those tools). Screenshots showcasing behavior before the fix:

image

image

Screenshot with field documentation in correct location:

image

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -35,9 +35,12 @@ export tuple RunnerInput =
   export Directory:   String
   export Stdin:       String
   export Resources:   List String
-  export Prefix:      String        # a unique prefix for this job
-  export Record:      Usage         # previous resource usage
-  export IsAtty:      Boolean       # should job run in psuedoterminal
+  # A unique prefix for this job
+  export Prefix:      String
+  # Previous resource usage
+  export Record:      Usage
+  # Determines if job should run in psuedoterminal
+  export IsAtty:      Boolean
 
 export tuple RunnerOutput =
   export Inputs:  List String
@@ -86,23 +89,40 @@ export data Persistence =
 
 # A Plan describes a not-yet-executed Job
 tuple Plan =
-  export Label:        String       # The label used when showing the command during execution
-  export Command:      List String  # The command-line arguments (the first is the command to run)
-  export Visible:      List Path    # Only these files should be available to the command
-  export Environment:  List String  # KEY=VALUE environment variables fed to the command
-  export Directory:    String       # The directory in which the command should be run
-  export Stdin:        String       # The file to which standard input should be connected
-  export Stdout:       LogLevel     # How should standard output be displayed during a build
-  export Stderr:       LogLevel     # How should standard error be displayed during a build
-  export Echo:         LogLevel     # Echo the command to this stream
-  export Persistence:  Persistence  # See Persistence table above
-  export LocalOnly:    Boolean      # Must run directly in the local workspace; no output detection performed
-  export Resources:    List String  # The resources a runner must provide to the job (licenses/etc)
-  export RunnerFilter: Runner => Boolean # Reject from consideration Runners which the Plan deems inappropriate
-  export Usage:        Usage        # User-supplied usage prediction; overruled by database statistics (if any)
-  export FnInputs:     (List String => List String) # Modify the Runner's reported inputs  (files read)
-  export FnOutputs:    (List String => List String) # Modify the Runner's reported outputs (files created)
-  export IsAtty:       Boolean      # Should job run in psuedoterminal
+  # The label used when showing the command during execution
+  export Label:        String
+  # The command-line arguments (the first is the command to run)
+  export Command:      List String
+  # Only these files should be available to the command
+  export Visible:      List Path
+  # KEY=VALUE environment variables fed to the command
+  export Environment:  List String
+  # The directory in which the command should be run
+  export Directory:    String
+  # The file to which standard input should be connected
+  export Stdin:        String
+  # How should standard output be displayed during a build
+  export Stdout:       LogLevel
+  # How should standard error be displayed during a build
+  export Stderr:       LogLevel
+  # Echo the command to this stream
+  export Echo:         LogLevel
+  # See Persistence table above
+  export Persistence:  Persistence
+  # Must run directly in the local workspace; no output detection performed
+  export LocalOnly:    Boolean
+  # The resources a runner must provide to the job (licenses/etc)
+  export Resources:    List String
+  # Reject from consideration Runners which the Plan deems inappropriate
+  export RunnerFilter: Runner => Boolean
+  # User-supplied usage prediction; overruled by database statistics (if any)
+  export Usage:        Usage
+  # Modify the Runner's reported inputs  (files read)
+  export FnInputs:     (List String => List String)
+  # Modify the Runner's reported outputs (files created)
+  export FnOutputs:    (List String => List String)
+  # Should job run in psuedoterminal
+  export IsAtty:       Boolean
 
 def isOnce: Persistence => Boolean = match _
   ReRun = False


bmitc:
This is also probably a good time for us to just jog over the documentation for each field and see if it could be updated or improved, so if you have time to read over the documentation, that would be great!

share/wake/lib/system/job.wake
```diff
@@ -86,23 +89,40 @@ export data Persistence =
...
+  # How should standard output be displayed during a build
+  export Stdout:       LogLevel
+  # How should standard error be displayed during a build
+  export Stderr:       LogLevel
+  # Echo the command to this stream
+  export Echo:         LogLevel
+  # See Persistence table above
+  export Persistence:  Persistence
+  # Must run directly in the local workspace; no output detection performed
+  export LocalOnly:    Boolean
```
bmitc, line 113: Could/should this be re-worded to `# Forces job to run in the local workspace; ...`? I'm mainly asking because I'm not sure of the cases where this is set to `True` but the job is not ran in the local workspace.

share/wake/lib/system/job.wake
```diff
@@ -86,23 +89,40 @@ export data Persistence =
...
-  export Persistence:  Persistence  # See Persistence table above
-  export LocalOnly:    Boolean      # Must run directly in the local workspace; no output detection performed
-  export Resources:    List String  # The resources a runner must provide to the job (licenses/etc)
-  export RunnerFilter: Runner => Boolean # Reject from consideration Runners which the Plan deems inappropriate
-  export Usage:        Usage        # User-supplied usage prediction; overruled by database statistics (if any)
-  export FnInputs:     (List String => List String) # Modify the Runner's reported inputs  (files read)
-  export FnOutputs:    (List String => List String) # Modify the Runner's reported outputs (files created)
-  export IsAtty:       Boolean      # Should job run in psuedoterminal
+  # The label used when showing the command during execution
+  export Label:        String
```
bmitc, line 93: Should we elaborate here on how the label is printed out with the job ID in parentheses and where this shows up? I know how the label shows up in `wake --last` and the `.in.json` and `.out.json` files and probably other places, but that isn't necessarily clear from the documentation. I don't know (off the top of my head at the moment) if we have it documented elsewhere.

bmitc (commented):
Just some questions about some of the fields that would help me understand things a bit better.

share/wake/lib/system/job.wake
mmjconolly: It also appears in the terminal progress bar

JakeSiFive: Also, I think one of the best features now is that you can find the output of a job by searching for its label now.

share/wake/lib/system/job.wake
```diff
@@ -86,23 +89,40 @@ export data Persistence =
...
+  # The command-line arguments (the first is the command to run)
+  export Command:      List String
+  # Only these files should be available to the command
```
JakeSiFive, line 96: and directories

share/wake/lib/system/job.wake
```diff
@@ -86,23 +89,40 @@ export data Persistence =
...
+  export Visible:      List Path
+  # KEY=VALUE environment variables fed to the command
+  export Environment:  List String
+  # The directory in which the command should be run
```
JakeSiFive, line 100: I'd calling out "working directory" specifically.

share/wake/lib/system/job.wake
```diff
@@ -86,23 +89,40 @@ export data Persistence =
...
+  export Directory:    String
+  # The file to which standard input should be connected
```
JakeSiFive, line 102: Might note that the empty string is interpreted as there being no Stdin file

share/wake/lib/system/job.wake
mmjconolly: I think longer term we'll remove `LocalOnly` entirely.

It was supposed to mean that it would run on the local machine, rather than a remote machine.
But it came to mean that it would run without fuse.
Then it came to mean it would run without fuse and without filesystem replacement.
Which effectively means that it runs the same a `runJobWith localRunner`.

I think the safest thing to say right now is `# Deprecated` without an explaination

JakeSiFive: 'LocalOnly' is named poorly. Right now in practice it means "run without sandboxing" but I think it was supposed to mean "run locally" as in "do not use remote execution"

share/wake/lib/system/job.wake
```diff
@@ -86,23 +89,40 @@ export data Persistence =
...
+  # How should standard error be displayed during a build
+  export Stderr:       LogLevel
+  # Echo the command to this stream
+  export Echo:         LogLevel
+  # See Persistence table above
+  export Persistence:  Persistence
+  # Must run directly in the local workspace; no output detection performed
+  export LocalOnly:    Boolean
+  # The resources a runner must provide to the job (licenses/etc)
+  export Resources:    List String
```
JakeSiFive, line 115: I'd clarify that these are un-interpreted strings and that they are only meaningful to the right runners.

share/wake/lib/system/job.wake
```diff
@@ -193,8 +213,10 @@ export def makePlan (label: String) (visible: List Path) (command: String): Plan
   makeShellPlan command visible
   | setPlanLabel label
 
-# Set reasonable defaults for all Plan arguments
+# Identity function
 def id x = x
+
+# Set reasonable defaults for all Plan arguments
```
bmitc, line 219: This change was done due to #1207.

As an aside, the `id` function is used both in the `Job.wake` and `Path.wake` files. While out of scope for this particular PR, what do we think about adding an exported identity function in Wake? I think I've created one every now and then myself, and it could be useful. I would probably recommend `def identity` instead of just `def id` though, if we did that, since "id" will be far more commonly used than "identity", as words and function and variable names.

share/wake/lib/system/job.wake
bmitc: @JakeSiFive @mmjconolly @ag-eitilt Should I additionally state a little blurb about how this field is discouraged or something to focus for what they're specifically meant for? My understanding is that over time and in some flows, the `Resources` field is used (abused?) as a way to get custom behavior/DSLs in plans.

share/wake/lib/system/job.wake
bmitc: [Clarified here](https://github.com/sifive/wake/pull/1206/commits/1fc38d4549433709dd823c8af2fd16751eb68308)

share/wake/lib/system/job.wake
mmjconolly: Not all usage of Resources is bad, they're just opaque strings that get passed through

I guess that's up to the authors of a `Runner` to decide, as they are the ultimate intepreters of `Resources`.

share/wake/lib/system/job.wake
bmitc: [Clarified here](https://github.com/sifive/wake/pull/1206/commits/1fc38d4549433709dd823c8af2fd16751eb68308)

share/wake/lib/system/job.wake
bmitc: [Clarified here](https://github.com/sifive/wake/pull/1206/commits/1fc38d4549433709dd823c8af2fd16751eb68308)

share/wake/lib/system/job.wake
bmitc: ["Deprecated" here](https://github.com/sifive/wake/pull/1206/commits/1fc38d4549433709dd823c8af2fd16751eb68308)

share/wake/lib/system/job.wake
bmitc: Ah, gotcha. I may have picked up my sentiment from a singular example of `Resources` being used in an unintended way at some point and extrapolated.

[Clarified here](https://github.com/sifive/wake/pull/1206/commits/1fc38d4549433709dd823c8af2fd16751eb68308)

share/wake/lib/system/job.wake
bmitc: I took a stab at this [here](https://github.com/sifive/wake/pull/1206/commits/ed11306e48d66f3b8554cd0d79b4aa09bdee9a51). Not married to this documentation implementation, so suggestions/corrections are welcomed.

bmitc:
@mmjconolly @JakeSiFive This is not a poke but just a comment to clarify that I'm done updating the documentation based upon the feedback and clarifications given. Thanks for those! Let me know if my updates are kosher or need further adjustments.

V-FEXrt (approved):

JakeSiFive (approved):

PR #1208, opened 2023-04-13
format: specifiy internal newlines based on structure

V-FEXrt:
Newlines are now emitted inside of a def/require body based on the following rules.

1) multi-line def/requires get a newline
2) the “return” item gets a newline
3) a commented item gets a newline
4) changing def<->require gets a newline
4) anything else doesn't

### DIFF
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -1275,6 +1275,10 @@ def tarball Unit =
         | s # a comment
 
     def x = y
+    def srcs = allSources # a comment
+    def x = y
+    def srcs = allSources # a comment
+    def x = y
 
     x
 
@@ -1284,5 +1288,28 @@ def tarball Unit =
         | s # a comment
 
     require x = y
+    require srcs = allSources # a comment
+    require x = y
+    require srcs = allSources # a comment
+    require x = y
+
+    require Pass binfiles2 = binfilesResult2
+    else
+        blah # a comment
+
+    require x = y
+
+    require Pass binfiles2 = binfilesResult2 # a comment
+    else
+        blah # a comment
+
+    require x = y
+
+    require Pass binfiles2 = binfilesResult2 # a comment
+    else blah
+
+    require srcs = allSources # a comment
+    require x = y
+    require srcs = allSources # a comment
 
     x
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -1268,3 +1268,21 @@ def doInstall dest kind =
     require Pass bin = val
 
     Pass outputs
+
+def tarball Unit =
+    def srcs =
+        allSources
+        | s # a comment
+
+    def x = y
+
+    x
+
+def tarball Unit =
+    require srcs =
+        allSources
+        | s # a comment
+
+    require x = y
+
+    x
--- a/tests/wake-format/basic/stdout
+++ b/tests/wake-format/basic/stdout
@@ -10,31 +10,25 @@ def five = 5
 
 def name Unit =
     def other = @here
+
     def other =
         "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+
     another
 
 def name Unit =
     def other = @here
     def other = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
-    another
 
-def name Unit =
-    def other = @here
-
-    # not floating
-
-    def other = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
     another
 
 def name Unit =
     def other = @here
 
     # not floating
-    # not floating
-    # not floating
 
     def other = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+
     another
 
 def name Unit =
@@ -44,10 +38,8 @@ def name Unit =
     # not floating
     # not floating
 
-    # not floating
-    # not floating
-
     def other = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+
     another
 
 def name Unit =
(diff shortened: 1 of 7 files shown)


tests/wake-format/basic/stdout
```diff
@@ -95,27 +106,32 @@ def name Unit =
...
+
     another
 
 def name Unit =
     def other = @here
+
     # wake-format off
     def other = here
-
     def other = @here
```
ag-eitilt, line 126: Is this due to the `# wake-format off` or to both of them being simple defs?  I'm expecting the second, but I'm curious if disabling the formatter would have an effect either way.  Also, weren't we planning on using the formatter to switch over to `@here`?

tests/wake-format/basic/stdout
```diff
@@ -558,61 +574,77 @@ def all variant =
...
     require Pass allSources = sources "." `.*`
+
     def srcs =
         allSources
         | filter (!matches `(debian/|wake.spec).*` _.getPathName) # omit packaging files
         | filter (!matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
         | filter (!matches `(.*/)*\.gitignore` _.getPathName) # omit git files
+
+
     def Pair testPaths wakePaths = splitBy (matches `tests/.*` _.getPathName) srcs
```
ag-eitilt, line 634: Is this double space deliberate?

tests/wake-format/basic/stdout
```diff
@@ -853,8 +899,8 @@ def a =
     # comment
 
     def c = 5
-
     def d = 5
+
     # comment
 
     result
```
ag-eitilt, line 906: Much better!  I think you've said previously that if the comment had been followed by two blank lines before `result` then both would stay in the output; is that still true?

tests/wake-format/basic/stdout
```diff
@@ -211,6 +227,7 @@ def t10 =
 def name Unit = # wake-format off
     def other = here
     def other = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+
     another
```
ag-eitilt, line 231: Should disabling the formatting disable the newline, or am I misunderstanding where the `# wake-format off` is bound to?

ag-eitilt (approved):
I like the formatting!  (Not looked at the C++)  It's not a set of rules I would have come up with easily, but it does seem to work out very nicely.

tests/wake-format/basic/stdout
JakeSiFive: I'm unclear, we don't expect anyone to use it like this so I don't think it matters but I think this is binding to the left so its saying "don't format the '=' token" lol. There's no place to bind the format off comment to the block itself so you'd have to turn off formatting for the whole def I think.

tools/wake-format/emitter.cpp
```diff
@@ -1572,33 +1641,66 @@ wcl::doc Emitter::walk_require(ctx_t ctx, CSTElement node) {
...
-               .token(TOKEN_KW_REQUIRE)
-               .ws()
-               .walk(WALK_NODE)
-               .consume_wsnlc()
-               .space()
-               .token(TOKEN_P_EQUALS)
-               .consume_wsnlc()
-               .join(rhs_fmt())
+               .join(pre_body_fmt)
+               // Returns true if the body should be separated from the currently require
```
JakeSiFive, line 1667: nit: "current"

tools/wake-format/emitter.cpp
```diff
@@ -1572,33 +1641,66 @@ wcl::doc Emitter::walk_require(ctx_t ctx, CSTElement node) {
...
+                     CSTElement copy = inner.firstChildElement();
+                     wcl::doc fmted =
+                         fmt().join(pre_body_fmt).compose(ctx.sub(builder), copy, token_traits);
+                     if (fmted->newline_count() > count_allowed_newlines(traits, inner) + 1) {
+                       return true;
+                     }
+
+                     return false;
+                   },
+                   fmt().newline().freshline(), fmt().freshline())
```
JakeSiFive, line 1703: I love "newline" and "freshline", I think those were very hard won and this shows how nice they are!

JakeSiFive (approved):
C++ looks good.

tests/wake-format/basic/stdout
V-FEXrt: This pushes in to the just the first def.
Its the same as 

```
def name Unit =
     # wake-format off
    def other = here
    def other = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

    another
```

so only the first def is "off", everything is formatted normally. FWIW if I see this is code review I'd ask it to be changed. This is more a sanity check than anything else

tests/wake-format/basic/stdout
V-FEXrt: Yes. `#comment` is bound to `result` and it captures all newlines along the way. We could possibly change that. I'm not sure which is better, but right now we are assuming the author had a reason for not putting them together

tests/wake-format/basic/stdout
V-FEXrt: no, nice catch!

tests/wake-format/basic/stdout
V-FEXrt: > Is this due to the # wake-format off or to both of them being simple defs? 

Simple defs, the format off only applies to `def other = here`, you need to put it at the top level if you want to disable the entire function

>  Also, weren't we planning on using the formatter to switch over to @here?

Yep, but we have to respect format off even for refactors. In this case we'd have to fix the (likely few) cases manually before removing `here`

ag-eitilt: > > Also, weren't we planning on using the formatter to switch over to @here?
> 
> Yep, but we have to respect format off even for refactors. In this case we'd have to fix the (likely few) cases manually before removing `here`

I completely missed the fact that was falling under the format-off switch -- saw it to ask about the line (my question was whether format-off for one or the other `def` would affect removing the linebreak between them or whether that's entirely controlled by the parent block), but then promply forgot it for the `here`.

tests/wake-format/basic/stdout
V-FEXrt: sorry took a bit to  fix this because I had to  add a new api mnemonic. But ultimately its a net positive so glad to have it.

Should be good now

JakeSiFive (approved):
LGTM with the addition of the comments we discussed offline, also seems like else blocks aren't quite handled correctly?

tests/wake-format/basic/stdout
```diff
@@ -1284,5 +1288,28 @@ def tarball Unit =
...
 
     require x = y
+    require srcs = allSources # a comment
+    require x = y
+    require srcs = allSources # a comment
+    require x = y
+
+    require Pass binfiles2 = binfilesResult2
+    else
+        blah # a comment
```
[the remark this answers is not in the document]
V-FEXrt: Yeah, I'll fix that in a follow up PR

JakeSiFive (approved):

PR #1210, opened 2023-04-17
format: Fixup failing cases

V-FEXrt:
After recent changes a couple formatting cases were asserting/emitting invalid code. This change fixes those cases and adds tests for them.

It also removes some unused code

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -114,18 +114,6 @@ static inline bool is_simple_literal(wcl::doc_builder& builder, ctx_t ctx, CSTEl
   return part.empty();
 }
 
-static inline bool contains_req_else(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node,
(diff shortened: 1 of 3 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -1263,6 +1250,12 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
     parts = collect_right_binary(op_token, node);
   }
 
+  if (ctx.explode_option == ExplodeOption::Prevent) {
+    MEMO_RET(select_best_choice({
```
JakeSiFive, line 1254: why use select_best_choice when there is only one option?

V-FEXrt: because I didn't thing about it lol

V-FEXrt: hey if you only get it one choice it makes the selection super easy

JakeSiFive (approved):

PR #1212, opened 2023-04-18
format: Small refactor for "human flat"

V-FEXrt:
Explicitly call out and document "human flat" which is an emitted doc where the only newlines are created by comments

### DIFF
--- a/tests/wake-format/basic/basic.wake
+++ b/tests/wake-format/basic/basic.wake
@@ -895,42 +895,6 @@ target writeImp inputs mode path content =
     | getJobOutput
 
 export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
-  def json =
-    # comment
-    ( 
-      JObject 
-      "label"       :-> JString label,
-      "command"     :-> command     | map JString | JArray,
-      "environment" :-> environment | map JString | JArray
-    )
-    (
-    "label"       :-> JString label,
-    "command"     :-> command     | map JString | JArray,
-    "environment" :-> environment | map JString | JArray,
-    "visible"     :-> visible | map (_.getPathName.JString) | JArray,
-    "directory"   :-> JString directory,
-    "stdin"       :-> JString stdin,
-    "resources"   :-> res | map JString | JArray,
-    "version"     :-> JString version,
-    "mount-ops"      :-> JArray (
-      JObject (
-        "type"        :-> JString "workspace",
-        "destination" :-> JString ".",
-        Nil
-      ),
-      Nil
-    ),
-    "usage"       :-> JObject (
-      "status"    :-> JInteger status,
-      "runtime"   :-> JDouble  runtime,
-      "cputime"   :-> JDouble  cputime,
-      "membytes"  :-> JInteger membytes,
-      "inbytes"   :-> JInteger inbytes,
-      "outbytes"  :-> JInteger outbytes,
-      Nil
-    ),
-    Nil
-    )
 
   def json = JObject (
     "label"       :-> JString label,
@@ -959,6 +923,51 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
       Nil
     ),
     Nil
+  )
+
+  def json =
+    # comment
+    ( 
+      JObject 
+      "label"       :-> JString label,
+      "command"     :-> command     | map JString | JArray,
+      "environment" :-> environment | map JString | JArray
+    )
+    (
+    "label"       :-> JString label,
+    "command"     :-> command     | map JString | JArray,
+    Nil
+    )
+
+  def json = (
+    JObject
+    (
+      "label"       :-> JString label,
+      "command"     :-> command     | map JString | JArray,
+      "environment" :-> environment | map JString | JArray,
+      Nil
+    )
+  )
+
+  def json = (
+    JObject #comment
+    (
+      "label"       :-> JString label,
+      "command"     :-> command     | map JString | JArray,
+      "environment" :-> environment | map JString | JArray,
+      Nil
+    )
+  )
+
+  def json = (
+    JObject
+    #comment
+    (
+      "label"       :-> JString label,
+      "command"     :-> command     | map JString | JArray,
+      "environment" :-> environment | map JString | JArray,
+      Nil
+    )
   )
 
   json
--- a/tests/wake-format/basic/basic.wake
+++ b/tests/wake-format/basic/basic.wake
@@ -895,6 +895,43 @@ target writeImp inputs mode path content =
     | getJobOutput
 
 export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
+  def json =
+    # comment
+    ( 
+      JObject 
+      "label"       :-> JString label,
+      "command"     :-> command     | map JString | JArray,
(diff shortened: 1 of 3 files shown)


tests/wake-format/basic/basic.wake
```diff
@@ -895,6 +895,43 @@ target writeImp inputs mode path content =
...
 export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
+  def json =
+    # comment
+    ( 
+      JObject 
+      "label"       :-> JString label,
+      "command"     :-> command     | map JString | JArray,
+      "environment" :-> environment | map JString | JArray
+    )
+    (
```
JakeSiFive, line 906: did you mean for a comma to be here or something? In fact none of this is particularly valid wake code lol. I'm not sure what you were trying to do.

tests/wake-format/basic/stdout
```diff
@@ -1052,6 +1051,34 @@ target writeImp inputs mode path content =
     | getJobOutput
 
 export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
+    def json =
```
JakeSiFive, line 1054: This looks *super* nice to me btw.

tools/wake-format/emitter.cpp
```diff
@@ -201,23 +201,54 @@ static size_t count_trailing_newlines(const token_traits_map_t& traits, const CS
...
+
+// Determines if a given node would emit a trailing comment if emitted.
+static bool has_trailing_comment(const CSTElement& node, const token_traits_map_t& traits) {
+  return count_trailing_newlines(traits, node) > 0;
+}
+
+// Determizes of a doc is "flat" as a human would judge it. Human's don't
+// consider leading/trailing comments as invalidating flatness even though they
+// necessite extra newlines. Thus leading/trailing newlines are allowed.
+static bool is_human_flat(const wcl::doc& doc, const CSTElement& node,
```
JakeSiFive, line 217: I think part of my issue with this is that there are actually 3 kinds of "flat", not just 2.

1) Actully flat meaning the docs has no (or 1 depending on how you define it) newlines
2) This definition which is good for defs/requires inside of blocks, or expressions inside of exploded operator lists...e.g. anything where we have a sequence we want to output vertically.
3) One where only trailing comments are not considered to ruin flatness but a leading comment would actully be cause to indent something.

For instance

The following should not be indented
```
require Some x = y
else failWithError "blarg" # fail fast
```

but this should be

The following should not be indented (in fact it has to be in order to be correct)
```
require Some x = y
else
    # fail fast
    failWithError "blarg"
```

That said the expression in the else case would be considered "human flat" as you described it. I think we need to refine our vocubulary here to consider all 3 cases and use them appropriately.

tools/wake-format/emitter.cpp
```diff
@@ -1057,19 +1092,42 @@ wcl::optional<wcl::doc> Emitter::combine_apply_constructor(ctx_t ctx,
...
+  // If the RHS has a leading comment then we must respect the regular format
+  // Json
+  // # comment
+  // ( ... )
+  //
+  // can't become
+  //
+  // Json # comment (
+  //   ...
+  // )
```
JakeSiFive, line 1112: Can you add this test?

tools/wake-format/emitter.cpp
```diff
@@ -1057,19 +1092,42 @@ wcl::optional<wcl::doc> Emitter::combine_apply_constructor(ctx_t ctx,
...
-  if (lhs_fmt->newline_count() != count_leading_newlines(token_traits, lhs)) {
+  // If the LHS has a trailing comment then we must respect the regualr format
+  // Json # comment
+  // ( ... )
+  //
+  // can't become
+  //
+  // Json # comment (
+  //   ...
+  // )
```
JakeSiFive, line 1125: ditto

tests/wake-format/basic/basic.wake
V-FEXrt: no comma, the goal was two have something multiline on the LHS of the apply.

so the main test is 

`Constructor ( some multiline RHS)`

while this bonus test is

`(some multiline constructor) (some multiline RHS)`

tools/wake-format/emitter.cpp
JakeSiFive: How about this

1) strict_flat (doc has no newlines)
2) vertically_flat (what you currently call human flat. this name hints that we want to use this notion of flat in vertical output cases like blocks or exploded expressions)
3) weak_flat (the 3rd, as yet unnamed notion of "flat")

V-FEXrt: Okay! This should be done! PTAL

tools/wake-format/emitter.cpp
```diff
@@ -201,23 +201,60 @@ static size_t count_trailing_newlines(const token_traits_map_t& traits, const CS
...
+
+// Determines if the doc is "weakly flat". Weakly flat is a flat doc
+// with a single trailing comment allowed. No other newlines my be emitted.
+static bool is_weakly_flat(const wcl::doc& doc, const CSTElement& node,
+                           const token_traits_map_t& traits) {
+  return !doc->has_newline() ||
+         (doc->newline_count() == 1 && count_trailing_newlines(traits, node) == 1);
+}
+
+// Determizes of a doc is "flat" as a human would judge it. Human's don't
```
JakeSiFive, line 222: "if" not "of"

tools/wake-format/emitter.cpp
```diff
@@ -201,23 +201,60 @@ static size_t count_trailing_newlines(const token_traits_map_t& traits, const CS
...
+// consider leading/trailing comments as invalidating flatness even though they
+// require extra newlines. Thus leading/trailing newlines are allowed.
```
JakeSiFive, line 224: I would change this comment to reflect the subtitles of what humans consider "flat", for instance this comment never matched my intuition. Something like this:

Determines if a doc is "vertically" flat. A vertically flat doc is "flat" if the only newlines in it come from comments. This is the notion of "flat" you would want to consider when arranging docs in a vertical list where only the "body" of the doc (e.g. not the leading or trailing comments) needs to be flat. Internal comments would however violate this property.

tools/wake-format/emitter.cpp
```diff
@@ -1449,7 +1450,7 @@ wcl::doc Emitter::walk_data(ctx_t ctx, CSTElement node) {
                    .join(fmt_members)
                    .format(ctx, node.firstChildElement(), token_traits);
 
-  if (is_vertically_flat(no_nl, node, token_traits)) {
+  if (is_vertically_flat(no_nl, node, token_traits) && !is_top_level) {
```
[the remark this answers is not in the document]
JakeSiFive: nevermind, I was just looking at the most recent commit without seeing all the commits at once.

JakeSiFive (approved):
My only request is for some comment changes, otherwise looks good to me!

tools/wake-format/emitter.cpp
JakeSiFive: Also "Determizes" is spelled wrong. I noticed the "z" but I didn't trust my own spelling enough honestly.

PR #1218, opened 2023-04-20
format: Disable splitting types across multiple lines

V-FEXrt:
For the MVP, I want to skip splitting types into multiple lines. It has some implications we haven't explored yet and technically isn't supported


JakeSiFive (approved):
Fine by me for MVP, we should come back to this and see if people want it.

V-FEXrt:
Agreed!

PR #1220, opened 2023-04-21
format: prepare wake code for formatter

V-FEXrt:
Certain code isn't supported and generally can't be  supported by the formatter.  There will always be cases (do to the infinite nature of recursive grammars) where the programmer know better. This change  handles those cases manually in sifive/wake

### DIFF
--- a/share/wake/lib/core/json.wake
+++ b/share/wake/lib/core/json.wake
@@ -423,6 +423,7 @@ def mergeValueList (describeCompat: Boolean) (values: List JValue): Result JValu
             if eqFn x y then
                 Pass x
             else
+                # wake-format off
                 compatibilityMessage "N" "n" "ot every type can combine multiple different values. The incompatible values are: {map wrapFn values | formatExamples}"
                 | failWithError
         match values
@@ -508,6 +509,7 @@ def mergeValueList (describeCompat: Boolean) (values: List JValue): Result JValu
             | JArray
             | Pass
         _ _ _ _ _ _ _ =
+            # wake-format off
             compatibilityMessage "V" "v" "alues of different types may not be combined. The incompatible values are: {formatExamples values}"
             | failWithError
 
--- a/share/wake/lib/core/print.wake
+++ b/share/wake/lib/core/print.wake
@@ -61,20 +61,20 @@ export def makeLogLevel2 (name: String) (colour: Option Colour) (intensity: Opti
     def set name code = prim "colour"
     def colourCode = match colour
         None   = 0
-        Some c = 8 + match c
-            Black   = 0
-            Red     = 1
-            Green   = 2
-            Yellow  = 3
-            Blue    = 4
-            Magenta = 5
-            Cyan    = 6
-            White   = 7
+        Some c = match c
+            Black   = 8
+            Red     = 9
+            Green   = 10
+            Yellow  = 11
+            Blue    = 12
+            Magenta = 13
+            Cyan    = 14
+            White   = 15
     def intensityCode = match intensity
         None   = 0
-        Some i = 16 * match i
-            Dim    = 1
-            Bright = 2
+        Some i = match i
+            Dim    = 16
+            Bright = 32
     def _ = set name (colourCode+intensityCode)
     LogLevel name
 
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -221,6 +221,9 @@ export def makePlan (label: String) (visible: List Path) (command: String): Plan
 # Identity function
 def id x = x
 
+def bToInt b =
+  if b then 1 else 0
+
 # Set reasonable defaults for all Plan arguments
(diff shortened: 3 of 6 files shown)


share/wake/lib/core/print.wake
```diff
@@ -61,20 +61,20 @@ export def makeLogLevel2 (name: String) (colour: Option Colour) (intensity: Opti
...
+            Cyan    = 14
+            White   = 15
     def intensityCode = match intensity
         None   = 0
-        Some i = 16 * match i
-            Dim    = 1
-            Bright = 2
+        Some i = match i
+            Dim    = 16
+            Bright = 32
```
ag-eitilt, line 77: What does the formatter turn this into currently?

share/wake/lib/core/json.wake
```diff
@@ -423,6 +423,7 @@ def mergeValueList (describeCompat: Boolean) (values: List JValue): Result JValu
             if eqFn x y then
                 Pass x
             else
+                # wake-format off
                 compatibilityMessage "N" "n" "ot every type can combine multiple different values. The incompatible values are: {map wrapFn values | formatExamples}"
```
ag-eitilt, line 427: Sorry about this!

ag-eitilt (approved):

share/wake/lib/core/print.wake
V-FEXrt:
```
export def makeLogLevel2 (name: String) (colour: Option Colour) (intensity: Option Intensity): LogLevel =
    def set name code = prim "colour"

    def colourCode = match colour
        None -> 0
        Some c ->
            8
            + match c
                Black -> 0
                Red -> 1
                Green -> 2
                Yellow -> 3
                Blue -> 4
                Magenta -> 5
                Cyan -> 6
                White -> 7

    def intensityCode = match intensity
        None -> 0
        Some i ->
            16
            * match i
                Dim -> 1
                Bright -> 2

    def _ = set name (colourCode + intensityCode)

    LogLevel name
```

I  don't mind it, but you mentioned it as being something you didn't like back on the OG bikeshed  PR

share/wake/lib/core/json.wake
JakeSiFive: I didn't know about that function but that's dedication to correct error messages and exactly the kind of stuff that makes i18n a giant fucking pain hahaha.

share/wake/lib/core/print.wake
JakeSiFive: Well that makes me want to puke personally so I'm glad you changed it. I don't think this pattern is that common but we should probably fix this after MVP. I think `8 + match c` shouldn't require indent

share/wake/lib/system/job.wake
```diff
@@ -241,7 +244,7 @@ export def localRunner: Runner =
       Fail e
     Pass (RunnerInput _ cmd vis env dir stdin _ _ predict isatty) =
         def Usage status runtime cputime mem in out = predict
-        def _ = launch job dir stdin env.implode cmd.implode status runtime cputime mem in out (if isatty then 1 else 0)
```
JakeSiFive, line 244: Does this not count as a "simple" if statement? Or is it just too long?

vendor/git.wake
```diff
@@ -67,12 +68,12 @@ def buildOn Unit = match (subscribe releaseOn)
...
-    #                            ^^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^  ^^^^
-    #                            |                          |        |
-    #                            |                          |        | The "rest", everything after the '-' after the commit number
-    #                            |                          | The number of commits since the tagged version, surrounded by '-'
     #                            | Matches the version tag, which should be 3 numbers separated by '.' with an optional leading 'v'
+    #                            |                          | The number of commits since the tagged version, surrounded by '-'
+    #                            |                          |        | The "rest", everything after the '-' after the commit number
+    #                            |                          |        |
+    #                            vvvvvvvvvvvvvvvvvvvvvvvvvv vvvvvvvv vvvv
+    def unsafe_version_regex = `^(v?[0-9]+\.[0-9]+\.[0-9]+)-([0-9]+)-(.*)`
```
JakeSiFive, line 76: This is dedication!

JakeSiFive (approved):
This PR made me giggle multiple times!

share/wake/lib/system/job.wake
V-FEXrt: So  the  problem  is that the rest of the line is too long so `explode` gets set which overrides the "simple" check. And we have to respect `explode` inside of if because `if/else if / else` needs it to be consistent (otherwise one part would explode while the rest is flat)

JakeSiFive (approved):

share/wake/lib/core/json.wake
ag-eitilt: You wouldn't know about it because I wrote that function specifically for that message...

PR #1221, opened 2023-04-21
format: always newline top level publish/target

V-FEXrt:
Since `target` and  `publish` can only be top level (per the  grammar) we just always newline after the `=`


JakeSiFive (approved):

ag-eitilt (approved):
Much more consistent, thanks!

PR #1222, opened 2023-04-21
format: | and $ shouldn't force explode in if/else

V-FEXrt:
`|` and `$` inside of an `if`/`else` condition shouldn't be considered a "root" binary operation otherwise it formats as

```
def x =
    if cmpFn
    | isGT then
        0
    else
        0
```

This change fixes that and also adds a `ChangeContextAction` instead of a new `binop` formatter function that way `binop` is  easier to delete in the future

### DIFF
--- a/tools/wake-format/actions.h
+++ b/tools/wake-format/actions.h
@@ -191,6 +191,19 @@ struct NestAction {
   }
 };
 
+template <class F, class FMT>
+struct ChangeContextAction {
+  F f;  // (ctx_t) -> ctx_t
+  FMT formatter;
+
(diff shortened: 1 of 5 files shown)


ag-eitilt (approved):
Good catch, thanks!

tools/wake-format/actions.h
```diff
@@ -191,6 +191,19 @@ struct NestAction {
   }
 };
 
+template <class F, class FMT>
+struct ChangeContextAction {
```
JakeSiFive, line 195: Neat!

JakeSiFive (approved):

PR #1225, opened 2023-04-25
format: fix binops with comments

V-FEXrt:


tools/wake-format/emitter.cpp
```diff
@@ -167,8 +167,9 @@ static size_t count_leading_newlines(const token_traits_map_t& traits, const CST
 }
 
 static size_t count_trailing_newlines(const token_traits_map_t& traits, const CSTElement& node) {
-  CSTElement token;
-  {
+  CSTElement token = node;
+
+  if (node.isNode()) {
```
JakeSiFive, line 172: What change is being made here?

JakeSiFive (approved):

tools/wake-format/emitter.cpp
V-FEXrt: The function didn't work on tokens (only nodes) for no specific reason. The change makes it so 
 
- if it's a token just get the answer
- if its a node, find the correct token, then get the answer

PR #1228, opened 2023-04-28
format: Fix various broken rules

V-FEXrt:
Certain repos were crashing the formatter or emitting invalid code on format. This change fixes (nearly) all the breakages.

Broken rules:

1) Missing a space in an ascribe caused an assert
2) Really long require patterns would explode (as expected) but if the pattern is a binop (like `require a, b, Nil = blah`) then exploding isn't legal syntax. Explode is now disable in requires for binops. In the same vein, target patterns are prevented from exploding
3) Multiline strings without indentation caused an assert because it was always assumed that a TOKEN_WS would exists. This case now probably indents
4) An overly restrictive assertion that CST_BLOCKs only contain CST_DEF was removed as they may also contain imports

There is a subtle but with `"% ... %"` strings that I haven't figured out yet. It seems that a `"` inside the string is parsed as `TOKEN_WS` which is incorrect and breaks the formatter. I'm not going to account for that with a formatter rule though. For now I'm just hand rewriting those specific cases to `""" ... """"` strings since they are identical minus this one bug. (I honestly think its just time to pull the plug on `"%%"` strings TBH

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -515,9 +515,10 @@ auto Emitter::pattern_fmt(cst_id_t stop_at) {
   //    isn't a good way to determine if it fits since the NL can be in many places. Instead
   //    just return flat, accepting in some very rare cases the output will exceed the max width
 
-  return fmt().fmt_try_else(
-      [](const wcl::doc_builder& builder, ctx_t ctx, wcl::doc doc) { return doc->has_newline(); },
-      all_flat, fmt().fmt_if_fits_all(all_flat, all_flat /*all_explode*/));
+  return all_flat;
+  // return fmt().fmt_try_else(
+  //     [](const wcl::doc_builder& builder, ctx_t ctx, wcl::doc doc) { return doc->has_newline();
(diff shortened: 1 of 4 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -1869,18 +1870,23 @@ wcl::doc Emitter::walk_require(ctx_t ctx, CSTElement node) {
...
-                          .join(rhs_fmt())
-                          .consume_wsnlc()
-                          .fmt_if(TOKEN_KW_ELSE, else_fmt);
+  auto pre_body_fmt =
+      fmt()
+          .freshline()
+          .token(TOKEN_KW_REQUIRE)
+          .ws()
+          .fmt_if_else(CST_BINARY,
+                       // Binops must not explode inside of a require unwrap.
```
JakeSiFive, line 1879: unwrap? Did you mean Pattern?

JakeSiFive (approved):

tools/wake-format/emitter.cpp
V-FEXrt: yeah, sorry

PR #1231, opened 2023-05-01
format: Don't force match to newline

V-FEXrt:


JakeSiFive (approved):
I'm a very big fan of this.

ag-eitilt (approved):
I am slightly less big a fan of this, but it's still very nice.

PR #1232, opened 2023-05-01
format: append a space to leading unary not

V-FEXrt:


ag-eitilt (dismissed):
That's better, thanks.  I do think Jake raised a good point about the stacking, though -- if we have both `-` and `~` as unary operators with different precedence (not quite sure what `~` represents, but it's an example), wouldn't `~(-(2))` theoretically be representable in some cases as `~ -2`?  And if so, does the current formatter guarantee it's not going to collapse into `~-2` which would be a separate unary operator `~-`?  (Heading out the door for an appointment, I won't see any response for a while.)

V-FEXrt:
> does the current formatter guarantee it's not going to collapse into ~-

nope, not at all. We probably need a deeper implementation here. Was throwing up for exactly this kind of convo :)

JakeSiFive:
Precedence won't be an issue since we maintain the user's parens. But I think stacking is a real issue

tools/wake-format/emitter.cpp
```diff
@@ -2060,13 +2060,25 @@ wcl::doc Emitter::walk_type(ctx_t ctx, CSTElement node) {
   //              .concat(ctx));
 }
 
+wcl::doc Emitter::walk_unary_op(ctx_t ctx, CSTElement node) {
+  MEMO(ctx, node);
+  FMT_ASSERT(node.id() == CST_OP, node, "Expected CST_OP");
+
+  MEMO_RET(
+      fmt()
+          .fmt_if_else(TOKEN_OP_INEQUAL, fmt().walk(WALK_TOKEN).space(), fmt().walk(WALK_TOKEN))
```
JakeSiFive, line 2069: The condition should check if the child token is a primitive expression or not.

Also this would apply to any unary operator in the same precedence bracket as `!` I think

ag-eitilt: That does seem like the better way of doing it.  `!True` reads well, even if it's completely worthless semantically.

Now that I think more about it, I don't really like that this is making an exception for `!` (and potentially the related class of operators).  We should come up with something that works for unary operators across the board, and "primitive (reusing simple?) expression" seems like a good division.

V-FEXrt: Why do we need to check if the child token is a primitive?

> Also this would apply to any unary operator in the same precedence bracket as ! I think

on it :)

V-FEXrt: > We should add a similar rule for unary post fix operators (e.g. unary left associative operators)
That breaks unary `,` no?

`def x = 5 ,` instead of `def x = 5,`?

ag-eitilt: Yeah, but `def xs = (2, fib 3,)` isn't.  Not quite what the best way to solve that is, actually.

V-FEXrt: also, looking at the quickref, it seems like `,` is the _only_ unary postfix operator

JakeSiFive: So you have two categories of operators

Left associative:
  Dot, Quantifier, MulDiv, AddSub, Test, And, Or, Comma (the docs say that comma is right associative but a single comma is left associative)

Right associative:
  Exponent, InEqual, Dollar, Ascription, Assignment, Comma (but not the unary case)

Their category determines if you render it before or after for a unary operator (you don't have to think about this with binary operators). Both cases need to check if the underlying expression is primitive or not to determine if it can avoid the space.

JakeSiFive: Ah sorry about that last comment. Something strange happened on my end.

JakeSiFive: The other annoyance is DOT expressions

```
!myTuple.myField
```

would be better than 

```
! myTuple.myField
```

We might say that application and DOT are also allowed as well as primitive expressions because there are no operators (and thus no unary operators) that are more tightly binding. This would solve most of the list ending issues but would not solve this case:

```
x, y, 3 + 2,
```

would still add the space if we don't special case `,`

JakeSiFive: Ashley and I spoke offline, turns out the docs are a bit off (which you can figure out if you think of more examples like 5 - 6 or -5). Each category has a separate associativity for unary and binary.

JakeSiFive (approved):
Awesome!

tools/wake-format/emitter.cpp
```diff
@@ -78,6 +78,27 @@ static inline bool is_expression(cst_id_t type) {
          type == CST_INTERPOLATE || type == CST_MATCH || type == CST_REQUIRE || type == CST_PRIM;
 }
 
+static inline bool is_primary_term(wcl::doc_builder& builder, ctx_t ctx, CSTElement& node,
+                                   const token_traits_map_t& traits) {
```
ag-eitilt, line 82: How does this relate to the simple-expression determination we were using for... `require` was it?  I know there was something which we were at least talking about allowing single lines for in some cases.  If that's a similar domain, it feels like it might be easier for people to reason about what format to use themselves if we use a consistent trigger for the exceptions -- is there anything critical there which we'd lose by switching to use `is_primary_term` across the board?

tests/wake-format/basic/stdout
```diff
@@ -635,9 +635,9 @@ def tarball Unit =
 
     def srcs =
         allSources
-        | filter (! matches `(debian/|wake.spec).*` _.getPathName) # omit packaging files
-        | filter (! matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
-        | filter (! matches `(.*/)*\.gitignore` _.getPathName) # omit git files
+        | filter (!matches `(debian/|wake.spec).*` _.getPathName) # omit packaging files
+        | filter (!matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
+        | filter (!matches `(.*/)*\.gitignore` _.getPathName) # omit git files
```
ag-eitilt: Looks like this implementation isn't actually solving the original problem?  Or is this considered a primary expression?

tests/wake-format/basic/stdout
```diff
@@ -1470,3 +1470,20 @@ def x =
     d,
     # Comment 4
     e,
+
+def y =
+    def x = | & ! ≤ - * √2, ,
+    def x = -2, , ,
+    def x = (1 + 2),
+    def x = -(1 + 2)
+    def x = -(1 + 2),
```
ag-eitilt, line 1479: This all looks good, though!

tests/wake-format/basic/stdout
ag-eitilt: This is apparently marked as "outdated" because the *first* commit is being reverted, not because the second is being overridden.

tools/wake-format/emitter.cpp
JakeSiFive: These are different. This is a syntactic category from the grammar. The "simple" category is more or less "single token" and has no exact analog in the grammar.

You can think of this category as the category of expressions which are clearly on inspection higher precedence than any prefix operator.

JakeSiFive (approved):

PR #1237, opened 2023-05-08
Remove macos from our CI

JakeSiFive:
The github mac builders are no longer working. It's time to remove it and drop macos support. At this time we don't plan on actively removing mac support but it isn't going to be tested for now.


V-FEXrt (approved):

ag-eitilt:
We probably want to make a note of the untested status in the readme.

PR #1238, opened 2023-05-08
Refine log level changes to maintain use cases

JakeSiFive:
This change is needed in order enable some use cases we forgot. The most important one is allowing wake to act as a build server via bsp-wake.

### DIFF
--- a/share/wake/lib/core/print.wake
+++ b/share/wake/lib/core/print.wake
@@ -30,8 +30,7 @@ export def format (anyType: a): String =
 
     p anyType
 
-# LogLevel is a user-opaque type.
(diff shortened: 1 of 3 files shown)


share/wake/lib/core/print.wake
```diff
@@ -32,11 +32,9 @@ export def format (anyType: a): String =
 
 # LogLevel is a user-opaque type.
 # To construct a LogLevel, use makeLogLevel.
```
V-FEXrt, line 34: Please update this comment

V-FEXrt (approved):

JakeSiFive:
ok Jack confirmed this works. Going to merge

PR #1245, opened 2023-05-11
First pass at batched file hashing in wake

JakeSiFive:
This optimizes wake's hashing in two ways

1) We now batch the hashing of multiple files so we have less process overhead
2) We still do things in parallel but we attempt to limit the number of threads to reduce thread overhead

TODO: For single file hashes this should in theory be worse since it starts up an OS thread. We should optimize that case

### DIFF
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -130,11 +130,39 @@ def dirHash =
 export def getPathParent (path: Path): Path =
     Path (simplify "{getPathName path}/..") dirHash
 
-def addhash (f: String): String =
-    def p = simplify f
+def computeHashes (files: List String): List String =
+    # Lots of jobs have no output
+    require _, _ = files
+    else Nil
+
+    def hashPlan cmd =
+        Plan "" cmd Nil Nil "." "" logNever logError logDebug ReRun True Nil (\_ True) hashUsage id id False
+
     def add f h = prim "add_hash"
 
-    add p (hashcode p)
+    require Pass stdout =
+        hashPlan ("lib/wake/wake-hash", files)
+        | setPlanLabel "batch-hash"
+        | runJobWith localRunner
+        | getJobStdout
+    else
+        def _ = map (\f add f "BadHash")
+        files
+
+    def _ =
+        stdout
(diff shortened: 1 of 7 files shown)


share/wake/lib/system/path.wake
```diff
@@ -130,11 +130,39 @@ def dirHash =
...
+        | zip files
+        | map (\(Pair file hash) add file hash)
+
+    files
+
+#def addhash (f: String): String =
+#    def p = simplify f
+#    def add f h = prim "add_hash"
+
+#    add p (hashcode p)
```
mmjconolly, line 165:
```suggestion
```

JakeSiFive:
Another thing that I need to add to this is a call to `get_hash` because otherwise we can't use the modtime optimization and would block people from using add_hash on their own for special cases.

jackkoenig:
For my main use case of groups of often thousands of files, this is a fantastic speedup!

JakeSiFive:
TODO:
1) We want to benchmark this on a more diverse sample size to make sure we're not slowing down other use cases too much
2) If other benchmarks are slow we should avoid starting up a thread at all if we're only going to start one thread anyway
3) If benchmarks are still slow we should consider adding this to shim wake (then there should be zero performance degradation between those two optimizations)

JakeSiFive:
From my bench-marking there is no determinable increase or decrease on our current typical flows so I say we land this.

tools/wake-hash/hash.wake
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 SiFive, Inc.
```
V-FEXrt, line 1: nit: 2023

V-FEXrt (approved):

mmjconolly:
Say if the hash fails, will it appear in `--failed`? As while it probably shouldn't appear in `--last` the failure could be helpful for people trying to understand their build failures

JakeSiFive:
> Say if the hash fails, will it appear in `--failed`? As while it probably shouldn't appear in `--last` the failure case could be helpful for people trying to understand their build failures

It will still appear. --failed doesn't filter noisy jobs at all.

PR #1250, opened 2023-05-15
Add basic fuzz testing for the shared cache

JakeSiFive:
This change adds fuzz testing for the shared cache and fixes a couple bugs that it found!

### DIFF
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -92,6 +92,7 @@ struct LRUEvictionPolicyImpl {
   job_cache::PreparedStatement insert_last_use;
   job_cache::PreparedStatement set_last_use;
   job_cache::PreparedStatement get_last_use;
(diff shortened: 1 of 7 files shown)


src/job_cache/eviction_policy.cpp
```diff
@@ -170,19 +175,28 @@ struct LRUEvictionPolicyImpl {
...
+      if (result == SQLITE_ROW) return;
+
+      // If there wasn't however we need to insert the result
+      if (result == SQLITE_DONE) {
         insert_last_use.bind_integer(1, job_id);
         insert_last_use.bind_integer(2, tp.tv_sec);
         insert_last_use.step();
         insert_last_use.reset();
+      } else {
+        log_fatal("get_last_use result was unexpected: %d", result);
```
V-FEXrt, line 199: nit: can you invert and early exit?

V-FEXrt (approved):

PR #1256, opened 2023-05-17
Remove `job` and `shellJob` wrappers around `runJob`

### LINKED ISSUE #1254: remove 'job' function
This function should be removed, it's not advisable to use, it doesn't add a label

```
export def job (cmd: List String) (visible: List Path): Job =
    makeExecPlan cmd visible
    | runJob
```
(from https://github.com/sifive/wake/blob/master/share/wake/lib/system/job.wake#L1013)

This would be a breaking api change, but the function has very few users and they can easily just migrate to directly calling makePlan/runJob.

ag-eitilt:
As part of moving away from encouraging label-less and implicit-runner jobs, the old `job` and `shellJob` wrappers are unnecessary and even undesirable.

There are a lot of places in the code which reference `job` objects, but all of those are shadowed names rather than referring to this function.  At this point, the only use of `job`-the-function is in the wake-format tests, which we want to keep frozen anyway to test the previous format.

Fixes #1254

### DIFF
--- a/share/doc/wake/tutorial.md
+++ b/share/doc/wake/tutorial.md
@@ -549,8 +549,10 @@ order to allow their execution to be cached.
 export def infoH _args =
     def cmdline =
         which "uname", "-sr", Nil
+    def plan =
+        makePlan "get kernel" Nil cmdline
     def os =
-        job cmdline Nil
+        runJob plan
     def str =
         getWhenFail "" os.getJobStdout
     def body =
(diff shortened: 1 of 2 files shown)


share/doc/wake/tutorial.md
```diff
@@ -1167,7 +1171,7 @@ def curl url extension =
         url,
         Nil
     def curl =
-        job cmdline Nil
+        makePlan "download {url}" Nil cmdline
```
mmjconolly, line 1174: does this need a runJob?

mmjconolly (approved):

share/doc/wake/tutorial.md
ag-eitilt: Yes, it does...  There was some strangeness with these files getting listed twice at different paths in the sidebar, so I'm going to pretend that got lost as part of that "do I need to copy this?" confusion.

PR #1257, opened 2023-05-17
Use default values with missing wakeroot

### LINKED ISSUE #1240: Directory with wake.db should be treated as an implicit .wakeroot
Previously we've been able to inspect a `wake.db` in an empty directory, where the presence of a `wake.db` indicated that this was the root directory for the wake workspace
```
mkdir $other
cp $other/wake.db $other
cd $other
wake --failed
```
But now we see `Failed to load .wakeroot: Failed  to read '.wakeroot'`.

V-FEXrt:
Fixes #1240 

```
> ls
wake.db
> ../wake/bin/wake --config
[WARNING] Failed to read '.wakeroot' using defaults instead.

Wake config:
  user_config = '/home/ashley/.config/wake.json' (Default)
  version = '' (Default)
  log_header = '[$stream] $source: ' (Default)
  log_header_source_width = '25' (Default)
  label_filter = '.*' (Default)
  max_cache_size = '26843545600' (Default)
  low_cache_size = '16106127360' (Default)
  log_header_align = 'false' (Default)
> ../wake/bin/wake --failed
[WARNING] Failed to read '.wakeroot' using defaults instead.

# pkg-config: utf8proc(644)
$ /usr/bin/pkg-config --short-errors --modversion utf8proc
No package 'utf8proc' found
```

Updated:

```
shell> ls
wake.db


shell> ../wake/bin/wake --config
Failed to load .wakeroot: Failed to read '.wakeroot'. Using default values instead.
Wake config:
  user_config = '/home/ashley/.config/wake.json' (Default)
  version = '' (Default)
  log_header = '[$stream] $source: ' (Default)
  log_header_source_width = '25' (Default)
  label_filter = '.*' (Default)
  max_cache_size = '26843545600' (Default)
  low_cache_size = '16106127360' (Default)
  log_header_align = 'false' (Default)


shell> ../wake/bin/wake --failed
Failed to load .wakeroot: Failed to read '.wakeroot'. Using default values instead.
# pkg-config: utf8proc(644)
$ /usr/bin/pkg-config --short-errors --modversion utf8proc
No package 'utf8proc' found
```

### DIFF
--- a/src/runtime/config.cpp
+++ b/src/runtime/config.cpp
@@ -160,7 +160,7 @@ static wcl::result<JAST, std::pair<ReadJsonFileError, std::string>> read_json_fi
   std::ifstream file(path);
   if (!file) {
     return wcl::make_error<JAST, std::pair<ReadJsonFileError, std::string>>(
-        ReadJsonFileError::BadFile, "Failed  to read '" + path + "'");
+        ReadJsonFileError::BadFile, "Failed to read '" + path + "'");
   }
 
   std::stringstream buff;
@@ -297,14 +297,17 @@ bool WakeConfig::init(const std::string& wakeroot_path, const WakeConfigOverride
   // to our default constructor. Thus we are forced to use `new`
   _config = std::unique_ptr<WakeConfig>(new WakeConfig());
 
+  JAST wakeroot_json(JSON_OBJECT);
+
   // Parse .wakeroot
   auto wakeroot_res = read_json_file(wakeroot_path);
   if (!wakeroot_res) {
-    std::cerr << "Failed to load .wakeroot: " << wakeroot_res.error().second << std::endl;
-    return false;
+    std::cerr << "[WARNING] " << wakeroot_res.error().second << " using defaults instead."
+              << std::endl
+              << std::endl;
+  } else {
+    wakeroot_json = std::move(*wakeroot_res);
   }
-
-  JAST wakeroot_json = std::move(*wakeroot_res);
 
   // Check for disallowed keys
   {
--- a/src/runtime/config.cpp
+++ b/src/runtime/config.cpp
@@ -302,9 +302,15 @@ bool WakeConfig::init(const std::string& wakeroot_path, const WakeConfigOverride
   // Parse .wakeroot
   auto wakeroot_res = read_json_file(wakeroot_path);
   if (!wakeroot_res) {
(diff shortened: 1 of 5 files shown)


mmjconolly:
say if you run `wake --failed` without the .wakeroot, does it show the failures?

src/runtime/config.cpp
```diff
@@ -297,15 +297,18 @@ bool WakeConfig::init(const std::string& wakeroot_path, const WakeConfigOverride
...
 
+  JAST wakeroot_json(JSON_OBJECT);
+
   // Parse .wakeroot
   auto wakeroot_res = read_json_file(wakeroot_path);
   if (!wakeroot_res) {
-    std::cerr << "Failed to load .wakeroot: " << wakeroot_res.error().second << std::endl;
-    return false;
+    std::cerr << "[WARNING] " << wakeroot_res.error().second << " using defaults instead."
+              << std::endl
```
mmjconolly, line 306:
```suggestion
```

JakeSiFive (approved):

JakeSiFive:
hmmm...I think I've changed my mind on this. I think commands like --failed should avoid parsing the wake root instead to solve this.

JakeSiFive (approved):
Ok I had a slight misunderstanding here. It turns out that we look for either a wake.db or a .wakeroot file to set our current working directory. This means that constructing a default is equivalent in the database case to supporting the use case mentioned in the ticket.

Additionally its not like running this is going to give people an easy way to run wake (non-database commands) unless they manually create a database using something like wake --init or moving the database or something. So I think this change actually does what we all want more or less (slightly different from what I asked for, but I'm happy with it now that I understand more)

V-FEXrt:
Okay so it turns out that the new 'empy wakeroot' test is picking up another wakeroot by happenstance which is messing up the test. I believe this is some what  due to the fact that its wake launched from within wake which is a super unsupported flow but not 100% sure. 

For now I'm just removing the test because 1) I really want this to go in with the next release, its a core feature for our users that we broken and 2) I'm very confident in the change 3) It is a very small surface area, for a very specific usecase with minimal impact if this does introduce an issue in the future

JakeSiFive:
Yeah if you run a test in a test directory without, the wake within wake will pick up the wakeroot from the repo. I think if you call `wake --init` before making any other wake invocations you can construct a database. If that doesn't work or you can't make it work you can use `sqlite3 --line wake.db '.tables'` to create an empty wake.db that wake will for sure pick up.

PR #1261, opened 2023-05-22
Audit, defensivelly convert, and canonicalize all paths

JakeSiFive:
This change aggressively converts file paths to adhere a uniform set of rules: (with one exception that we'll go over in a bit)

1. All paths stored in the database are sandbox-absolute paths
2. All sandbox paths are sandbox-absolute paths everywhere in the code
3. When the server is reading a file from the disk, it always uses a client-absolute path
4. When the server is writing a file to the disk, it always uses a client-absolute path
5. In general when client paths are stored in a data structure, they're stored as client-absolute paths
6. We accept inputs to the server and client as both relative and absolute but we canonicalize all relative inputs to absolute paths immediately upon parsing them.
7. When sending a response back to the client, we use client-relative paths only

The way we implement 7 is also the exception. The MatchingJob object can either be a sandbox-absolute object or a client-relative object. We read its contents from the database at first which means that it will be sandbox-absolute but we then convert it to client-relative before sending it back to the client. Rule #7 ensures that wake has fewer means to ever find out its own absolute workspace path.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -362,7 +362,9 @@ target pwd Unit =
     | getJobStdout
     | rmap (replace `\n` "")
 
-export def mkJobCacheRunner ((Runner name score baseDoIt): Runner): Runner =
+# wakeroot is the absolute sandbox-path from which input and output files will
+# be interpreted as being relative to if they're in fact relative.
+export def mkJobCacheRunner (wakeroot: String) ((Runner name score baseDoIt): Runner): Runner =
     def virtual job stdout stderr status runtime cputime membytes ibytes obytes = prim "job_virtual"
     def badlaunch job error = prim "job_fail_launch"
     def job_cache_read str = prim "job_cache_read"
@@ -381,19 +383,11 @@ export def mkJobCacheRunner ((Runner name score baseDoIt): Runner): Runner =
 
             def jobCacheVisible = JArray (map mkVisJson vis)
 
-            # TODO: Allow this to be configurable
-            def workspace = "/workspace"
-
-            def abs_dir =
-                if matches `/.*` dir then
-                    dir
-                else
-                    "{workspace}/{dir}"
-
             def jobCacheJsonIn =
                 prettyJSON
                 $ JObject (
-                    "cwd" :-> JString abs_dir,
+                    "wakeroot" :-> JString wakeroot,
+                    "cwd" :-> JString dir,
                     "command_line" :-> JString cmd.implode,
                     "envrionment" :-> JString env.implode,
                     "stdin" :-> JString stdin,
@@ -498,7 +492,7 @@ export def mkJobCacheRunner ((Runner name score baseDoIt): Runner): Runner =
             def inputsTree = listToTree scmpCanonical inputs
 
             def mkOutputFileJson src =
-                JObject ("src" :-> JString src, "path" :-> JString "/workspace/{src}",)
+                JObject ("src" :-> JString src, "path" :-> JString "{workspace}/{src}",)
 
             def jobCacheOutputFiles = JArray (map mkOutputFileJson outputs)
 
@@ -513,7 +507,8 @@ export def mkJobCacheRunner ((Runner name score baseDoIt): Runner): Runner =
             def jobCacheAddJson =
                 prettyJSON
                 $ JObject (
-                    "cwd" :-> JString abs_dir,
+                    "wakeroot" :-> JString wakeroot,
+                    "cwd" :-> JString dir,
                     "command_line" :-> JString cmd.implode,
                     "envrionment" :-> JString env.implode,
                     "stdin" :-> JString stdin,
@@ -928,7 +923,14 @@ export def defaultRunner: Runner =
     require Some _ = getenv "WAKE_EXPERIMENTAL_JOB_CACHE"
     else fuseRunner
 
-    mkJobCacheRunner fuseRunner
+    # The fuseRunner does not actully mount over / and instead uses the
+    # the host root as the sandbox root. This means that wakeroot will
+    # change from dpending on where wake is being run from. As a hack
+    # to work around this we abuse the fact that the fuseRunner only
+    # works in relative paths to make different runs consistent. Ideally
+    # you'd have a more complex sandbox that kept the wakeroot at a
+    # consistent place across runs.
+    mkJobCacheRunner "/workspace" fuseRunner
 
 # A plan describing how to construct a JSONRunner
 # RawScript: the path to the script to run jobs with
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -393,7 +393,7 @@ export def mkJobCacheRunner (wakeroot: String) ((Runner name score baseDoIt): Ru
                     "stdin" :-> JString stdin,
                     "input_files" :-> jobCacheVisible,
                     "client_cwd" :-> JString client_cwd,
-                    "dir_redirects" :-> JObject (workspace :-> JString "./",),
+                    "dir_redirects" :-> JObject (wakeroot :-> JString "./",),
                 )
 
             require Pass cacheResult =
@@ -492,7 +492,7 @@ export def mkJobCacheRunner (wakeroot: String) ((Runner name score baseDoIt): Ru
             def inputsTree = listToTree scmpCanonical inputs
 
             def mkOutputFileJson src =
-                JObject ("src" :-> JString src, "path" :-> JString "{workspace}/{src}",)
+                JObject ("src" :-> JString src, "path" :-> JString "{wakeroot}/{src}",)
 
             def jobCacheOutputFiles = JArray (map mkOutputFileJson outputs)
 
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -925,7 +925,7 @@ export def defaultRunner: Runner =
 
     # The fuseRunner does not actully mount over / and instead uses the
     # the host root as the sandbox root. This means that wakeroot will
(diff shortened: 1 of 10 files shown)


share/wake/lib/system/job.wake
```diff
@@ -381,16 +381,25 @@ export def mkJobCacheRunner ((Runner name score baseDoIt): Runner): Runner =
...
                 $ JObject (
-                    "cwd" :-> JString dir,
+                    "cwd" :-> JString abs_dir,
                     "command_line" :-> JString cmd.implode,
                     "envrionment" :-> JString env.implode,
                     "stdin" :-> JString stdin,
                     "input_files" :-> jobCacheVisible,
                     "client_cwd" :-> JString client_cwd,
-                    "dir_redirects" :-> JObject ("/workspace" :-> JString "./",),
+                    "dir_redirects" :-> JObject (workspace :-> JString "./",),
```
mmjconolly, line 402: `workspace` is already an absolute path

src/job_cache/daemon_cache.cpp
```diff
@@ -484,10 +484,11 @@ class SelectMatchingJobs {
 
       // Ok this is the job, it matches *exactly* so we should
       // expect running it to produce exaxtly the same result.
+      // TODO, clean this shit up?
```
V-FEXrt, line 487: Can you elaborate on how you'd like this cleaned up?

src/job_cache/daemon_cache.cpp
```diff
@@ -763,10 +773,7 @@ FindJobResponse DaemonCache::read(const FindJobRequest &find_request) {
     log_warning("Failed to send eviction update: %s", strerror(errno));
   }
 
-  // TODO: We should really return a different thing here
-  //       that mentions the *output* locations but for
-  //       now this is good enough and we can assume
-  //       workspace relative paths everywhere.
+  // The response we give back sends client-absolute paths but we'll
```
V-FEXrt, line 776: but we'll ....?

src/job_cache/types.cpp
```diff
@@ -107,62 +107,112 @@ JAST JobOutputInfo::to_json() const {
...
 }
 
-MatchingJob::MatchingJob(const JAST &json) {
+MatchingJob::MatchingJob(const std::string &client_cwd, const JAST &json) {
   output_info = JobOutputInfo(json.get("output_info"));
 
   for (const auto &output_file_json : json.get("output_files").children) {
-    output_files.push_back(CachedOutputFile(output_file_json.second));
+    CachedOutputFile output_file(output_file_json.second);
+    // Canoncolize matching jobs to use client-relative paths.
```
V-FEXrt, line 115: sp: Canonicalize

src/job_cache/types.cpp
```diff
@@ -107,62 +107,112 @@ JAST JobOutputInfo::to_json() const {
...
+    if (wcl::is_absolute(output_file.path)) {
+      output_file.path = wcl::relative_to(client_cwd, output_file.path);
+    }
+    output_files.push_back(output_file);
   }
 
   for (const auto &output_dir_json : json.get("output_dirs").children) {
-    output_dirs.push_back(CachedOutputDir(output_dir_json.second));
+    CachedOutputDir output_dir(output_dir_json.second);
+    // Canoncolize matching jobs to use client-relative paths.
```
V-FEXrt, line 124: sp: Canonicalize

src/job_cache/types.cpp
```diff
@@ -107,62 +107,112 @@ JAST JobOutputInfo::to_json() const {
...
+    if (wcl::is_absolute(output_dir.path)) {
+      output_dir.path = wcl::relative_to(client_cwd, output_dir.path);
+    }
+    output_dirs.push_back(output_dir);
   }
 
   for (const auto &output_symlink_json : json.get("output_symlinks").children) {
-    output_symlinks.push_back(CachedOutputSymlink(output_symlink_json.second));
+    CachedOutputSymlink output_sym(output_symlink_json.second);
+    // Canoncolize matching jobs to use client-relative paths.
```
V-FEXrt, line 133: sp: Canonicalize

src/job_cache/types.cpp
```diff
@@ -107,62 +107,112 @@ JAST JobOutputInfo::to_json() const {
...
+    if (wcl::is_absolute(output_sym.path)) {
+      output_sym.path = wcl::relative_to(client_cwd, output_sym.path);
+    }
+    output_symlinks.push_back(output_sym);
   }
 
   for (const auto &input_file_json : json.get("input_files").children) {
-    input_files.push_back(input_file_json.second.value);
+    // Canoncolize matching jobs to use client-relative paths.
```
V-FEXrt, line 141: sp: Canonicalize

src/job_cache/types.cpp
```diff
@@ -107,62 +107,112 @@ JAST JobOutputInfo::to_json() const {
...
+    std::string input_path = input_file_json.second.value;
+    if (wcl::is_absolute(input_path)) {
+      input_path = wcl::relative_to(client_cwd, input_path);
+    }
+    input_files.push_back(input_path);
   }
 
   for (const auto &input_dir_json : json.get("input_dirs").children) {
-    input_dirs.push_back(input_dir_json.second.value);
+    // Canoncolize matching jobs to use client-relative paths.
```
V-FEXrt, line 150: sp: Canonicalize

V-FEXrt: lol  this is in here a bunch, i'll stop making new comments but it should be fixed everywhere

src/job_cache/types.cpp
```diff
@@ -490,16 +628,16 @@ JAST FindJobRequest::to_json() const {
...
   if (found.kind != JSON_TRUE) {
     match = {};
     return;
   }
-  match = wcl::make_some<MatchingJob>(json.get("match"));
+  match = wcl::make_some<MatchingJob>(client_cwd, json.get("match"));
 }
 
-JAST FindJobResponse::to_json() const {
+JAST FindJobResponse::to_json(const std::string &client_cwd) const {
```
V-FEXrt, line 640: Can `client_cwd` here be stored in the constructor since it's provided there?

src/job_cache/types.h
```diff
@@ -82,8 +82,8 @@ struct MatchingJob {
   JobOutputInfo output_info;
 
   MatchingJob() = default;
-  explicit MatchingJob(const JAST &json);
-  JAST to_json() const;
+  MatchingJob(const std::string &client_cwd, const JAST &json);
```
V-FEXrt, line 85: Same thought here: re storing in the constructor for use in to_json

V-FEXrt (approved):

share/wake/lib/system/job.wake
```diff
@@ -919,7 +923,14 @@ export def defaultRunner: Runner =
     require Some _ = getenv "WAKE_EXPERIMENTAL_JOB_CACHE"
     else fuseRunner
 
-    mkJobCacheRunner fuseRunner
+    # The fuseRunner does not actully mount over / and instead uses the
+    # the host root as the sandbox root. This means that wakeroot will
+    # change from dpending on where wake is being run from. As a hack
```
V-FEXrt, line 928: typo: dpending

src/job_cache/types.cpp
JakeSiFive: Ah that was also something I realized was wrong mid-writing. Comment removed!

src/job_cache/types.cpp
JakeSiFive: oh good question, let me think about that.

JakeSiFive: yep that worked. I was trying to avoid adding info that didn't need to be there but it doesn't really matter too much because we're just sending back the thing that the user told us so they can ignore it if they want. It might be useful for debugging in the future too.

share/wake/lib/system/job.wake
```diff
@@ -919,7 +923,14 @@ export def defaultRunner: Runner =
...
 
-    mkJobCacheRunner fuseRunner
+    # The fuseRunner does not actully mount over / and instead uses the
+    # the host root as the sandbox root. This means that wakeroot will
+    # change from depending on where wake is being run from. As a hack
+    # to work around this we abuse the fact that the fuseRunner only
+    # works in relative paths to make different runs consistent. Ideally
+    # you'd have a more complex sandbox that kept the wakeroot at a
+    # consistent place across runs.
+    mkJobCacheRunner "/workspace" fuseRunner
```
JakeSiFive, line 933: @mmjconolly I think after our discussion this is perhaps the main source of confusion since 1) I had this wrong and only working accidentally before but 2) I think it by happen stance works out to be our best option still. With the edits made in this change and this comment added for the one remaining super confusing issue. Does this make more sense?

share/wake/lib/system/job.wake
JakeSiFive: Ok I figured out what you were talking about here because of a bug I had to fix. I had shadowed the magic workspace variable to be something else not this. After I got rid of that my code still compiled! That path was previously a sandbox path not as I think you understood it a client path.

JakeSiFive:
@mmjconolly Are you ok with landing this for now? Ashley and I wanted to do a patch release with a few of the recent changes we've made in it and it would be nice to get this in as well.

mmjconolly:
> @mmjconolly Are you ok with landing this for now? Ashley and I wanted to do a patch release with a few of the recent changes we've made in it and it would be nice to get this in as well.

Sure

JakeSiFive:
We'll just make it version 34 I think. There is no specific desire to make it a patch release specifically, just to get a few small updates out the door.

V-FEXrt:
Well I really wanted it to be 33.0.1 for aesthic purposes, but we'll just do 34 instead

PR #1265, opened 2023-05-24
Make stdin a psuedoterminal when isatty is set

JakeSiFive:
Sometimes programs check isatty(STDIN_FILENO) in addition to stderr and stdout to determine if they're using a terminal or not. We should support that case as well as we can. We'd need to perform some surgery on shim-wake to support the case where stdin is a file input but it *ALSO* has input. So for now we treat setting stdin to a file to be just that and not a pseudo-terminal.

### DIFF
--- a/src/runtime/job.cpp
+++ b/src/runtime/job.cpp
@@ -835,7 +835,7 @@ static void launch(JobTable *jobtable) {
       stdin_file_str = task.stdin_file;
     } else if (task.is_atty) {
       create_psuedoterminal(stdin_stream);
-      stdin_file_str = "#" + std::to_string(stdin_stream[0]);
+      stdin_file_str = "#" + std::to_string(stdin_stream[1]);
     }
     std::stringstream prelude;
     prelude << find_execpath() << "/../lib/wake/shim-wake" << '\0'
--- a/src/runtime/job.cpp
+++ b/src/runtime/job.cpp
@@ -829,9 +829,17 @@ static void launch(JobTable *jobtable) {
     jobtable->imp->pipes[stdout_stream[0]] = entry;
     jobtable->imp->pipes[stderr_stream[0]] = entry;
     clock_gettime(CLOCK_REALTIME, &entry->job->start);
+    std::string stdin_file_str = "/dev/null";
+    int stdin_stream[2];
+    if (!task.stdin_file.empty()) {
+      stdin_file_str = task.stdin_file;
+    } else if (task.is_atty) {
+      create_psuedoterminal(stdin_stream);
+      stdin_file_str = "#" + std::to_string(stdin_stream[0]);
+    }
     std::stringstream prelude;
     prelude << find_execpath() << "/../lib/wake/shim-wake" << '\0'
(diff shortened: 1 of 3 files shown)


tools/shim-wake/main.c
```diff
@@ -147,10 +147,17 @@ int main(int argc, char **argv) {
     return 127;
   }
 
-  stdin_fd = open(argv[1], O_RDONLY);
-  if (stdin_fd == -1) {
-    fprintf(stderr, "open: %s: %s\n", argv[1], strerror(errno));
-    return 127;
+  // Because we want to be able to read in both file descirptors and
```
V-FEXrt, line 150: typo: descirptors

tools/shim-wake/main.c
```diff
@@ -147,10 +147,17 @@ int main(int argc, char **argv) {
...
+  // files for stdin, we use a '#' to denote a file descriptor.
+  const char* stdin_file = argv[1];
+  if (stdin_file[0] == '#') {
+    stdin_fd = atoi(stdin_file + 1);
+  } else {
+    stdin_fd = open(stdin_file, O_RDONLY);
+    if (stdin_fd == -1) {
+      fprintf(stderr, "open: %s: %s\n", argv[1], strerror(errno));
+      return 127;
```
V-FEXrt, line 159: Please move this `if` after the `if/else` in case `atoi` fails for some reason

src/runtime/job.cpp
```diff
@@ -829,9 +829,17 @@ static void launch(JobTable *jobtable) {
     jobtable->imp->pipes[stdout_stream[0]] = entry;
     jobtable->imp->pipes[stderr_stream[0]] = entry;
     clock_gettime(CLOCK_REALTIME, &entry->job->start);
+    std::string stdin_file_str = "/dev/null";
+    int stdin_stream[2];
```
V-FEXrt, line 833: I'm not going to block, but I really think this should be explicit via a tuple in wakelang.

V-FEXrt (approved):

src/runtime/job.cpp
JakeSiFive: What do you mean exactly?

V-FEXrt: Mentioned in chat but something like 

```
data StdInOption =
  DevNull
  Interactive
  File (filename)


def job = 
  makeExecPlan ......
  | setPlanStdinOption DevNull
  | runJob
```

tools/shim-wake/main.c
JakeSiFive: atoi does not tell you if it fails so we can't catch the error :(

V-FEXrt: I just meant something like 

```
if (stdin_file[0] == '#') {
    stdin_fd = atoi(stdin_file + 1);
  } else {
    stdin_fd = open(stdin_file, O_RDONLY);
}

if (stdin_fd == -1) {
      fprintf(stderr, "open: %s: %s\n", argv[1], strerror(errno));
      return 127;
}
```

PR #1269, opened 2023-05-30
Fix failure to handle large messages

JakeSiFive:
This change does a couple things

1) It makes communication between the daemon and client blocking. It wasn't blocking before because we based this code off of some code that I wrote ages ago that didn't use epoll at first and instead would just busy read from each client until it could get something. I switched to epoll forever ago in that code but the non-blocking accept stayed with it for some reason. This doesn't present an issue unless you fill up the kernel's buffer for the socket which never happens on small messages.
2) The code on both sides is now hardened so that in theory its compatible with non-blocking sockets. Hopefully we don't have to rely on that code as it essentially busy waits until it gets a read which is not great. I'm not sure if Linux will reschedule on blocking read with a non-blocking socket or not. So it could be anywhere from a total waste to negligible since the reads and writes are fairly tightly linked a single reschedule should clear everything up...either way we should avoid it but handle the error correctly so that we don't have to think about this sort of issue in code abstracted from the type of file descriptor being used.

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -101,8 +101,7 @@ static void create_file(const char *tmp_path, const char *final_path, const char
   }
 }
 
-// Create a *blocking* domain socket. We intend on using
-// epoll later so we don't need anything to be non-blocking.
+// Create a *blocking* domain socket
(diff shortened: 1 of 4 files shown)


mmjconolly (approved):
rubber stamp

PR #1272, opened 2023-05-30
Always ship wake with debug information

JakeSiFive:
There's no real reason not to have debug information on every type of build. It increases binary size drastically (~20x in the case of wake) but Wake is only ~2.5MB so increasing to ~50MB really just isn't that big of a deal and it saves us a huge amount of headache debugging things sometimes.

Size increases:
wake: 2.53MB -> 49.3MB
wakebox: 0.275MB -> 5.35MB
wake-format: ? -> 40.4MB
fuse-waked: 0.235MB -> 3.37MB
lsp-wake: ? -> 26.6MB
shim-wake: 32.1KB -> 0.159MB
wake-hash: 0.226MB -> 0.803MB
bsp-wake: ? -> 4.59MB

So the total package size will be dominated by the binaries and will be around ~135MB which is still a very acceptable size IMO. If need be there are settings that we can play with to reduce debug information size. We can also split debug information from these binaries and ship it separately if need be. Luckily this kind of thing used to be my day job! That all said, its by far the easiest thing to just keep them colocated which is what this change does.

### DIFF
--- a/share/wake/lib/gcc_wake/gcc.wake
+++ b/share/wake/lib/gcc_wake/gcc.wake
@@ -31,16 +31,16 @@ export def debugLFlags =
     ("-lpthread", "-g", "-pg", Nil)
 
 export def releaseCFlags =
-    ("-Wall", "-Wno-format-security", "-O2", Nil)
+    ("-Wall", "-Wno-format-security", "-g", "-O2", Nil)
 
 export def releaseLFlags =
     ("-lpthread",)
 
 export def staticCFlags =
-    ("-Wall", "-Wno-format-security", "-O2", "-flto", Nil)
+    ("-Wall", "-Wno-format-security", "-g", "-O2", "-flto", Nil)
 
 export def staticLFlags =
-    ("-lpthread", "-flto", "-static", "-s", Nil)
+    ("-lpthread", "-flto", "-static", "-g", Nil)
 
 target pwd _ =
     makeExecPlan ("pwd",) Nil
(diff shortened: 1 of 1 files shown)


mmjconolly:
Do we have something to catch the segfault and print a stack trace?

mmjconolly (approved):

JakeSiFive:
> Do we have something to catch the segfault and print a stack trace?

No but I can totally add that. One sec.

JakeSiFive:
> Do we have something to catch the segfault and print a stack trace?
After I thought about this, I think its actually somewhat tricky but maybe there's some linux/gdb magic to get around this issue that I'm not aware of. The issue is that if you catch the signal in order to automatically print a backtrace, you wind up handling the error. Now you *could* swap back to the default handler and re-emit the same signal to your own process but I'm not exactly sure at this time what that would do to debug-ability of the core dump. Let me look into this for a bit. I don't want to give up the improved debug-ability of the proper core dump just to get a backtrace even if that's all I need 95% of the time.

mmjconolly:
Sure whatever works, just having the symbols is great if we can get the core files from people when theres an issue

ag-eitilt:
> Wake is only ~2.5MB so increasing to ~50MB really just isn't that big of a deal

Oh, the wonders of modern computing.

mmjconolly:
> 50MB

Nothing a double-height 3.5" drive plugged into as ISA sound card cant handle

PR #1273, opened 2023-05-30
remove Plan's LocalOnly and RunnerFilter

mmjconolly:
This removes 2 fields from `Plan`
* `LocalOnly`
* `RunnerFilter`

This is a breaking change.

Some jobs relied on using `LocalOnly`, they have been transitioned to using `runJobWith localRunner` and some were removed.

Documentation received some very light updates.

The `runJob` score function becomes unable to select the localRunner, as it's no longer published to the runner topic.
A user who still needs this could in theory decide to publish a wrapper of the localRunner to that topic, that looks at some otherwise opaque Resource strings to determine if it should be used instead of the fuseRunner.

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -75,9 +75,8 @@ export def installAs (dest: String) (file: Path): Result Path Error =
     makeExecPlan cmd inputs
     | setPlanLabel "installAs: {dest} {file.getPathName}"
     | setPlanEnvironment Nil
-    | setPlanLocalOnly True
     | setPlanFnOutputs foutputs
-    | runJob
+    | runJobWith localRunner
     | getJobOutput
 
 # Copy a file from one directory subtree to another
--- a/tests/tests.wake
+++ b/tests/tests.wake
@@ -139,9 +139,8 @@ export def runUnitTests _: Result Unit Error =
         | setPlanLabel "testing: {testName}"
         | setPlanStdout logNever
         | setPlanStderr logNever
-        | setPlanLocalOnly True
         | setPlanShare False
-        | runJob
+        | runJobWith localRunner
 
     def removeCarriageReturns = replace `\r` ""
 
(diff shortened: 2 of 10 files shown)


tests/tests.wake
```diff
@@ -219,12 +218,11 @@ def runTest (testScript: Path): Result Unit Error =
         makeExecPlan ("./{testScript.getPathName.inTestDir}", wakeDir.inTestDir, Nil) (visibleFiles ++ wakeVisible)
         | setPlanDirectory testDirectory
         | setPlanLabel "testing: {testName}"
-        | setPlanLocalOnly True # On OS/X you cannot mount fuse within fuse
         | setPlanEnvironment ("HOME={testDirectory}",)
         | setPlanStdout logNever
         | setPlanStderr logNever
         | setPlanShare False
-        | runJob
+        | runJobWith localRunner # On OS/X you cannot mount fuse within fuse
```
ag-eitilt, line 225: This might not even need the `localRunner` if we're committing to actively not supporting Mac.

share/wake/lib/system/io.wake
```diff
@@ -75,9 +75,8 @@ export def installAs (dest: String) (file: Path): Result Path Error =
     makeExecPlan cmd inputs
     | setPlanLabel "installAs: {dest} {file.getPathName}"
     | setPlanEnvironment Nil
-    | setPlanLocalOnly True
     | setPlanFnOutputs foutputs
-    | runJob
+    | runJobWith localRunner
```
ag-eitilt, line 79: I'm sure I'm just missing something, but why does this have to be `localRunner`?  The `file` is already a Path so it shouldn't have any issues being sandboxed, and the parent `destDir` is created with `mkdir` and already in the visible list.

Side note, I think the `mkdir "{dest}/.."` might be buggy if anyone tries to `installAs` something directly into the workspace root?

ag-eitilt (approved):
Nice!  Noticed a couple places where I think we can sandbox a bit further, but that's probably a followup rather than things to add to this one.

share/wake/lib/system/io.wake
mmjconolly: I blindly converted this, but looking at it now I see what you mean though. I believe the rational was that some 'basic' level operations could be expected to behave well without the overhead of starting a sandbox, such as in this case calling `cp`.
Additionally it uses the `which` function which implies that we're expecting to resolve the path to `cp` from the host's `$PATH` instead of within a sandbox

tests/tests.wake
mmjconolly: Sure, I'll leave that for someone else in a 'remove macos' PR

JakeSiFive: "someone else"

...oh yeah....

PR #1274, opened 2023-05-31
Make batch hashing work for arbitrarily large numbers of files

JakeSiFive:
This change is to fix an issue where the new batch hashing cannot handle extremely large numbers of files. This change makes it so that we generate a file if the number of files to hash is large enough but use the command line otherwise. The file is unlinked as the last step if it was created.

Additionally while I was at it, I converted to a work stealing scheme instead of dividing things up "evenly". This should make much much better use of available cores and spread the total work load much more evenly. There are still some optimizations we could do to hashing like using async io to create a more generic work queue so that we can work steal multiple hash digests at a time off a single thread. I think if we do that we should probably instead implement the hashing-in-fuse optimization

### DIFF
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -130,8 +130,8 @@ def dirHash =
 export def getPathParent (path: Path): Path =
     Path (simplify "{getPathName path}/..") dirHash
 
-# Previouslly we were using ++ to concat the hashed and non-hashed files
-# but this would cause non-determinitic behavoir with the hash order being
+# Previously we were using ++ to concat the hashed and non-hashed files
+# but this would cause non-deterministic behavior with the hash order being
 # different on different runs depending on which files were touched. By
 # merging them this way we can maintain the existing order.
 def mergeSelect (existing_hashes: List Boolean) (hashed: List String) (not_to_hash: List String): List String =
@@ -212,7 +212,7 @@ def computeHashes (prefix: String) (files: List String): List String =
     require True = len hash_lines == to_hash_len
     else unreachable "wake-hash returned {format hash_lines} lines but we expected {str to_hash_len} lines"
 
-    # Finally actully add all the hashes
+    # Finally actually add all the hashes
     def hashed =
         hash_lines
         | zip to_hash
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -130,48 +130,101 @@ def dirHash =
 export def getPathParent (path: Path): Path =
     Path (simplify "{getPathName path}/..") dirHash
 
-def computeHashes (files: List String): List String =
+# Previouslly we were using ++ to concat the hashed and non-hashed files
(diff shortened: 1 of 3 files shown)


mmjconolly (approved):
this is a rubber stamp

share/wake/lib/system/path.wake
```diff
@@ -130,48 +130,101 @@ def dirHash =
 export def getPathParent (path: Path): Path =
     Path (simplify "{getPathName path}/..") dirHash
 
-def computeHashes (files: List String): List String =
+# Previouslly we were using ++ to concat the hashed and non-hashed files
```
V-FEXrt, line 133: typo: Previouslly

share/wake/lib/system/path.wake
```diff
@@ -130,48 +130,101 @@ def dirHash =
...
 
+    # These numbers don't appear to be well documented in any easily computeable fashion for Linux.
+    # The problem is that the same error is returned for multiple cases, and many hard to compute
+    # things cause us to go over. These numbers appear to be safe in practice (found exact via binary search).
+    # To account for any other variance I backed off these numbers quite a bit.
+    def use_file = to_hash_file_bytes >= 1900000 || to_hash_len >= 12000
     def add f h = prim "add_hash"
 
+    def hashPlan cmd vis =
+        Plan "<hash>" cmd vis Nil "." "" logNever logError logDebug ReRun Nil hashUsage id id False
```
V-FEXrt, line 182: This plan helper should add a label to the plan

share/wake/lib/system/path.wake
```diff
@@ -130,48 +130,101 @@ def dirHash =
...
+    # that are returned
+    def hash_lines =
         stdout
         | tokenize `\n`
+        | filter (_ !=* "")
+
+    require True = len hash_lines == to_hash_len
+    else unreachable "wake-hash returned {format hash_lines} lines but we expected {str to_hash_len} lines"
+
+    # Finally actully add all the hashes
```
V-FEXrt, line 215: typo: actually

tools/wake-hash/main.cpp
```diff
@@ -184,38 +185,65 @@ static Hash256 do_hash(const char* file) {
...
+    for (size_t i = 0; i < num_threads; ++i) {
+      // In each thread we work steal a thing to hash
+      to_join.emplace_back(std::async([&counter, &hashes, &files_to_hash]() {
+        while (true) {
+          size_t idx = counter.fetch_add(1);
+          // No more work to do so we exit
+          if (idx >= files_to_hash.size()) {
+            return;
+          }
+          // Output the result directlty into the output location. This
```
V-FEXrt, line 229: typo: directly

share/wake/lib/system/path.wake
JakeSiFive: The first argument *is* the label. It's `"<hash>"`

V-FEXrt: ope missed that lol

PR #1275, opened 2023-05-31
use job numbers for wakebox spec and result jsons

mmjconolly:
If you set `WAKEBOX_KEEP_IO_FILES=1` you'd see files in a `.build` directory that look like
``` 
$ ls .build/
2272028.29.in.json
2272028.29.out.json
2272028.301.in.json
2272028.301.out.json
2272028.302.in.json
....
```
These files correspond to the input specification (`*.in.json`) and result (`*.out.json`) for each wakebox job.

The first element is the process id of wake on the current machine, this is often not useful information to a user.
The second element is the job id from the database, that users see in `wake --last/--failed`
In this PR: 
* remove the PID from the filenames
* remove in/out.json filename suffixes, replace with spec/result
* the tracking of the files are adjusted to allow `wake --clean` to remove them
* `WAKEBOX_KEEP_IO_FILES` option is removed, the spec/result files now always kept

After changes:
```
$ ls .build
result-29.json
result-301.json
result-302.json
spec-29.json
spec-301.json
spec-302.json
....
```

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -978,8 +978,7 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
 
             Pair (Pass proxy) inFile
 
-    def resultPath specPath =
-        replace `spec-` "result-" specPath
+    def resultPath specPath = replace `spec-` "result-" specPath
 
     def post = match _
         Pair (Fail f) _ = Fail f
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -97,7 +97,7 @@ tuple Plan =
     # The label used when showing the command during execution. This is recommended for efficient
     # debugging and locating job information. For example, the label will show up in the terminal during
     # job execution, in a job's progress bar in the terminal, when running the `wake --last` command,
-    # and in the `<..>.spec` and `<...>.result` files in the `.build` directory.
+    # and in the `spec-<jobid>.json` and `result-<jobid>.json` files in the `.build` directory.
     export Label: String
     # The command-line arguments (the first is the command to run)
     export Command: List String
@@ -963,14 +963,14 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
                 | rmap getPathName
             else Pair (Fail (makeError "Failed to 'mkdir .build'.")) ""
 
-            def specFilePath = "{build}/{prefix}.spec"
+            def specFilePath = "{build}/spec-{prefix}.json"
 
             require Pass inFile =
                 write specFilePath (prettyJSON json)
                 | rmap getPathName
             else Pair (Fail (makeError "Failed to 'write {specFilePath}'.")) ""
 
-            def outFile = "{build}/{prefix}.result"
+            def outFile = resultPath inFile
             def cmd = script, "-I", "-p", inFile, "-o", outFile, extraArgs
 
             def proxy =
@@ -978,12 +978,15 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
 
             Pair (Pass proxy) inFile
 
+    def resultPath specPath =
+        replace `spec-` "result-" specPath
+
     def post = match _
         Pair (Fail f) _ = Fail f
         Pair (Pass (RunnerOutput _ _ (Usage x _ _ _ _ _))) inFile if x != 0 =
             Fail (makeError "Non-zero exit status ({str x}) for JSON runner {script} on {inFile}")
         Pair (Pass _) inFile =
-            def outFile = replace `\.spec$` ".result" inFile
+            def outFile = resultPath inFile
             def json = parseJSONFile (Path outFile "BadHash")
 
             match json
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -97,7 +97,7 @@ tuple Plan =
     # The label used when showing the command during execution. This is recommended for efficient
     # debugging and locating job information. For example, the label will show up in the terminal during
     # job execution, in a job's progress bar in the terminal, when running the `wake --last` command,
-    # and in the `<..>.in.json` and `<...>.out.json` files in the `.build` directory.
+    # and in the `<..>.spec` and `<...>.result` files in the `.build` directory.
     export Label: String
     # The command-line arguments (the first is the command to run)
     export Command: List String
@@ -536,9 +536,6 @@ export def virtualRunner: Runner =
 
     Runner "virtual" (\_ Pass 0.0) doit
 
-def pid =
-    prim "pid"
-
 def implode l =
     cat (foldr (_, "\0", _) Nil l)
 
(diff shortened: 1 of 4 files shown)


mmjconolly:
The test failure on this is kind of interesting, it reproduces locally easily enough via `make test`

```
Fail (Error "Test failed (Exited 1) with 'Fail (Error \"Was expecting a END, but got a BCLOSE at .build/3.result:1:168\" Nil)'" Nil)
```

```
$ cat tests/job-cache/basic-lru/.build/3.result
{"usage":{"status":0,"membytes":4116480,"inbytes":0,"outbytes":22,"runtime":0.022625,"cputime":0.002634},"inputs":[],"outputs":["one.txt"],"run-host":"pe01.internal.sifive.com"}}} 
```
There are extra `}` right at the end?

share/wake/lib/system/job.wake
```diff
@@ -97,7 +97,7 @@ tuple Plan =
     # The label used when showing the command during execution. This is recommended for efficient
     # debugging and locating job information. For example, the label will show up in the terminal during
     # job execution, in a job's progress bar in the terminal, when running the `wake --last` command,
-    # and in the `<..>.in.json` and `<...>.out.json` files in the `.build` directory.
+    # and in the `<..>.spec` and `<...>.result` files in the `.build` directory.
```
JakeSiFive, line 100: Can we keep the ".json" part? It's useful for syntax highlighting and the like.

share/wake/lib/system/path.wake
```diff
@@ -200,3 +200,19 @@ target hashcode (f: String): String =
         match job.isJobOk hash
             True (x, Nil) = x
             _ _ = "BadHash"
+
+# Allow an untracked file to be removed via `wake --clean`
+export target markFileCleanable (filepath: String): Result Unit Error =
```
JakeSiFive, line 205: Can this just be implemented as `claim filepath | rmap (\_ Unit)` instead?

share/wake/lib/system/job.wake
```diff
@@ -963,39 +958,38 @@ export def makeJSONRunner (plan: JSONRunnerPlan): Runner =
...
-            def outFile = replace `\.in\.json$` ".out.json" inFile
+            def outFile = replace `\.spec$` ".result" inFile
             def json = parseJSONFile (Path outFile "BadHash")
-            def _ = unlink inFile
 
             match json
                 Fail f = Fail f
                 Pass content =
-                    def _ = unlink outFile
+                    def _ = markFileCleanable outFile
```
JakeSiFive, line 992: This is quite clever btw

JakeSiFive (approved):

share/wake/lib/system/job.wake
mmjconolly: I'd actually rather have no extension, but wanted to separate out the spec and result. I see people putting the encoding type as file extensions a lot, but it's just implementation detail. Syntax highlighting for json specifically is pretty low value compared to having it pretty-printed.

Perhaps we could swap the name order to become `spec-{jobid}` and `result-{jobid}`

JakeSiFive: I'm fine with no json extension. Does that give us some extra value or is it just a preference?

mmjconolly: It's a preference, as a lot of people end up call things in discussions the $thing json rather than $thing

share/wake/lib/system/path.wake
ag-eitilt: `claim` might indeed be better, since this method will also trigger "File output by multiple jobs" errors for anyone who *does* try to claim the file, and is not necessarily an easy label to track down.  (By the same token, this won't solve the bug I was running into previously.)

share/wake/lib/system/job.wake
ag-eitilt: +1 on dropping the extension.  It's unfortunate that programs/platforms treat that as the sole marker, but especially since we don't expect the average user to be editing these, we shouldn't be constrained by implementation details.

ag-eitilt (approved):

share/wake/lib/system/job.wake
ag-eitilt: No, the assumption is correct: "what a file's format is **intended but not guaranteed** to be".  If it makes sense to use a standard file extension in some situation, then there's no need to be going off looking for something unique.  For things like JSON, though, where there's a (maybe implicit) schema *in addition to* the syntax, and especially in this case where we have two different schemas being used in the same directory, it sometimes makes more sense to treat that as the primary thing to be naming files after rather than the agnostic format.  (On the other hand, if you're using a file as interchange between different programs with a looser/no schema, then you definitely *should* be using the format extension.)

mmjconolly: > What is the point of file extensions?

On unix-like systems, there is no inherent purpose/mechanism other than visually communicating to the user. It's just bytes.

It's expected that if something really does need to know file-type, it should inspect it via something like `file`.
On a graphical system, some use programs like `xdg-open` to guess associated applications for use with file extensions like on windows

JakeSiFive: @bmitc what does vscode do if there are no extensions? I seem to recall it assuming that some of my log files without extensions were yaml files for some reason.

JakeSiFive: Also I would still like to say, I mostly consider this a bike shed but my preference for the color of the bike shed at the nuclear facility is "with extension"

share/wake/lib/system/path.wake
mmjconolly: Claimed files don't get removed via `wake --clean`

```
$ echo foo > baa
$ wake -x 'claim "baa"'
Pass (Path "baa" "20590a52c4f00588c500328b16d466c982a26fabaa5fa4dcc83052dd0a84f233")
$ wake --clean
$ ls baa
baa
```

ag-eitilt: Ah, that's probably a good thing, now I think about it.

mmjconolly:
The earlier CI issue was that we weren't open()ing the wakebox result file with truncation.
We didn't notice this before, as we used wake's process id to create different filenames even when a job number was re-used. In theory if wake received the same PID a second time it could have run into this issue

mmjconolly:
Changes since last push are in 7cc3aada72f74ad2cb84bed4e22e727a9b2c336c

`.json` suffix is kept, but now with `spec-` and `result-` prefixes

tools/wakebox/main.cpp
```diff
@@ -193,7 +193,7 @@ int run_batch(const char *params_path, bool has_output, bool use_stdin_file, boo
   }
 
   // Open the output file
-  int out_fd = open(result_path, O_WRONLY | O_CREAT | O_CLOEXEC, 0664);
+  int out_fd = open(result_path, O_WRONLY | O_CREAT | O_CLOEXEC | O_TRUNC, 0664);
```
JakeSiFive, line 196: Small nit for future PRs, I think its better to create a new inode than it is to truncate. The only way I know to do this is to create a temp file and rename it into place.

PR #1278, opened 2023-06-05
Update `write` documentation

bmitc:
This is a small change to clarify the behavior of `write` to mention that it does no processing. Specifically, it does not write a final terminating `\n` character. This should make it a small amount more clear to users.

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -46,7 +46,9 @@ target writeImp inputs mode path content =
     | runJobWith writeRunner
     | getJobOutput
 
-# Create all directories and the named file
+# Create all directories and the named file. The `content` string is written verbatim with no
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):
More documentation is always better, but FWIW the fact that its called `write` vs `writeLine` makes it pretty clear to me personally that it doesn't change the input.

JakeSiFive:
Having a new line at the end of files is a really silly thing IMO. Most places I've worked have not required it in source files despite it technically being the "correct way" to do this for text files according to posix. That's a rather silly requirement if you ask me. It's issues like that where a bad blanket statement was applied to all files by standards that mean that there's even a difference between a binary file and a non-binary file in common file io interfaces.

If it were up to me \r\n would die, and so would this requirement lol

PR #1279, opened 2023-06-06
implement retry for cache read with recovery

V-FEXrt:


src/job_cache/job_cache.cpp
```diff
@@ -167,28 +167,26 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
...
+}
 
-  // connect to the daemon with backoff.
-  // TODO: Put this into a function so that we can call it again if
-  // the daemon fails unexpectedly.
+// Connect to the job cache daemon with backoff.
+void Cache::backoff_try_connect() {
   wcl::xoshiro_256 rng(wcl::xoshiro_256::get_rng_seed());
   useconds_t backoff = 1000;
   for (int i = 0; i < 10; i++) {
```
JakeSiFive, line 188: While you're here, can you change this to 14 instead of 10? Nathan has reported this error-ing out and I think it was just because the machine was slow. Putting this up to 14 will allow for up to ~16 seconds of hang before killing itself while right now we wait ~1 second

src/job_cache/job_cache.cpp
```diff
@@ -219,16 +231,20 @@ FindJobResponse Cache::read(const FindJobRequest &find_request) {
     MessageParserState state = parser.read_messages(messages);
 
     if (state == MessageParserState::StopFail) {
-      // TODO: Try to reconnect to the daemon, launching our own if need be.
-      //       if that fails then depending on user preference either fail
-      //       or return a cache miss.
+      if (false) {
```
JakeSiFive, line 234: I starred at this for so long before I realized this was dead code lol.

src/job_cache/job_cache.cpp
```diff
@@ -219,16 +231,20 @@ FindJobResponse Cache::read(const FindJobRequest &find_request) {
...
+
+      wcl::log::info("retry failed. msg = %s", err_msg);
       wcl::log::fatal("Cache::read(): failed receiving message");
     }
 
     if (state == MessageParserState::StopSuccess && messages.empty()) {
-      // TODO: Try to reconnect to the daemon, launching our own if need be.
-      //       if that fails then depending on user preference either fail
-      //       or return a cache miss.
+      if (false) {
```
JakeSiFive, line 243: luckily I had learned my lesson by this one ;)

src/job_cache/job_cache.cpp
JakeSiFive: Actually I realized that we also want exponential back off on the read/write retries so I think we want this to be configurable so that on start up we give it the old college try but on retries we try to limit the wait to about 1 second.

src/job_cache/job_cache.cpp
V-FEXrt: yeah, this is where the config check for "cache miss instead of fail" goes

JakeSiFive: In that case this logic should be moved outside of this function so that we can declare it a failure only once we've tried reconnecting multiple times.

src/job_cache/job_cache.cpp
```diff
@@ -215,49 +224,81 @@ FindJobResponse Cache::read(const FindJobRequest &find_request) {
...
-    wcl::log::fatal("Cache::read(): failed to parse daemon response");
+    wcl::log::error("Cache::read(): failed to parse daemon response");
+    return wcl::result_error<FindJobResponse>(FindJobError::FailedParseResponse);
+  }
+
+  return wcl::result_value<FindJobError>(FindJobResponse(json));
+}
+
+FindJobResponse Cache::read(const FindJobRequest &find_request) {
+  for (int i = 0; i < 10; i++) {
```
JakeSiFive, line 282: I think we want exponential back off here, but we also want a static variable here that counts the total number of times we've failed to connect so we don't spend too much time in the "no-fail" config case. We can wait to add that once we have an actual no-fail config case though but maybe a TODO here would be good.

This flag will also prevent us from spamming them with errors/warnings when they're in a long run that they want to keep running.

src/job_cache/job_cache.cpp
```diff
@@ -167,28 +167,26 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
...
 
     wcl::log::fatal("exec(%s): %s", job_cache.c_str(), strerror(errno));
   }
+}
 
-  // connect to the daemon with backoff.
-  // TODO: Put this into a function so that we can call it again if
-  // the daemon fails unexpectedly.
+// Connect to the job cache daemon with backoff.
+void Cache::backoff_try_connect(int attempts) {
```
JakeSiFive, line 185: I think this now has to return a result too :(

src/job_cache/job_cache.cpp
```diff
@@ -167,28 +167,26 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
...
 
-Cache::Cache(std::string dir, uint64_t max, uint64_t low) {
-  mkdir_no_fail(dir.c_str());
-
-  // launch the daemon
-  if (daemonize(dir.c_str())) {
-    // We are the daemon, launch the cache
-
+// Launch the job cache daemon
+void Cache::launch_daemon() {
```
JakeSiFive, line 171: this also has to return a result :(

src/job_cache/job_cache.cpp
```diff
@@ -215,49 +224,81 @@ FindJobResponse Cache::read(const FindJobRequest &find_request) {
...
-      // TODO: Try to reconnect to the daemon, launching our own if need be.
-      //       if that fails then depending on user preference either fail
-      //       or return a cache miss.
-      wcl::log::fatal("Cache::read(): failed receiving message");
+      // TODO: Add config var to determine if fail is a cache miss
+      if (false) {
+        return wcl::result_value<FindJobError>(FindJobResponse(wcl::optional<MatchingJob>{}));
+      }
+
+      wcl::log::error("Cache::read(): failed receiving message");
```
JakeSiFive, line 237: These non-fatal errors should be shown to the human while running so that they know what could be causing the slow down...and they should possibly be warnings I'm not sure.

mmjconolly:
Please put descriptions on all pull requests, it's hard to have context for what this change is without it

V-FEXrt:
> Please put descriptions on all pull requests, it's hard to have context for what this change is without it

Right, sorry. This started out as a draft PR mostly there for checking direction then became a serious PR without me updating the description. I'll fix it on the "real" PR https://github.com/sifive/wake/pull/1281

src/job_cache/job_cache.cpp
V-FEXrt: I still don't agree that *all* `log::error`s should be displayed to the user. Which is why we need the filtered logger in https://github.com/sifive/wake/pull/1284.

Can we punt this comment until that lands?

JakeSiFive: We can punt but I think if this message comes up and user's have the config flag enabled to not fail they need to see these error messages.

V-FEXrt:
Closing this now that all the comments are covered in the other PR

PR #1280, opened 2023-06-07
Add more keywords to VS Code syntax highlighting

bmitc:
Some newer keywords like `@here`, `@file`, etc. as defined in https://github.com/sifive/wake/blob/master/src/parser/lexer.re were missing from the VS Code extension's syntax highlighting for Wake.

I had noticed that `@here` wasn't highlighted correctly in a Wake file:

image

No testing was done by me. I simply edited the regex and checked that `!` and `@` were not special regex operators, and as far as I can tell, they are not and are just literal character matches.

### DIFF
--- a/extensions/vscode/syntaxes/wake.tmLanguage.json
+++ b/extensions/vscode/syntaxes/wake.tmLanguage.json
@@ -29,7 +29,7 @@
 		"keyword": {
 			"patterns": [{
 				"name": "keyword.control.wake",
-				"match": "\\b(_|def|tuple|data|global|target|publish|subscribe|prim|if|then|else|here|match|require|package|import|export|from|type|topic|unary|binary)\\b"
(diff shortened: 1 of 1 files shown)


JakeSiFive (approved):
Thanks for this!

V-FEXrt:
The new extension should be downloadable from here https://github.com/sifive/wake/actions/runs/5201522774?pr=1280 if you can give it a test @bmitc

extensions/vscode/syntaxes/wake.tmLanguage.json
```diff
@@ -152,4 +152,4 @@
 			]
 		}
 	}
-}
\ No newline at end of file
+}
```
V-FEXrt, line 155: can't remember what the policy on trailing newlines is, but looks like it was added here when it was previously missing

V-FEXrt (approved):
LGTM % testing with the built artifact.

Thanks!

extensions/vscode/syntaxes/wake.tmLanguage.json
mmjconolly: non-"binary"-data files like this one should have the last character of the file be a newline

PR #1281, opened 2023-06-07
Increase robustness of job cache client

V-FEXrt:
This is a repush of https://github.com/sifive/wake/pull/1279 rebased and re-targeted to master. There was a bunch of merge conflict I didn't want to deal with so I'll just close that PR after comments from it are resolved here.

Job cache client now attempts to recover from several failure modes related to the job cache daemon. Launching and connecting to the daemon is reattempted several times with backoff. Cache read requests are also retried including relaunching and reconnecting to the daemon with backoff. After a set number of attempts the retries will give up and fail. Currently on failure wake will exit with code `1` however the scaffolding is in place so that wake can later be configured to just return a cache miss on failure

### DIFF
--- a/src/job_cache/job_cache.cpp
+++ b/src/job_cache/job_cache.cpp
@@ -278,12 +278,22 @@ wcl::result<FindJobResponse, FindJobError> Cache::read_impl(const FindJobRequest
   return wcl::result_value<FindJobError>(FindJobResponse(json));
 }
 
+// TODO: track total number of times we've failed in order to
+// fail fast after some threshold when the users requests a
+// cache miss on fail
 FindJobResponse Cache::read(const FindJobRequest &find_request) {
+  wcl::xoshiro_256 rng(wcl::xoshiro_256::get_rng_seed());
+  useconds_t backoff = 1000;
+
   for (int i = 0; i < 10; i++) {
     auto response = read_impl(find_request);
     if (response) {
       return *response;
     }
+
+    std::uniform_int_distribution<useconds_t> variance(0, backoff);
+    usleep(backoff + variance(rng));
+    backoff *= 2;
 
     // Retry
     wcl::log::info("Relaunching the daemon.");
--- a/src/job_cache/job_cache.cpp
+++ b/src/job_cache/job_cache.cpp
@@ -167,28 +167,26 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
   return wcl::make_some<wcl::unique_fd>(std::move(socket_fd));
 }
 
-Cache::Cache(std::string dir, uint64_t max, uint64_t low) {
-  mkdir_no_fail(dir.c_str());
-
-  // launch the daemon
-  if (daemonize(dir.c_str())) {
-    // We are the daemon, launch the cache
-
+// Launch the job cache daemon
+void Cache::launch_daemon() {
+  // We are the daemon, launch the cache
+  if (daemonize(cache_dir.c_str())) {
     std::string job_cache = wcl::make_canonical(find_execpath() + "/../bin/job-cache");
-    std::string low_str = std::to_string(low);
-    std::string max_str = std::to_string(max);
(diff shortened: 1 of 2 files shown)


src/job_cache/job_cache.cpp
```diff
@@ -167,28 +167,29 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
...
-    std::string max_str = std::to_string(max);
-    execl(job_cache.c_str(), "job-cached", dir.c_str(), low_str.c_str(), max_str.c_str(), nullptr);
+    std::string low_str = std::to_string(low_threshold);
+    std::string max_str = std::to_string(max_size);
+    execl(job_cache.c_str(), "job-cached", cache_dir.c_str(), low_str.c_str(), max_str.c_str(),
+          nullptr);
 
-    wcl::log::fatal("exec(%s): %s", job_cache.c_str(), strerror(errno));
+    wcl::log::error("exec(%s): %s", job_cache.c_str(), strerror(errno));
+    return wcl::make_some<ConnectError>(ConnectError::FailedToLaunchDaemon);
```
JakeSiFive, line 181: I just realized that this error is confusing to handle since it only gets returned in the child process. Maybe this function shouldn't attempt to return an error? Like we just try to demonize and hope for the best and the other retries and logging messages will smooth user experience and debugging over?

src/job_cache/job_cache.cpp
```diff
@@ -201,11 +202,34 @@ Cache::Cache(std::string dir, uint64_t max, uint64_t low) {
...
+  max_size = max;
+  low_threshold = low;
+
+  mkdir_no_fail(cache_dir.c_str());
+
+  // TODO: Add config var to determine if fail is a cache miss
+  wcl::optional<ConnectError> error;
+
+  error = launch_daemon();
+  if (error && true) {
```
JakeSiFive, line 222: Did you get a compiler error with just `if(error)`?

src/job_cache/job_cache.cpp
```diff
@@ -215,49 +239,95 @@ FindJobResponse Cache::read(const FindJobRequest &find_request) {
...
+    wcl::log::error("Cache::read(): failed to parse daemon response");
+    return wcl::result_error<FindJobResponse>(FindJobError::FailedParseResponse);
+  }
+
+  return wcl::result_value<FindJobError>(FindJobResponse(json));
+}
+
+// TODO: track total number of times we've failed in order to
+// fail fast after some threshold when the users requests a
+// cache miss on fail
```
JakeSiFive, line 288: We should probably do this now. Should just be a simple static counter inside Cache::read. I think 100 attempts or something like that is a fair number?

src/job_cache/job_cache.cpp
```diff
@@ -167,28 +167,29 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
...
-  mkdir_no_fail(dir.c_str());
-
-  // launch the daemon
-  if (daemonize(dir.c_str())) {
-    // We are the daemon, launch the cache
-
+// Launch the job cache daemon
+wcl::optional<ConnectError> Cache::launch_daemon() {
+  // We are the daemon, launch the cache
+  if (daemonize(cache_dir.c_str())) {
```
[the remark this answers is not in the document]
V-FEXrt: yeah lol, that's a can of worms that we should probably punt

src/job_cache/job_cache.cpp
JakeSiFive: If it does happen though, we'll have the log, but then the child process will keep on running which is very odd.

src/job_cache/job_cache.cpp
JakeSiFive: I see, can you add a TODO: there then?

JakeSiFive (approved):

src/job_cache/job_cache.cpp
V-FEXrt: ohhh yeah that is pretty weird

PR #1287, opened 2023-06-12
Shift wake log to be more fluent

V-FEXrt:
The internal logging api was already starting to explode. Instead of a million subtlety different helper functions this PR converts wake logging into a more fluent api that can be used to layer in more information.

```C++
wcl::log::event("msg: %s", "var").warning().urgent().pid().time()();
```

There are still some helper functions for the most common use cases 

```C++
wcl::log::warning("msg: %s", "var")();
```

the trailing `()` is the signal to "consume and trigger the event" enabling flexible usage like the following

```C++
Event e = wcl::log::warning("msg").info();
e.items["A"] = "B";
e();
```

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -73,6 +73,7 @@ static void lock_file(const char *lock_path) {
   if (errno == EAGAIN || errno == EACCES) {
     wcl::log::info("fcntl(F_SETLK, %s): %s -- assuming another daemon exists, closing", lock_path,
                    strerror(errno))();
+    exit(0);
   }
 
   // Something went wrong trying to grab the lock
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -47,7 +47,8 @@ using namespace job_cache;
 static int open_fd(const char *str, int flags, mode_t mode) {
   int fd = open(str, flags, mode);
   if (fd == -1) {
-    wcl::log::fatal("open(%s): %s", str, strerror(errno));
+    wcl::log::error("open(%s): %s", str, strerror(errno)).urgent()();
+    exit(1);
   }
   return fd;
 }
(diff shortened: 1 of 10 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -70,13 +71,13 @@ static void lock_file(const char *lock_path) {
 
   // Some other process has the lock, they are the daemon
   if (errno == EAGAIN || errno == EACCES) {
-    int pid = getpid();
```
V-FEXrt, line 73: I've removed `pid` from several places since its already included in the log

src/job_cache/daemon_cache.cpp
```diff
@@ -70,13 +71,13 @@ static void lock_file(const char *lock_path) {
...
-    wcl::log::exit("fcntl(F_SETLK, %s): %s -- assuming another daemon exists, closing pid=%d",
```
V-FEXrt, line 74: I'm missing an `exit(0);` here

src/wcl/tracing.cpp
```diff
@@ -174,24 +154,5 @@ void FormatSubscriber::receive(const Event& e) {
   }
 }
 
-void FatalEventSubscriber::receive(const Event& e) {
```
V-FEXrt, line 177: This better version of this comes back with https://github.com/sifive/wake/pull/1284

src/wcl/tracing.h
```diff
@@ -59,6 +58,20 @@ struct Event {
...
+  Event urgent() && __attribute__((warn_unused_result));
+  Event time() && __attribute__((warn_unused_result));
+  Event pid() && __attribute__((warn_unused_result));
+
+  Event info() && __attribute__((warn_unused_result));
+  Event warning() && __attribute__((warn_unused_result));
+  Event error() && __attribute__((warn_unused_result));
+
+  void operator()() &&;
+  // TODO: initializer list operator()
```
V-FEXrt, line 74: Still need to do this

tools/wake/main.cpp
```diff
@@ -316,8 +316,7 @@ int main(int argc, char **argv) {
   std::ofstream log_file("wake.log", std::ios::app);
   auto log_file_defer = wcl::make_defer([&log_file]() { log_file.close(); });
   wcl::log::subscribe(std::make_unique<wcl::log::FormatSubscriber>(log_file.rdbuf()));
-  wcl::log::subscribe(std::make_unique<wcl::log::FatalEventSubscriber>(std::cerr.rdbuf()));
-  wcl::log::info("Initialized logging");
+  wcl::log::info("Initialized logging")();
```
JakeSiFive, line 319: Don't we still need a subscriber to display urgent messages?

src/wcl/tracing.h
```diff
@@ -59,6 +58,20 @@ struct Event {
...
+  Event message(const char* fmt, ...) && __attribute__((format(printf, 2, 3)));
+  Event message(const char* fmt, va_list args) &&;
+
+  Event urgent() && __attribute__((warn_unused_result));
+  Event time() && __attribute__((warn_unused_result));
+  Event pid() && __attribute__((warn_unused_result));
+
+  Event info() && __attribute__((warn_unused_result));
+  Event warning() && __attribute__((warn_unused_result));
+  Event error() && __attribute__((warn_unused_result));
```
JakeSiFive, line 71: Instead of these, we should instead have `level` that takes a string as input. These won't generally be used so I think a more orthogonal, generic method is better.

tools/wake/main.cpp
V-FEXrt: Yeah, I want to do that as part of #1286 since it does a slightly different version of this already. 

Mainly that's just to minimize the merge conflicts, and make this PR a bit smaller

JakeSiFive (approved):
LGTM!

PR #1297, opened 2023-06-15
Make job cache more self-healing

JakeSiFive:
This change adds two functions to the job cache

1) If a corrupted job is detected, instead of just returning a cache miss, we clean up that job so future checks won't fail
2) If a folder storing the hashed files of a job exists in the folder but does not exist in the database, we clean up the folder

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -623,7 +623,7 @@ void DaemonCache::remove_corrupt_job(int64_t job_id) {
   if (!dir_res) {
     // We can keep going even with this failure but we need to at least log it
     wcl::log::error("cleaning corrupt job: wcl::directory_range::open(%s): %s", job_dir.c_str(),
-                    strerror(dir_res.error()));
+                    strerror(dir_res.error()))();
     return;
   }
 
@@ -632,7 +632,7 @@ void DaemonCache::remove_corrupt_job(int64_t job_id) {
     if (!entry) {
       // It isn't critical that we remove this so just log the error and move on
       wcl::log::error("cleaning corrupt job: bad entry in %s: %s", job_dir.c_str(),
-                      strerror(entry.error()));
+                      strerror(entry.error()))();
       return;
     }
     to_delete.emplace_back(wcl::join_paths(job_dir, entry->name));
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -282,6 +282,7 @@ class JobTable {
   std::shared_ptr<job_cache::Database> db;
   PreparedStatement add_job;
   PreparedStatement add_output_info;
+  PreparedStatement remove_job;
 
  public:
   static constexpr const char *insert_query =
@@ -293,10 +294,16 @@ class JobTable {
       "(job, stdout, stderr, ret, runtime, cputime, mem, ibytes, obytes)"
       "values (?, ?, ?, ?, ?, ?, ?, ?, ?)";
 
+  static constexpr const char *remove_job_query = "delete from jobs where job_id = ?";
+
   JobTable(std::shared_ptr<job_cache::Database> db)
-      : db(db), add_job(db, insert_query), add_output_info(db, add_output_info_query) {
+      : db(db),
+        add_job(db, insert_query),
+        add_output_info(db, add_output_info_query),
+        remove_job(db, remove_job_query) {
     add_job.set_why("Could not insert job");
     add_output_info.set_why("Could not add output info");
+    remove_job.set_why("Could not remove job");
   }
 
   int64_t insert(const std::string &cwd, const std::string &cmd, const std::string &env,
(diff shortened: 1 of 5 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -596,6 +609,44 @@ int DaemonCache::run() {
   return 0;
 }
 
+void DaemonCache::remove_corrupt_job(int64_t job_id) {
+  // First remove this job from the database so that we don't get hung up on it anymore
+  impl->jobs.remove(job_id);
+
+  // Find this job directory so we can remove all the file
```
V-FEXrt, line 616: nit: `files`

src/job_cache/daemon_cache.cpp
```diff
@@ -596,6 +609,44 @@ int DaemonCache::run() {
...
+  uint8_t group_id = job_id & 0xFF;
+  std::string job_dir = wcl::join_paths(wcl::to_hex(&group_id), std::to_string(job_id));
```
V-FEXrt, line 618: I might be wrong, but I though we wrote a helper func for getting the job_dir?

src/job_cache/daemon_cache.cpp
```diff
@@ -596,6 +609,44 @@ int DaemonCache::run() {
...
+    to_delete.emplace_back(wcl::join_paths(job_dir, entry->name));
+  }
+
+  // Unlink them all
+  for (const auto &file : to_delete) {
+    unlink_no_fail(file.c_str());
+  }
+
+  // Remove the files
+  rmdir_no_fail(job_dir.c_str());
```
V-FEXrt, line 647: per the comments above. Do we want to have the Daemon die if unlink/rmdir fail?

src/job_cache/eviction_policy.cpp
```diff
@@ -219,9 +220,84 @@ struct LRUEvictionPolicyImpl {
...
+  // Find all the entries to remove
+  for (const auto& entry : *dir_res) {
+    if (!entry) {
+      // If one entry has a failure we can just keep going to try and remove more entries
+      wcl::log::error("cleaning corrupt job: bad entry in %s: %s", job_dir.c_str(),
+                      strerror(entry.error()))();
+      continue;
+    }
+    std::string file = wcl::join_paths(job_dir, entry->name);
+    unlink_no_fail(file.c_str());
```
V-FEXrt, line 242: same re `_no_fail`

src/job_cache/eviction_policy.cpp
```diff
@@ -219,9 +220,84 @@ struct LRUEvictionPolicyImpl {
...
+  }
+}
+
+static void garbage_collect_orphan_folders(std::shared_ptr<job_cache::Database> db) {
+  constexpr const char* all_jobs_q = "select job_id from jobs";
+  PreparedStatement all_jobs(db, all_jobs_q);
+  Transaction transact(db);
+  std::unordered_set<int64_t> jobs;
+  // uint8_t group_id = job_id & 0xFF;
+  // std::string job_dir = wcl::join_paths(wcl::to_hex(&group_id), std::to_string(job_id));
```
V-FEXrt, line 280: commented code

src/job_cache/eviction_policy.cpp
```diff
@@ -219,9 +220,84 @@ struct LRUEvictionPolicyImpl {
...
+    garbage_collect_group(jobs, group_id);
+  }
+}
+
 void LRUEvictionPolicy::init(const std::string& cache_dir) {
-  auto db = std::make_unique<job_cache::Database>(cache_dir);
-  impl = std::make_unique<LRUEvictionPolicyImpl>(cache_dir, std::move(db));
+  std::shared_ptr<job_cache::Database> db = std::make_unique<job_cache::Database>(cache_dir);
+  impl = std::make_unique<LRUEvictionPolicyImpl>(cache_dir, db);
+  std::async([db]() { garbage_collect_orphan_folders(db); });
```
V-FEXrt, line 300: potential race with adding jobs while garbage collecting

V-FEXrt (changes requested):

src/job_cache/daemon_cache.cpp
JakeSiFive: I would say we don't want it to die at all in this loop.

src/job_cache/daemon_cache.cpp
JakeSiFive: We did not thus far. I don't want to go re-factor other code to do that. I looked at doing it just now and it was a bit less than trivial since both the group_id and the job_dir wind up getting used in a couple places.

src/job_cache/eviction_policy.cpp
JakeSiFive: Thanks for catching this!!! I solved it by only removing jobs that are less than or equal to the jobs that we saw at startup so anything newer will be left alone.

V-FEXrt (approved):

PR #1299, opened 2023-06-16
Log when an eviction occurs

JakeSiFive:
This change causes the shared cache .stderr file to log evictions. It would be nicer to log to .stdout but that is being used as a communication channel between the job-cache daemon and the eviction child process. We should be able to switch the eviction child process over to a thread now which would resolve this issue.

### DIFF
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -52,6 +52,7 @@ struct LRUEvictionPolicyImpl {
   job_cache::PreparedStatement find_least_recently_used;
   job_cache::PreparedStatement remove_least_recently_used;
   job_cache::Transaction transact;
+  std::thread cleaning_thread;
 
   // TODO: Right now we're using `obytes` as a proxy for the size of a job
   //       but this is an over aproximation of the storage cost really.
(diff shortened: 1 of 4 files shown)


src/job_cache/eviction_policy.cpp
```diff
@@ -213,9 +222,10 @@ struct LRUEvictionPolicyImpl {
     // Now we remove the threads in the background.
     // TODO: Figure out how many cores we actully have and use a multiple of that
     // NOTE: We don't wait for this to finish
-    std::async(std::launch::async, [dir = cache_dir, jobs_to_remove = std::move(jobs_to_remove)]() {
-      remove_backing_files(dir, jobs_to_remove, 8, 128);
-    });
+    // std::async(std::launch::async, [dir = cache_dir, jobs_to_remove =
+    // std::move(jobs_to_remove)]() {
```
V-FEXrt, line 226: Can you explain / fix the commented code here?

src/job_cache/eviction_policy.cpp
```diff
@@ -213,9 +222,10 @@ struct LRUEvictionPolicyImpl {
...
+    remove_backing_files(cache_dir, jobs_to_remove, 8, 128);
```
V-FEXrt, line 227: Where did `cache_dir` come from? I don't see a new var introduced

src/job_cache/job_cache_impl_common.cpp
```diff
@@ -251,11 +255,17 @@ void remove_backing_files(const std::string &dir, int64_t job_id) {
...
     }
-    unlink_no_fail(entry->name.c_str());
+    files_to_remove.push_back(wcl::join_paths(job_dir, entry->name));
+  }
+  for (const auto &file : files_to_remove) {
+    unlink_no_fail(file.c_str());
   }
+
+  rmdir_no_fail(job_dir.c_str());
 }
```
V-FEXrt, line 265: do you need to rebase?

V-FEXrt (changes requested):

src/job_cache/eviction_policy.cpp
JakeSiFive: Oh right. This is the "async's destructor blocks" issue. I should probably put a bit of work into using an std::thread like I did before instead.

src/job_cache/eviction_policy.cpp
JakeSiFive: previously we had a generalized capture of `dir = cache_dir`

src/job_cache/job_cache_impl_common.cpp
JakeSiFive: Nope! I found bugs in the removal while writing this change. This fixes them.

V-FEXrt (approved):

PR #1300, opened 2023-06-16
Never cache a failed job

JakeSiFive:
We had the quite silly behavior of caching job failures before which makes no sense. This change eliminates that.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -511,6 +511,7 @@ export def mkJobCacheRunner (wakeroot: String) ((Runner name score baseDoIt): Ru
 
                 # Caching a failed job is a waste of space, never do that
                 require True = status == 0
+                else Pass ""
 
                 job_cache_add jobCacheAddJson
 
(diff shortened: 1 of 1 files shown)


V-FEXrt (changes requested):

share/wake/lib/system/job.wake
```diff
@@ -509,6 +509,10 @@ export def mkJobCacheRunner (wakeroot: String) ((Runner name score baseDoIt): Ru
 
                     True
 
+                # Caching a failed job is a waste of space, never do that
+                require True = status == 0
+                else Pass ""
```
ag-eitilt, line 514:
```recommendation
                require 0 = status
                else Pass ""
```

ag-eitilt (commented):
I didn't see this go up earlier, but I'm slightly concerned about flakiness from this, mainly involving race conditions.  Say you have two `def` which nominally are executed in parallel; one tries to `write` to a path, the other tries to `claim` from that same location.  Depending on which gets reached first (assuming a clean directory each time), the `claim` might successfully fail sometimes when it triggers on a nonexistent file, then the `write` comes along and happily passes since there's no indication the `claim` touched the path; OR the `write` might happen first and pass, but then the `claim` is evaluated, finds the file and wake aborts early due to a "file output by multiple jobs".

There's no question that the code described is *bad*, but I see failure tracking -- when we know what files to expect -- as a mitigation for surfacing known-bad cases *consistently* rather than letting them through to generate flakiness down the line.

PR #1302, opened 2023-06-16
Add logInteractive, remove echo from no-tty, update docs

V-FEXrt:
- logInteractive is now a log stream that only displays when a tty is set, usually implying a human is running the command
- logEcho doesn't get printed by default when --no-tty is set, most folks don't want echo when setting --no-tty, and CI can add it back manually if they want it
- The logging docs had drifted a bit, they are now updated


JakeSiFive (approved):
@jackkoenig might want to be made aware of this

mwachs5 (commented):
thanks for updating the docs!

PR #1304, opened 2023-06-16
CI: add Fedora 38

ErikParawell-SiFive:
I end up using Fedora a lot with Wake and I assume others do as well. This will help make sure things don't break in Fedora.

This PR might fail at first as @V-FEXrt and @JakeSiFive have not merged in the Fedora 38 related patches to master.


JakeSiFive:
I think the lemon fix to be .cpp is our best bet personally. It's quick and easy and virtually all C code is also C++ code (which is why the copy worked in the first place).

V-FEXrt (approved):

PR #1306, opened 2023-06-21
Add runner hash

JakeSiFive:
This change allows runner authors to add in some kind of hash or other key (I call it a "runner hash" but it is allowed to be any string) to allow runner authors to make the cache aware of differences in jobs that may not be otherwise apparent. This also allows less conservative use of shared caching with the trade off of being more conservative about what is considered to be "matching" and puts this trade off in the runner author's control.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -366,7 +366,7 @@ export def mkJobCacheRunner (hashFn: Result RunnerInput Error => Result String E
                     "stdin" :-> JString stdin,
                     "input_files" :-> jobCacheVisible,
                     "client_cwd" :-> JString workspace,
-                    "hash" :-> JString hashKey,
+                    "runner_hash" :-> JString hashKey,
                     "dir_redirects" :-> JObject (wakeroot :-> JString "./",),
                 )
 
@@ -486,7 +486,7 @@ export def mkJobCacheRunner (hashFn: Result RunnerInput Error => Result String E
                     "command_line" :-> JString cmd.implode,
                     "envrionment" :-> JString env.implode,
                     "stdin" :-> JString stdin,
-                    "hash" :-> JString hashKey,
+                    "runner_hash" :-> JString hashKey,
                     "input_files" :-> jobCacheReadFiles,
                     "input_dirs" :-> JArray Nil, # TODO: This will need some fuse work to make good on
                     "output_files" :-> jobCacheOutputFiles,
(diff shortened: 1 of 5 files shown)


share/wake/lib/system/job.wake
```diff
@@ -337,7 +337,7 @@ def getPath input =
 
 # wakeroot is the absolute sandbox-path from which input and output files will
 # be interpreted as being relative to if they're in fact relative.
-export def mkJobCacheRunner (wakeroot: String) ((Runner name score baseDoIt): Runner): Runner =
+export def mkJobCacheRunner (hashFn: Result RunnerInput Error => Result String Error) (wakeroot: String) ((Runner name score baseDoIt): Runner): Runner =
```
mmjconolly, line 340: Could this be a `String`? The `RunnerInput` is already inspect-able by the cache runner

JakeSiFive: What specifically do you want to be a string?

mmjconolly: Like this:
```suggestion
export def mkJobCacheRunner (hash: String) (wakeroot: String) ((Runner name score baseDoIt): Runner): Runner =
```

JakeSiFive: You can do that with a function that always returns the same string instead. This is a more generic interface.

mmjconolly: ok

src/job_cache/schema.sql
```diff
@@ -20,8 +20,9 @@ create table if not exists jobs(
   commandline  blob    not null,
   environment  blob    not null,
   stdin        text    not null,
-  bloom_filter integer);
-create index if not exists job on jobs(directory, commandline, environment, stdin);
+  bloom_filter integer not null,
+  runner_hash  text    not null);
+create index if not exists job on jobs(directory, commandline, environment, stdin, runner_hash);
```
V-FEXrt, line 25: discussed offline but we should try to do a migration here instead so folks don't loose their current caches

Something like:
1) add column if not exists
2) drop index if exists
3) add index if not exists

V-FEXrt (approved):

src/job_cache/schema.sql
JakeSiFive: As I went to go fix this, it turns out that sqlite3 does not have the required operations to do this. We will just give people a one-time command to run if they want to migrate. Eventually we'll have global caching and we can migrate that via the proper mechanisms and we can repopulate nuked local caches from that.

PR #1316, opened 2023-06-26
CI: add rocky 9

mmjconolly:
Add Rocky 9 to CI

Only github actions CI is added, as that is used for generating releases, circle CI is ignored

This should be compatible with RHEL9, but it's not feasible to use RHEL9 directly.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,6 +119,14 @@ jobs:
               test_cmd: rpm -i x86_64/*.rpm && cd tests && wake runTests
               install_src_glob: build/x86_64/*.rpm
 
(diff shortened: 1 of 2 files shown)


JakeSiFive (approved):

V-FEXrt (approved):

.circleci/dockerfiles/rocky-9
```diff
@@ -0,0 +1,11 @@
+FROM rockylinux/rockylinux:9
```
V-FEXrt, line 1: need to move this file under the new Github directory

mmjconolly: ah right

PR #1317, opened 2023-06-26
CI: Remove CircleCI

mmjconolly:
The rational for keeping CircleCI was that they provided virtual machine images running an old enough kernel to pair with some of the older docker image distros that we were testing.

It looks like we're not actually using an old enough virtual machine anymore, so I think this is redundant.

There is a webhook in github's `Settings` tab for this repository that would need to be disabled if we were to merge this PR.


JakeSiFive:
Is there something else we need to remove in our github config to make this go away?

V-FEXrt:
Working on the updates right now. Give me a minute

V-FEXrt:
Okay, I updated the required checks and deleted the circle ci webhook. Should be good to land this now and we can see if CircleCI creates errors for future PRs

.github/workflows/build.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
     docker_builds:
       runs-on: ubuntu-latest
       needs: tarball
-      name: Docker Build (${{ matrix.target }})
```
V-FEXrt, line 86: I think (?) this name is load bearing in terms of the 'required' checks for GitHub actions. It doesn't seem to be effecting this PR so it's probably okay but we may need to tweak the settings after this lands

V-FEXrt (approved):

PR #1318, opened 2023-06-26
Add initial setup for a basic shared cache server

JakeSiFive:
This change is just to get us started down the development path for a shared caching server that can be used to repopulate the local shared caches.


mmjconolly:
It could help to have some discriptions for what these are
What is `entity`, `gsc`, `migration` and so on

JakeSiFive:
Oh sure, "entity" and "migration" are the standard names used by sea-orm to make sea-orm work. The "entity" package is generated by sea-orm from an active database but can be modified by hand easily enough as well. Each "entity" is a rust type that corresponds to a table in the database schema. An instance of an entity is equivalent to a row in the database. The "migration" crate is partially managed by sea-orm and partially not. The individual migrations have to be written by hand. Each migration has an "up" and a "down" to go back and forth between them and there's a list of migrations that is kept in chronological order. This lets you move a database up and down and give good error messages when you can't migrate all the way to the latest version expected by your current service. This is also how we plan on not loosing terrabytes of data in the shared cache when we inevitably update the schema.

The "gsc" crate is short for "global shared cache" and implements a JSON RPC web service.

V-FEXrt:
FWIW: `Entity`  and `Migration` are standard terms in (web) API/backend development. We _can_ change those but we probably shouldn't

rust/entity/src/job.rs
```diff
@@ -0,0 +1,17 @@
+//! `SeaORM` Entity. Generated by sea-orm-codegen 0.11.3
+
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "job")]
+pub struct Model {
```
V-FEXrt, line 7: should `Model` here be `Job` or similar? We will eventually have several "Models"

rust/entity/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
V-FEXrt, line 1: we just need the Entity crate for the Models right? I think we should be able to delete the main.rs here

rust/gsc/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+entity = { path = "../entity" }
+migration = { path = "../migration" }
+sea-orm = { version = "0.11.3", features = ["sqlx-postgres", "runtime-tokio-native-tls"]}
+serde = "1.0.164"
+tokio = { version = "1.28.2", features = ["full"] }
+# tower-http = "0.4.1"
```
V-FEXrt, line 12: Commented code

rust/gsc/src/main.rs
```diff
@@ -0,0 +1,51 @@
+use axum::{
+    // http::StatusCode,
```
V-FEXrt, line 2: here too

rust/gsc/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+    env: String,
+}
+
+async fn add_job(Json(payload): Json<AddJobPayload>, conn: Arc<DatabaseConnection>) {
+    let insert_job = job::ActiveModel {
+        id: ActiveValue::NotSet,
+        cmd: ActiveValue::Set(payload.cmd),
+        env: ActiveValue::Set(payload.env),
+    };
+    insert_job.insert(conn.as_ref()).await.unwrap();
```
V-FEXrt, line 25: We should handle the error here instead of `unwrap()`ing

rust/gsc/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // connect to our db
+
+    let connection = sea_orm::Database::connect("postgres://127.0.0.1/test").await?;
```
V-FEXrt, line 32: not for this PR, but `"postgres://127.0.0.1/test"` will eventually need to be sourced from an env var

rust/gsc/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+    let app = Router::new().route(
+        "/job",
+        post({
+            let shared_state = state.clone();
+            move |body| add_job(body, shared_state)
+        }),
+    );
+
+    // run it with hyper on localhost:3000
+    axum::Server::bind(&"127.0.0.1:3000".parse().unwrap())
```
V-FEXrt, line 46: `"127.0.0.1:3000"` same here, but it should be env vars, one for URL, one for PORT

rust/gsc/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
```
V-FEXrt, line 49: Same comment about handling the error here, though in this case I think we can get away with changing the return type of `main()` and doing 

```
  ...
  .serve(...)
  .await?;
```

rust/migration/src/m20220101_000001_create_table.rs
```diff
@@ -0,0 +1,42 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
```
V-FEXrt, line 9: This comment can be deleted

V-FEXrt (changes requested):

rust/entity/src/main.rs
JakeSiFive: Yeah, I think this was generated by `cargo init`, I'll delete it!

rust/gsc/Cargo.toml
JakeSiFive: I intentionally left this here as we're likely to use this in the future. I can get rid of it now if you'd like however.

rust/gsc/src/main.rs
JakeSiFive: So the only options for error handling here are 1) to return a particular StatusCode (which we should probably do) or 2) to use tracing to log the error (which we should probably *ALSO* do) which is a bit unconventional to the normal `.map_err(...)?` that you might otherwise do.

rust/gsc/src/main.rs
JakeSiFive: yep, same with respect to conflicts.

rust/gsc/src/main.rs
JakeSiFive: I need to look into this as its a bit strange. The error returned is a hyper error that does not implement std::error::Error for some reason, otherwise I would not have added this unwrap.

rust/gsc/Cargo.toml
JakeSiFive: removed

rust/gsc/src/main.rs
JakeSiFive: ok added this but the tracing only outputs to stdout right now, eventually we can add an append only log of some kind.

rust/gsc/src/main.rs
JakeSiFive: Actully it did and I was just being dumb. I'm not sure why I was getting an error on that before.

V-FEXrt (approved):

rust/gsc/src/main.rs
mmjconolly: I realise this is early scaffold, but for later:

This relies on the host machine being able to route between the loopback device and other devices, and that the loopback hasn't been assigned to a WAN address. Can this be a less specific address like `0.0.0.0` or `::`, or to take the interface and/or ip address/port from a configuration file

JakeSiFive: The plan is to take the ip address and port from an environment variable

Issue #1309, opened 2023-06-23
Can lsp-wake work with Emacs lsp?


davidmlw:
I followed the Emacs tutorial https://emacs-lsp.github.io/lsp-mode/page/adding-new-language/ to add lsp-wake like this, but it does not work. Can lsp-wake work with Emacs lsp, what kind of Emacs lisp need to be added to support lsp-wake. I suppose lsp-wake follows the standard lsp protocol.

```
(with-eval-after-load 'lsp-mode
  (add-to-list 'lsp-language-id-configuration
               '(wake-mode . "lsp-wake"))

(lsp-register-client
 (make-lsp-client :new-connection (lsp-stdio-connection '("/usr/local/bin/wake" "--lsp"))
                  :activation-fn (lsp-activate-on "lsp-wake")
                  :server-id 'lsp-wake)))

;; or

(lsp-register-client
 (make-lsp-client :new-connection (lsp-stdio-connection "/usr/local/lib/wake/lsp-wake"))
                  :activation-fn (lsp-activate-on "lsp-wake")
                  :server-id 'lsp-wake)))

```

lsp-wake:
https://github.com/sifive/wake/blob/3cd5ce7547384f64d1d36cccc45cfe343358f00b/tools/lsp-wake/main.cpp#L639

wake-mode:
https://github.com/sifive/wake/blob/3cd5ce7547384f64d1d36cccc45cfe343358f00b/share/doc/wake/syntax/emacs/wake-mode.el#L192

mmjconolly:
> but it does not work.

It would be helpful to describe what messages or failures you see when trying to use it

V-FEXrt:
It _should_ be possible but as @mmjconolly mentioned we would need to see the error messages to help. Most editors have an LSP log for the client side so having that would be useful.

One other thing is that most LSP clients require you to specify where the "project root" is. For neovim this is done like this 

```lua
function find_wakeroot(name)
    dir = string.gsub(name, "/[^/]*$", "")
    if dir == "" then 
        return false 
    end
    if file_exists(dir .. "/.wakeroot") or file_exists(dir .. "/wake.db") then
        return dir
    else
        return find_wakeroot(dir)
    end
end

local configs = require('lspconfig.configs')
if not configs.wakelsp then
  configs.wakelsp = {
    default_config = {
      cmd = { '/path/to/wake', '--lsp' },
      filetypes = { 'wake' },
      root_dir = find_wakeroot,
      settings = {},
    },
  }
end
```

[...]

PR #1324, opened 2023-06-28
try workspaceFolders.uri and rootUri

davidmlw:
I think it works for https://github.com/sifive/wake/issues/1309

### DIFF
--- a/tools/lsp-wake/main.cpp
+++ b/tools/lsp-wake/main.cpp
@@ -331,8 +331,14 @@ class LSPServer {
     }
 
     isInitialized = true;
-    std::string workspaceUri =
-        receivedMessage.get("params").get("workspaceFolders").children[0].second.get("uri").value;
+    auto uri = receivedMessage.get("params").get_opt("workspaceFolders");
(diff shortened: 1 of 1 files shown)


V-FEXrt:
Per the spec https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize, `rootPath` and `rootUri` are both deprecated. We initially chose not to implement them because of that.

Does your emacs lsp client still rely on them? Is it possible to update it if so? Actually could you send a link to your emacs client so I can take a look at it?

davidmlw:
https://github.com/emacs-lsp/lsp-mode/blob/05c88ea1bd1ac56bf2a386541047a96240676047/lsp-mode.el#L7628

From this line, I think emacs is using the deprecated lsp specification. And I filed an issue there.

You are right that the best way is that Emacs lsp catch up the lsp spec. But it is also good to help Emacs users use wake-lsp now with a workaround.

davidmlw:
I found how emacs support this.

https://github.com/emacs-lsp/lsp-mode/issues/4081

PR #1326, opened 2023-06-28
Fix race between previous daemon close, and next daemon create

JakeSiFive:
There is a race that can happen if you run a wake process directly after another wake process that looks like this:

1. A wake process runs and successfully creates and connects to a shared cache daemon
2. That wake process exits
3. That shared cache daemon sees that it has no more connections which sends it into a shutdown sequence but the shutdown sequence doesn't complete before...
4. A new wake process starts up, trys to create a daemon, that daemon sees that the cache lock is still locked so it exits
5. The shutdown sequence from step 3 finishes and releases the lock
6. The process from step 4 keeps trying to connect but there is no shared cache to connect to

Aside from that race being an issue, there are all sorts of reasons why a shared cache daemon might fail while we're trying to connect so this is a good change to make regardless.

### DIFF
--- a/src/job_cache/job_cache.cpp
+++ b/src/job_cache/job_cache.cpp
@@ -178,6 +178,7 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
 
 // Launch the job cache daemon
 void Cache::launch_daemon() {
+  wcl::log::info("Relaunching the daemon.")();
   // We are the daemon, launch the cache
   if (daemonize(cache_dir.c_str())) {
     std::string job_cache = wcl::make_canonical(find_execpath() + "/../bin/job-cache");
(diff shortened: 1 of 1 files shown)


src/job_cache/job_cache.cpp
```diff
@@ -196,6 +197,13 @@ wcl::optional<ConnectError> Cache::backoff_try_connect(int attempts) {
   wcl::xoshiro_256 rng(wcl::xoshiro_256::get_rng_seed());
   useconds_t backoff = 1000;
   for (int i = 0; i < attempts; i++) {
+    // We normally connect in about 3 tries on fresh connect so
+    // if we haven't connected at this point its a good spot to start
+    // start trying.
+    if (attempts > 3) {
```
V-FEXrt, line 203: This should be `i` not `attempts`. Attempts never changes :)

also nit: maybe `(i + 1) % 3 == 0`?

V-FEXrt (changes requested):

src/job_cache/job_cache.cpp
JakeSiFive: I didn't want to try just every 3rd time, as the time between invocations is going up exponentially, after a certain point we should just try again every-time. The 4th wait takes as long as all prior waits combined (in expectation).

V-FEXrt (approved):

PR #1327, opened 2023-06-28
Always coredump on crash if we're allowed to

JakeSiFive:
This change should just set the limit on the core size to what ever the maximum allowed is. When the size is 0 (often the default on various systems) no core dump will be produced.

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -21,6 +21,7 @@
 
 #include <inttypes.h>
 #include <stdlib.h>
+#include <sys/resource.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <wcl/defer.h>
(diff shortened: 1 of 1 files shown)


mmjconolly (approved):

tools/wake/main.cpp
```diff
@@ -190,6 +191,14 @@ class TerminalReporter : public DiagnosticReporter {
...
 
 int main(int argc, char **argv) {
+  // Make sure we always get core dumps but don't fail
+  // if that fails for some reason.
+  struct rlimit core_lim;
+  getrlimit(RLIMIT_CORE, &core_lim);
+  core_lim.rlim_cur = core_lim.rlim_max;
+  setrlimit(RLIMIT_CORE, &core_lim);
+
+  // Geth the start time for wake
```
V-FEXrt, line 201: typo:  `Get`

V-FEXrt (approved):

tools/wake/main.cpp
```diff
@@ -190,6 +191,14 @@ class TerminalReporter : public DiagnosticReporter {
 };
 
 int main(int argc, char **argv) {
+  // Make sure we always get core dumps but don't fail
+  // if that fails for some reason.
+  struct rlimit core_lim;
+  getrlimit(RLIMIT_CORE, &core_lim);
```
mmjconolly, line 197: I'm wondering if we should restore the original limit after we next `fork()` (where ever that might be), as here we enable core dumps for all payload commands run inside wakebox

JakeSiFive: I'm not personally very apposed to getting core dumps for everything but if people want that I can make shim-wake reset the limit or reset it just before invoking shim wake.

PR #1331, opened 2023-06-29
lsp: Add lsp integration test

V-FEXrt:
This change adds a basic integration test to ensure that changing the LSP doesn't cause any serious regressions. It can be expanded on in the future but I wanted to get something out there


rust/test-lsp-wake/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    let res: tower_lsp::jsonrpc::Response = serde_json::from_str(&json).unwrap();
+    let value = res.result().unwrap();
+    let inner: R::Result = serde_json::from_value(value.clone()).unwrap();
+    return Ok(inner);
+}
+
+async fn exchange_request_response<R>(
+    stdin: &mut ChildStdin,
+    reader: &mut Lines<BufReader<ChildStdout>>,
```
JakeSiFive, line 111: I think you should put these two things inside of a struct and make these impl methods instead.

rust/test-lsp-wake/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+where
+    R: tower_lsp::lsp_types::request::Request,
+    R::Params: serde::Serialize,
+    R::Result: serde::de::DeserializeOwned,
+{
+    let _length = reader.next_line().await?;
+    let _end_header = reader.next_line().await?;
+    let Some(json) = reader.next_line().await? else {
+        todo!();
+    };
```
JakeSiFive, line 101: nit abut reading this correctly.

rust/test-lsp-wake/test.wake
```diff
@@ -0,0 +1,3 @@
+package foo
```
JakeSiFive, line 1: Can you put test data in a separate folder?

JakeSiFive (commented):
I think these high level functions look good, the only high level feedback I have is about putting stdin and reader into a struct.

ag-eitilt (approved):
Seems good so far as I'm able to parse it.

PR #1332, opened 2023-06-29
wakebox: signal when parent exits

mmjconolly:
In this PR:
* If the `wakebox` process exits, the `wb-mount-ns` process will be sent `SIGTERM`.
* If the `wb-mount-ns` process exits, the `wb-pid-ns` process will be sent `SIGTERM`.

In an earlier PR:
* When the `wb-pid-ns` process terminates, the whole pid namespace collapses and all its children terminate.

So, if you have a process tree that looks like:
```
$ pstree -p 2815889
bash(2815889)─┬─fuse-waked(2826391)
              └─wakebox(2826388)───wb-mount-ns(2826394)───wb-pid-ns(2826395)───mypayload(2826396)
```
Then if you send a signal to _any_ of `wakebox`/`wb-mount-ns`/`wb-pid-ns`/`mypayload`, then everything except the root `bash` process will exit.

### DIFF
--- a/src/wakefs/fuse.cpp
+++ b/src/wakefs/fuse.cpp
@@ -148,6 +148,12 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
     std::vector<std::string> command = args.command;
     std::vector<std::string> envs_from_mounts;
 #ifdef __linux__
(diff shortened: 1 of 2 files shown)


JakeSiFive (approved):
LGTM

PR #1334, opened 2023-06-30
mkdir_with_parents: skip existing directories

mmjconolly:
Previously we relied on `mkdir` claiming that a (parent) directory already exists.
There are some circumstances where the kernel will report other issues _ahead_ of saying that the directory already exists.
For example, a read-only filesystem error could be reported even when the directory exits.

This PR checks if the directory exists before trying to create it.


JakeSiFive (approved):
Interesting, I hadn't ever thought about that.

PR #1335, opened 2023-07-06
Enrich job type

JakeSiFive:
This adds read/write (but write is not blocked by an API key) and adds in all the info the that local shared cache has per job.


JakeSiFive:
Keep in mind that the entity crate is automatically generated and should not be reviewed. The migration and the gsc code *should* be reviewed.

rust/gsc/src/main.rs
```diff
@@ -44,13 +337,21 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            let shared_state = state.clone();
-            move |body| add_job(body, shared_state)
-        }),
-    );
+    let app = Router::new()
+        .route(
+            "/job",
+            post({
+                let shared_state = state.clone();
+                move |body| add_job(body, shared_state)
```
V-FEXrt, line 345: Is there a  rusty reason this can't be 

`move |body| add_job(body, state.clone())`

rust/gsc/src/main.rs
```diff
@@ -44,13 +337,21 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            }),
+        )
+        .route(
+            "/job/matching",
```
V-FEXrt, line 349: Feel free to rename this as you see fit we don't have to go with this route name

rust/migration/src/m20220101_000001_create_table.rs
```diff
@@ -17,25 +29,227 @@ impl MigrationTrait for Migration {
...
+                    .col(ColumnDef::new(Job::IsAtty).boolean().not_null())
+                    .col(ColumnDef::new(Job::HiddenInfo).ezblob())
+                    .col(ColumnDef::new(Job::Stdout).ezblob())
+                    .col(ColumnDef::new(Job::Stderr).ezblob())
+                    .col(ColumnDef::new(Job::Status).integer().not_null())
+                    .col(ColumnDef::new(Job::Runtime).double().not_null())
+                    .col(ColumnDef::new(Job::Cputime).double().not_null())
+                    .col(ColumnDef::new(Job::Memory).big_unsigned().not_null())
+                    .col(ColumnDef::new(Job::IBytes).big_unsigned().not_null())
+                    .col(ColumnDef::new(Job::OBytes).big_unsigned().not_null())
```
V-FEXrt, line 46: Is this missing "Runner hash"?

V-FEXrt (changes requested):

rust/gsc/Cargo.toml
```diff
@@ -12,3 +12,5 @@ tokio = { version = "1.28.2", features = ["full"] }
 tracing = "0.1"
 tracing-subscriber = "0.3"
 axum = "0.6.18"
+serde_bytes = "0.11.9"
+blake3 = "1.4.0"
```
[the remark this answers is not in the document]
JakeSiFive: Right now (but not later) the global shared cache is agnostic to the hashing algorithm that the client uses, but the clients that add jobs will have to agree with the clients that read jobs and to prevent corruption I plan on using blake3 to detect the corruption. I'll eventually migrate wake clients to use blake3 for everything as well.

In this change blake3 is just used for the hash of the request which is completely internal to the implementation and is not exposed in anyway to the client.

rust/gsc/src/main.rs
```diff
@@ -1,27 +1,320 @@
...
+use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use tracing;
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Serialize)]
+struct VisibleFile {
+    path: String,
+    hash: [u8; 32],
+}
```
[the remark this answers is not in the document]
JakeSiFive: heh, figured you might say that lol

rust/gsc/src/main.rs
```diff
@@ -1,27 +1,320 @@
...
+                    .filter(job::Column::Hash.eq(hash))
+                    .one(txn)
+                    .await?
+                else {
+                  return Ok(ReadJobResponse::NoMatch);
+                };
+
+                let output_files = output_file::Entity::find()
+                    .filter(output_file::Column::JobId.eq(matching_job.id))
+                    .all(txn)
```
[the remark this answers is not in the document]
JakeSiFive: I'll give a shot!

rust/gsc/src/main.rs
JakeSiFive: `state` would be moved in that case and then I wouldn't be able to share the connection with the other routes. If I didn't have the `move` keyword (which would take a reference to `state` instead) then the function wouldn't implement Send/Sync (or would have some bonkers lifetime shenanigans).

rust/gsc/src/main.rs
JakeSiFive: I'm very ok with whatever route name you want. My gripe is with other RESTful stuff.

rust/migration/src/m20220101_000001_create_table.rs
JakeSiFive: hidden info is a more generic name for `runner_hash` that we can use to cover more things.

rust/gsc/src/main.rs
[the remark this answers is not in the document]
JakeSiFive: ok so this is slightly more complicated it turns out. We spoke offline about it but I'll leave it here for later. TL;DR I was able to use lazy-loading via `find_related` to clean up the code a bit but the performance will not be different (Should be the same queries even). The reason that eager-loading is a bad idea is because it will duplicate the stdin/stdout which could be quite large. There are however ways to efficently do this in 2 qurries but its a bit of a headache so we don't want to do that right now. There are much lower hanging performance gains to be made when we decide to optimize.

rust/gsc/src/read_job.rs
```diff
@@ -0,0 +1,92 @@
...
+            Box::pin(async move {
+                let Some(matching_job) = job::Entity::find()
+                    .filter(job::Column::Hash.eq(hash))
+                    .one(txn)
+                    .await?
+                else {
+                  return Ok(ReadJobResponse::NoMatch);
+                };
+
+                let output_files = matching_job
```
V-FEXrt, line 31: can this be

```Rust
let stuff = matching_job
  .find_related(output_files::Entity)
  .find_related(output_symlink::Entity)
  .find_related(output_dir::Entity)
  .all(txn)
  .await?
```
?

I'm guessing the `.map` is the headache there?

JakeSiFive: I wasn't able to get that to work but maybe you can give it a try later.

PR #1337, opened 2023-07-07
lsp: Fix VSCode wrapper hack around lsp server

V-FEXrt:
In https://github.com/sifive/wake/pull/1328, I fixed an out of spec response. This caused an issue in the VSCode wrapper for the lsp as we detected the out-of-spec response and rewrote it to be in-spec.

Now that we are in-spec this hack is no longer needed and should be deleted


JakeSiFive (approved):
This is just upsetting

PR #1338, opened 2023-07-07
Add api key management

JakeSiFive:
So far nothing is doing authentication with this but we want the ability to deny writes to users that don't have an API key and we want to be able to manage those API keys which this does. Note that API keys are not meant to be a security mechanism. It is simply a means of managing the access of otherwise trusted users, revoke capabilities for problematic clients, etc...


rust/gsc/src/gsc/main.rs
```diff
@@ -15,7 +16,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            ErrorKind::Other,
+            format!(
+                "This gsc version expects {:?} additional migrations to be applied",
+                pending_migrations.len()
+            ),
+        );
+        tracing::error! {%err, "unperformed migrations, please apply these migrations before starting gsc"};
+        Err(err)?;
+    }
+    //Migrator::up(&connection, None).await?;
```
JakeSiFive: Oh didn't quite realize it but I also added this feature. Please just consider this an extra and don't make me split it out lol.

rust/gsc/src/gsc_tool/main.rs
```diff
@@ -0,0 +1,209 @@
...
+    if results.len() == 0 {
+        println!("{} is not a valid key", key);
+        std::process::exit(2);
+    }
+
+    // We only want to prompt the user if the user can type into the terminal
+    if atty::is(Stream::Stdin) {
+        let should_delete = Confirm::new("Are you sure you want to delete this key?")
+            .with_default(false)
+            .with_help_message(format!("key = {}, desc = {:?}", key, results[0].desc).as_str())
```
V-FEXrt, line 95: `results[0]` makes an implicit dependence on the fact that the key is unique in the DB. That's probably fine but if that ever changes then this code would delete more keys than the user expects.

May be  good to add another check fore `results.len() == 1`

PR #1339, opened 2023-07-07
Added use tracking of jobs

JakeSiFive:
This change adds tracking of uses, as is this is intended for eviction but we will probably want to add another table for statistics that will not have a forgin key in order to keep track of all historical uses.


rust/gsc/src/read_job.rs
```diff
@@ -1,12 +1,24 @@
...
-    ColumnTrait, DatabaseConnection, DbErr, EntityTrait, ModelTrait, QueryFilter, TransactionTrait,
+    ActiveModelTrait, ActiveValue, ColumnTrait, DatabaseConnection, DbErr, EntityTrait, ModelTrait,
+    QueryFilter, TransactionTrait,
 };
 use std::sync::Arc;
 use tracing;
 
-use crate::types::{Dir, File, ReadJobPayload, ReadJobResponse, Symlink};
+#[tracing::instrument]
+async fn record_use(job_id: i32, conn: Arc<DatabaseConnection>) {
```
V-FEXrt, line 13: I think functions like this are better  written as extentions (I think in Rust they are called Impls?). Basically add this as a member function of  `Job`. Then you can do `matching_job.record_use(conn)`

Ideally we should avoid passing  around parts of a model and instead just act directly on the model

rust/gsc/src/read_job.rs
```diff
@@ -18,14 +30,14 @@ pub async fn read_job(
 
     let result = conn
         .as_ref()
-        .transaction::<_, ReadJobResponse, DbErr>(|txn| {
+        .transaction::<_, (i32, ReadJobResponse), DbErr>(|txn| {
```
V-FEXrt, line 33: This would become`.transaction::<_, (job::Entity, ReadJobResponse), DbErr>`

rust/gsc/src/read_job.rs
```diff
@@ -62,25 +74,39 @@ pub async fn read_job(
...
     match result {
-        Ok(result) => Json(result),
+        Ok((job_id, response)) => {
+            // If we get a match we want to record the use but we don't
+            // want to block sending the response on it so we spawn a task
+            // to go do that.
+            if let ReadJobResponse::Match { .. } = response {
+                let shared_conn = conn.clone();
+                tokio::spawn(async move {
+                    record_use(job_id, shared_conn).await;
```
V-FEXrt, line 105: and this would be `job.record_use(shared_conn)`

rust/migration/src/m20230707_144317_record_uses.rs
```diff
@@ -0,0 +1,53 @@
...
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(JobUses::Table)
+                    .if_not_exists()
```
V-FEXrt, line 15: I may have missed this before but you shouldn't use `.if_not_exists()` in a migration if that table wouldn't have been created already by a previous migration. If this table already exists then we should fail hard because the migrator is the only one with authority to create this table and we can look at the migration history and assert it hasn't been created in the chain

V-FEXrt: (this is another one of those things that won't ever actually come up but its another safety rail)

V-FEXrt (changes requested):

rust/gsc/src/read_job.rs
JakeSiFive: Ok, I have to use a trait to do this but this is a not-uncommon pattern to use (this is why you see traits that end in Ext a lot)

JakeSiFive: So ...actually... this is difficult but it is doable....but like wait a second.

1) You can't define an impl outside of the crate that the struct was defined in
2) You *can* define a trait however and this is a not-uncommon pattern. You'll often see traits that end in `Ext` that do exactly this. but....
3) You can't add async functions inside of traits because `async fn foo(...) -> T` desugars into `fn foo(...) -> impl Future<Output = T>` which works fine in an `impl` because no associated type is needed but that's an implicit associated type if its in a trait buuut.....
4) There's a trait called `async-trait` that handles this for you and makes a bespoke associated type for each method aaaaaaand...
5) Tyler and the rest of the compiler team are working on getting async traits in by default so that you don't need a macro for it buuuuut....
6) I don't think it will be stable for a while

We could do the async-trait thing and wait for a proper language feature OR we could avoid all that hassle and just pass a `job::Model&` instead of part of it....however either way we have to consider something else:

There's a second issue to contend with:
1) `i32` implements copy but `job::Model` does not (nor should it). It should however implement `clone` but `clone` could be extremely expensive because of stdout/stderr
2) I could however move this into an arc (cheap enough) and then call this in the spawned task as `record_use(matching_job_arc.as_ref(), new_conn)`

I think making it a regular function instead of a method and going with option 2) following that is the best compromise

Funny how rust complicates things like this huh?

rust/gsc/src/read_job.rs
JakeSiFive: Arc<job::Model> if we take my compromise above I think...actully even if we don't compromise it would need to be an Arc

rust/migration/src/m20230707_144317_record_uses.rs
JakeSiFive: Yeah that makes sense. I can go back and modify them to fit that in a follow up. For some reason it was the sea-orm way to do things

rust/gsc/src/read_job.rs
V-FEXrt: I see. hmmm. Rust does complicate this a bit.

I think I want to avoid polluting stuff with arc just yet and the copy would be too expensive. 

This is easily fixed in the future so lets just leave this as is until things have shaken  out a bit in `async-trait` and also we more generally have a feeling for what rust-in-wake looks like

rust/migration/src/m20230707_144317_record_uses.rs
V-FEXrt: SGTM

rust/gsc/src/read_job.rs
JakeSiFive: Also reading Tyler's blog post it turns out that I was slightly wrong on point 4 there. async-trait actully creates a `Pin<Box<dyn Future<Output = T>>>` for each method which is super-duper slow. That probably isn't too big of a cost however given that this is a spawned write task.

PR #1340, opened 2023-07-10
lsp: Fix subtle lsp bugs

V-FEXrt:
https://github.com/sifive/wake/pull/1307 introduces a couple subtle bugs that snuck through code review and aggressively broke the VSCode extension.

The VSCode extension has several more layers of indirection, a different build target and other subtle differences that direct clients of the native lsp don't exercise. This means its not sufficient to test the extension against native clients. It must also always be tested against VSCode

### DIFF
--- a/tools/lsp-wake/main.cpp
+++ b/tools/lsp-wake/main.cpp
@@ -93,7 +93,7 @@ class LSPServer {
   };
 
   MethodResult processRequest(const std::string &requestString) {
-    wcl::log::info(requestString.c_str())({{"rpc", "rx"}});
+    wcl::log::info("%s", requestString.c_str())({{"rpc", "rx"}});
 
     // Parse that requestString as JSON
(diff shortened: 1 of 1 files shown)


tools/lsp-wake/main.cpp
```diff
@@ -93,7 +93,7 @@ class LSPServer {
   };
 
   MethodResult processRequest(const std::string &requestString) {
-    wcl::log::info(requestString.c_str())({{"rpc", "rx"}});
+    wcl::log::info("%s", requestString.c_str())({{"rpc", "rx"}});
```
JakeSiFive, line 96: I think we should understand 1) why the warning attributes we added didn't tell us about this and 2) Why this is even a bug in WASM in the first place as it should (at runtime anyway) be valid.

JakeSiFive (approved):

PR #1342, opened 2023-07-10
Add a robust configuration system for the GSC

JakeSiFive:
This change adds the ability to configure the GSC in 3 ways

1) via environment variables
2) overriding environment variables via a config file (which can be specified on the command line)
3) overriding either of those two options via command line arguments

When deploying it is intended that 1) be used, for developing it is intended that 2) be the primary mechanism and a default .config.json has been added for this purpose. For special cases 3) can be used


V-FEXrt (approved):
Not required, but would be nice to have a debug option to dump the current calculated config

PR #1345, opened 2023-07-11
Remove VSCode hack for setting the stdLib location

V-FEXrt:
Previously we had a middleman capture, save, and replay requests to the server in order to ask the client for the stdlib path before actually setting up the server. This is strange and certainly not what the LSP Spec intends.

Instead we should use the optional `initializationOptions` key spelled out by the spec for this exact purpose. This PR makes that change. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

### DIFF
--- a/extensions/vscode/lsp-server/src/common.ts
+++ b/extensions/vscode/lsp-server/src/common.ts
@@ -66,7 +66,6 @@ const InternalError  = -32603;
 
 export function prepareConnection(connection: Connection, isWeb: Boolean) {
     const getResponse = async <T, E>(request: string): Promise<T | ResponseError<E>> => {
-        console.log("-> " + request);
         let c_str = await lock.use(async () => { // synchronise requests to wasm
             return await lspModule?.processRequest(request) ?? null;
         });
@@ -81,8 +80,6 @@ export function prepareConnection(connection: Connection, isWeb: Boolean) {
 
         let response = lspModule.toString(c_str);
         lspModule._free(c_str); // free the string mallocced by c++
-
-        console.log("<- " + response);
 
         let result = JSON.parse(response);
 
@@ -104,22 +101,10 @@ export function prepareConnection(connection: Connection, isWeb: Boolean) {
     // Create lsp handlers
     connection.onInitialize(async (params: InitializeParams): Promise<InitializeResult | ResponseError<InitializeError>> => {
         lspModule = await wakeLspModule(connection);
-
-        // instantiate the server with a stdLib falling back to the default
-        // when none is provided
-        let path = params.initializationOptions?.stdLibPath;
-        if (path === null) {
-            lspModule?._instantiateServer();
-        } else {
-            // TODO: We should move the handling of the stdlib location
-            // directly into the LSP, but that is a significant refactor
-            lspModule?.instantiateServerCustomStdLib(path);
-        }
+        lspModule?._instantiateServer();
 
         let response: InitializeResult | ResponseError<InitializeError> = await getResponse(createLspJson('initialize', params));
         // TODO: impl & test invalid stdlib
-        //
-
         return response;
     });
 
--- a/extensions/vscode/lsp-server/src/common.ts
+++ b/extensions/vscode/lsp-server/src/common.ts
@@ -65,63 +65,26 @@ function createLspJsonNoParams(method: string): string {
 const InternalError  = -32603;
 
 export function prepareConnection(connection: Connection, isWeb: Boolean) {
-    let initializeRequest = '';
-
     const getResponse = async <T, E>(request: string): Promise<T | ResponseError<E>> => {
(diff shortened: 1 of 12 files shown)


extensions/vscode/lsp-server/src/common.ts
```diff
@@ -140,7 +100,12 @@ export function prepareConnection(connection: Connection, isWeb: Boolean) {
 
     // Create lsp handlers
     connection.onInitialize(async (params: InitializeParams): Promise<InitializeResult | ResponseError<InitializeError>> => {
-        return await getResponse(createLspJson('initialize', params));
+        lspModule = await wakeLspModule(connection);
+        lspModule?._instantiateServer();
+
+        let response: InitializeResult | ResponseError<InitializeError> = await getResponse(createLspJson('initialize', params));
+        // TODO: impl & test invalid stdlib
```
V-FEXrt, line 107: I think I still need to do this, unclear if VSCode will extract the notification for us or if it needs to be manually extracted as it was before

tools/lsp-wake/main.cpp
```diff
@@ -323,29 +320,42 @@ class LSPServer {
...
+      }
+    }
+
+    bool stdLibValid = is_readable((stdLibPath + "/core/boolean.wake").c_str());
+    if (!stdLibValid) {
+      notifyAboutInvalidSTDLib(methodResult, stdLibPath);  // set notification
       methodResult.response = JSONConverter::createInitializeResultInvalidSTDLib(receivedMessage);
-    } else {
-      methodResult.response = JSONConverter::createInitializeResultDefault(receivedMessage);
+      return methodResult;
```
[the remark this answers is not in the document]
JakeSiFive: Wake doesn't currently function at all without a stdlib because we use things like `List` and `JValue` in our prims so you can do which ever is easiest to implement but I agree that we should do one or the other.

tools/lsp-wake/main.cpp
```diff
@@ -641,13 +617,12 @@ int main(int argc, const char **argv) {
...
-    instantiateServer(argv[1]);
-  } else {
-    instantiateServer(wcl::make_canonical(find_execpath() + "/../../share/wake/lib"));
-  }
+  instantiateServerImpl();
+
   // Process requests until something goes wrong
   lspServer->processRequests();
+
+  delete lspServer;
```
JakeSiFive, line 625: 👀 something something unique_ptr or something something defer?

extensions/vscode/lsp-server/src/common.ts
JakeSiFive: I'm not sure what this code is doing, could you explain it a a bit more?

tools/lsp-wake/main.cpp
V-FEXrt: Normally yes, but there isn't a good owner of the unique_ptr. There isn't really an entry point for this code in VSCode

tools/lsp-wake/main.cpp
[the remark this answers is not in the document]
V-FEXrt: alright perfect, I'll just not support it then. Which is what this PR does

extensions/vscode/lsp-server/src/common.ts
V-FEXrt: `await getResponse(createLspJson('initialize', params));` is just exchanging a message with the lsp server.

`response` _may_ have a `notification` set if the stdLib being sent to the server isn't valid. In that case we need to show the notification to the user.

In the old code we manually extracted the notification and replayed it to the user. My comment here is just that I need to test that out and see if we actually need to do the replay or if that was some weird quirk of the hack. In _theory_, I'd expect VSCode to just handle the notification automatically but I need to confirm

V-FEXrt: Okay, I just tested this and it actually just works without the manual replay. So we are actually removing two hacks in the PR :)

tools/lsp-wake/main.cpp
JakeSiFive: is `lspServer` global? Where does it get allocated?

JakeSiFive: Then it can be a global std::unique_ptr

V-FEXrt: :pikachu-face:

V-FEXrt: done :)

JakeSiFive (approved):

V-FEXrt:
Forgot to call it out, but this is technically a breaking change for a _very_ small  subset of users.

Anyone who
 - uses the LSP
 - outside of VSCode
 - with a custom stdlib

will need to move the setting from being a command line argument to being an initialization option. 

In reality, I don't think anyone falls into that category but it is possible

PR #1346, opened 2023-07-11
Move the job cache log to a dedicated file for cleanup

V-FEXrt:
Moves log outputs from `.stdout`/`.stderr` to `.cache.<YYYY-MM-DD>.log`. This was done to make cleaning up old logs easier but also has  the side effect of making our other logging code simplier as we don't have to worry about polluting stdout/stderr

This PR doesn't implement the actual cleanup, that will be a separate PR

### DIFF
--- a/src/job_cache/job_cache.cpp
+++ b/src/job_cache/job_cache.cpp
@@ -71,6 +71,8 @@ static bool daemonize(std::string dir) {
     return false;
   }
 
+  wcl::log::info("fork1: success")();
+
   {
     // Replace stdin with /dev/null so we can't receive input
     auto null_fd = wcl::unique_fd::open("/dev/null", O_RDONLY);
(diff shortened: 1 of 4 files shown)


src/job_cache/job_cache.cpp
```diff
@@ -105,10 +107,6 @@ static bool daemonize(std::string dir) {
     replace_fd(STDERR_FILENO, err_log_fd->get());
   }
 
-  wcl::log::clear_subscribers();
-  wcl::log::subscribe(std::make_unique<wcl::log::FormatSubscriber>(std::cout.rdbuf()));
-  wcl::log::info("Reinitialized logging for job cache daemon")();
-
```
JakeSiFive, line 111: Are we sure we're not changing what gets logged where with this?

JakeSiFive: Ok I thought through it and yeah it does mean that many error logs might show up in wake.log instead of in the shared cache but I think that's ok since this is launching code.

JakeSiFive (approved):
Ok looks good. It took me a while to think through how the location of the logs would change here but I think I understand it all now.

PR #1347, opened 2023-07-11
Create a log viewer; switch to structured logging

JakeSiFive:
It's been getting harder and harder to understand the wake logs as shared caching has gotten more complex, and has seen larger more real world use cases. This PR creates a tool (shipped as wake-log-viewer in a wake install) that can take several log files (more on the log file format in a second) and interleave them by their timestamp, and color them so that we know the process from which a given log line came from.

This tool can be used in combination with `grep` (ideally with `--color=never`) and `less -r` in order to efficiently search the logs. Later I'll add more filtering options to the log viewer but this solves the immediate issues we've been having of it being difficult to make sense of all the logs so I figured I'd put this up so we can always have it.

Additionally this changes our logging (at least for non-user facing logs) to output to json instead. This means that its hard (but not impossible) for a human to read the logs without a tool like this one. Additionally logs now come with both the time in seconds but also in nano seconds so that logs can be precisely ordered.

Lots of future features could be added but this does like 80% to 90% of what we want it to do.

Example of what collated logs look like over multiple wake runs:

Screen Shot 2023-07-11 at 3 51 55 PM


rust/log_viewer/test.json
```diff
@@ -0,0 +1,20 @@
+{"pid": 23, "time": "2023-07-10T11:11:11", "level":"error", "recv_pid":"this is a message", "message":"this is a log message to be viewed"}
```
JakeSiFive, line 1: this was an accident and needs to be deleted

build.wake
```diff
@@ -103,10 +103,25 @@ def all variant =
 
     Pass (flatten x)
 
+def buildWakeLogViewer Unit =
+    require Some path = getenv "PATH"
+    else failWithError "Could not find cargo on path"
+
+    require Pass log_viewer =
+        makeExecPlan (whichIn path "cargo", "build", "--release",) Nil
```
JakeSiFive, line 111: Its quite annoying depending on PATH here like this but I'm not sure what else to do right now. Ideally we'd download cargo within wake as part of the build I suppose? That might alleviate the docker head aches I'm having right now anyway.

tools/job-cache/main.cpp
```diff
@@ -20,6 +20,7 @@
 #define _POSIX_C_SOURCE 200809L
 
 #include <job_cache/daemon_cache.h>
+#include <json/json5.h>
```
JakeSiFive, line 23: Left over from the rebase

src/wcl/tracing.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "tracing.h"
 
+#include <chrono>
```
JakeSiFive, line 24: left over from trying to use C++'s chrono lib

src/job_cache/job_cache.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "job_cache.h"
 
+#include <json/json5.h>
```
JakeSiFive, line 24: Left over from rebase

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+use gumdrop::Options;
+use palette::{convert::FromColorUnclamped, encoding::Srgb, rgb::Rgb, Lab};
+use serde::{Deserialize, Serialize};
+use serde_aux::prelude::*;
+use serde_json::Value;
+use std::fs::File;
+use std::io::BufRead;
+use std::{collections::HashMap, io::BufReader};
+
+// TODO: Shorten messages to fit the screen
```
V-FEXrt, line 13: Done right?

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+// While on one hand this is extremly simple, on another hand it uses
+// some obtuse math concepts. There's a color space called CIELAB
+// that behaves a bit better when sampling from it. If you sample two points
+// a certain distance apart in CIELAB, the perceived distance is roughly proportional
+// to the distance between those points. So if we can sample a lot of points that stay
+// relativelly distinct from each other we can consequently get nice distinct colors.
+// Additionally CIELAB has a sort of "brightness" field so that we can keep a relativelly
```
V-FEXrt, line 45: typo: relativelly

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+// consistent brightness.
+//
+// In order to sample distinct points we use quasi-monte color methods, specifically
+// low-discrepancy sequences. These give us consistently spaced out values in any dimensions
+// and the minimum area/volume between anytime sampled points is inversely related to the
+// number of points sampled. So if we only sample a few points, our colors will be very
+// distinct, as we sample more and more points they will get closer and closer togethor but
```
V-FEXrt, line 52: typo: togethor

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+// they should do a quite good job of staying as far apart as possible. We're using an additive
+// recurence which does quite a good job for this use case and is very simple.
+fn generate_distinct_colors() -> impl Iterator<Item = Color> {
+    //let params = JoeKuoD6::minimal();
```
V-FEXrt, line 56: commented code

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+    AdditiveRec {
+        x: 0.0f32,
+        y: 0.0f32,
+        z: 0.0f32,
+    }
+    .map(|point| {
+        // Vary luminosity from 60 to 90. I just played around with it
+        // and found this to be a good range
+        let l = point.0 * 30.0f32 + 60.0f32;
+        //let l = 75.0f32;
```
V-FEXrt, line 67: here too

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+        // to crossterm's color type
+        Color::Rgb {
+            r: rgb.red,
+            g: rgb.green,
+            b: rgb.blue,
+        }
+    })
+}
+
+fn neg_one() -> i32 {
```
V-FEXrt, line 94: This could use a comment

rust/log_viewer/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+    let Some(level) = level
+    else {
+      return Color::White;
+    };
+
+    match level {
+        x if x == "info" => Color::Grey,
+        x if x == "warning" => Color::Yellow,
+        x if x == "error" => Color::Red,
+        _ => Color::White,
```
V-FEXrt, line 122: This seems like a really weird way to spell this out. does 

```

    match level {
        "info" => Color::Grey,
        "warning" => Color::Yellow,
        "error" => Color::Red,
        _ => Color::White,
```

not work?

rust/log_viewer/test.json
V-FEXrt: or turned into a unittest :)

rust/log_viewer/src/main.rs
JakeSiFive: I tried that first and there was an issue with it. I didn't look into the specifics of how rust does that and instead just did this because I knew it would work 🤷

build.wake
```diff
@@ -103,10 +103,52 @@ def all variant =
...
+        whichIn path "cargo"
+
+    "{cargo_home}/bin/cargo"
+
+# Builds a cargo project, the Cargo.toml file must be listed as well as the outputs of intrest.
+# All source files must be listed as well. Cargo will not reuse work because the sandbox
+# will not contain its extra info, inside of target.
+# TODO: Add an environment variable to opt-in to localRunner cargo builds if we detect
+#       existing outputs in {pathToCrate}/target/release
+def buildWithCargo (sources: List Path) (pathToCrate: String) (outputs: List String) =
```
mmjconolly, line 121: Should this go in `share/wake/lib/rust_wake/cargo.wake`

build.wake
```diff
@@ -103,10 +103,52 @@ def all variant =
...
+        | setPlanLabel "cargo: {pathToCrate}"
+        | setPlanDirectory pathToCrate
+        | setPlanFnOutputs (\_ map ("{pathToCrate}/target/release/{_}") outputs)
+        | setPlanEnvironment ("PATH={path}",)
+        | runJobWith defaultRunner
+        | getJobOutputs
+
+    Pass outputs
+
+def buildWakeLogViewer Unit =
```
mmjconolly, line 136: Should this go in `rust/log_viewer/build.wake`?

build.wake
```diff
@@ -103,10 +103,52 @@ def all variant =
...
+def buildWithCargo (sources: List Path) (pathToCrate: String) (outputs: List String) =
+    require Some path = getenv "PATH"
+    else failWithError "PATH was not set; cargo will not be able to find rustc"
+
+    require Pass outputs =
+        makeExecPlan (findCargoPath Unit, "build", "--release",) sources
+        | setPlanLabel "cargo: {pathToCrate}"
+        | setPlanDirectory pathToCrate
+        | setPlanFnOutputs (\_ map ("{pathToCrate}/target/release/{_}") outputs)
+        | setPlanEnvironment ("PATH={path}",)
```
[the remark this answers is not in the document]
JakeSiFive: We set path to a constant (which is good because setting it to an env-var would introduce non-determinism)

```
# A topic used to globally add directories to the default PATH of a Plans
export topic path: String

# Default search path for new Plans
export def path: String =
    catWith ":" (subscribe path | distinctBy scmp)

# Add PATH to the default environment
publish environment =
    "PATH={path}", "TERM=xterm-256color", Nil
```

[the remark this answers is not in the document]
mmjconolly: Rereading, I think this is going to give the job whatever the exterior-to-wake PATH is

Could we instead inject only the bit to find cargo?
```suggestion
        | prependPlanPath (dirname (findCargoPath Unit))
```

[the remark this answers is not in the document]
JakeSiFive: Yeah that would be more hygienic. I'm still working on getting all the docker stuff situated unfortunately.

JakeSiFive:
ok @mmjconolly @V-FEXrt this is ready for review again and is actually working 😭 the largest contribution here was apparently not the log viewer itself but getting rust working in CI

.github/workflows/dockerfiles/alpine
```diff
@@ -1,9 +1,16 @@
 FROM alpine:3.14.0
 
-RUN apk add m4 g++ make pkgconf git tar xz gmp-dev re2-dev sqlite-dev fuse-dev ncurses-dev dash sqlite-static ncurses-static linux-headers
+RUN apk add m4 g++ make pkgconf git tar xz gmp-dev re2-dev sqlite-dev fuse-dev ncurses-dev dash sqlite-static ncurses-static linux-headers curl
```
mmjconolly, line 3: Is curl used?

JakeSiFive: oh good catch, I think I was at one point but then I switch to wget, let me go purge curl

share/wake/lib/rust_wake/cargo.wake
```diff
@@ -0,0 +1,45 @@
...
+
+target findCargoPath Unit =
+    require Some cargo_home = getenv "CARGO_HOME"
+    else
+        require Some path = getenv "PATH"
+        else "cargo"
+
+        whichIn path "cargo"
+
+    "{cargo_home}/bin/cargo"
```
mmjconolly, line 26: I think we should fail if we cant find `cargo` anywhere

.github/workflows/dockerfiles/rocky-8
```diff
@@ -3,7 +3,14 @@ FROM rockylinux/rockylinux:8
...
 RUN rpmdev-setuptree
 
+
 WORKDIR /build
+ENV VERSION=1.70.0
+ENV RUST_FOLDER="rust-$VERSION-x86_64-unknown-linux-gnu"
+ENV RUST_TAR="$RUST_FOLDER.tar.gz"
+RUN wget "https://static.rust-lang.org/dist/$RUST_TAR"
+RUN tar -xf $RUST_TAR
+RUN sh ./$RUST_FOLDER/install.sh --verbose
```
mmjconolly, line 16: Inconsistent end-of-file for some dockerfiles

share/wake/lib/rust_wake/cargo.wake
```diff
@@ -0,0 +1,45 @@
...
+
+package rust
+
+target findCargoPath Unit =
+    require Some cargo_home = getenv "CARGO_HOME"
+    else
+        require Some path = getenv "PATH"
+        else failWithError "neither $PATH nor $CARGO_HOME is set; cannot find cargo"
+
+        Pass (whichIn path "cargo")
```
mmjconolly, line 24: What does `whichIn` do when `cargo` is not on the `$PATH`?

share/wake/lib/rust_wake/cargo.wake
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package rust
+
+target findCargoPath Unit =
+    require Some cargo_home = getenv "CARGO_HOME"
```
mmjconolly, line 19: This makes `CARGO_HOME` be taken in preference to `PATH`.
While unlikely, I think that would be unexpected behavior to shell users if they were inconsistent

share/wake/lib/rust_wake/cargo.wake
JakeSiFive: It just returns "cargo" which is the behavior that I was mimicking before. You wind up getting an error later on from execve

share/wake/lib/rust_wake/cargo.wake
JakeSiFive: This is the whole point of CARGO_HOME though, it's used to override your cargo location in the cargo echo-system.

mmjconolly:
I haven't only read the infra changes (ie, not rust/c++) and they look ok to me

share/wake/lib/rust_wake/cargo.wake
JakeSiFive: To match the home crate we'd actually prefer $HOME/.cargo over $PATH: https://docs.rs/home/latest/home/fn.cargo_home.html and this is what the cargo crate recommends as well: https://doc.rust-lang.org/cargo/guide/cargo-home.html

(the home crate is a part of the cargo/rust-lang project and can also be used to find rustup/rustc in the default expected way)

rust/log_viewer/src/main.rs
V-FEXrt: still a typo. should be `relatively`

rust/log_viewer/src/main.rs
V-FEXrt: hmmmmmm we should probably look into this or at least have an idea _why_ it doesn't work. Doesn't have to be this PR but I'm not super happy with working around compiler errors without knowing why

share/wake/lib/rust_wake/cargo.wake
```diff
@@ -0,0 +1,45 @@
...
+    else
+        require Some path = getenv "PATH"
+        else failWithError "neither $PATH nor $CARGO_HOME is set; cannot find cargo"
+
+        Pass (whichIn path "cargo")
+
+    Pass "{cargo_home}/bin/cargo"
+
+# Builds a cargo project, the Cargo.toml file must be listed as well as the outputs of intrest.
```
V-FEXrt, line 28: typo: `intrest` -> `interest`

share/wake/lib/rust_wake/cargo.wake
```diff
@@ -0,0 +1,45 @@
...
+# All source files must be listed as well. Cargo will not reuse work because the sandbox
+# will not contain its extra info, inside of target.
+# TODO: Add an environment variable to opt-in to localRunner cargo builds if we detect
+#       existing outputs in {pathToCrate}/target/release
+export def buildWithCargo (sources: List Path) (pathToCrate: String) (outputs: List String) =
```
V-FEXrt, line 33: nit: maybe `cargoBuild` instead of  `buildWithCargo`? Not sure, don't feel super strongly but I'm not the biggest fan of `with` in the name

V-FEXrt (approved):

rust/log_viewer/src/main.rs
JakeSiFive: I just needed to add `level.as_str()` all fixed now

PR #1348, opened 2023-07-12
Cleanup old logs on cache start

V-FEXrt:
Deletes any files in `cache_dir` of shape `.cache.<YYYY-MM-DD>.log` where `YYYY-MM-DD` is more that 4 days from the current date

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -73,7 +73,7 @@ static std::unique_ptr<std::ofstream> initialize_logging() {
   std::vector<std::string> to_delete;
   for (const auto &entry : *res) {
     if (!entry) {
-      wcl::log::warning("bad /proc/self/fd/ entry: %s\n", strerror(entry.error()))();
+      wcl::log::warning("bad file entry: error = %s\n", strerror(entry.error()))();
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -47,17 +47,66 @@ namespace {
...
+  auto res = wcl::directory_range::open(".");
+  if (!res) {
+    wcl::log::warning("Failed to open cwd to cleanup log files")();
+    return log_file;
+  }
+
+  std::vector<std::string> to_delete;
+  for (const auto &entry : *res) {
+    if (!entry) {
+      wcl::log::warning("bad /proc/self/fd/ entry: %s\n", strerror(entry.error()))();
```
JakeSiFive, line 76: This warning needs to be changed, I think you took this from shim-wake lol

src/job_cache/daemon_cache.cpp
```diff
@@ -47,17 +47,66 @@ namespace {
...
+
+    std::string day = entry->name.substr(filename_prefix_len, str_fmt_len);
+    struct tm prev_tm = {0};
+    strptime(day.c_str(), str_fmt, &prev_tm);
+
+    double diff_secs = difftime(today, mktime(&prev_tm));
+
+    int four_days = 60 /*secs*/ * 60 /*mins*/ * 24 /*hours*/ * 4 /*days*/;
+    if (diff_secs > four_days) {
+      to_delete.push_back(entry->name);
```
JakeSiFive, line 100: nit: if you have time to get to this modtime would be better because 1) it would actually be 3-days since it was touched and 2) it would be more robust than parsing from the file name

JakeSiFive (approved):
Happy to land as is (modulo the /proc/self/fd thing) but I do have a nit you could look at if you have time, if not don't worry about it.

PR #1352, opened 2023-07-18
Increase local shared cache robustness to daemon being killed

JakeSiFive:
This change is meant to do the following

1) If a write to a client fails, the daemon keeps running and just closes its connection to that client.
2) If a write to the daemon fails, the client treats that as it would read failures and the client should restart

I don't have a good way to do auto-mated testing of this yet but I manually killed the job-cache many times and everything continued to work perfectly. I don't consider this change complete without automated testing but its hard and I wanted to get these fixes out there first so we can do another release with them in it. All existing tests pass, and manually killing the daemon repeatedly causes no issues that I can observe yet

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -1083,15 +1083,15 @@ void DaemonCache::handle_new_client() {
   wcl::log::info("new client connected: %d", accept_fd)();
 }
 
-void DaemonCache::close_client(int clinet_fd) {
-    wcl::log::info("closing client fd = %d", client_fd)();
-    poll.remove(client_fd);
-    close(client_fd);
-    message_parsers.erase(client_fd);
-    if (message_parsers.empty()) {
-      exit_now = true;
-      wcl::log::info("All clients disconnected, exiting.")();
-    }
+void DaemonCache::close_client(int client_fd) {
+  wcl::log::info("closing client fd = %d", client_fd)();
+  poll.remove(client_fd);
+  close(client_fd);
+  message_parsers.erase(client_fd);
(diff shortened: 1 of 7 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -1111,7 +1122,16 @@ void DaemonCache::handle_msg(int client_fd) {
...
       FindJobRequest req(json.get("params"));
       FindJobResponse res = read(req);
-      send_json_message(client_fd, res.to_json());
+      auto write_error = send_json_message(client_fd, res.to_json());
+      if (write_error) {
+        wcl::log::error("DaemonCache::handle_msg(): send_json_message(%d): %s", client_fd,
+                        strerror(*write_error))
+            .urgent()();
+        wcl::log::error(
+            "DaemonCache::handle_msg(): failed to send client a response, closing cleint")
```
V-FEXrt, line 1131: nit:   `cleint` -> `client`

src/job_cache/job_cache.cpp
```diff
@@ -16,11 +16,10 @@
  */
 
 // Open Group Base Specifications Issue 7
+#include "types.h"
```
V-FEXrt, line 19: nit: This should be below 

```c++
#define _XOPEN_SOURCE 700
```

no?

V-FEXrt (approved):
LGTM  % a couple nits

src/job_cache/job_cache.cpp
JakeSiFive: For some reason my editor keeps using the wrong clang-format invocation that does this but re-doing clang-format doesn't fix this, its annoying.

PR #1356, opened 2023-07-20
Don't force a split inside of interpolated strings

V-FEXrt:
When a `|` was inside a string interpolation is  was being forced into multilining. Don't do that.

Before 

```
"{one | two} three"
```
->
```
"{one 
   | two} three"
```

After:

```
"{one | two} three"
```
->
```
"{one | two} three"
```

### DIFF
--- a/tools/wake-format/emitter.cpp
+++ b/tools/wake-format/emitter.cpp
@@ -1666,7 +1666,7 @@ wcl::doc Emitter::walk_import(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_interpolate(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  MEMO_RET(walk_placeholder(ctx.binop(), node));
 }
 
 wcl::doc Emitter::walk_kind(ctx_t ctx, CSTElement node) {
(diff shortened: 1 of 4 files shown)


mmjconolly (approved):
rubber stamp!

(this is more of a show of support, I don't actually understand how it fixes it)

tools/wake-format/emitter.cpp
```diff
@@ -1666,7 +1666,7 @@ wcl::doc Emitter::walk_import(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_interpolate(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
+  MEMO_RET(walk_placeholder(ctx.binop(), node));
```
JakeSiFive, line 1669: lol this is kind of a dirty trick, we should add a TODO: to generalize this to a "don't split" member instead because thats what it really is.

JakeSiFive (approved):
Approved with a nit

tools/wake-format/emitter.cpp
V-FEXrt: Yeah, the  name is incorrect, and pretty much was from the beginning 😓

PR #1358, opened 2023-07-21
Maintain `String` null termination invariant

V-FEXrt:
Wake's `String` type internally maintains that its raw string pointer is a null terminated string so that the pointer may be used directly. However in some cases the internal pointer is filled manually outside of the class. In this case the author must remember to manually null  terminate the string or it will be inconsistent to various parts of the runtime. This may manifest itself as a segfault but may alternatively just be a string with a couple extra bytes at the end if a convenient NULL byte is nearby. 

This change fixes a case where the null terminator was forgotten but in the long term we should consider preventing this case all together by making the internal pointer read-only

### DIFF
--- a/src/runtime/string.cpp
+++ b/src/runtime/string.cpp
@@ -210,6 +210,8 @@ static PRIMFN(prim_read) {
       runtime.heap.reserve(need_pass);
 
       String *out = String::claim(runtime.heap, size);
+      out->c_str()[size] = 0;
+
       t.seekg(0, t.beg);
       t.read(out->c_str(), out->size());
       if (t) RETURN(claim_result(runtime.heap, true, out));
--- a/src/runtime/value.cpp
+++ b/src/runtime/value.cpp
@@ -83,7 +83,9 @@ String *String::claim(Heap &h, const std::string &str) {
 }
 
 String *String::claim(Heap &h, const char *str, size_t length) {
-  return new (h.claim(reserve(length))) String(str, length);
+  String* out = new (h.claim(reserve(length))) String(str, length);
+  out->c_str()[length] = 0;
+  return out;
 }
 
 String *String::alloc(Heap &h, size_t length) {
@@ -95,7 +97,9 @@ String *String::alloc(Heap &h, const std::string &str) {
 }
 
 String *String::alloc(Heap &h, const char *str, size_t length) {
-  return new (h.alloc(reserve(length))) String(str, length);
+  String* out = new (h.alloc(reserve(length))) String(str, length);
+  out->c_str()[length] = 0;
+  return out;
 }
 
 String *String::alloc(Heap &h, const char *str) { return alloc(h, str, strlen(str)); }
--- a/src/runtime/value.cpp
+++ b/src/runtime/value.cpp
@@ -83,7 +83,7 @@ String *String::claim(Heap &h, const std::string &str) {
 }
 
 String *String::claim(Heap &h, const char *str, size_t length) {
-  String* out = new (h.claim(reserve(length))) String(str, length);
+  String *out = new (h.claim(reserve(length))) String(str, length);
   out->c_str()[length] = 0;
   return out;
 }
@@ -97,7 +97,7 @@ String *String::alloc(Heap &h, const std::string &str) {
 }
 
 String *String::alloc(Heap &h, const char *str, size_t length) {
-  String* out = new (h.alloc(reserve(length))) String(str, length);
+  String *out = new (h.alloc(reserve(length))) String(str, length);
   out->c_str()[length] = 0;
   return out;
 }
--- a/src/runtime/value.cpp
+++ b/src/runtime/value.cpp
@@ -25,6 +25,7 @@
 #include <re2/re2.h>
 #include <string.h>
 
+#include <algorithm>
 #include <sstream>
 
 #include "optimizer/ssa.h"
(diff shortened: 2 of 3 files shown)


mmjconolly (approved):

src/runtime/string.cpp
```diff
@@ -210,6 +210,8 @@ static PRIMFN(prim_read) {
       runtime.heap.reserve(need_pass);
 
       String *out = String::claim(runtime.heap, size);
+      out->c_str()[size] = 0;
+
       t.seekg(0, t.beg);
       t.read(out->c_str(), out->size());
```
JakeSiFive, line 216: You should put this in this claim function

```
String *String::claim(Heap &h, size_t length) {
  return new (h.claim(reserve(length))) String(length);
}
```

turns into

```
String *String::claim(Heap &h, size_t length) {
  auto out = new (h.claim(reserve(length))) String(length);
  out->c_str()[length] = '\0';
  return out;
}
```

like-wise for this function
```
String *String::alloc(Heap &h, size_t length) {
  return new (h.alloc(reserve(length))) String(length);
}
```

The other claim/alloc implementations appear to account for these issues.

V-FEXrt: This is closer to an "API" fix. I personally think the correct API fix is to instead make `out->c_str()` `const` only and to add a `fill(src,len)`/`strcpy(src,len)` method to String that ensures the null byte is always written when externally filling a `String`. All the `String` constructors correctly that upkeep the null byte and we could make fill/strcpy to so as well. The null termination seems a bit unrelated to claim/alloc

JakeSiFive: The contract for string that these uphold is that there is always a null-terminator as long as you don't write outside of the given range. These make it so that the claim/alloc functions always keep that promise.

V-FEXrt: Right, but you shouldn't even be allowed to write outside the given range. Making `c_str` a editable pointer is a clear footgun without any protections. What is the cost of a dedicated `fill` function that makes writing to the raw pointer preferable?

I'm imagining

```
int fill(const char* src, size_t count) {
  assert(count < length);
  memcpy(c_str(), src, count);
  c_str()[count] = 0
}
```

which is also more correct as the null byte always immediately follows the end of the string. If you go the `claim`/`alloc` route and write a string smaller than the reserved space you end up with the opportunity for garbage at the end of the string

JakeSiFive: I'm not against that, but you still need the uninit constructor and that constructor should also maintain the invariant. If you do that you still want my suggested change as well.

V-FEXrt: Which constructor do you mean? I spent time going through them all on Friday and they all maintained the invariant. The only thing that could mess stuff was manually editing the underlying pointer (which `fill` prevents)

JakeSiFive: Spoke off line to clarify, by "constructor" I mean the claim/alloc static member functions. Ashley pointed out that they were just in-place new constructing from an actual constructor and we agreed that the invariant should be maintained there but it wasn't currently being maintained. We slightly disagreed about what the exact invariant should be but we both liked doing a 0 fill so that we could eliminate the complicated case of having uninitialized strings entirely which greatly simplifies the desired invariant.

src/runtime/string.cpp
```diff
@@ -210,6 +210,8 @@ static PRIMFN(prim_read) {
       runtime.heap.reserve(need_pass);
 
       String *out = String::claim(runtime.heap, size);
+      out->c_str()[size] = 0;
+
```
JakeSiFive, line 214: nit: You don't need this anymore

src/runtime/value.cpp
```diff
@@ -65,50 +66,39 @@ std::string HeapObject::to_str() const {
...
 
 String *String::alloc(Heap &h, size_t length) {
   return new (h.alloc(reserve(length))) String(length);
 }
 
 String *String::alloc(Heap &h, const std::string &str) {
-  String *out = alloc(h, str.size());
-  memcpy(out->data(), str.c_str(), str.size() + 1);
-  return out;
+  return new (h.alloc(reserve(str.length()))) String(str.c_str(), str.length());
```
JakeSiFive, line 94: On all these, `length` is non-standard and doesn't work on anything but std::string. `size()` is the standard in C++ and what you'll see used with std::string

src/runtime/string.cpp
V-FEXrt: I want to keep it for code consistency until we make `c_str()` const only. Every other case in this file still has this

src/runtime/value.cpp
V-FEXrt: yee, I went back and forth on that lol. I'll change it

src/runtime/value.cpp
```diff
@@ -65,50 +66,39 @@ std::string HeapObject::to_str() const {
...
 }
 
 String *String::alloc(Heap &h, const char *str, size_t length) {
-  String *out = alloc(h, length);
-  memcpy(out->c_str(), str, length);
-  out->c_str()[length] = 0;
-  return out;
+  return new (h.alloc(reserve(length))) String(str, length);
 }
```
JakeSiFive, line 99: I wonder if this was being used somewhere where the `out->c_str()[length] = 0;` was required because `str` didn't have a null term in some use case?

src/runtime/value.cpp
```diff
@@ -65,50 +66,39 @@ std::string HeapObject::to_str() const {
...
-  String *out = claim(h, str.size());
-  memcpy(out->c_str(), str.c_str(), str.size() + 1);
-  return out;
+  return new (h.claim(reserve(str.length()))) String(str.c_str(), str.length());
 }
 
 String *String::claim(Heap &h, const char *str, size_t length) {
-  auto out = claim(h, length);
-  memcpy(out->c_str(), str, length);
-  out->c_str()[length] = 0;
```
JakeSiFive, line 85: see comment below about alloc here as well.

src/runtime/value.cpp
JakeSiFive: omg....what a roller coaster. This upsets me.

JakeSiFive (approved):

Issue #1288, opened 2023-06-13
`wake-format` removes relative indentation in some cases of multiline strings


bmitc:
In some cases, `wake-format` will remove relative indentation between lines inside a multiline string `"""`. Note: I have not tested the below with Wake's other multiline string syntax of `"%...%"` since it's my understanding that that will eventually be obsoleted in favor of `"""`.

This is best seen via some reproducible examples. Each example code block should be viewed as an entire `.wake` file. This does matter in terms of the issue surfacing or not.

This code:

```wake
    def cmd1 =
        """
        if [ -f %{path} ]; then
            echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
        fi
        """

def cmd2 =
    """
    if [ -f %{path} ]; then
        echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
    fi
    """

def test =
    def cmd =
        """
        if [ -f %{path} ]; then
            echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
        fi
        """

    cmd

export def writeIfChanged =
    require Pass filehash =
        def cmd =
            """
            if [ -f %{path} ]; then
                echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
            fi
            """
        Pass cmd

    Pass filehash

```

will be formatted as:

```wake
def cmd1 =
    """
        if [ -f %{path} ]; then
        echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
        fi
    """

def cmd2 =
    """
        if [ -f %{path} ]; then
        echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
        fi
    """

def test =
    def cmd =
        """
            if [ -f %{path} ]; then
            echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
            fi
        """

    cmd

export def writeIfChanged =
    require Pass filehash =
        def cmd =
            """
                if [ -f %{path} ]; then
                echo -n $(b2sum -z -l 256 %{path} | awk '{ print $1 }')
                fi
            """

        Pass cmd

    Pass filehash

```

[...]

PR #1360, opened 2023-07-25
format: Fix indentation loss in multiline strings

V-FEXrt:
Fixes https://github.com/sifive/wake/issues/1288

Some indentation would be lost when a multiline string contained interpolation. This is because indentation is tracked at a lower level than interpolation. This change updates it so that interpolation will track slightly more global indentation and forward that info down to the literal level where multiline strings are actually processed


richardxia (approved):
Thanks for fixing this! I can't provide a competent review of the actual changes to the code formatter, but the added tests look good to me!

tools/wake-format/types.h
```diff
@@ -47,6 +47,7 @@ enum class ExplodeOption {
 
 struct ctx_t {
   size_t nest_level = 0;
+  size_t prefix_length = 0;
```
JakeSiFive, line 50: This needs a comment and a better name. "prefix" is too generic for how specific this appears to be. As is I have no idea what this is for other than some highly specific thing for fixing this issue that I don't yet understand.

JakeSiFive (commented):
Can you show me was happening before and what is happening now so I can get a feel for this? I'm not sure I understand whats being fixed

V-FEXrt:
> Can you show me was happening before and what is happening now so I can get a feel for this? I'm not sure I understand whats being fixed

It's mentioned in the bug https://github.com/sifive/wake/issues/1288 but tl;dr if a multiline string contained interpolation and relative indentation then it would loose said indentation.

```
def _ =
    """
    if %{x}; then
        %{x}
    fi
    """
```

would be formatted as 

```
def _ =
    """
    if %{x}; then
    %{x}
    fi
    """
```

tools/wake-format/types.h
V-FEXrt: fair

JakeSiFive:
Why was the MultiLineString thing behaving that way in the first place? Like why is this the fix and not fixing our multi line string indentation handling?

V-FEXrt:
> Why was the MultiLineString thing behaving that way in the first place? Like why is this the fix and not fixing our multi line string indentation handling?

We talked offline about this,  but here was the extra context needed to explain the issue.

```
0: TOKEN_NL (r: 1, c: 1, i: 0)
0: CST_DEF
1:   TOKEN_KW_DEF (r: 2, c: 1, i: 0)
1:   TOKEN_WS (r: 2, c: 4, i: 0) (1)
1:   CST_HOLE
2:     TOKEN_P_HOLE (r: 2, c: 5, i: 0)
1:   TOKEN_WS (r: 2, c: 6, i: 0) (1)
1:   TOKEN_P_EQUALS (r: 2, c: 7, i: 0)
1:   TOKEN_NL (r: 2, c: 8, i: 0)
1:   TOKEN_WS (r: 3, c: 1, i: 0) (4)
1:   CST_INTERPOLATE
2:     CST_LITERAL
3:       TOKEN_MSTR_BEGIN (r: 3, c: 5, i: 4) -> """
3:       TOKEN_NL (r: 3, c: 8, i: 4)
3:       TOKEN_WS (r: 4, c: 1, i: 4) (4)
3:       TOKEN_MSTR_CONTINUE (r: 4, c: 5, i: 4) -> x
3:       TOKEN_NL (r: 4, c: 6, i: 4)
3:       TOKEN_WS (r: 5, c: 1, i: 4) (6)
3:       TOKEN_MSTR_PAUSE (r: 5, c: 7, i: 6) -> %{
2:     CST_ID
3:       TOKEN_ID (r: 5, c: 9, i: 6) -> x
2:     CST_LITERAL
3:       TOKEN_MSTR_RESUME (r: 5, c: 10, i: 6) -> }
3:       TOKEN_NL (r: 5, c: 11, i: 6)
3:       TOKEN_MSTR_END (r: 6, c: 1, i: 6) ->     """
1:   TOKEN_NL (r: 6, c: 8, i: 6)
def _ =
    """
    x
      %{x}
    """
```
 
`CST_INTERPOLATE` is a parent node to `CST_LITERAL`

Multiline strings without interpolation are just a single `CST_LITERAL`, with interpolation its something ala `CST_INTERPOLATE(CST_LITERAL, <expr>, CST_LITERAL, <expr>, ...)`

The way we calculate the real indentation of a multiline string is via the "prefix", which is defined as the number of whitespace preceding the leftmost non whitespace character

Once we have the prefix, any extra whitespace on a line after the prefix is real whitespace that must go into the resulting string.
The formatter needs to know the prefix in order to safely change the absolute indentation of a line in a multiline string without changing the context of said string. Previously we only accounted for the prefix at the `CST_LITERAL` level, but when nested inside of a `CST_INTERPOLATE` we have to consider the prefix across all of the constituent `CST_LITERAL`s

The important observation is that
```
    """
    x
      %{x}
    """
```
is actually two `CST_LITERAL`s and when looked at independently they have `prefix = 4` and `prefix = 6` respectively meaning all their leading whitespace gets deleted

JakeSiFive (approved):
Thanks for making this change!

PR #1362, opened 2023-07-25
Add chaos monkey style testing for shared caching

JakeSiFive:
This change adds a new kind of shared cache fuzz testing where the test harness will actively kill or briefly pause processes to test fault tolerance.

Additionally this change adds some long needed features to the unit testing framework to allow filtering of the tests that are run. You can now filter by test name prefix, or you can filter by tags. Any tests which matches any prefix will be considered for running and if no prefix is specified its as if the empty prefix was used so all tests are considered. Among considered tests only those tests who's tags are a subset of the ones the user specified will run. I've found these two controls to allow me to get almost any subset of tests I want to run to run.

@mmjconolly Your expertise is needed in reviewing the code for pid namespaces, forking, traversal of /proc, etc...

@V-FEXrt if you could review what remains or any parts you understand that should cover the rest

### DIFF
--- a/tools/wake-unit/fuzz_test_job_cache.cpp
+++ b/tools/wake-unit/fuzz_test_job_cache.cpp
@@ -1,5 +1,6 @@
 #include <json/json5.h>
 #include <sched.h>
+#include <sys/mount.h>
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
@@ -325,8 +326,8 @@ bool run_as_init_proc(F f) {
   }
 
   // Now that we're in an isolated child, we unshare our namespaces
-  if (unshare(CLONE_NEWUSER | CLONE_NEWPID) != 0) {
-    wcl::log::error("unshare(CLONE_NEW_USER | CLONE_NEWPID): %s", strerror(errno))();
+  if (unshare(CLONE_NEWUSER | CLONE_NEWNS | CLONE_NEWPID) != 0) {
+    wcl::log::error("unshare(CLONE_NEW_USER | CLONE_NEWNS | CLONE_NEWPID): %s", strerror(errno))();
     exit(1);
   }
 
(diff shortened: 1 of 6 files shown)


tools/wake-unit/fuzz_test_job_cache.cpp
```diff
@@ -293,6 +298,75 @@ struct Pool {
...
+
+    // Relay errors up to the top
+    if (WIFEXITED(status) && WEXITSTATUS(status) == 0) {
+      return true;
+    }
+    return false;
+  }
+
+  // Now that we're in an isolated child, we unshare our namespaces
+  if (unshare(CLONE_NEWUSER | CLONE_NEWPID) != 0) {
```
mmjconolly, line 328: optional: I don't think it costs much to add a CLONE_NEWNS here and then a mount of /proc between lines 344-345, it probably makes things easier to reason about if you ever need to debug these in detail

tools/wake-unit/fuzz_test_job_cache.cpp
```diff
@@ -293,6 +298,75 @@ struct Pool {
...
+  if (pid == -1) {
+    wcl::log::error("run_as_init_proc: fork failure: %s", strerror(errno))();
+    exit(1);
+  }
+
+  // In the child process we just want to run the function.
+  // We assure the user that the code executed by f() will be
+  // executed as the init process of its own namespace.
+  if (pid == 0) {
+    int retcode = f();
```
mmjconolly, line 345: A quirk of being PID1 is that other members of the pid namespace can not send you signals unless you choose to recieve them, so if you happen to be sending signals around it's something to be mindful of

mmjconolly (approved):
I only read `run_as_init_proc`, seems good to me

tools/wake-unit/fuzz_test_job_cache.cpp
JakeSiFive: wait...I thought I had that code and it was mounted, I think somewhere in here I lost my CLONE_NEWNS and proc mount

tools/wake-unit/fuzz_test_job_cache.cpp
JakeSiFive: It should be fine, pid 1 shouldn't need to mess about with anything

JakeSiFive:
@mmjconolly I'm fixing the missing /proc mount, I somehow lost that change and it might break something but if you could look at `fuzz_many_with_ns` as well that would be helpful as that has a lot of subtle posix stuff happening inside of it as well.

V-FEXrt (approved):
Didn't look super closely at the PID namespace stuff but all in all seems good

PR #1363, opened 2023-07-25
Make JsonSubscriber more atomic

JakeSiFive:
This change is an attempt to make log corruption due to concurrent writers less likely but it doesn't fully solve the problem. It does two things

1) It removes the biggest offender for large log lines
2) It makes sure that we write log lines in a single call to write (ignoring errors)

This should hopefully eliminate all of our log parsing errors...I should probably also make the log viewer handle corrupted lines better than by just failing on them but this PR doesn't do that yet.

### DIFF
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -237,10 +237,12 @@ static void garbage_collect_job(std::string job_dir) {
   wcl::log::info("found orphaned job folder: %s", job_dir.c_str())();
   auto dir_res = wcl::directory_range::open(job_dir);
   if (!dir_res) {
-    // We can keep going even with this failure but we need to at least log it
+    // It's not an error if this directory doesn't exist
     if (dir_res.error() == ENOENT) {
       return;
     }
+
+    // We can keep going even with this failure but we need to at least log it
     wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
                     job_dir.c_str(), strerror(dir_res.error()))();
     return;
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -238,6 +238,9 @@ static void garbage_collect_job(std::string job_dir) {
   auto dir_res = wcl::directory_range::open(job_dir);
   if (!dir_res) {
     // We can keep going even with this failure but we need to at least log it
+    if (dir_res.error() == ENOENT) {
+      return;
+    }
     wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
                     job_dir.c_str(), strerror(dir_res.error()))();
     return;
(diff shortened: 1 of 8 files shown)


src/job_cache/eviction_policy.cpp
```diff
@@ -238,6 +238,9 @@ static void garbage_collect_job(std::string job_dir) {
   auto dir_res = wcl::directory_range::open(job_dir);
   if (!dir_res) {
     // We can keep going even with this failure but we need to at least log it
+    if (dir_res.error() == ENOENT) {
```
V-FEXrt, line 241: nit: this code placement makes the previous comment confusing

src/json/json5.cpp
```diff
@@ -144,11 +147,27 @@ std::ostream &operator<<(std::ostream &os, const JAST &jast) {
...
     out.add(item.first, item.second);
   }
 
-  s << out << std::endl;
+  std::stringstream ss;
+  ss << out << std::endl;
+  std::string line = ss.str();
+
+  if (line.size() > 4095) {
+    write(to_append.get(), warning_msg.data(), warning_msg.size());
```
V-FEXrt, line 170: this doesn't actually safely split the line unless you are saying "may be split due to interleaving"

src/json/json5.h
```diff
@@ -26,13 +26,20 @@
...
+  wcl::unique_fd to_append;
+
+  explicit JsonSubscriber(wcl::unique_fd &&f) : to_append(std::move(f)) {}
 
  public:
-  JsonSubscriber(std::streambuf *rdbuf) : s(rdbuf) {}
+  JsonSubscriber(const JsonSubscriber &) = delete;
+  JsonSubscriber(JsonSubscriber &&) = default;
+  JsonSubscriber() = delete;
+  static wcl::result<JsonSubscriber, wcl::posix_error_t> create(const char *log_path);
```
V-FEXrt, line 41: It's unclear to be why we need a static create function here.

Can you not just have the public api be  `explicit JsonSubscriber(wcl::unique_fd f)` then create the `unique_fd` where you create the `JsonSubscriber` then `std::move()` it in. Requiring a `path` here restricts the api (can't use stdin/stdout) and this feels like its doing more than it needs to

JakeSiFive: Ah good point, you're right!

JakeSiFive: It also makes error handling more specific to the context that its in

src/json/json5.cpp
JakeSiFive: oh I see what you mean, you mean that I'm not adding newlines.

JakeSiFive: I'm going to change the warning message to better reflect my intent and say "corrupted" rather than "split"

src/json/json5.h
JakeSiFive: Ok it turned out this was a bit more complex and we spoke off line. We don't want to alloy invalid constructions and it turns out there's only one valid way to construct the unique_fd so this does turn out to make sense

V-FEXrt: Adding slightly more context here. I think the real problem was with the leaking error handling. I'm not going to block on it but ideally I'd like to see either the requirement encoded in the type or a less leaky error type returned

V-FEXrt (approved):

PR #1364, opened 2023-07-26
Optionally discover wake files via manifest

V-FEXrt:
Wake is prohibitively slow when using it in a large repo on machines with slow disks (HHDs) because it hits every single inode.

This change allows those users to specify a `.wakemanifest` file that explicitly enumeration all wake files relative to the workspace root in the following format

```
a.wake
b.wake
# comment

# empty lines are also allowed

path/to/something/c.wake
path/to/other/thing/d.wake
```

This should make file discovery much faster at the cost of upkeeping the manifest file. This feature is opt-in and if a `.wakemanifest` is not set then the previous file scanning method will be used.


.wakemanifest
```diff
@@ -0,0 +1,69 @@
+src/types/types.wake
```
mmjconolly, line 1: For testing, is this relying on the .wakemanifest being present before the tarball CI step, but not afterwards?

V-FEXrt: The tarball case is an interesting thing to consider, but not fully sure what you are asking here. If the manifest doesn't exist then wake should fallback to the old method.

Honestly, I'm tempted to have wake still use the old method since it doesn't have the speed problem and let other repos load test the manifest method

mmjconolly:
If a filename is deleted from the .wakemanifest (but not the filesystem), then we (maybe) see a type error for a missing symbol. This seems like what we expect

If I delete the file from the filesystem, but not the manifest, we still see a type error, rather than a "could not find file"

.wakemanifest
mmjconolly: Yeah the wake repo using the search mechanism seems reasonable. I was asking if you were using the unusal structure of the CI to test this feature as on/off, but it could be explicitly tested instead

V-FEXrt:
Yeah, we should add a pass to make sure all the files exist. I just threw something together quickly

.wakemanifest
V-FEXrt: no, I wasn't doing any intentional testing here. To do a good test I think we'd need some secret flag for dumping all discovered wake files

V-FEXrt: Might be able to do a unit test but IIRC files user `src/runtime` are more of a pain to unit test than `src/wcl`

src/parser/wakefiles.cpp
```diff
@@ -454,6 +454,64 @@ static std::vector<std::string> filter_wakefiles(std::vector<std::string> &&wake
...
+  StringSegment segment = file.segment();
+  std::stringstream in;
+  in.write(reinterpret_cast<const char *>(segment.start), segment.size());
+
+  std::string line;
+  while (std::getline(in, line)) {
+    // strip trailing whitespace (including windows CR)
+    size_t found = line.find_last_not_of(" \t\f\v\n\r");
+    if (found != std::string::npos) {
+      line.erase(found + 1);
```
JakeSiFive, line 502: I hate how the string API is just subtly different from all the other container APIs. I had to go look this up to make sure it didn't just remove a single character.

src/parser/wakefiles.cpp
```diff
@@ -454,6 +454,64 @@ static std::vector<std::string> filter_wakefiles(std::vector<std::string> &&wake
...
+    } else {
+      line.clear();  // entirely whitespace
+    }
+
+    // allow empty lines and comments
+    if (line == "" || line[0] == '#') continue;
```
JakeSiFive, line 508: Very nice touch

JakeSiFive (approved):
LGTM

JakeSiFive:
>Yeah, we should add a pass to make sure all the files exist. I just threw something together quickly

Oh I thought this was being done already, I'm realizing now that we only did the "is_readable" check on the manifest and not each file we added to the list. I also thought we'd handle the file open correctly and not just skip the error.

src/parser/wakefiles.cpp
V-FEXrt: as much as I want to take credit for this, its just the impl of .wakeignore minus `!` operator

.wakemanifest
```diff
@@ -0,0 +1,69 @@
...
+share/wake/lib/core/boolean.wake
+share/wake/lib/core/vector.wake
+share/wake/lib/core/double.wake
+share/wake/lib/core/integer.wake
+share/wake/lib/core/tree.wake
+share/wake/lib/core/types.wake
+share/wake/lib/core/result.wake
+share/wake/lib/core/string.wake
+share/wake/lib/core/map.wake
+share/wake/lib/core/syntax.wake
```
JakeSiFive, line 43: I think we should omit these from the wake repo as they will also be omitted in other repos. Wakes shared library finding technique should work inside the wake repo as long as wake executables sit inside of the bin directory.

V-FEXrt: When wake notices the stdlib in the current workspace it chooses to use that instead (by falling back to the process for finding all wake files in the workspace). Since we replace that mechanism with the manifest and we have the stdlib in our workspace, we have to use the manifest to locate these files

build.wake
```diff
@@ -154,11 +154,40 @@ def tarball Unit =
...
-        | filter (! matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
-        | filter (! matches `(.*/)*\.gitignore` _.getPathName) # omit git files
+        | splitBy (matches regex _.getPathName)
+
+    def _ =
+        def files = (skips | map (\x (("  ", x.getPathName, Nil) | cat)))
+        def msg = "Files below are excluded from tarball", files
+
+        msg
+        | map (println _)
```
JakeSiFive, line 190: Do you want to leave this in or was it a debugging print?

build.wake
```diff
@@ -154,11 +154,40 @@ def tarball Unit =
...
+    require Pass regex =
+        "(({skipFolders}).*|({skipFiles})|(.*/)*({skipFilesRecursive}))"
+        | stringToRegExp
+
+    def Pair skips srcs =
         allSources
-        | filter (! matches `(debian/|wake.spec).*` _.getPathName) # omit packaging files
-        | filter (! matches `(\.circleci/|\.github/).*|\.dockerignore` _.getPathName) # omit CI files
-        | filter (! matches `(.*/)*\.gitignore` _.getPathName) # omit git files
+        | splitBy (matches regex _.getPathName)
```
JakeSiFive, line 183: And then down here it would be `splitBy (\x exists (matches _ x.getPathName) toExclude)`

.wakemanifest
JakeSiFive: I think we should instead have wake always use `{execpath}/../share/lib` as the standard lib personally.

build.wake
V-FEXrt: I wanted to leave it in as an info since `make tarball` is pretty much only called in CI  where its okay to be a bit noiser. Could also dump it to logEcho if preferred

build.wake
```diff
@@ -154,11 +154,40 @@ def tarball Unit =
     # Identify those sources files to include in the tarball
     require Pass allSources = sources "." `.*`
 
-    def srcs =
+    def skipFolders =
+        (".circleci/", "debian/", ".github/", ".vscode/", Nil)
```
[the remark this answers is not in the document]
V-FEXrt: I actually explicitly chose not to do something like this because regexes are hard to read.

My variable names above make it immediately clear  when its just a singe file, any instance of that filename, or the entire folder whereas listing out all the regexes requires mental overhead to figure out what each one does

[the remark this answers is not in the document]
JakeSiFive: ok well its still incorrect and you have to fix that. I also find the regex construction below to be a bit confusing personally.

What about a list of globs, you convert each glob to a regex and you check if any of those regexes match?

.wakemanifest
JakeSiFive: (not in this change to be clear, just how I wish it worked)

src/runtime/sources.cpp
```diff
@@ -254,131 +254,66 @@ static std::string slurp(int dirfd, const char *const *argv, bool &fail) {
...
-    failed = true;
+  const char *tok = fileStr.data();
+  const char *end = tok + fileStr.size();
+
+  for (const char *scan = tok; scan != end; ++scan) {
+    if (*scan == 0 && scan != tok) {
+      std::string path(tok);
+      tok = scan + 1;
+
+      if (!is_directory(path)) {
```
JakeSiFive, line 287: We decided not to filter this out didn't we? Also this means that you stat every source up front when you could just wait for the inode to be touched when a user calls `source` or `sources` instead. The only downside is that empty directories from inactive sub-modules will be in `sources` but that seems fine to me.

tools/wake/main.cpp
```diff
@@ -597,12 +597,12 @@ int main(int argc, char **argv) {
...
-      wakefilenames = find_all_wakefiles(enumok, clo.workspace, clo.verbose, libdir, ".", user_warn);
-      gettimeofday(&stop, 0);
-      double delay = (stop.tv_sec - start.tv_sec) + (stop.tv_usec - start.tv_usec) / 1000000.0;
-      wcl::log::info("Find wakefile took %f seconds", delay).urgent()();
+    struct timeval start, stop;
+    gettimeofday(&start, 0);
+    wakefilenames = find_all_wakefiles(enumok, clo.workspace, clo.verbose, libdir, ".", user_warn);
+    gettimeofday(&stop, 0);
+    double delay = (stop.tv_sec - start.tv_sec) + (stop.tv_usec - start.tv_usec) / 1000000.0;
+    wcl::log::info("Find wakefile took %f seconds", delay).urgent()();
```
JakeSiFive: Why did the formatting here change?

src/runtime/sources.cpp
V-FEXrt: oh yeah,  sorry. Need to remove this

tools/wake/main.cpp
V-FEXrt: I'm pretty sure I manually "formatted" it and that manual format was just incorrect

V-FEXrt:
At this point, this PR is getting quite heafty. I think I'll split it into 4 PRs next week

1) adding startup profiling that gets  put into wake.log
2) removing extra files from the wake  tarball
3) using  git ls-files instead of recursively searching for source files 
4) using either the manifest or git ls-files instead of recursively searching for wake files

build.wake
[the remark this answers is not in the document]
ag-eitilt: It makes the `catWith` a bit trickier (probably needs to be re-written as a `foldl`), but there's also `map quote`.

build.wake
ag-eitilt: I'd prefer `info`; it's distinctly *not* the command line itself as is intended for `echo`.  However, it's currently being sent to `report` -- which isn't a bad place either, but it might be noisier than you intended.

ag-eitilt (commented):
I'm not sure where I'd be looking for it to answer the question for myself, but do we already exclude `.wakeignore` (and maybe even `.gitignore`) from the recursive search?

build.wake
V-FEXrt: the specific function is very rarely called by users, instead its only called by CI so we can be more tolerance to noise here but I don't feel super strongly and I see that I'm in the minority so I'll move it over to `info`

ag-eitilt: I don't actually mind it being a `report`, I just want it to be either that or `info` rather than `echo`. Sorry for the confusion!

V-FEXrt:
Okay! Everything has been pulled out of this PR into various other PRs. Feel free to take another look though the PR should be  at the same state it was when @JakeSiFive last approved it

JakeSiFive (approved):
Still LGTM

PR #1365, opened 2023-07-26
build: Fix bad PR race between #1362 and #1363

V-FEXrt:
https://github.com/sifive/wake/pull/1362 and https://github.com/sifive/wake/pull/1363 were incompatible together but passed CI separately. This broke the build on master, this change fixes the build.


JakeSiFive (approved):
Fix looks good, thanks for figuring out that it was a PR race!

PR #1369, opened 2023-07-31
Profile wake startup times

V-FEXrt:
This is the first PR of several being split out from  https://github.com/sifive/wake/pull/1364

This change profiles wake startup time and dumps it into `wake.log` so that optimizations can be verified.

In the long term, it would be interesting to build out some kind of macro or C++ template for a `ProfiledStep` that accepts a low cost function and wraps in it profiling but I don't think that is needed just yet

### DIFF
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -594,8 +594,16 @@ int main(int argc, char **argv) {
 
   bool enumok = true;
   std::string libdir = wcl::make_canonical(find_execpath() + "/../share/wake/lib");
-  auto wakefilenames =
-      find_all_wakefiles(enumok, clo.workspace, clo.verbose, libdir, ".", user_warn);
+  std::vector<std::string> wakefilenames;
+  {
+    struct timeval start, stop;
+    gettimeofday(&start, 0);
+    wakefilenames = find_all_wakefiles(enumok, clo.workspace, clo.verbose, libdir, ".", user_warn);
+    gettimeofday(&stop, 0);
+    double delay = (stop.tv_sec - start.tv_sec) + (stop.tv_usec - start.tv_usec) / 1000000.0;
+    wcl::log::info("Find all wakefiles took %f seconds", delay)();
+  }
+
   if (!enumok) {
     if (clo.verbose) std::cerr << "Workspace wake file enumeration failed" << std::endl;
     // Try to run the build anyway; if wake files are missing, it will fail later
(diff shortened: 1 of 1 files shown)


tools/wake/main.cpp
```diff
@@ -622,57 +639,64 @@ int main(int argc, char **argv) {
...
+
+    struct timeval starttv, stop;
+    gettimeofday(&starttv, 0);
+
+    for (size_t i = 0; i < wakefilenames.size(); i++) {
+      auto &wakefile = wakefilenames[i];
+
+      auto now = std::chrono::steady_clock::now();
+      if (!clo.quiet && is_stdout_tty &&
+          std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() > 1000) {
```
JakeSiFive, line 658: nit: its very strange that we're using both chrono and gettimeofday

V-FEXrt: yeah, :/ this was lifted from some other profiling code we have. I can see about just using `chrono`

tools/wake/main.cpp
```diff
@@ -594,8 +594,16 @@ int main(int argc, char **argv) {
...
-  auto wakefilenames =
-      find_all_wakefiles(enumok, clo.workspace, clo.verbose, libdir, ".", user_warn);
+  std::vector<std::string> wakefilenames;
+  {
+    auto start = std::chrono::steady_clock::now();
+    wakefilenames = find_all_wakefiles(enumok, clo.workspace, clo.verbose, libdir, ".", user_warn);
+    auto stop = std::chrono::steady_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(stop - start).count();
+    wcl::log::info("Find all wakefiles took %f seconds",
+                   static_cast<double>(duration) / 1000000.0)();
```
JakeSiFive, line 604: nit (And I stress this is a *nit*, you can land this as is):

```
    auto duration = std::chrono::duration_cast<std::chrono::seconds, double>(stop - start).count();
    wcl::log::info("Find all wakefiles took %f seconds", duration)();
```

tools/wake/main.cpp
```diff
@@ -604,7 +612,15 @@ int main(int argc, char **argv) {
...
   Profile tree;
   Runtime runtime(clo.profile ? &tree : nullptr, clo.profileh, heap_factor);
-  bool sources = find_all_sources(runtime, clo.workspace);
+  bool sources = false;
+  {
+    auto start = std::chrono::steady_clock::now();
+    sources = find_all_sources(runtime, clo.workspace);
+    auto stop = std::chrono::steady_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(stop - start).count();
+    wcl::log::info("Find all sources took %f seconds", static_cast<double>(duration) / 1000000.0)();
```
JakeSiFive, line 621: ditto nit

JakeSiFive (approved):

tools/wake/main.cpp
V-FEXrt: its trivial to  fix, and strictly better. No reason to not just fix it

PR #1370, opened 2023-07-31
Filter tarball files with globs; filter more files

V-FEXrt:
Part two of splitting up https://github.com/sifive/wake/pull/1364

This change 
 - removes more files from the  tarball that have snuck in over time
 - uses globs  for exclusion instead of regexes
 - prints the excluded files to `info`

### DIFF
--- a/share/wake/lib/core/regexp.wake
+++ b/share/wake/lib/core/regexp.wake
@@ -81,28 +81,6 @@ export def regExpToString (regExp: RegExp): String =
 export def matches (testRegExp: RegExp) (str: String): Boolean =
     (\_ \_ prim "match") testRegExp str
 
-# Test if a String matches all provided regular expressions
-# matchesAll (`a*`, `*b`) "ab" = True
-# matchesAll (`a*`, `*b`) "a" = False
-export def matchesAll (testRegExps: List RegExp) (str: String): Boolean = match testRegExps
-    Nil = True
-    head, tail =
-        if (matches head str) then
-            matchesAll tail str
-        else
-            False
-
-# Test if a String matches any provided regular expressions
-# matchesAny (`a*`, `*b`) "ac" = True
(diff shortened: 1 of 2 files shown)


share/wake/lib/core/regexp.wake
```diff
@@ -81,6 +81,28 @@ export def regExpToString (regExp: RegExp): String =
...
+# Test if a String matches all provided regular expressions
+# matchesAll (`a*`, `*b`) "ab" = True
+# matchesAll (`a*`, `*b`) "a" = False
+export def matchesAll (testRegExps: List RegExp) (str: String): Boolean = match testRegExps
+    Nil = True
+    head, tail =
+        if (matches head str) then
+            matchesAll tail str
+        else
+            False
```
JakeSiFive, line 93: This is too generic to be in regex, use forall/exist in list.wake

share/wake/lib/core/regexp.wake
```diff
@@ -81,6 +81,28 @@ export def regExpToString (regExp: RegExp): String =
...
+
+# Test if a String matches any provided regular expressions
+# matchesAny (`a*`, `*b`) "ac" = True
+# matchesAny (`a*`, `*b`) "c" = False
+export def matchesAny (testRegExps: List RegExp) (str: String) = match testRegExps
```
JakeSiFive, line 98: ditto

share/wake/lib/core/regexp.wake
V-FEXrt: ah ye. makes sense. fixed!

ag-eitilt (approved):

V-FEXrt:
@JakeSiFive I'm going to go ahead and land this since I've addressed your comments and this PR is fairly small/uncontroversial now. If you have any changes you want when you get  back lmk and I'll make them asap

PR #1372, opened 2023-07-31
Determine source files from recursive `git ls-files`

V-FEXrt:
Our previous implementation of `find_all_sources` would use `git ls-files`  non-recursively requiring the exploration of every directory in the workspace which is slow on very large projects.  

This change now relies on `git ls-files --recurse-submodules` to report files in sub modules which drastically reduces the number of inodes that must be explored.

### DIFF
--- a/src/runtime/sources.cpp
+++ b/src/runtime/sources.cpp
@@ -39,7 +39,6 @@
 #include <iostream>
 #include <sstream>
 
-#include "compat/readable.h"
 #include "parser/wakefiles.h"
 #include "prim.h"
 #include "types/data.h"
@@ -287,7 +286,8 @@ bool find_all_sources(Runtime &runtime, bool workspace) {
   std::vector<std::string> sources;
 
   // Source discovery requires git
-  if (is_readable(".git")) {
+  struct stat unused;
+  if (stat(".git", &unused) == 0) {
     sources = scan_git(dirfd);
     std::sort(sources.begin(), sources.end());
   }
--- a/src/runtime/sources.cpp
+++ b/src/runtime/sources.cpp
@@ -254,134 +254,48 @@ static std::string slurp(int dirfd, const char *const *argv, bool &fail) {
   return str.str();
 }
 
-static bool scan(std::vector<std::string> &files, std::vector<std::string> &submods,
-                 const std::string &path, int dirfd) {
-  auto dir = fdopendir(dirfd);
-  if (!dir) {
-    fprintf(stderr, "Failed to fdopendir %s: %s\n", path.c_str(), strerror(errno));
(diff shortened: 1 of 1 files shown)


JakeSiFive (approved):

src/runtime/sources.cpp
```diff
@@ -283,8 +284,13 @@ bool find_all_sources(Runtime &runtime, bool workspace) {
   int flags, dirfd = open(".", O_RDONLY);
   if ((flags = fcntl(dirfd, F_GETFD, 0)) != -1) fcntl(dirfd, F_SETFD, flags | FD_CLOEXEC);
 
-  std::vector<std::string> sources = scan_git(dirfd);
-  std::sort(sources.begin(), sources.end());
+  std::vector<std::string> sources;
+
+  // Source discovery requires git
+  if (is_readable(".git")) {
```
JakeSiFive, line 290: Does `is_readable` work on folders?

JakeSiFive (approved):

src/runtime/sources.cpp
V-FEXrt: I'm pretty sure it does. It  was working for me when I was testing it.

Under the hood its just calling `access(path, R_OK)` https://man7.org/linux/man-pages/man2/access.2.html

ag-eitilt: In that case, it might be missing a `X_OK` check as well?  I don't know exactly where the line between reading and executing a directory falls, but I know there's some muddiness between them.

V-FEXrt: I think for true correctness, we actually do need to check `X_OK` on the directory. I'm think its okay to not do that here though since it would be very strange to have `R_OK` but not `X_OK` on the `.git` repo and a I'd have to make another function or just call `access` directly here since `is_readable` actively shouldn't check `X_OK`

V-FEXrt: hmmm that said, directly calling `access` here isn't arduous. I should probably just do that

mmjconolly: I haven't read all of this, but something to be careful of is that sometimes the `.git` is a file rather than a directory

JakeSiFive: I think we should just state the inode honestly. If the inode exists, we assume the user is using git

ag-eitilt: That's actually a good point.  All we care about is whether git is tracking things or not -- if the workspace is set up to use git but git is broken, then it's not an issue on our side and git will be able to report the problem much better than we are.

V-FEXrt: Done!

JakeSiFive (approved):

PR #1376, opened 2023-08-02
Add --auto option to automatically pick files to format

V-FEXrt:
This change introduces the `--auto`  flag which uses the built in wake mechanism to discover which files should be formatted.

This means that 

`wake-format --auto --dry-run` is sufficient to ensure all wake files discovered by wake will be formatted and `wake-format --auto --in-place` will format all said files.


Makefile
```diff
@@ -56,7 +56,7 @@ install:	all
 # It assumes clang is available on the PATH and will fail otherwise
 formatAll:
 	@clang-format -i --style=file $(shell ./scripts/which_clang_files all)
-	@bin/wake-format.native-cpp14-release -i $(shell ./scripts/which_wake_files all)
```
JakeSiFive, line 59: Can we delete that script now?

JakeSiFive: Also was this script missing build.wake and tests.wake?

build.wake
```diff
@@ -187,35 +187,41 @@ def tarball Unit =
     def wakeSourcesString =
         wakePaths
         | map (_.getPathName.format)
-        | foldr (_, ",\n  ", _) Nil
+        | foldr (_, ",\n    ", _) Nil
```
JakeSiFive, line 190: Why is this space being added?

V-FEXrt: See the other comment; tl;dr it ensures `manifest.wake` is formatted

JakeSiFive: Ah ok gotcha

Makefile
JakeSiFive: Ah gotcha

JakeSiFive (approved):

PR #1378, opened 2023-08-03
Switch shared cache comunication over to non-blocking io

JakeSiFive:
This is a rather massive change to the shared cache. It doesn't change the protocol, the directory layout, the read or write implementations etc... but it changes the nitty gritty specifics of how messages are sent/received and how blocking is handled. Specifically the following changes have been made It should be impossible for a colleciton of clients and a daemon to deadlock now because neither side ever blocks and there's no shared state between the two sides. Additionally connection closing happens much more aggressively to simplify functionality. This is all in the name of stopping deadlock from occuring.

This change also adds tests that combine large messages with parallelism which is where I think the bug was hiding previously but I've yet to run those tests on a previous commit to see if they triggered deadlock (will do that soon).

1) Unit tests now always output information to wake.log
2) The client now uses non-blocking IO with leveled triggered polling and timeouts in order to ensure that it never blocks for longer than the time
3) The daemon now uses non-blocking IO with *edge* triggered polling and timeouts in order to ensure that it never blocks if there is any other work that it can do instead. Previously each read and each write had to happen synchronously and would potentially block under many circumstances. Now even writes are non-blocking and can happen out of order.
4) If a client times out, the daemon closes the connection
5) Every request is made over 1 domain socket so that there is little to no state shared between the daemon and a client
6) Previously the daemon would close itself after a disconnect form a client if it discovered that it had no further connected clients. Now the daemon exits after 10 minutes of no activity regardless of how many clients are connected. However you can get back the old functionality by using an environment variable to enable fast closing.
7) As part of 3, we have both MessageParser AND MessageSender which is like MessageParser except it allows writes to be chunked and sent out of order as possible just like MessageParser allows the same for reading.
8) There's a new wrapper around epoll because I didn't want to mess with the old Poll wrapper that drives the heart of wake here. This new wrapper is not implementable by non-epoll interfaces and is thus Linux only.

### DIFF
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -243,8 +243,8 @@ static void garbage_collect_job(std::string job_dir) {
     }
 
     // We can keep going even with this failure but we need to at least log it
-    // wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
-    //                job_dir.c_str(), strerror(dir_res.error()))();
+    wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
+                    job_dir.c_str(), strerror(dir_res.error()))();
     return;
   }
 
@@ -275,8 +275,8 @@ static void garbage_collect_group(const std::unordered_set<int64_t> jobs, int64_
   auto dir_res = wcl::directory_range::open(group_dir);
   if (!dir_res) {
     // We can keep going even with this failure but we need to at least log it
-    // wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
-    //                group_dir.c_str(), strerror(dir_res.error()))();
+    wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
+                    group_dir.c_str(), strerror(dir_res.error()))();
     return;
   }
 
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -243,8 +243,8 @@ static void garbage_collect_job(std::string job_dir) {
     }
 
     // We can keep going even with this failure but we need to at least log it
-    wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
-                    job_dir.c_str(), strerror(dir_res.error()))();
+    //wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
+    //                job_dir.c_str(), strerror(dir_res.error()))();
     return;
   }
 
@@ -275,8 +275,8 @@ static void garbage_collect_group(const std::unordered_set<int64_t> jobs, int64_
   auto dir_res = wcl::directory_range::open(group_dir);
   if (!dir_res) {
     // We can keep going even with this failure but we need to at least log it
-    wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
-                    group_dir.c_str(), strerror(dir_res.error()))();
+    //wcl::log::error("garbage collecting orphaned folders: wcl::directory_range::open(%s): %s",
+    //                group_dir.c_str(), strerror(dir_res.error()))();
     return;
   }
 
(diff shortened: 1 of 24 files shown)


src/job_cache/job_cache.cpp
```diff
@@ -326,7 +319,11 @@ FindJobResponse Cache::read(const FindJobRequest &find_request) {
       return *response;
     }
 
-    if (miss_on_failure && misses_from_failure > 30) {
+    if (response.error() == FindJobError::CouldNotConnect) {
```
V-FEXrt, line 322: super  small nit that  I'm waffling on

`failed_on_connect |= response.error() == FindJobError::CouldNotConnect`

src/job_cache/message_parser.h
```diff
@@ -55,13 +67,19 @@ struct MessageParser {
...
-        // only occur if the user gave us a non-blocking socket.
-        if (errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK) continue;
+        // On EINTR we should just retry until we get EAGAIN/EWOULDBLOCK
+        if (errno == EINTR) {
+          continue;
+        }
+        // If we hit EAGAIN/EWOULDBLOCK then we might have more work to do but
+        // we can't do that work just yet and need to continue.
+        if (errno == EAGAIN || errno == EWOULDBLOCK)
+            return MessageParserState::Continue;
```
V-FEXrt, line 77: I personally  prefer wrapping in  `{}`  but at  the very least the brackets should be consistent. Right now this function has three  different styles

src/job_cache/message_parser.h
```diff
@@ -55,13 +67,19 @@ struct MessageParser {
...
+
         return MessageParserState::StopFail;
       }
 
+      // Here we split this up
```
V-FEXrt, line 82: This comment doesn't seem super useful

src/job_cache/message_sender.h
```diff
@@ -0,0 +1,123 @@
...
+  std::string data;
+  std::string::iterator start;
+  time_t deadline;
+  int fd;
+  MessageSenderState state = MessageSenderState::Continue;
+
+ public:
+  MessageSender() = delete;
+  MessageSender(const MessageSender&) = delete;
+  MessageSender(MessageSender&& sender) {
```
V-FEXrt, line 56: do we actually use/need the move constructor here?

V-FEXrt (approved):
Overall looks quite good. Lots of small things to cleanup (log spam,  wasm+epoll) but I don't see  any API/architecture changes 

Thanks!

src/job_cache/eviction_policy.cpp
```diff
@@ -367,7 +367,8 @@ LRUEvictionPolicy::~LRUEvictionPolicy() {}
 int eviction_loop(const std::string& cache_dir, std::unique_ptr<EvictionPolicy> policy) {
   policy->init(cache_dir);
 
-  MessageParser msg_parser(STDIN_FILENO);
+  // Famous last words "a timeout of a year is plenty"
+  MessageParser msg_parser(STDIN_FILENO, 60 * 60 * 24 * 30 * 12);
```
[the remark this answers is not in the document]
JakeSiFive: Yeah at a few levels 1) I checked the fuzz tests after running to ensure their size and that eviction was running at all and 2) We have the basic-lru test that just sort of smoke tests that its working at all and sets the sizes so that it can intentionally cause lru-evictions at specific points.

src/job_cache/job_cache.cpp
JakeSiFive: I don't think that's worth doing personally but if you want it I can do it.

src/job_cache/message_parser.h
JakeSiFive: Yeah I'll add the brackets

src/job_cache/message_sender.h
JakeSiFive: Yes, we need the move constructor in several places. 1) To store an object in a resizable container (e.g. a hash table) you have to be moveable. 2) We construct a message sender and then move construct it via `emplace` when we add a message sender to message_senders.

src/job_cache/eviction_policy.cpp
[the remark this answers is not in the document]
JakeSiFive: I think we could add something to the fuzz testing to ensure 1) A certain hit rate occurs and 2) that the size never goes over what we expect it to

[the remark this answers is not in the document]
JakeSiFive: That should be in a later PR I think however, we've tested it ok-ishly well here.

src/job_cache/job-cache.wake
```diff
@@ -18,9 +18,7 @@ package build_wake
 from wake import _
 from gcc_wake import _
 
-target jobCacheLib variant = match variant
-    Pair "wasm-cpp14-release" _ -> Pass (SysLib "" Nil Nil Nil Nil)
```
V-FEXrt: Can you cleanup `fuse-waked.wake` too? Maybe just generally revert this PR. https://github.com/sifive/wake/pull/983/files

Will make it easier later if we don't leave cruft around

JakeSiFive: sorry just not found this, I'll do this in a follow up PR, I promise.

PR #1380, opened 2023-08-04
Remove clients that have timed out

JakeSiFive:
In the previous non-blocking io change I forgot to evict clients that had timed out. This wound up not mattering much but I think its a good feature to keep in mind. All tests still pass.

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -36,6 +36,7 @@
 
 #include <ctime>
 #include <fstream>
+#include <unordered_set>
 
 #include "db_helpers.h"
 #include "eviction_command.h"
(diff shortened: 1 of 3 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -716,6 +717,24 @@ int DaemonCache::run() {
         wcl::log::info("Unrecognized event on %d: events = %d", event.data.fd, event.events)();
       }
     }
+
+    // Check for timeouts
+    std::unordered_set<int> clients_to_close;
+    for (auto &client : message_senders) {
```
V-FEXrt, line 723: We should be removing these from `message_senders` and `message_parsers` no? If not we could close something more than once

JakeSiFive: That's why I use an unordered_set and close them in a loop after both.

V-FEXrt: right, but you never remove the old element from the variables: `message_senders` and `message_parsers`. If these loops ever run twice without resetting them then all the closed clients would get closed again.

It looks like this only happens right before exiting `DaemonCache::run` and we _shouldn't_ ever be calling that function twice but that's not actually enforced anywhere

JakeSiFive: close_client handles that clean up automatically because there are many spots where we close a client. Also this is run every 5 seconds or after every IO event

V-FEXrt: ah, right

V-FEXrt (approved):

PR #1384, opened 2023-08-08
parser: Correctly handle paths when using manifest

V-FEXrt:
https://github.com/sifive/wake/pull/1364 allowed wake files to be explicitly listed via a manifest file. These paths were directly returned to the calling function.

For `wake` this is always correct as the wake tool will `cwd` into the workspace before calling the function. This means that the relative paths returned are correct.

However `lsp-wake` doesn't `cwd` into the workspace and instead relies on absolute paths. When a relative path is returned it assumes it is absolute then fails to find the file later.

This change fixes the broken invariant that the paths returned from `find_all_wakefiles` must be directly readable given the callers cwd.

### DIFF
--- a/src/parser/wakefiles.cpp
+++ b/src/parser/wakefiles.cpp
@@ -487,7 +487,8 @@ std::vector<std::string> find_workspace_wakefiles_via_search(bool &ok, bool verb
   return filter_wakefiles(std::move(workfiles), workdir, verbose);
 }
 
-std::vector<std::string> find_workspace_wakefiles_via_manifest(const std::string &manifest_path) {
+std::vector<std::string> find_workspace_wakefiles_via_manifest(const std::string &manifest_path,
+                                                               const std::string &workdir) {
   std::vector<std::string> workfiles;
   std::vector<std::string> unreadable_files;
   DiagnosticIgnorer ignorer;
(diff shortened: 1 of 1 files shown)


src/parser/wakefiles.cpp
```diff
@@ -496,6 +497,8 @@ std::vector<std::string> find_workspace_wakefiles_via_manifest(const std::string
   std::stringstream in;
   in.write(reinterpret_cast<const char *>(segment.start), segment.size());
 
+  std::string prefix = (workdir == ".") ? "" : (workdir + "/");
```
JakeSiFive, line 500: This should be "is workdir relative" not "is workdir this one specific relative path" I think.

JakeSiFive (approved):
LGTM, I have a recommendation however.

src/parser/wakefiles.cpp
V-FEXrt: I think the check has to be this way, we aren't really asking is it relative, we are asking is it relative to cwd.

if `workdir` was `../workspace` and there was a file `build.wake` we should return `../workspace/build.wake` not `build.wake`.

Honestly it might just be acceptable to _always_ return `workdir + "/" + filename` as `./build.wake` is the same as `build.wake`

PR #1386, opened 2023-08-09
Require explode when a binop ends with a unary comma

V-FEXrt:
Trailing unary comma now tells the formatter to newline a list. 

Ex:

```
def x = a, b, c,
```
becomes 
```
def x = 
    a,
    b,
    c,
```

instead of being flat.

This is something that @ag-eitilt and @bmitc have been asking for for awhile and I think its a good tradeoff now that I've seen a bunch of use cases in internal repos that would either use this or `# wake-format off`

### DIFF
--- a/build.wake
+++ b/build.wake
@@ -51,12 +51,18 @@ def bootstrapTarget targetFn variant =
     else Pass paths
 
     def cppSuffixRegex =
-        (".", variant.getPairFirst,)
+        (
+            ".",
+            variant.getPairFirst,
+        )
         | cat
         | quote
 
     def cSuffixRegex =
-        (".", variant.getPairSecond,)
+        (
+            ".",
+            variant.getPairSecond,
+        )
         | cat
         | quote
 
--- a/build.wake
+++ b/build.wake
@@ -51,18 +51,12 @@ def bootstrapTarget targetFn variant =
     else Pass paths
 
     def cppSuffixRegex =
-        (
-            ".",
-            variant.getPairFirst,
-        )
+        (".", variant.getPairFirst, Nil)
         | cat
         | quote
 
     def cSuffixRegex =
-        (
-            ".",
-            variant.getPairSecond,
-        )
+        (".", variant.getPairSecond, Nil)
         | cat
         | quote
 
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -231,7 +231,7 @@ export def getPlanHash (plan: Plan): Integer =
     def isAttyStr = if isAtty then "true" else "false"
 
     def sig =
-        cat (
+        (
             isAttyStr,
             "\0\0",
             implode cmd,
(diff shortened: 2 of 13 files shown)


build.wake
```diff
@@ -51,12 +51,18 @@ def bootstrapTarget targetFn variant =
     else Pass paths
 
     def cppSuffixRegex =
-        (".", variant.getPairFirst,)
```
JakeSiFive, line 54: This should be made to end in `Nil` instead now, all these small cases should work that way

share/wake/lib/system/job.wake
```diff
@@ -229,7 +229,19 @@ export def editPlanShare (f: Boolean => Boolean): Plan => Plan =
 export def getPlanHash (plan: Plan): Integer =
     def Plan _ cmd _ env dir stdin _ _ _ _ _ _ _ _ isAtty = plan
     def isAttyStr = if isAtty then "true" else "false"
-    def sig = cat (isAttyStr, "\0\0", implode cmd, "\0\0", implode env, "\0\0", dir, "\0\0", stdin,)
+
+    def sig =
```
JakeSiFive, line 233: This one looks better like this I think.

tools/lsp-wake/lsp.wake
```diff
@@ -30,5 +30,31 @@ target buildLSP variant: Result (List Path) Error = match variant
             "--post-js", postJS.getPathName,
             "--pre-js", preJS.getPathName,
 
-        tool @here Nil variant "lib/wake/lsp-wake" (dst, util, wcl,) (postJS, preJS,) extraLFlags
-    _ = tool @here Nil variant "lib/wake/lsp-wake" (dst, util, wcl,) Nil Nil
+        tool
```
JakeSiFive, line 33: Yeah this look awful

JakeSiFive (commented):
I think this is clashing with cases where code was written without this being intended.

tools/wake-format/emitter.cpp
```diff
@@ -1365,6 +1365,22 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
     MEMO_RET(*doc);
   }
 
+  // Lists with trailing unary commas should always exlode
+  // Ex:
+  //   a, b, c, => explode
+  //   a, b, c  => prefer flat when possible
+  if (parts.back().id() == CST_UNARY) {
```
JakeSiFive, line 1372: This code itself looks good to me. `back()` should always be valid because this is a binary op so you'll get at least 2, the case of no binary op and just a trailing comma will be un-exploded which is fine by me, all the checks look good to me.

share/wake/lib/system/job.wake
V-FEXrt: lol we pretty much agreed on each case. I'm working on cleaning these up now

JakeSiFive (approved):
Approved with cleanup so that the wake code in the repo doesn't look so bad

PR #1387, opened 2023-08-09
Improve logging yet again, focusing on hangs, and distributed case

JakeSiFive:
This change just tweaks logging in lots of small ways

1) We add a bunch of before/after logging of important events in order to better understand if a hang is occuring and if so to help pin down where
2) I removed a debug line printing out the full message
3) We log before and after comunication with the eviction loop in case that ever hangs
4) We log when a job statrts and when it ends so we can detect job hangs
5) hostnames are always added so that logs can be collated across multiple machines
6) We now print out if a message received was a cache hit or not.

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -945,8 +945,11 @@ FindJobResponse DaemonCache::read(const FindJobRequest &find_request) {
   std::string msg = cmd.serialize();
   msg += '\0';
 
+  wcl::log::info("Sending Read command to eviction loop")();
   if (write(evict_stdin, msg.data(), msg.size()) == -1) {
     wcl::log::warning("Failed to send eviction update: %s", strerror(errno))();
+  } else {
+    wcl::log::info("Successfully sent eviction the job")();
   }
 
   return FindJobResponse(wcl::make_some<MatchingJob>(std::move(result)));
@@ -1031,8 +1034,11 @@ void DaemonCache::add(const AddJobRequest &add_request) {
   std::string msg = cmd.serialize();
   msg += '\0';
 
+  wcl::log::info("Sending Write command to eviction loop")();
   if (write(evict_stdin, msg.data(), msg.size()) == -1) {
     wcl::log::warning("Failed to send eviction update: %s", strerror(errno))();
+  } else {
+    wcl::log::info("Sucessfully sent eviction add update")();
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -945,8 +945,11 @@ FindJobResponse DaemonCache::read(const FindJobRequest &find_request) {
   std::string msg = cmd.serialize();
   msg += '\0';
 
+  wcl::log::info("Sending Read command to eviction loop")();
```
V-FEXrt, line 948: Super long term idea not for this PR, but it would be a cool optimization to hash static messages then emit the hash everywhere we see the message + a message->hash header. Then the log viewer could reconstruct the actual log

src/job_cache/job_cache.cpp
```diff
@@ -307,6 +308,8 @@ wcl::result<FindJobResponse, FindJobError> Cache::read_impl(const FindJobRequest
 }
 
 FindJobResponse Cache::read(const FindJobRequest &find_request) {
+  wcl::log::info("Starting cache read")();
```
V-FEXrt, line 311: nit: for these style messages I think `Cache::read enter` and `Cache::read exit` are a tiny bit more concise and a tiny bit clearer

src/runtime/job.cpp
```diff
@@ -861,6 +863,7 @@ static void launch(JobTable *jobtable) {
     sigprocmask(SIG_UNBLOCK, &set, 0);
     pid_t pid = wake_spawn(cmdline[0], cmdline, environ);
     sigprocmask(SIG_BLOCK, &set, 0);
+    wcl::log::info("Spawned job: label = '%s', pid = %d", job_label_str->c_str(), pid)();
```
V-FEXrt, line 866: This will cause log sizes to grow super fast. I'm a bit concerned about users having huge `wake.log`s that they don't think to delete

src/runtime/job.cpp
```diff
@@ -1848,6 +1855,8 @@ static PRIMFN(prim_job_cache_add) {
   std::string result_json_str = "successfully added job";
   size_t need = String::reserve(result_json_str.size()) + reserve_result();
   runtime.heap.reserve(need);
+
+  wcl::log::info("Cahce::add successfully called")();
```
V-FEXrt, line 1859: type: `Cahce` ->`Cache`

V-FEXrt (approved):

src/runtime/job.cpp
JakeSiFive: Ok I'll remove it for now then, if I still want it I can add in a flag for it.

JakeSiFive: I can get this info from the database so its still there, I just have to find it.

PR #1390, opened 2023-08-15
Shared cache env-var tweaks

JakeSiFive:
This change does the following

1) Declares the wake shared cache no longer experimental and changes the env-var used for it to be WAKE_LOCAL_JOB_CACHE
2) Adds the ability to configure some config options via environment variables. In particular the user config can now be specified via an env-var which allows for greatly increased flexibility. I did however additionally allow env-vars for cache miss on failure, and cache size manipulation in case that's helpful but we anticipate only
3) I went ahead and added a command line override for user-config so that you can locally fix what might become a shell-global override of .wakeroot

### DIFF
--- a/src/runtime/config.cpp
+++ b/src/runtime/config.cpp
@@ -335,14 +335,26 @@ bool WakeConfig::init(const std::string& wakeroot_path, const WakeConfigOverride
     }
   }
 
+  // The priority of config sources is the following from lowest priority to highest:
+  // 1) .wakeroot
+  // 2) user config
+  // 3) environment variables
+  // 4) command line options
+  //
+  // When parsing the user config, the user config path can't be in the user config
+  // but it can be anywhere else so before parsing the user config we parse wakeroot
+  // and then the other two sources. Once we parse the user config though we'll have
+  // go overwrite anything from the user config that should be form an env-var or
+  // a command line option.
+
   // Parse values from .wakeroot
   _config->set_all<WakeConfigProvenance::WakeRoot>(wakeroot_json);
 
   // Sometimes we need to the user_config with an env-var so we check env-vars first here
   _config->set_all_env_var();
 
-  // Further more users may choose to override the user config at the command line level so
-  // we run that too only to run it again later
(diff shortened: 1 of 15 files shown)


src/runtime/config.cpp
```diff
@@ -338,6 +338,13 @@ bool WakeConfig::init(const std::string& wakeroot_path, const WakeConfigOverride
   // Parse values from .wakeroot
   _config->set_all<WakeConfigProvenance::WakeRoot>(wakeroot_json);
 
+  // Sometimes we need to the user_config with an env-var so we check env-vars first here
+  _config->set_all_env_var();
```
V-FEXrt, line 342: The priority of each source is a bit unclear to me. Can we document that directly somewhere?

src/runtime/config.h
```diff
@@ -82,12 +87,14 @@ struct UserConfigPolicy {
   static constexpr bool allowed_in_userconfig = false;
   type user_config;
   static constexpr type UserConfigPolicy::*value = &UserConfigPolicy::user_config;
-  static constexpr Override<input_type> override_value = nullptr;
+  static constexpr Override<input_type> override_value = &WakeConfigOverrides::user_config;
+  static constexpr const char* env_var = "WAKE_USER_CONFIG";
```
V-FEXrt, line 91: Can you update  the  config tests? I'd like to see a test where a value is sourced from the env

V-FEXrt (changes requested):

V-FEXrt (approved):

JakeSiFive:
1) Declares the wake shared cache no longer experimental and changes the env-var used for it to be WAKE_LOCAL_JOB_CACHE
2) Adds the ability to configure some config options via environment variables. In particular the user config can now be specified via an env-var which allows for greatly increased flexibility. I did however additionally allow env-vars for cache miss on failure, and cache size manipulation in case that's helpful but we anticipate only
2) I went ahead and added a command line override for user-config so that you can locally fix what might become a shell-global override of .wakeroot

PR #1392, opened 2023-08-16
Add bulk logging env var for wake

JakeSiFive:
This change adds the WAKE_BULK_LOGGING_DIR environment variable. If turned on, wake will emit a copy all logs to that directory, with each file having its hostname, pid, the current day as part of the file name. This ensures that even over NFS, logging should stay atomic. NFS has no append atomicity but because wake is single threaded and the pid and hostname are included things should be atomic.

The day in the file name is mostly so that tools and scripts for parsing logs can pre-filter the logs they want to view based on the time frames without parsing all of the logs. In theory if a multi-day process is running on the same host with the same pid, or a pid gets recycled some days later this would also be midly helpful but those are quite rare cases.

Log cleanup can be done by deleting files with an expiered modtime but that system is not implemented in wake itself for bulk logging like it is for cache logging.

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -50,7 +50,7 @@ namespace {
 
 using namespace job_cache;
 
-static void initialize_logging() {
+static void initialize_logging(std::string bulk_dir) {
   const char *str_fmt = "%Y-%m-%d";
   const size_t str_fmt_len = 10;         // count("XXXX-XX-XX")
   const size_t filename_prefix_len = 7;  // count (".cache.")
@@ -72,8 +72,7 @@ static void initialize_logging() {
   }
   wcl::log::subscribe(std::make_unique<JsonSubscriber>(std::move(*log_file_res)));
 
-  const char *bulk_dir = getenv("WAKE_BULK_LOGGING_DIR");
-  if (bulk_dir) {
+  if (!bulk_dir.empty()) {
     std::string pid = std::to_string(getpid());
     char buf[512];
     if (gethostname(buf, sizeof(buf)) != 0) {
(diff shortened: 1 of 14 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -72,6 +72,29 @@ static void initialize_logging() {
   }
   wcl::log::subscribe(std::make_unique<JsonSubscriber>(std::move(*log_file_res)));
 
+  const char *bulk_dir = getenv("WAKE_BULK_LOGGING_DIR");
```
V-FEXrt, line 75: Not blocking on this, but I'd really like everything like this eventually get moved into WakeConfig. Would be nice to have a singular thing to audit when trying to figure out what goes into wake outside of the wake files

src/job_cache/daemon_cache.cpp
```diff
@@ -72,6 +72,29 @@ static void initialize_logging() {
...
+  if (bulk_dir) {
+    std::string pid = std::to_string(getpid());
+    char buf[512];
+    if (gethostname(buf, sizeof(buf)) != 0) {
+      std::cerr << "urgent warning: Could not init logging: gethostname(): " << strerror(errno)
```
V-FEXrt, line 80: lol "urgent warning" instead of "error" is a funny concept but I get why its called that. The program doesn't exit.

tools/wake/main.cpp
```diff
@@ -333,6 +344,27 @@ int main(int argc, char **argv) {
   }
   wcl::log::subscribe(std::make_unique<JsonSubscriber>(std::move(*res)));
 
+  // Bulk logging
+  const char *bulk_dir = getenv("WAKE_BULK_LOGGING_DIR");
```
V-FEXrt, line 348: yeah, the fact that we have this in two places makes me feel more strongly about using WakeConfig

V-FEXrt (approved):

tools/wake/main.cpp
JakeSiFive: I can put this in WakeConfig, I probably should have but I was being lazy about it

JakeSiFive: added

src/job_cache/daemon_cache.cpp
JakeSiFive: ok done!

V-FEXrt (approved):

V-FEXrt:
Thanks!

PR #1394, opened 2023-08-21
wakebox: set loopback device to UP

mmjconolly:
I'm not sure if we should merge this as-is or not, as we could implement this manually rather than shelling out to `ip`/`ifconfig`

This PR sets the loopback device to `UP` when wakebox is asked to isolate the network namespace.
A command could choose to make a tcp connection with itself, or other running processes within the namespace.
Currently they would have to raise the loopback interface themselves to use it.

From:
```
sh-4.4# ip addr  
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN group default qlen 1000   
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00  
```
To:
```
sh-4.4# ip addr      
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000  
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00  
    inet 127.0.0.1/8 scope host lo    
       valid_lft forever preferred_lft forever   
    inet6 ::1/128 scope host  
       valid_lft forever preferred_lft forever 
```


mmjconolly:
This change is not exercised by CI, as it would have failed on which ever docker containers don't have `ip` installed

mmjconolly:
Added a test so we can at least see it fail on any docker containers that don't have `ip` installed

JakeSiFive:
I think this failed not because the program itself failed but because the tool didn't output "UP" as a result, it output something else for some reason.

mmjconolly:
It has an empty output, because it didn't find an `UP`, I've neglected to get it to print the error-output

mmjconolly:
This passes CI now
This will only affect jobs that try to isolate networking

I'm somewhat inclined to have it continue to just shell out, rather than manually implementing the link-up logic which seems to have some significant boilerplate. If we really want to avoid the fork/exec, there is some LGPL code that I saw that might be a reasonable candidate to use as a library

Later when we start using `slirp4netns` we'd be shelling out to setup the really custom networking, so it's longer term there's no avoiding shelling out.

JakeSiFive:
I'm happy shelling out, I don't think the cost is significant after the ip executable is cached in memory. `ip` seems quite standard as well, I'd expect it to be on any Debian, Arch, Fedora based distro and its on Alpine too. That covers a lot of distros and I'm sure more have it. If we have a need to support a distro that doesn't have `ip` for some reason we can discuss it then.

JakeSiFive (approved):
I checked every linux distro on this list up until "Linux from scratch" and they all seem to use ip over ifconfig: https://distrowatch.com/search-mobile.php?ostype=All&category=All&origin=All&basedon=Independent&notbasedon=None&desktop=All&architecture=All&package=All&rolling=All&isosize=All&netinstall=All&language=All&defaultinit=All&status=Active#simple

mmjconolly:
Yeah `ifconfig` is really old now

JakeSiFive:
In general I prefer only relaying on libc and kernel features but we're already not doing that partly with fuse and I feel like `ip` falls under the same category as fuse

PR #1396, opened 2023-08-22
Cleaning up some mistakes in rust code

JakeSiFive:
This change does the following:

1) Changes "gsc" to "rsc" in accordance with our choice to rename "global shared cache" to "remote shared cache"
2) Switches from the "atty" library to the "is_terminal" library (for checking if a given file-like thing is a terminal)
3) Switches from the "gumdrop" library to the "clap" library (for command line argument parsing)
4) Switches from the "color_eyre" library to "miette" (for nicely rendering errors from many soruces)

The change from atty to is_terminal is just because miette already uses is_terminal but they're essentially equivalent libraries.

The change from gumdrop to clap is because a) clap is a richer library for command line arguments but also b) after we consider all the things that miette uses, they overlap almost entirely with what miette uses or other libraries we've independently decided to use. So using gumdrop doesn't really save us any deps in actuality.

The change from color_eyre to miette is because a) miette is a richer library for error rendering but also b) after considering the entire tree of deps that we must have the extra deps of miette over color_eyre, we don't save anything by using color_eyre. (same as for clap)

TL;DR we avoided using the richer and more standard miette and clap because they had so many deps, in actuality after we built things up and looked at the essential deps we needed, miette and clap don't add very much as long as we stick to some of their choices of sub libraires (like owo-colors over other color libraries)


rust/log_viewer/src/main.rs
```diff
@@ -211,10 +206,10 @@ fn render(color_map: HashMap<i32, Color>, events: Vec<Event>) -> Result<(), Repo
     Ok(())
 }
 
-fn main() -> Result<(), Report> {
-    color_eyre::install()?;
+fn main() -> Result<(), Error> {
+    //color_eyre::install()?;
```
V-FEXrt, line 210: commented code

rust/rsc/src/rsc_tool/main.rs
```diff
@@ -117,82 +117,73 @@ async fn remove_api_key(
...
-    #[options(command)]
+    #[command(subcommand)]
     api_key_command: Option<ApiKey>,
 }
 
-#[derive(Debug, Options)]
+#[derive(Debug, Subcommand)]
 enum ApiKey {
-    #[options(help = "list all api keys in the database")]
+    //#[options(help = "list all api keys in the database")]
```
V-FEXrt, line 144: here too

V-FEXrt (approved):
overall LGTM, but there is commented out code that needs to be deleted

PR #1397, opened 2023-08-22
Cleanup what runOnce considers as required for re-run

JakeSiFive:
We agreed on the following changes for runOnce

1) The visible list should cause a re-run. This will what is today a warning into an error that says that two different jobs had the same output rather than saying that the same job was requested with different auxiliary params
2) Likewise, isatty was always supposed to be part of the primary key and is considered as such in the database already
3) If a different runner is used, the behavoir of the program will change accordingly so this is now part of the primary key of runOnce.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -656,7 +656,7 @@ def runAlways cmd env dir stdin res uusage finputs foutputs vis keep run echo st
                 Pair Nil last = confirm False last (build Unit)
 
 # Only run if the first four arguments differ
-target runOnce cmd env dir stdin \ res usage finputs foutputs vis keep run echo stdout stderr label isatty =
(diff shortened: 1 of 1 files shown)


mmjconolly:
Is there any follow-on changes need in the sqlite database to handle this? Or does the db not know about this level of abstraction?

JakeSiFive:
The answer is different in each case but we don't need to do anything in the database. The database was already handing isatty correct despite runOnce not handling it correctly. The database currently only records the job inputs and not all the visible files so that behavior will remain unchanged. Eventually as noted I'd like to make the database and runOnce behavoir match by making the database track all inputs as visible. The runner itself is not info that the database knows about.

The runner case is curious though. It could occur that two separate jobs run that from the eyes of the database look the same but such that no error would occur at runtime because the runner sufficiently changed the launch parameters.

mmjconolly (approved):

mmjconolly:
I ran a single large build using this and it didn't complain about anything

PR #1398, opened 2023-08-23
Fix `in` documentation

bmitc:
The documentation mentioned `join` and not `in`, so the documentation has been corrected. `join` is the more common name here, so I'd be amenable to that change. But this PR just makes a start at addressing this by correcting the documentation to the currently defined name. It appears that `in` is intended to be used in a pipeline, as in `<path> | in <directory>`, which is why the name is slightly off for normal use of joining two paths.

### DIFF
--- a/share/wake/lib/system/path.wake
+++ b/share/wake/lib/system/path.wake
@@ -27,10 +27,10 @@ export def sortPaths (paths: List Path): List Path =
     sortBy fn paths
 
 # Concatenate two paths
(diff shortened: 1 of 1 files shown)


JakeSiFive (commented):
The next wake is going to be a major version anyway so I think its fine to rename this to `join` but please run some large build to ensure people aren't using this for something else in a conflicting way.

JakeSiFive (approved):
Approving in case you don't want to rename it right now

PR #1399, opened 2023-08-23
Improve wake-hash errors

JakeSiFive:
Recently we've had a lot of issues with people seeing error messages like this:

```
UNREACHABLE: wake hash expected X lines but received Nil
```

This isn't really unreachable if you hit some sort of issue with fnoutputs however but previously this code assumed that no such issues would ever occur (this is manifestly false).

To fix this I do the following

1) If the exit code of wake-hash is non-zero for any reason all hashed files from that job are considered BadHash and will thus later trigger a failure when `getJobOutputs` is called but the error will not point blame at wake-hash and will instead just say that it couldn't hash some random file (which one is chosen is up to which BadHash is checked first)
2) If any specific file cannot be hashed, wake-hash will return BadHash for it and thus that specific file will later give an error with getJobOutputs which should make people's lives easier
3) If for some reason wake still gets this "expected X lines but received Nil" error after those two cases, we now call it at "PANIC" instead of "UNREACHABLE" (I cleaned up one other such unreachable that we commonly see as well)

Additionally it recently came to light that sometimes we wish to have a job "hash" a domain socket similar to how we "hash" directories. So we now consider anything that can be opened as read only that isn't a symlink, directory, or regular file to be "exotic" and to have a hash of 1 (similar to how directories have a hash of 0). This will allow jobs to output exotic "files" if they wish to but this should be avoided by 99% of users.

### DIFF
--- a/share/wake/lib/core/syntax.wake
+++ b/share/wake/lib/core/syntax.wake
@@ -97,6 +97,12 @@ export def wait (f: a => b) (x: a): b =
 # 200
 # ```
 export def unreachable (reason: String): a =
-    def f x = prim "unreachable"
+    def f x = prim "panic"
 
-    f reason
+    f "REACHED UNREACHABLE CODE: {reason}"
+
+# Like unreachable but with a different error message
+export def panic (reason: String): a =
+   def f x = prim "panic"
+
+   f "PANIC: {reason}"
--- a/share/wake/lib/core/syntax.wake
+++ b/share/wake/lib/core/syntax.wake
@@ -97,12 +97,12 @@ export def wait (f: a => b) (x: a): b =
 # 200
 # ```
 export def unreachable (reason: String): a =
-    def f x = prim "panic"
+    def f x = prim "unreachable"
 
     f "REACHED UNREACHABLE CODE: {reason}"
 
 # Like unreachable but with a different error message
 export def panic (reason: String): a =
-   def f x = prim "panic"
+    def f x = prim "panic"
 
-   f "PANIC: {reason}"
+    f "PANIC: {reason}"
--- a/src/runtime/exception.cpp
+++ b/src/runtime/exception.cpp
@@ -53,16 +53,16 @@ static PRIMFN(prim_stack) {
   RETURN(claim_list(runtime.heap, objs.size(), objs.data()));
 }
 
-static PRIMTYPE(type_unreachable) {
+static PRIMTYPE(type_panic) {
   return args.size() == 1 && args[0]->unify(Data::typeString);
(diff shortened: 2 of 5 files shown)


share/wake/lib/core/syntax.wake
```diff
@@ -97,6 +97,12 @@ export def wait (f: a => b) (x: a): b =
 # 200
 # ```
 export def unreachable (reason: String): a =
-    def f x = prim "unreachable"
+    def f x = prim "panic"
```
richardxia, line 100: Should this stay as `prim "unreachable"`, since the comment you wrote in the primitive function registration code says that we do want to differentiate between panic and unreachable?

tools/wake-hash/main.cpp
```diff
@@ -117,9 +118,18 @@ struct Hash256 {
   bool operator!=(Hash256 other) { return !(*this == other); }
 };
 
-static Hash256 hash_dir() { return Hash256(); }
+// If a file handle is not a symlink, directory, or regular file
+// then we consider it "exotic". This includes block devices, 
+// character devices, FIFOs, and scokets.
```
richardxia, line 123:
```suggestion
// character devices, FIFOs, and sockets.
```

richardxia (commented):
I don't think I'm familiar enough with the internals of the code to give a competent review, but your high level description of what the new behavior is looks good to me!

share/wake/lib/core/syntax.wake
JakeSiFive: Yep! Good catch!

src/runtime/exception.cpp
```diff
@@ -85,7 +85,9 @@ static PRIMFN(prim_true) {
 void prim_register_exception(PrimMap &pmap) {
   // These should not be evaluated in const prop, but can be removed
   prim_register(pmap, "stack", prim_stack, type_stack, PRIM_ORDERED);
-  prim_register(pmap, "unreachable", prim_unreachable, type_unreachable, PRIM_ORDERED);
+  prim_register(pmap, "panic", prim_panic, type_panic, PRIM_ORDERED);
```
V-FEXrt, line 88: The parameters on one of these should be different right? To do the whole dead code elim allowed/denied?

JakeSiFive: unreachable is treated specially. It's not that ordered makes it special, its that it has the name "unreachable" that makes it special.

JakeSiFive: no wait, crap you're right.

V-FEXrt (approved):

PR #1401, opened 2023-08-23
Fix wake-format find_all_files flake 

V-FEXrt:
Previously `ok` is uninitialized and never set to true on success.  If the underlying memory happened to be set to `0` then it would incorrectly report a failure even though none had occurred. Set the default to true to avoid that case

### DIFF
--- a/tools/wake-format/main.cpp
+++ b/tools/wake-format/main.cpp
@@ -198,7 +198,7 @@ int main(int argc, char **argv) {
   FILE *user_warn = stdout;
   user_warn = fopen("/dev/null", "w");
   if (auto_find_files) {
-    bool ok;
+    bool ok = true;
     wakefiles = find_all_wakefiles(ok, true, false, ".", ".", user_warn);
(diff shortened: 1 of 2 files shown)


tools/wake-format/main.cpp
```diff
@@ -198,7 +198,7 @@ int main(int argc, char **argv) {
   FILE *user_warn = stdout;
   user_warn = fopen("/dev/null", "w");
   if (auto_find_files) {
-    bool ok;
+    bool ok = true;
     wakefiles = find_all_wakefiles(ok, true, false, ".", ".", user_warn);
```
JakeSiFive, line 202: Can you set it to true inside the function as well? I don't like functions with such difficult and hidden pre-reqs. We should set it true in both places so that neither side of the API is making an assumption.

JakeSiFive (changes requested):

JakeSiFive (approved):

PR #1403, opened 2023-08-25
Make getJobStd{err,out} check job status before returning

JakeSiFive:
This change makes it so that getJobStdout and getJobStderr first check the job status before returning a Pass, and return a Fail if its non-zero.

To match the old behavoir getJobFailedStdout and getJobFailedStderr were added.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -494,8 +494,8 @@ export def mkJobCacheRunner (hashFn: Result RunnerInput Error => Result String E
 
                 JArray (map mkVisJson readPaths)
 
-            require Pass stdout = getJobStdout job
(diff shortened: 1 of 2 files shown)


share/wake/lib/system/job.wake
```diff
@@ -777,6 +789,16 @@ export def getJobStderr (job: Job): Result String Error =
 
     Pass (stderr | filterTerminalCodes)
 
+export def getJobFailedStdout (job: Job): Result String Error =
```
mmjconolly, line 792: minor: please add comments for humans and LSP

mmjconolly (approved):

V-FEXrt (approved):

PR #1409, opened 2023-08-30
Use lock files when building rust code

JakeSiFive:
Previously I did not use the generated lock files and had git ignore them so all libraries were pinned at latest. Sometimes a library can become incompatible with an older version of rust however Now they're checked in for all rust code and its recommended that you use `cargo build --locked` to build code unless you intend on changing a lock file.

These lock files assume rust/cargo 1.70.0


mmjconolly:
> 1.70.0

We had generally been using 1.69.0, is there something that we need from 1.70.0?

JakeSiFive:
> We had generally been using 1.69.0, is there something that we need from 1.70.0?

I've been using 1.70.0 locally and most of the code has not been exercised by 1.69.0 as a consequence.

JakeSiFive:
one sec let me update the docker files too, whoops

JakeSiFive:
Oh no those are update to date I just messed up the formatting. So yeah wake has been using 1.70.0 exclusively. I avoided updating to 1.71.0 but we will want to upgrade to 1.72.0 when it comes out for a couple things.

mmjconolly (approved):

PR #1412, opened 2023-09-07
Rewrite match `=` as `->` inside of `match`

V-FEXrt:
Updates `wake-format` rule to always emit `->` as the match case separator which shows the correct data flow direction.

### DIFF
--- a/tests/tests.wake
+++ b/tests/tests.wake
@@ -37,12 +37,12 @@ export topic wakeTestBinary: Unit => Result (Pair String (List Path)) Error
 export topic wakeUnitTestBinary: (variant: Pair String String) => Result (List Path) Error
 
 def wakeToTestDir Unit = match (subscribe wakeTestBinary)
-    buildTestWake, Nil =
+    buildTestWake, Nil ->
         require Pass (Pair path visible) = buildTestWake Unit
 
         Pass (Pair (simplify "{path}/..") visible)
-    Nil = Pass (Pair "{wakePath}" Nil)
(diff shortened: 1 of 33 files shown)


tools/wake-format/emitter.cpp
```diff
@@ -1591,15 +1591,13 @@ wcl::doc Emitter::walk_flag_global(ctx_t ctx, CSTElement node) {
 
 wcl::doc Emitter::walk_guard(ctx_t ctx, CSTElement node) {
   MEMO(ctx, node);
-  MEMO_RET(walk_placeholder(ctx, node));
```
ag-eitilt, line 1594: I assume `walk_placeholder` is used elsewhere and so doesn't become dead code?

tests/tests.wake
ag-eitilt, line 1: This file is the only one which seems to be related to testing -- don't the unit/golden tests have to be updated to reflect this as well?

ag-eitilt (commented):
So far as I'm able to read the formatter code, this looks good to me -- the `*.wake` changes certainly seem to be doing the right things.

tools/wake-format/emitter.cpp
V-FEXrt: yeah, its the main fallback function

tests/tests.wake
V-FEXrt: they do, I missed them

JakeSiFive (approved):
let 'er rip

PR #1413, opened 2023-09-08
Rewrite all multiline `"%` strings to be `"""` strings

V-FEXrt:
Replaces `"% ... %"` with the equivalent `""" ... """` in preparation for eventual removal


JakeSiFive (approved):
A couple more releases and we can remove the code for those too!

V-FEXrt:
yus that's the goal!

PR #1417, opened 2023-09-12
Add primitive for stat to avoid a job or string parsing

JakeSiFive:
This change adds a fast way to get the following information for a file that's already been hashed

1) inode permissions (e.g.
2) inode type (e.g. regular file, symlink, directory, or something more exotic like a pipe or socket)
3) file size

This is exposed via a new function in the standard library called "stat" that returns a `Result Stat Error` which in turn (if its a Pass) contains that information.

Other information contained in the posix stat struct is volatile and inconsistent across machines so I did not add it but these should be relatively stable.

### DIFF
--- a/share/wake/lib/core/result.wake
+++ b/share/wake/lib/core/result.wake
@@ -95,13 +95,12 @@ export def rmap (fn: a => b): Result a fail => Result b fail = match _
     Pass a -> Pass (fn a)
     Fail f -> Fail f
 
-# rmapError: apply a function to a Failing-ing result
+# rmapError: apply a function to a Fail-ing result
 #
 #   rmapError (_+1) (Pass 123) = Pass 123
 #   rmapError (_+1) (Fail 123) = Fail 124
-export def rmapError (fn: a => b): Result c a => Result c b = match _
-    Pass a -> Pass a
-    Fail f -> Fail (fn f)
+export def rmapError (fn: a => b): Result c a => Result c b =
+    rmapFail (fn _ | Fail)
 
 # rmapPass: apply a Fail-able function to a Pass-ing result
 # If you find yourself using the function, consider using require instead.
--- a/share/wake/lib/core/result.wake
+++ b/share/wake/lib/core/result.wake
@@ -100,8 +100,8 @@ export def rmap (fn: a => b): Result a fail => Result b fail = match _
 #   rmapError (_+1) (Pass 123) = Pass 123
 #   rmapError (_+1) (Fail 123) = Fail 124
 export def rmapError (fn: a => b): Result c a => Result c b = match _
-    Pass a = Pass a
-    Fail f = Fail (fn f)
+    Pass a -> Pass a
+    Fail f -> Fail (fn f)
 
 # rmapPass: apply a Fail-able function to a Pass-ing result
 # If you find yourself using the function, consider using require instead.
--- a/share/wake/lib/core/result.wake
+++ b/share/wake/lib/core/result.wake
@@ -95,6 +95,14 @@ export def rmap (fn: a => b): Result a fail => Result b fail = match _
     Pass a -> Pass (fn a)
     Fail f -> Fail f
 
+# rmapError: apply a function to a Failing-ing result
+#
+#   rmapError (_+1) (Pass 123) = Pass 123
+#   rmapError (_+1) (Fail 123) = Fail 124
+export def rmapError (fn: a => b): Result c a => Result c b = match _
+    Pass a = Pass a
+    Fail f = Fail (fn f)
+
 # rmapPass: apply a Fail-able function to a Pass-ing result
 # If you find yourself using the function, consider using require instead.
 export def rmapPass (fn: a => Result b fail): Result a fail => Result b fail = match _
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -30,9 +30,23 @@ export tuple Permission =
     Execute: Boolean
 
 export def perm2bits (perm: Permission) =
-    def b3 = if getPermissionRead perm then 4 else 0
-    def b2 = if getPermissionWrite perm then 2 else 0
-    def b1 = if getPermissionExecute perm then 1 else 0
+    def b3 =
+        if getPermissionRead perm then
+            4
+        else
+            0
+
+    def b2 =
+        if getPermissionWrite perm then
+            2
+        else
+            0
+
+    def b1 =
+        if getPermissionExecute perm then
+            1
+        else
+            0
 
     b1 + b2 + b3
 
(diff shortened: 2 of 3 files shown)


share/wake/lib/core/result.wake
```diff
@@ -95,6 +95,14 @@ export def rmap (fn: a => b): Result a fail => Result b fail = match _
     Pass a = Pass (fn a)
     Fail f = Fail f
 
+# rmapError: apply a function to a Failing-ing result
```
V-FEXrt, line 98: nit: `Failing-ing` -> `Fail-ing`

share/wake/lib/system/io.wake
```diff
@@ -15,6 +15,40 @@
...
+    PathTypeDirectory
+    PathTypeCharacterDevice
+    PathTypeBlockDevice
+    PathTypeFifo
+    PathTypeSymlink
+    PathTypeSocket
+
+export tuple Stat =
+    Type: PathType
+    Mode: Integer
```
V-FEXrt, line 29: Can we/should we provide a better wrapper around mode? Just because linux uses a int here doesn't mean we have to. 

I could imagine something like 

```
export tuple Permission =
    Read: Bool
    Write: Bool
    Execute: Bool

export def Mode = 
    User: Permission
    Group: Permission
    Other: Permission
```

src/runtime/string.cpp
```diff
@@ -192,6 +192,86 @@ static PRIMFN(prim_breadcrumb) {
...
+  size_t max_error = path->size() + 100;
+
+  struct stat buf;
+  if (stat(path->c_str(), &buf) < 0) {
+    std::stringstream str;
+    str << "stat " << path->c_str() << ": " << strerror(errno);
+    std::string s = str.str();
+
+    size_t len = std::min(s.size(), max_error);
+    String *out = String::claim(runtime.heap, s.c_str(), len);
```
V-FEXrt, line 224: Can you confirm `claim` will add the expected `\0` in the case that `s.size() > max_error`?

src/runtime/string.cpp
```diff
@@ -192,6 +192,86 @@ static PRIMFN(prim_breadcrumb) {
...
+    case S_IFLNK:
+      type = 5;
+      break;
+    case S_IFSOCK:
+      type = 6;
+      break;
+  }
+
+  runtime.heap.reserve(reserve_result() + Integer::reserve(mode) + Integer::reserve(type) +
+                       Integer::reserve(buf.st_size) + 2 * reserve_tuple2());
```
V-FEXrt, line 265: I think the size reservations here could use a comment.

its not imeditaly clear that a fake tuple3 (`tuple2(thing, tuple2(...,..))`) takes up the space of 2 tuple2

share/wake/lib/system/io.wake
JakeSiFive: I thought the same thing but I got lazy and didn't do it. I'll work on it.

src/runtime/string.cpp
JakeSiFive: This is what we do for all error messages so I just cargo cult-ed it.

src/runtime/string.cpp
JakeSiFive: They match the calls below. No where do I say that I'm making a fake tuple3. I just make two pairs. I happen to do this to avoid creating a new record type.

mmjconolly:
I wonder if we could have the stat information as a field within the `Path` datatype, would this be too expensive?

share/wake/lib/system/io.wake
```diff
@@ -15,6 +15,40 @@
...
+    Size: Integer
```
mmjconolly, line 30: perhaps
```suggestion
    SizeBytes: Integer
```

JakeSiFive:
> I wonder if we could (additionally) have the stat information as a field within the Path tuple, would this be too expensive?
I think the cheaper way to do it would be to make `stat` a `target` so that it saves the result and then to make it "feel" like a field we can give it a name like `getPathStat` but I'd rather just use the name `stat` instead for that. By being a target it won't be re-computed unless required but at the same time we won't stat paths that we don't have to. The other issue is that we don't hash the information in the stat struct right now so while it shouldn't be possible with well behaved jobs to have non-deterministic stat info, it would be possible to change a field somewhere and have what is today the same Path show up as distinct.

We should probably discuss the merits of hashing the stat info as part of the Path hash before we store it in the Path itself. I've long wondered if we should be hashing stat info for directories. The other thing that occured to me is that a symlink to foo/bar/baz will hash to the same thing as a file containing the string foo/bar/baz today which is quite awkward if you ask me.

share/wake/lib/core/result.wake
```diff
@@ -95,6 +95,14 @@ export def rmap (fn: a => b): Result a fail => Result b fail = match _
     Pass a -> Pass (fn a)
     Fail f -> Fail f
 
+# rmapError: apply a function to a Failing-ing result
+#
+#   rmapError (_+1) (Pass 123) = Pass 123
+#   rmapError (_+1) (Fail 123) = Fail 124
+export def rmapError (fn: a => b): Result c a => Result c b = match _
+    Pass a -> Pass a
+    Fail f -> Fail (fn f)
```
ag-eitilt, line 104: This is definitely a more comfortable signature than `rmapFail`, but not everything would be an `Error` given the freer typing, so the name's a bit misleading.  It might also be nice to define it in terms of `rmapFail (fn _ | Fail)`, or if you keep the raw `match` for simplicity please at least make the `Pass a` binding use the same letter as the `Result c x` annotation.

share/wake/lib/system/io.wake
```diff
@@ -15,6 +15,92 @@
...
+    def otherBits = perm2bits mode.getModeOther
+    def groupBits = perm2bits mode.getModeGroup << 3
+    def userBits = perm2bits mode.getModeUser << 6
+
+    otherBits + groupBits + userBits
+
+export tuple Stat =
+    Type: PathType
+    Mode: Mode
+    SizeBytes: Integer
```
ag-eitilt, line 82: This feels like a good interface; the number of files we'll have to `stat` is much less than the number of `Path`s we produce, so it's likely unnecessary overhead to add to every call.  That said, though, your symlink-vs-standard-file hash conflict does make it seem like we should be taking this into account anyway (and directories created with different modes can have very noticeable differences when trying to write files within them), and if we're `stat`ing everything to be able to differentiate the hashes then we might want to go the `getPathType`/`*Mode`/`*SizeBytes` route to keep the conceptually-related info centralized.

JakeSiFive: Yeah I think in an ideal system we would be hashing the mode and type and as long as we're doing that we should also have the size info available (hashing the size is a nop since hash functions of content are already size sensitive)

JakeSiFive (commented):
lol, I think neither of us can approve this? Well I approve so go ahead ad land it after it passes tests.

V-FEXrt (approved):

PR #1423, opened 2023-09-21
cache: Fix bug in runner hash accounting

V-FEXrt:
The runtime has an accounting bug regarding the wakelang provided `runner_hash`. This lead to situations where wakelang would correctly declare two jobs as different yet the shared cache would still see them as identical.

This bug caused some jobs to resolve with stale/incorrect outputs.

The accounting error has been corrected and a test case was added that ensures a change in the runner hash results in a cache miss.

### DIFF
--- a/src/job_cache/types.cpp
+++ b/src/job_cache/types.cpp
@@ -446,6 +446,7 @@ AddJobRequest::AddJobRequest(const JAST &json) {
   ibytes = std::stoull(json.get("ibytes").value);
   obytes = std::stoull(json.get("obytes").value);
   client_cwd = json.get("client_cwd").value;
+  runner_hash = json.get("runner_hash").value;
   if (wcl::is_relative(client_cwd)) {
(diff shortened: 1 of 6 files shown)


mmjconolly (approved):
rubber stamp

JakeSiFive (approved):
Awesome! Thanks for fixing this while I'm out. Clever idea using time instead of two separate jobs.

JakeSiFive:
If `date` is the issue on alpine you can do this instead

```
def makeTestPlan (s: String): Plan =
    makePlan <insert stuff here>
    | setResourceString s

def resourceAwareCacheRunner =
    <do all the stuff to make a cache runner with the hash function>

def runTestJob (s: String) =
    makeTestPlan s
    | runJobWith resourceAwareCacheRunner
    | isJobOk

export def test1 = runTestJob "test1"
export def test2 = runTestJob "test2"
```

And then instead of calling test twice you first call test1 and then test2

It's less clever but it has fewer deps.

PR #1427, opened 2023-10-03
Delete spaces trailing after a comment

V-FEXrt:
When parsing a TOKEN_COMMENT node everything until a newline (including trailing whitespace) gets slurped into the 'contents' of the comment. We don't want to re-emit this whitespace so it has to be manually removed when emitted comments.

An alternative implementation would be to manually edit the CST after parsing to remove trailing whitespace from any TOKEN_COMMENT node but editing the CST seems ill advised if its even possible.

### DIFF
--- a/tools/wake-format/actions.h
+++ b/tools/wake-format/actions.h
@@ -28,7 +28,7 @@
 
 static inline wcl::doc rtrim_comment(const CSTElement& node) {
   std::string comment = node.fragment().segment().str();
-  comment.erase(comment.find_last_not_of(" ") + 1);
+  comment.erase(comment.find_last_not_of(" \t") + 1);
   return wcl::doc::lit(comment);
 }
 
--- a/tools/wake-format/actions.h
+++ b/tools/wake-format/actions.h
@@ -26,6 +26,13 @@
 #include "predicates.h"
 #include "types.h"
 
+
+static inline wcl::doc rtrim_comment(const CSTElement& node) { 
+    std::string comment = node.fragment().segment().str();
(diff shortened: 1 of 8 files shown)


tools/wake-format/actions.h
```diff
@@ -101,7 +101,10 @@ struct TokenReplaceAction {
         if (n.id() == TOKEN_COMMENT) {
           // Realign indent before writing the comment
           freshline(builder, ctx);
-          builder.append(n.fragment().segment().str());
+
+          std::string comment = n.fragment().segment().str();
+          comment.erase(comment.find_last_not_of(" ") + 1);
```
JakeSiFive, line 106: You should add tab here as well

tools/wake-format/emitter.cpp
```diff
@@ -580,7 +580,7 @@ wcl::doc Emitter::walk(ctx_t ctx, CSTElement node) {
       is_floating_comment,
       fmt()
           .fmt_while(TOKEN_COMMENT,
-                     fmt().token(TOKEN_COMMENT).freshline().fmt_if(TOKEN_NL, fmt().next()))
+                     fmt().walk(WALK_TOKEN).freshline().fmt_if(TOKEN_NL, fmt().next()))
```
JakeSiFive, line 583: Why do we need this to be duplicated? Can't we just use the TokenReplaceAction?

V-FEXrt: ah it shouldn't be. I only thought it was needed because I missed some cases in TokenReplaceAction

JakeSiFive (approved):

PR #1428, opened 2023-10-03
Fix tokenize segfault when too many tokens are output

JakeSiFive:
The output array from which we construct the token list in prim tokenize was incorrectly allocated. This switches it to use a vector instead. This issue was never seen before because no one tried to do a split this big until now apparently. A user encountered this issue in the wild and I'm still not 100% sure that this will resolve their issues but I'm going to try that next.


JakeSiFive:
I've confirmed that without this change the following causes a segfault, but with this change it does not:

```
wake -x 'seq 20000000 | map (\_ "") | catWith "a" | tokenize `a`'
```

V-FEXrt (approved):

PR #1429, opened 2023-10-04
Sync master to 39.* releases

V-FEXrt:
Merges release-39.0.0 back into master which branch for various hotflixes. This will pull through the correct updates to the debian changelog

### DIFF
--- a/debian/changelog.in
+++ b/debian/changelog.in
@@ -2,6 +2,12 @@ wake (@VERSION@-1) unstable; urgency=medium
 
   * Auto-generated release package.
 
+ -- Ashley Coleman <ashley.coleman@sifive.com>  Fri, 22 Sep 2023 10:00:00 -0800
+
+wake (39.0.0-1) unstable; urgency=medium
+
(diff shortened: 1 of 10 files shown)


V-FEXrt:
I'm not sure whats going on with the diff here. I'd expect only the changelog.in to be different since all the other changes where cherry-picked from master into the release branch

JakeSiFive:
I think just do a normal release, no need to merge right?

mmjconolly:
I'd have assumed all the patches that went out on the patch releases would have already been put in`master` branch

V-FEXrt:
All the patches are already on `master` which is the strange part. The only thing not on master is the changes to `debian/changelog.in` that is what I want to pull back in

V-FEXrt:
> I think just do a normal release, no need to merge right?

I'm not trying to do a release here, just trying to sync master and the release branch

mmjconolly:
> All the patches are already on master which is the strange part

We'd expect that as we landed the PRs there first, before cherry-picking them into release branches

PR #1431, opened 2023-10-05
Don't exit LSP when manifest error occurs

V-FEXrt:
When editing wake code with a running LSP an author may end up in a position where a file exists in the editor but not on disk. (rename or unsaved new file) that they then (correctly) add to the `.wakemanifest` file. This creates an issue where to the user the file exists but to wake source discovery it doesn't. When source discovery sees a file listed that doesn't exist it will exit with an error which is correct for wake, but not the LSP. 

This change adds a strict flag to source discovery that the LSP can set to false. This allows the LSP to keep going even if the manifest doesn't fully agree.

In the long term, we should do a better job of managing the source lifetime within the LSP and we should report the missing file as an LSP notification but that change will be difficult without a significant refactor

### DIFF
--- a/src/parser/wakefiles.cpp
+++ b/src/parser/wakefiles.cpp
@@ -488,7 +488,8 @@ std::vector<std::string> find_workspace_wakefiles_via_search(bool &ok, bool verb
 }
 
 std::vector<std::string> find_workspace_wakefiles_via_manifest(const std::string &manifest_path,
-                                                               const std::string &workdir) {
+                                                               const std::string &workdir,
+                                                               bool strict) {
   std::vector<std::string> workfiles;
   std::vector<std::string> unreadable_files;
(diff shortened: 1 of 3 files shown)


V-FEXrt:
@ag-eitilt @bmitc FYI

mmjconolly:
Does the LSP consider the currently-editing file as part of the manifest even if it's not listed yet?

V-FEXrt:
No, we don't do anything smart with the currently edited file. If its not on disk and not part of the manifest the LSP features aren't enabled

We could possibly make unsaved+not-in-manifest files have LSP features but once you save the file its treated as an "on disk" file instead of an "in memory" file and all on-disk files must be listed in the manifest

JakeSiFive (approved):

PR #1432, opened 2023-10-06
Generate prim.wake representing internal types

V-FEXrt:
This is a somewhat hacky (and quite slow) script that I've been wanting for a long time. 

Basically the script asks `wake` a bunch of questions about the installed `prim`s then outputs typed `def`s representing them. 

This should alleviate some of the "string typing" that the prims have in wake source.

run with `python3 scripts/dump_prims.py > share/wake/lib/prim/prim.wake`

Once this goes it, it would be official policy that `prim "name"` not be allowed in wake source outside of the stdlibs `prim.wake`

### DIFF
--- a/scripts/dump_prims.py
+++ b/scripts/dump_prims.py
@@ -1,0 +1,63 @@
+# Copyright 2023 SiFive, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You should have received a copy of LICENSE.Apache2 along with
+# this software. If not, you may obtain a copy at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+
+def prim_to_expr(prim_name: str, param_count: int) -> str:
+    expr = '('
+    for _ in range(param_count):
+        expr += '\\_ '
+    expr += 'prim "'
+    expr += prim_name
+    expr += '")'
+    return expr
+
+def determine_expr_type(expr: str) -> str:
+    result = subprocess.run(['./bin/wake', '-q', '--print-expr-type', '-x', expr], stderr=subprocess.DEVNULL, stdout=subprocess.PIPE)
+    return result.stdout.decode('utf-8').splitlines()[0].strip()
+
+def determine_number_of_params(prim_name: str) -> int:
+    for i in range(100):
+        result = subprocess.run(['./bin/wake', '-x', prim_to_expr(prim_name, i)], stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+        if result.returncode == 0:
+            return i
+
+    print("Unable to determine number of params under 100 tries, does prim have >100 parameters?")
+    exit(1)
+
+def prim_to_def(name: str, expr: str, type: str) -> str:
+    def_str = "export def prim_" + name + ": " + type + " =" + "\n"
+    def_str += "    " + expr
+    return def_str
+
+
+def main():
+    result = subprocess.run(['./bin/wake', '--list-prims'], stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
+
+    print("package prim_wake")
+    print("")
+
+    for prim in result.stdout.decode('utf-8').splitlines():
+        prim = prim.strip()
+        if prim == "":
+            continue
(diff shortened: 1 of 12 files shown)


scripts/dump_prims.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2023 SiFive, Inc.
```
mmjconolly, line 1:
```suggestion
#!/usr/bin/env python3

# Copyright 2023 SiFive, Inc.
```

mmjconolly: I wondering if we actually need this script to be retained, as it's rare to add a new prim, and it could be manually added to prim.wake at that time

share/wake/lib/prim/prim.wake
```diff
@@ -0,0 +1,401 @@
+package prim_wake
+
+export def prim_abs: Integer => Integer =
```
mmjconolly, line 3: I hate to be that guy, but we should probably write comments for these

V-FEXrt: This file is generated from the script and isn't supposed to directly used. Instead they are supposed to be wrapped by other properly commented stdlib functions. 

I don't think we could generate the comments for these but I suppose we could consider the script a "run once" script then manually update it moving forward. 

Regardless I don't personally have the context for everything here and would need help if we decided that this should be commented

scripts/dump_prims.py
V-FEXrt: Yeah, fair. I certainly had that thought too

share/wake/lib/prim/prim.wake
```diff
@@ -0,0 +1,401 @@
...
+    (\_ prim "json_str")
+
+export def prim_lcat: List String => String =
+    (\_ prim "lcat")
+
+export def prim_lcm: Integer => Integer => Integer =
+    (\_ \_ prim "lcm")
+
+export def prim_level: Integer =
+    (prim "level")
```
JakeSiFive, line 247: No idea what this does

share/wake/lib/prim/prim.wake
```diff
@@ -0,0 +1,401 @@
...
+    (\_ \_ prim "root")
+
+export def prim_sNFC: String => String =
+    (\_ prim "sNFC")
+
+export def prim_sNFKC: String => String =
+    (\_ prim "sNFKC")
+
+export def prim_scaseNFKC: String => String =
+    (\_ prim "scaseNFKC")
```
JakeSiFive, line 313: No clue what these do

share/wake/lib/prim/prim.wake
```diff
@@ -0,0 +1,401 @@
...
+    (\_ prim "unlink")
+
+export def prim_unreachable: String => a =
+    (\_ prim "unreachable")
+
+export def prim_use: a => a =
+    (\_ prim "use")
+
+export def prim_vcat: String =
+    (prim "vcat")
```
JakeSiFive, line 379: Should probably drop variadic things

share/wake/lib/prim/prim.wake
```diff
@@ -0,0 +1,401 @@
...
+    (prim "version")
+
+# export def prim_vget: Array a => Integer => a =
+#     (\_ \_ prim "vget")
+
+# export def prim_vnew: Integer => Array a =
+#     (\_ prim "vnew")
+
+# export def prim_vset: Array a => Integer => a => Unit =
+#     (\_ \_ \_ prim "vset")
```
JakeSiFive, line 391: The problem with `Array a` is that its a primitive type so we can't define it in wake but it has a parameter so we don't define it like `Integer` either.

It might be sufficient to create an unconstructable array type so that `prim_vnew` is the only way to construct it but I'm not sure how that would go exactly. I'd rather just drop these than try to shove it in.

JakeSiFive (commented):
Ideally the prims would be grouped by their use case (e.g. prim/double.wake, prim/string.wake, etc...)

No super easy way to do that I suppose

share/wake/lib/prim/prim.wake
mmjconolly: unicode normalization

share/wake/lib/prim/prim.wake
mmjconolly: > supposed to be wrapped by other properly commented stdlib functions.

I'm not sure I understand the use-case then, I had assumed this was to replace ad-hoc usage of `prim` in regular user code, as there are some times the need to use prim for cases that the stdlib doesnt handle

V-FEXrt:
Hmm  based on comments and DMs sounds like folks aren't super happy with this change unless we went in and hand wrote `prim.wake`. That's not something I have time for, was just trying to knock something out quickly as I thought it would be slightly better than the scatterings of `prim "blah"` across the source files.

Going to go ahead and close this for now but we can use the generated `prim.wake` as a guide if we ever decide to hand write it.

PR #1435, opened 2023-10-10
Add timeout to wakebox commands

### LINKED ISSUE #1177: Wakebox --timeout option
Can we add a`--timeout` option  to `wakebox` to terminate any child processes, except fuse-waked, after x time,. This would help ensure fuse-waked is able to report out the state of the wakebox, including its filesystem, upon termination.

I have a scenario in which wakebox is running in a cgroup and is sent a `SIGTERM` after a given timeout. Wakebox exits and does forward `SIGTERM` to the children, but before the children are able to handle the `SIGTERM`, the children are `SIGTERM` by the Parent process. This means fuse-waked is not able to report back any results. 

Perhaps another option `--kill-after` would be useful in that `--timeout` refers to when a `SIGTERM` is sent and `--kill-after` refers to how long after `--timeout` a `SIGKILL` should be sent.

I see some parallels in the following PR in ensuring fuse-waked is kept alive. 
https://github.com/sifive/wake/pull/1127

V-FEXrt:
This PR adds the key `command_timeout` to spec.json. When the key is set to an integer greater than 0, wakebox will kill the running command after that amount of seconds (plus some tiny overhead) has elapsed if the command hasn't completed.

Examples

```
$ cat spec.json          
{
  "command": [ "/bin/sh", "-c", "sleep 30 && python3 --version" ],
  "command_timeout": 3
}
$ wakebox -p spec.json
$ echo $?
124
```

```
$ cat spec.json          
{
  "command": [ "/bin/sh", "-c", "sleep 30 && python3 --version" ],
  "command_timeout": 45
}
$ wakebox -p spec.json
Python 3.10.12
$ echo $?
0
```

Fixes #1177


V-FEXrt:
I  still need to add the cli  option  but wanted y'all  to take a look before I got too far in the implementation

src/wakefs/fuse.cpp
```diff
@@ -217,9 +227,35 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
   (void)close(STDOUT_FILENO);
   (void)close(STDERR_FILENO);
 
-  do waitpid(pid, &status, 0);
-  while (WIFSTOPPED(status));
```
V-FEXrt, line 221: I'm not sure what the `WIFSTOPPED` is supposed to be doing here, so deleting it may be problematic. @JakeSiFive @mmjconolly

src/wakefs/fuse.cpp
```diff
@@ -217,9 +227,35 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+  int pidfd = syscall(SYS_pidfd_open, pid, 0);
```
mmjconolly, line 230: `SYS_pidfd_open` was added in linux kernel 5.3 which is a bit more recent than some of the systems we try to support. I suspect that impacted trying to find a libc wrapper too

src/wakefs/fuse.cpp
mmjconolly: I think its saying that if the child was signaled, just keep waiting on it anyway

src/wakefs/fuse.cpp
V-FEXrt: I was poking around and this seemed like the only way to do it with epoll. @JakeSiFive any alternative ideas? If not we might have to just use sleep

mmjconolly: What does /usr/bin/timeout do? We could even just add it to the `execve_wrapper` to prepend it?

JakeSiFive: I think it probably uses something like timer create but I'll go read the source code tomorrow. https://man7.org/linux/man-pages/man2/timer_create.2.html

JakeSiFive: I guess you could do this:

1) Create your payload process
2) Create a timer process that sleeps and then exits
3) wait on both of them, if the timer process exits first then you kill the payload process

src/wakefs/fuse.cpp
JakeSiFive: I suspect there's actully a bug here where if a child is stopped via SIGSTOP, and then continued wakebox will treat the process as having been terminated when it hasn't. I think the correct condition is `!WIFEXITED(status)`

src/wakefs/fuse.cpp
```diff
@@ -227,35 +222,41 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
-  if (args.command_timeout <= 0) {
-    events = epoll.wait(nullptr, nullptr);
-  } else {
-    struct timespec timeout;
-    timeout.tv_sec = args.command_timeout;
-    timeout.tv_nsec = 0;
-    events = epoll.wait(&timeout, nullptr);
+    if (timeout_pid == 0) {
+      std::string sleep = "sleep " + std::to_string(args.command_timeout);
+      execve_wrapper({"/bin/sh", "-c", sleep}, {});
```
JakeSiFive, line 238: just use the sleep function and then exit

src/wakefs/fuse.cpp
```diff
@@ -227,35 +222,41 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
-    kill(pid, SIGKILL);
-    std::cerr << "wakebox: Timed out waiting for process to complete" << std::endl;
-    exit(124);
-  }
+  pid_t wait_pid;
+  while ((wait_pid = wait(&status)) != -1) {
+    if (wait_pid == timeout_pid && WIFEXITED(status)) {
+      kill(payload_pid, SIGKILL);
+      std::cerr << "wakebox: Timed out waiting for process to complete" << std::endl;
+      exit(124);
```
JakeSiFive, line 247: I think we still need to output the results so it doesn't suffice to exit here. Data still has to be collected or Nathan's use case is not covered.

src/wakefs/fuse.cpp
```diff
@@ -227,35 +222,41 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+    }
 
-  epoll.remove(pidfd);
-  close(pidfd);
+    if (wait_pid == payload_pid && !WIFSTOPPED(status)) {
+      if (has_timeout) {
+        kill(timeout_pid, SIGKILL);
+      }
+      // Note: must not wait on timeout pid in this codepath
+      // otherwise it'll pollute rusage.
```
JakeSiFive, line 255: Good comment, we should clarify in RUsage call below what "waiting on a child" means here. At first we both assumed that it meant getRUsageChildren would return the usage of *every* child regardless of what had been collected by the wait call.

mmjconolly:
If we already need to fork to do the timeout, it's probably not too much more to exec the `timeout` command prefixing the real command before we run `execve_wrapper`

```diff
diff --git a/src/wakefs/fuse.cpp b/src/wakefs/fuse.cpp
index cb1a4d1d..ef7423b3 100644
--- a/src/wakefs/fuse.cpp
+++ b/src/wakefs/fuse.cpp
@@ -198,6 +198,10 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
       }
     }
 
+    if (args.timeout > 0) {
+        command.push_front("/usr/bin/timeout", std::string(args.timeout));
+    }
+
 #ifdef __linux__
     if (args.isolate_pids) {
       pidns_args nsargs = {command, args.environment};

```
(I haven't tested this)

V-FEXrt:
That would probably work, though one thing that's unclear is how to tell apart "job failed with error 124" and "job timed out"

V-FEXrt:
Okay, I can confirm that the `/usr/bin/timout` wrapper also works and  is the simplest to implement. It does mean that if a wrapped job returns status code `124` the slurm runner will assume the job timed out

mmjconolly:
Could use `alarm()` too

JakeSiFive (approved):
This looks good to me.

src/wakefs/fuse.cpp
```diff
@@ -217,16 +224,56 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+  // Launch timer process
+  if (has_timeout) {
+    timeout_pid = fork();
+    if (timeout_pid < 0) {
+      std::cerr << "wakebox: failed to fork timout process" << std::endl;
+      exit(1);
+    }
+
+    if (timeout_pid == 0) {
+      execve_wrapper({"/usr/bin/sleep", std::to_string(args.command_timeout)}, {});
```
JakeSiFive, line 239: nit: sleep and then exit would be slightly better

src/wakefs/fuse.cpp
```diff
@@ -217,16 +224,56 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+    }
+  }
+
+  pid_t wait_pid;
+  while ((wait_pid = wait(&status)) != -1) {
+    if (wait_pid == timeout_pid && WIFEXITED(status)) {
+      kill(payload_pid, SIGKILL);
+      std::cerr << "wakebox: Timed out waiting for process to complete" << std::endl;
```
mmjconolly, line 247: I think we should avoid printing, as this will end up in `wake --failed` as text, but it's already being exposed as data in the result json which the `Runner` type in wake can handle

V-FEXrt:
I'm going to skip the CLI flag for now, will add it later if it turns out too difficult to not have

src/wakefs/fuse.cpp
```diff
@@ -217,16 +224,56 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+  while ((wait_pid = wait(&status)) != -1) {
+    if (wait_pid == timeout_pid && WIFEXITED(status)) {
+      kill(payload_pid, SIGKILL);
+
+      struct timeval stop;
+      gettimeofday(&stop, 0);
+      std::string output;
+      args.daemon.disconnect(output);
+      RUsage usage = {};
+      return collect_result_metadata(output, start, stop, payload_pid, 124, usage, true,
```
mmjconolly, line 254: It doesn't really matter what the specific number is, as anything that care about knowing the real reason can check the result json

mmjconolly: Actually, I think we should return zero when isolate-retcode is set, as wakebox itself didn't fail, just the payload

JakeSiFive: I think pretending its 124 is somewhat helpful for debugging potentially. isolate-retcode is just meant to make wakebox return 0 as long as it doesn't fail right? I think this number is the return code of the payload isn't it?

mmjconolly: oh my bad, this is the path that threads through into the result json, rather than the exterior

mmjconolly:
It could also be nice to add a unit test for this in https://github.com/sifive/wake/tree/master/tests/wakebox

V-FEXrt:
Done!

src/wakefs/fuse.h
```diff
@@ -53,6 +53,7 @@ struct daemon_client {
 
 struct json_args {
   std::vector<std::string> command;
+  int command_timeout = -1;  // timeout in seconds.
```
JakeSiFive, line 56: nit: should be wcl::optional

JakeSiFive (approved):
I'm happy to land, I did have a nit on wcl::optional if you wanted to fix that before landing

V-FEXrt:
PTAL, but should be ready to merge  now

src/wakefs/fuse.cpp
```diff
@@ -217,16 +233,56 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
   (void)close(STDOUT_FILENO);
   (void)close(STDERR_FILENO);
 
-  do waitpid(pid, &status, 0);
-  while (WIFSTOPPED(status));
+  bool has_timeout = (bool)args.command_timeout;
```
mmjconolly, line 236: Does this cast somehow figure out if the option is a some/none ?

src/wakefs/fuse.cpp
```diff
@@ -217,16 +233,56 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+  // Launch timer process
+  if (has_timeout) {
+    timeout_pid = fork();
+    if (timeout_pid < 0) {
+      std::cerr << "wakebox: failed to fork timeout process" << std::endl;
+      exit(1);
+    }
+
+    if (timeout_pid == 0) {
+      sleep(*args.command_timeout);
```
mmjconolly, line 248:
```suggestion
      prctl(PR_SET_NAME, "wb-timer", 0, 0, 0);
      sleep(*args.command_timeout);
```

tests/wakebox/timeout-exceeded/input.json
```diff
@@ -0,0 +1,17 @@
...
+    "HOME=/root",
+    "PATH=/usr/bin:/bin:/usr/sbin:/sbin"
+  ],
+  "directory": ".",
+  "stdin": "",
+  "user-id": 0,
+  "group-id": 0,
+  "mount-ops": [],
+  "visible": [],
+  "command_timeout":  1
```
mmjconolly, line 16: We've been using `-` rather than `_` for these fields

src/wakefs/fuse.cpp
```diff
@@ -121,6 +136,7 @@ static bool collect_result_metadata(const std::string daemon_output, const struc
 
   result_jast.add("inputs", JSON_ARRAY).children = std::move(from_daemon.get("inputs").children);
   result_jast.add("outputs", JSON_ARRAY).children = std::move(from_daemon.get("outputs").children);
+  result_jast.add_bool("timed_out", timed_out);
```
mmjconolly, line 139: similar here for -/_

src/wakefs/fuse.cpp
mmjconolly: Example of inspecting using the thread names

Running this spec:
```
{
  "command": [
    "/usr/bin/sleep", "5"
  ],
  "command_timeout": 50,
  "isolate-pids": true
}
```

In one terminal:
```
$ echo $$
605615
$ unshare -mrp --mount-proc -f ./bin/wakebox -p spec.json
```
Then while that runs, in a second:
```
$ pstree -p 605615
bash(605615)───unshare(1903513)───wakebox(1903514)─┬─fuse-waked(1903517)
                                                   ├─wb-mount-ns(1903521)───wb-pid-ns(1903523)───sleep(1903524)
                                                   └─wb-timer(1903522)
```

src/wakefs/fuse.cpp
V-FEXrt: yeah exactly. We implement the "bool operator" here https://github.com/sifive/wake/blob/master/src/wcl/optional.h#L190

src/wakefs/fuse.cpp
```diff
@@ -61,6 +61,20 @@ bool json_as_struct(const std::string &json, json_args &result) {
...
+  JAST timeout_entry = jast.get("command_timeout");
+  if (timeout_entry.kind == JSON_INTEGER) {
+    int timeout = std::stoi(timeout_entry.value);
+    if (timeout <= 0) {
+      std::cerr << "timeout must be be an integer value greater than 0" << std::endl;
+      return false;
+    }
+
+    result.command_timeout = wcl::make_some<int>(timeout);
+  } else if (timeout_entry.kind != JSON_NULLVAL) {
```
mmjconolly, line 73: We probably don't care too much, but this does mean that `"isolate-pids": null` passes validation

src/wakefs/fuse.cpp
V-FEXrt: That said, this var isn't as helpful/as needed now just I'll just delete it

src/wakefs/fuse.cpp
V-FEXrt: Yeah, the `get`+`kind` API doesn't distinguish between "key not set" and "key set to null"

tests/wakebox/timeout-exceeded/input.json
```diff
@@ -0,0 +1,17 @@
+{
+  "command": [
+    "/bin/sh", "-c", "sleep 2 && true"
+  ],
```
mmjconolly, line 4: One of the motivations for the timeout is that any already-written files are tracked by the fuse workspace even when the timeout otherwise halts progress.

We could check this like

```suggestion
  "command": [
    "/bin/sh", "-c", "touch hello; sleep 2 && true"
  ],
  "mount-ops": [
      {
        "type": "workspace",
        "destination": "."
      }
   ],

```
and then seeing if `hello` is in the outputs file

V-FEXrt: > and then seeing if `hello` is in the outputs file

Before I resolve this, do you be if its in the outputs _key_ in the JSON? Because it doesn't seem to be. But the file does exist/get created

tests/wakebox/timeout-exceeded/pass.sh
```diff
@@ -0,0 +1,13 @@
...
+#
+# It's not valid to call wakebox with an empty PATH.
+# So we fill PATH with some typical values.
+export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
+
+${1}/wakebox -p input.json -o result.json
+
+trap "rm result.json hello" EXIT
+
+[ "$(cat result.json | jq '."timed-out"')" = "true" ] && [ -f "hello" ] && exit 0
```
mmjconolly, line 11: This checks that fuse-waked wrote the file to disk, which it would have anyway
We care that the json itself is populated so that if any other tools read it they know about the file that was written

something like:
```suggestion
[ "$(cat result.json | jq '."timed-out"')" = "true" ] && exit 0
[ "$(cat result.json | jq -r .outputs[])" = "hello" ] && exit 0
```

tests/wakebox/timeout-exceeded/input.json
mmjconolly: oh I've just commented on the other file, yeah the thing we're looking for is that the json key for the outputs is populated

mmjconolly: When I use this spec:

```
$ cat spec.json
{
  "command": [
    "/usr/bin/bash", "-c", "echo hello > yeah; sleep 10"
  ],
  "command_timeout": 9,
  "isolate-pids": true,
  "mount-ops": [
      {
        "type": "workspace",
        "destination": "."
      }
   ]
}
```
I see a file written:
```
$ ./bin/wakebox -p spec.json -o result.json
$ cat result.json | jq -r .outputs[]
yeah
```

V-FEXrt: got it. Alright, I'll have to poke around more. Probably won't have another commit for today as the day is wrapping up and someone is coming to work on the basement.

V-FEXrt: @JakeSiFive lmk if you have any ideas on what might be going wrong here

tests/wakebox/timeout-exceeded/input.json
```diff
@@ -0,0 +1,23 @@
...
+  "environment": [
+    "USER=root",
+    "HOME=/root",
+    "PATH=/usr/bin:/bin:/usr/sbin:/sbin"
+  ],
+  "directory": ".",
+  "stdin": "",
+  "user-id": 0,
+  "group-id": 0,
+  "mount-ops": [],
```
mmjconolly, line 14:
```suggestion
```
We had a second `mount-ops` here

src/wakefs/fuse.cpp
```diff
@@ -217,16 +233,56 @@ bool run_in_fuse(fuse_args &args, int &status, std::string &result_json) {
...
+  pid_t wait_pid;
+  while ((wait_pid = wait(&status)) != -1) {
+    if (wait_pid == timeout_pid && WIFEXITED(status)) {
+      kill(payload_pid, SIGKILL);
+
+      struct timeval stop;
+      gettimeofday(&stop, 0);
+      std::string output;
+      args.daemon.disconnect(output);
+      RUsage usage = {};
```
mmjconolly, line 262: It's probably nice to report the real usage back regardless of the timeout, rather than `{}`, it also covers non-time usage like file io bytes written

mmjconolly (approved):

src/wakefs/fuse.cpp
V-FEXrt: There isn't actually a good way to get the real info out. RUsage reports back the usage for all terminated child processes that have been `wait()`ed on. We could wait on the payload process after killing it but at that point the timeout process would have terminated+waited so the usage would be higher than expected. 

My thought was that it was better to give no info than to give incorrect info. But I don't feel strongly either way

JakeSiFive: I suppose seeing the usage on a timeout could be helpful for debugging. Like "oh yeah, that was writing massive amounts out to the disk and it was using 10x more memory than we wanted it to"

V-FEXrt: Unless there is significant push back/disagreement, I'd like to just land this PR since its been dragging on this we can make a separate one to explore usage

JakeSiFive:
I'm happy to land

src/wakefs/fuse.cpp
JakeSiFive: I'm happy to land. I don't think the usage is required.

mmjconolly: We should at least put in `runtime` which we can just copy straight in from the timeout value

V-FEXrt: We still insert runtime. That is derived from `gettimeofday(&stop, 0);` not `usage`

V-FEXrt:
```
$ cat result.json | jq                
{
  "usage": {
    "status": 124,
    "membytes": 0,
    "inbytes": 0,
    "outbytes": 0,
    "runtime": 1.001355,
    "cputime": 0
  },
  "inputs": [],
  "outputs": [],
  "timed-out": true,
  "run-host": "ashleycoleman-1518"
}
```

mmjconolly: ah ok

PR #1437, opened 2023-10-12
Adding create time for jobs in job cache

JakeSiFive:
This change adds a create time for jobs in the job cache which will invalidate the schema. This will be needed both to implement the TTL eviction policy and to implement the needed fixes to the LRU eviction policy so it seems an easy thing to jsut go ahead and add.

### DIFF
--- a/src/job_cache/daemon_cache.cpp
+++ b/src/job_cache/daemon_cache.cpp
@@ -385,7 +385,8 @@ class JobTable {
 
  public:
   static constexpr const char *insert_query =
-      "insert into jobs (directory, commandline, environment, stdin, bloom_filter, runner_hash, create_time)"
+      "insert into jobs (directory, commandline, environment, stdin, bloom_filter, runner_hash, "
+      "create_time)"
       "values (?, ?, ?, ?, ?, ?, ?)";
 
   static constexpr const char *add_output_info_query =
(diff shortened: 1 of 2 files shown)


src/job_cache/daemon_cache.cpp
```diff
@@ -1001,6 +1004,11 @@ void DaemonCache::add(const AddJobRequest &add_request) {
   std::string tmp_job_dir = "tmp_" + rng.unique_name();
   mkdir_no_fail(tmp_job_dir.c_str());
 
+  // Get the time so we know when this job was created
+  timespec tp;
+  clock_gettime(CLOCK_REALTIME, &tp);
+  int64_t time = 1000000ll * int64_t(tp.tv_sec) + tp.tv_nsec / 1000;
```
V-FEXrt, line 1010: nit is there not some magic "get unix epoc" function defined somewhere?

V-FEXrt: Its also worth considering SQLite3's datatime support for this https://www.sqlite.org/datatype3.html

V-FEXrt (commented):
We need to update other data types for this right? Like the wake runtime manually creates a json object representing this type, or am I misremembering?

src/job_cache/daemon_cache.cpp
JakeSiFive: This is the easiest way I'm aware of to get microseconds

mmjconolly: is per-second resolution good enough?

V-FEXrt (approved):

src/job_cache/daemon_cache.cpp
JakeSiFive: No its not. It makes testing for LRU stuff super hard and makes fuzz testing for eviction annoying as well because you have to sleep a whole second per in order to get the evictions to happen in the right order. But at the microsecond level everything just works because it takes around 100x to 1000x that long to do  a cache read.

PR #1440, opened 2023-10-13
fix makePlan examples in the tutorial and description

sifive-benjamin-morse:
# Description
This fixes `makePlan` usage and description in the tutorial.

# Notes
As a side note, running this on a popular system I get the following error message:

```
Fail (Error "No GitHub releases found for sifive/wake" Nil)
```

Which is a result of the fetched `releases.json` having these contents:

```
{"message":"API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

### DIFF
--- a/share/doc/wake/tutorial.md
+++ b/share/doc/wake/tutorial.md
@@ -551,7 +551,7 @@ from plan_scorer import _
 
 export def infoH _args =
     def cmdline =
-        which "uname", "-sr", Nil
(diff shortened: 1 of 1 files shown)


JakeSiFive (approved):
Thanks for this!

PR #1441, opened 2023-10-13
Support globbing for all param based inspection

V-FEXrt:
This is the second PR on the path to  tag inspection of the database. After this PR any inspection flag that accepts a parameter can use the glob operators `*` and `?` as expected.

Example:
`wake --input "*.c" --output "*.o" --job "???"`
which says match any job that takes a `c` file as input,  outputs an `o` file, and has a 3 digit job id


JakeSiFive (approved):
awesome idea!!!

PR #1445, opened 2023-10-18
Fix squashfuse issues

ErikParawell-SiFive:
Fix issues caught by @V-FEXrt after #1439 was merged

Also built a test case for wakebox squashfuse. In doing so I discovered that I was not invoking squashfuse correctly so that is corrected now.
As a side note the test case sleeps for 3 seconds before it exits because it is possible that fuse is still unmounting when the `rm -rf` cleanup happens. Maybe `wakebox` needs a `-o notify_pipe` 😆 

List of changes:

- Added a test case to test squashfsfuse in wakebox.
- Added `bash` to Alpine Dockerfile because I think that is a small requirement.
- Updated the package lists in the README.md to include squashfuse related ones.
- Addressed the post merge comments from #1439


JakeSiFive (commented):
We should add a test for this so that we don't hit this issue in the future.

A simple wakebox unit test that just runs cats a file out of a squash fuse mount would suffice

ErikParawell-SiFive:
I agree we should have a test case for this.

JakeSiFive:
Me neither but the fact that this bug slipped passed CI indicates to me that it must be untested

ErikParawell-SiFive:
Made a test case and update description:

> Also built a test case for wakebox squashfuse. In doing so I discovered that I was not invoking squashfuse correctly so that is corrected now.
As a side note the test case sleeps for 3 seconds before it exits because it is possible that fuse is still unmounting when the `rm -rf` cleanup happens. Maybe `wakebox` needs a `-o notify_pipe` 😆

JakeSiFive (approved):

V-FEXrt (approved):

JakeSiFive:
I'll take over this PR so it can land soon

JakeSiFive:
I'm going to wait to debug the rest of this. Issues like this are best debugged using a local podman/docker container but I've had serious issues getting docker to work correctly with namespacing since the kernel they use on the VM to run the containers on doesn't have things configured correctly. There is a way to use your own kernel on Mac docker but I'm lazy and would rather wait until I'm back home with my linux dev laptop that I can run podman on.

PR #1448, opened 2023-10-18
Add unsafe type to wake core

JakeSiFive:
This PR introduces the Unsafe type. The type on its own will be useful for users to self opt-in to marking their functions as unsafe but going forward we can also discuss wrapping things like local runner, which, claimFileAsPath, claim, etc...

We can also instrument these functions to see if any of them are being used in CI and having wake fail when this happens if we wanted to enforce that.

Candidates in standard library for unsafe status:
files, which, claim, claimFileAsPath, localRunner, setPlanFnOutputs, editPlanFnOutputs, getJobId, wakePath, workspace

My current ranking of these in terms of priority:
1. files and localRunner
2. which, workspace, and wakePath
3. claimFileAsPath
4. claim
5. getJobId (no one uses this as far as I'm aware? Maybe we can just delete it instead)
6. setPlanFnOutputs and editPlanFnOutputs

The functions from 1 and 2 are all instances of things we've seen real issues arise from people using them. The claim functions are interesting but I am not personally aware of anyone using them in a way that caused an issue. It would be nice to check that no one is using claim or claimFileAsPath in CI so we might have separate checks for that. One could argue that claim and claimFileAsPath don't need to be marked unsafe. getJobId might just be removoable.

setPlanFnOutputs and editPlanFnOutputs are intresting cases. People certainly use them in ways that cause problems from time to time but it's not totally clear to me that they violate safety on their own. They would only violate safety if the functions *themselves* were unsafe I think.


JakeSiFive:
I think personally I'd like to do the following:

* make files, localRunner, which, workspace, wakePath all unsafe
* Add an env-var WAKE_PANIC_ON_CLAIM to the claim functions so that we can make wake fail if they're used in CI
* Delete getJobId
* go ahead and mark setPlanFnOutputs, and editPlanFnOutputs as unsafe so that we can simplify what we tell people and just say "don't use unsafe code without consulting an expert"

richardxia:
I think it'll be hard to know exactly what the impact of modifying these API to return/accept `Unsafe`, since they may taint a large number of things, and we definitely want to avoid users getting in the habit of just calling `unsafe_avoid_using_run` every time they get an `Unsafe`. It might be nice to do an initial test of taking one of the APIs that you listed, changing its signature to have `Unsafe`, and seeing how badly it impacts SiFive's internal codebase, just to get a sense of what it'll be like.

Also, regarding the naming of the methods here, you've clearly taken a lot of inspiration from Haskell, but I think in general, Wake's naming conventions for functor/monadic functions don't really follow Haskell's, and I think we generally try to come up with synonyms that are more intuitive to users with less Haskell-like FP experience. `unsafeMap` I think is probably fine, but I'm almost certain that all of the rest will be very foreign to most people. We may not need to remove the functions you have, though, as long as the commonly-used ones have more accessible names. My intuition is that `unsafeMap`, `unsafeBind`, and `unsafeLiftA2` are probably that core set. Maybe rename `unsafeBind` to `unsafeFlatMap`, and `unsafeLiftA2` to `unsafeMap2`?

Lastly, here are a few more stdlib candidates for marking as unsafe: `cwd`, the entire first half of `environment.wake`, most of the Job metrics, like CPU and Memory usage. I agree with most of your ranking, and I think I would suggest `cwd` and `getenv` at the same level as `which`, `workspace`, and `wakePath`. Maybe `prim` itself should always return an `Unsafe`? Users shouldn't be calling `prim` directly anyway.

mmjconolly:
I'd be interesting in seeing a usage example

JakeSiFive:
Here's a usage example in a non-trivial case

```
# This API makes it possible to include a non-source file into a wake build.
# Generally, one should obtain Paths from sources or as the output of Jobs.
# This API is useful for things like a command-line-supplied input file.
export def claimFileAsPath (existingFile: String) (desiredWorkspacePath: String): Unsafe (Result Path Error) =
    def fileName = replace `.*/` "" desiredWorkspacePath

    extract `(.*)/.*` desiredWorkspacePath
    | head
    | getOrElse "."
    | mkdir
    | makeUnsafe
    | unsafeRMap (claimFileAsPathIn _ existingFile fileName)
```

JakeSiFive:
cwd for sure should be considered unsafe, I just missed it.

getenv is unclear. You could argue that the environment is an intended input but at the same time we know that people's environments are going to differ quite a lot so its probably best to mark them unsafe.

V-FEXrt:
echoing @richardxia, the function names here seem out-of-alignment with other wake stdlib names. The big one that stood out for me was `unsafeMap` should be `umap` which is congruent to `omap` and `rmap`

ag-eitilt:
> We can also instrument these functions to see if any of them are being used in CI and having wake fail when this happens if we wanted to enforce that.

I'm not sure that's something we can enforce, at least so long as we're marking `claim` and `claimFileAsPath` as unsafe.  The latter is *designed* to be used for "adopting" command-line arguments, and unless we enforce that all projects using Wake/opting into the new strict mode write their *entire* CI purely in Wake to wrap files generated from other sources (which I agree is ideal, but I don't think is something which can be made policy on the open-source side) then there's a fair bit of CI which stops being possible.  There's also the case of mixed entry points, where for example I only `claim` data if certain CLI arguments are passed, and while I'd be happy enough saying those arguments aren't allowed in CI the type system is never going to be robust enough to say "this is only unsafe in these specific conditions".  Yeah, both of those cases can explicitly unwrap the monad, but at that point what purpose did it serve in the first place?

I think my biggest concern is that this never says what, exactly, makes something unsafe.  You clearly have an idea in mind, but I look at the list and see `claimFileAsPath` as a completely separate set of concerns than `localRunner`.  I can't reason about the impact and future direction if I don't know *why* this function is marked Unsafe and that isn't.

A secondary concern is that Haskell makes this work very smoothly through `do` notation and judicious use of operators.  Wake obviously doesn't have the syntactic sugar, and (in keeping with the language conventions) you've only defined prefixing functions.  I'm not sure this is actually going to be reasonable to write any significant code in, which effectively restricts anything marked unsafe to the uppermost levels of code -- which isn't reasonable for something like `localRunner` -- or goes straight back to unwrapping the monad as soon as possible and not letting it serve a purpose.  That might be a baseless fear once we actually see code adapted to this interface, but I'm not overly comfortable until I see it in use.

V-FEXrt:
> but at that point what purpose did it serve in the first place?

By marking code as unsafe, we highlight it in code review. If a reviewer sees someone unwrapping an `Unsafe` it is a cue that they need to review the code extra closely/extra diligently 

> I think my biggest concern is that this never says what, exactly, makes something unsafe. 

Something is unsafe if it has the possibility of creating an unreproducible build when used incorrectly

ag-eitilt:
> > but at that point what purpose did it serve in the first place?
> 
> By marking code as unsafe, we highlight it in code review. If a reviewer sees someone unwrapping an `Unsafe` it is a cue that they need to review the code extra closely/extra diligently

That'll do, then.  I think we can make the unwrapping function slightly (only slightly) less ugly if we're relying on code review and not on mechanical means, and I'm still very concerned about ergonomics, but the biggest potential issues stem from automation so *not* automating things feels better.

> > I think my biggest concern is that this never says what, exactly, makes something unsafe.
> 
> Something is unsafe if it has the possibility of creating an unreproducible build when used incorrectly

And then we need to define what "unreproducible" means.  We can't be purely looking at hashes, or `claim`/`claimFileAsPath` would be just as fine as passing a non-path `String` on the command line -- or `source` for that matter which doesn't care if a file's been edited since the git checkout.  Why is `getJobId` more of an issue than `ReRun`, when most uses of the latter are for *explicitly* unreproducible Jobs?  I definitely agree with the theory and the impetus behind this, but it's something we're going to have to be very thorough in documenting.

richardxia:
> And then we need to define what "unreproducible" means.

We'll never be able to prevent/detect all possible sources of unreproducibility (nothing stops a user from writing nondeterministic code in another language that a Wake job calls), but I think there certain classes of things that we know that make a job unreproducible that we can flag as "unsafe", especially in wake-land. If we wanted to more formally define what we mean by reproducible and which specific invariants we want "safe" code to have, then I would suggest reading up on some existing articles and tools online that also attempt to define these things:

https://reproducible-builds.org/
https://bazel.build/basics/hermeticity

ag-eitilt:
Agreed, yeah.  My comment is more that we're talking about drawing a line here between `Unsafe a` and `a`, and we need to be very certain of what line we're drawing and why it's drawn where we draw it.

JakeSiFive:
Here's how I think about safety in a very very formal sense:

1) In a pure functional language without effects the following laws hold

Dead Code Elimination: `let x = A; B` == `B` if x is not free in B (for strict languages you must also stipulate that A always terminates)
Leibniz Rule: if `A` == `B` then `C(A) == C(B)` (C is a second order schematic expression so A/B can appear multiple times)
CSE: `A(B)` == `let x = B; A(x)` (not that A is a second order schematic variable so B may appear multiple times in A(B))
(Note that CSE implies a kind idempotentcy:)
Idempotent: `let x = A; let y = A; C` = `let x = A; y = x; C`
Commutative: `let x = A; let y = B; C` = `let y = B; let x = A; C` (assuming that x is not free in B)

Wake strives to make all of these true *except* dead code elimination because wake has effects. However its effects are all idempotent (unless they're not reproducible in which case...they are not idempotent). The only snafu here is that wake has a weaker notion of commutativity that you can phrase like this:

Weakly Commutative: `let x = A; let y = B; C` = `let y = B; let x = A; C` assuming that `let x = A; let y = B; C` does not evaluate to bottom

For instance in wake if you first create foo.txt by echoing out "bar" and then a second job creates it by echoing out "baz" then those files will conflict and wake will panic thus the program evaluates to bottom. So wake is only *weakly* communicative.

What is not captured about this case however are things like claim, claimFileAsPath, and getenv. The question we have to ask is "what is the input to a wake program?" Is it only the sources that wake has access to? (via git or via a manifest.wake like file) Is it the environment plus those sources? Is it the sources AND the non-sources in a workspace? Is it the sources AND all files accessible on the system (claimFileAsPath for instance allows this)?

This is not well defined at this time but I'd argue that anything that isn't a source file might be deemed unsafe because a change in one of those non-source details would cause the same program to behave differently. They all act like global variables which is problematic. Technically the sources that wake operates on ALSO have this problem because the same wake program run on a different workspace will behave differently (this intended however). So the specifics of cases that do not otherwise violate these issues are up for grabs.

So what is unsafe? Anything that violates a wake safety law (Leibniz, CSE, Idempotency, Weak Commutativity) OR that is not a blessed input to wake.

files: returns different results for the same invocation at different points in the program
which: same thing as files
claim: claim is up for grabs. Assuming no outside forces are modifying files in your workspace then claim violates no safety laws
claimFileAsPath: Also up for grabs but with a weaker claim since we have to assume that nothing modifies this file while a build accessing this file is using it
localRunner: This lets you do literally anything and can implement all of the unsafe functions
setPlanFnOutputs: This violates none of the safety laws or input laws but I want to catch it in review
editPlanFnOutputs: ditto
getJobId: as you run wake many times with the same database this will change for the same job
wakePath: this will change depending on what version of wake you use or where your wake binary came from
workspace: this will be different even if both workspaces have the same checkout of a repo
cwd: this will vary based on where wake is run from
getenv: different users will have different environments but we might consider some special WAKE prefixed variables to be inputs to the build and thus its ok that this changes
sysname: We now only support Linux so maybe this is safe but we have yet to enforce that
machine: We do support Arm and x86 and I'm sure we could support Risc-v without too much effort. I don't think anyone is going to use this so I think the point is moot.

richardxia:
> So what is unsafe? Anything that violates a wake safety law (Leibniz, CSE, Idempotency, Weak Commutativity) OR that is not a blessed input to wake.

I'm struggling a little bit to figure out exactly which part of your text to quote, but I think one important thing that you haven't mentioned explicitly is _determinism_ and _reproducibility_. I believe it's true that the Wake language itself is deterministic — ignoring `prim` calls — and so the program transformations (safety laws) you've cited that rely on referential transparency hold. However, wake is a DSL and build tool, and it's meant to shell out to operating system to run arbitrary programs, where some of those programs may _not_ be deterministic, i.e. that two calls to the same program with the exact same inputs (file hashes included) may not return the same result. This is especially the case if you include things like timing in the definition of "result".

I would even argue that under your definition of safety, not even the fully sandboxed `runJob` is safe, since the program you execute may not produce the same result even if all the inputs are the same (a couple of trivial ways to break this include making a syscall to get the current time, or even just running a multithreaded program whose output is sensitive to thread scheduling).

I think we may need to back up a bit and define what the safety _goals_ of wake are. Defining safety in terms of equivalence of code after certain transformations is I think missing part of the point; although those are useful for reasoning about the semantics of a program and for the wake compiler to know which program optimizations will preserve semantics, it's completely missing the properties of wake that hold across different _run_ of wake with slightly different inputs.

I'm not going to attempt to formally define this, but I think the following is an important statement about multiple runs to wake: Between two different runs of wake, minimize the amount of incremental work done. If I have jobs A and B, and I only change the inputs to B, then wake should only rerun B.

This gets even more interesting with shared caching, since we'd like to allow for reuse of jobs between different workspaces, and that means that we need to eliminate or at least minimize the impact of the differences of the workspaces on the inputs and outputs of jobs.

I think unlike a language like Rust, we're don't really want to define safety in terms of whether you will get a "correct" result or not. Even under your examples of "unsafe" operations, wake will produce a "correct" result, but I think what we all don't like about them is that they are either nondeterministic/unreproducible or that they unnecessarily prevent incremental builds/shared caching.

Note that the underlying shared caching system must not rely on users correctly making use of `Unsafe`. I think `Unsafe` is really just a hint that misuse of the value could unnecessarily prevent cache hits.

JakeSiFive:
> I'm struggling a little bit to figure out exactly which part of your text to quote, but I think one important thing that you haven't mentioned explicitly is determinism and reproducibility.

Semantically determinism is assumed here and non-determinism is modeled as a hidden input which falls under the "what counts as an input" section. Viewed another way, we want to ask what global inputs we're willing to allow programs to depend on and which we're not. Inputs that we would call "non-deterministic" are not inputs that we should avoid depending on in wake. A random number generator from the operating system is the easiest non-determnistic input to model. The input is an infinite sequence of numbers in that case. Note that I mean a *specific* sequence of numbers such as 1, 2, 3, ... etc... we would like our wake programs to be invariant over such inputs. Time is modeled as an increasing sequence of numbers etc...

So safety is both about these laws that some violate, as well as determining what "inputs" are considered safe. What inputs we consider safe is something I've found to be a thing we have to be quite lenent on. For instance while we'd like to not depend on most environment variables, we WOULD like to depend on some of them. There is no principal other than "I think WAKE_LOCAL_JOB_CACHE is unlikely to be anything unintended" that dictates that I consider that a "safe" environment variable. The same goes for sandboxed jobs.

On the other hand we have functions like `files` for which the observable effects are defined for any specific order of evaluation but for which the behavior changes if you re-order the program. This is not correctly modeled as an input.

richardxia:
> Semantically determinism is assumed here and non-determinism is modeled as a hidden input which falls under the "what counts as an input" section. Viewed another way, we want to ask what global inputs we're willing to allow programs to depend on and which we're not. Inputs that we would call "non-deterministic" are not inputs that we should avoid depending on in wake. A random number generator from the operating system is the easiest non-determnistic input to model. The input is an infinite sequence of numbers in that case. Note that I mean a _specific_ sequence of numbers such as 1, 2, 3, ... etc... we would like our wake programs to be invariant over such inputs. Time is modeled as an increasing sequence of numbers etc...

I understand the idea of modeling nondeterminism via hidden inputs, but for the purposes of this discussion, I think we need to focus on inputs that are _observable_ by the wake language. I agree that the infinite sequence of numbers is a way of modeling RNGs, but that is not the way it is modeled by wake today, and instead it is this shared, global, mutable state owned by the OS that any wake Job can tap into. More pragmatically, the RNG is not reified in wake, and therefore you cannot wrap in an `Unsafe` because there is quite literally nothing to wrap today. You could argue that the _outputs_ of a function that returns random values should be marked `Unsafe`, but outputs are not the same thing as inputs, and I think we need to be more precise about this so that we have a common and consistent definition of `Unsafe`.

> So safety is both about these laws that some violate, as well as determining what "inputs" are considered safe. What inputs we consider safe is something I've found to be a thing we have to be quite lenent on.

To continue with my line of questioning above, when you say "input" here, are you just referring to values that are reified in wake, or are you also including values that are not represented by wake at all? With your environment variable example, I would consider that to be reified in wake, since the set of environment variables and their values is immutable once the `wake` executable starts running. The OS RNG I would not consider reified because even if you were to extract its internal state, subsequent calls to it are not idempotent.

---

Just to be clear, on the whole, I think I'm in rough agreement with your assessment of which functions should be considered unsafe, and at a high level I roughly agree with your rules for how to objectively assess whether a particular function or input should be considered unsafe, but I'm just trying to push for more rigorous definitions.

I'm still curious what your take is on my earlier claim that even `runJob` should be considered "unsafe" under your definition of safety, since even without `localRunner`, `runJob` can produce nondeterministic output (let's say by calling `getrandom()` reading from `/dev/urandom`, to go along with your RNG example). I think it's probably impractical to mark the result of a `getJob` call as `Unsafe`, since that would mean that almost everything in a wake program would be unsafe.

V-FEXrt:
> I'm still curious what your take is on my earlier claim that even runJob should be considered "unsafe" under your definition of safety,

Disregarding formality and rigor, for me personally, I see `Unsafe` as a way to highlight something that needs extra close review. In terms of jobs that are themselves doing things unsafe there isn't any extra value in marking them as needing extra close review. Either 1) they are a shell script obviously doing something dangerous like reading from `/dev/urandom` or they are an opaque tool/blob that we can't do anything  about in the review anyways.

JakeSiFive:
>Just to be clear, on the whole, I think I'm in rough agreement with your assessment of which functions should be considered unsafe, and at a high level I roughly agree with your rules for how to objectively assess whether a particular function or input should be considered unsafe, but I'm just trying to push for more rigorous definitions.

Always welcome!

> I'm still curious what your take is on my earlier claim that even runJob should be considered "unsafe" under your definition of safety

TL;DR Yeah its unsafe, but based on what I meant when I wrote down it turns out not even `localRunner` is unsafe! Still I think we should mark `localRunner` as unsafe, sandbox'd runJob's as safe...but we should also consider marking ReRun as unsafe perhaps?

I was not making a distinction between reified and non-reified inputs in wake. Non-reified inputs are just very obtuse to access inputs. Maybe the thing that's missing is if in practice we expect the non-deterministic inputs to change between wake invocations within the same shell script. For time and OS RNG the answer is 100% yes. In this sense sandboxed jobs are 100% unsafe by my definition. Within a single wake job they're only unsafe if ReRun is used (which made me realize that ReRun is itself unsafe). Previously I was considering only a single wake run and was only considering non-ReRun jobs. I think that oversight has caused this discussion. ReRun causes what you're pointing out to be a problem *within* a single wake invocation which I hadn't realized until now.

So in a sense its not `localRunner` that's unsafe, its `ReRun` that's unsafe. But we have to treat all jobs like they might be ReRun because they will either be run again in CI with a fresh database OR they might fall out of the shared cache if that's being used. Also I don't consider it safe to paper over non-determinism with with memoization/caching; that's a recipe for a very fragile build. I explicitly want us to never do that. You could ask if a wake program behaves the same with or without job memoization in order to justify job memoization as non-fragile.

I also agree that sandboxed jobs should not be marked as unsafe. We should should just pretend that they're safe. I'd still rather have a good model of what is safe/unsafe however and break the rules of the model occasionally personally. Just saying "I know unsafe code when I see it" doesn't really answer Sam's question. On the other hand the model I've laid out with the caveat that we don't always follow it gives a more complete understanding.

richardxia:
> Still I think we should mark `localRunner` as unsafe, sandbox'd runJob's as safe...but we should also consider marking ReRun as unsafe perhaps?

Yes, I agree with each of these points.

> So in a sense its not `localRunner` that's unsafe, its `ReRun` that's unsafe. But we have to treat all jobs like they might be ReRun because they will either be run again in CI with a fresh database OR they might fall out of the shared cache if that's being used.

I think there's an important distinction to make here, I think especially between the two usages of `ReRun` in the two sentences here. It's not that `ReRun` is intrinsically unsafe, but that it is a way for users to explicitly mark the underlying Job as not referentially transparent with respect to the reified inputs of the Job. The reason that `ReRun` needs to do this is because normal, fully sandboxed `runJob` is technically unsafe, but for pragmatic reasons, we don't want to always mark it as `Unsafe`. So, in a roundabout way, I do agree that we probably want to mark all usages of `Rerun` as `Unsafe`, but the reason is not that rerunning a job is intrinsically unsafe but that `Rerun` means that the job itself unsafe to cache.

The reason I am calling out this distinction is due to your second sentence here, specifically the first half of it mentioning a CI run with a fresh database. Just because we will rerun all Jobs in a clean CI run doesn't necessarily mean that all jobs are treated as if they might be `ReRun` (I'm using capitalization and monospace very specifically to differentiate "rerun" from `ReRun`). A user calling `setPlanPersistence ReRun` is intentionally signaling that a Job should not be cached and that it should be rerun because the user is intentionally asking for either nondeterminism or for side effects. Any pure, deterministic job should be safe to cache _and_ to rerun, and the act of rerunning a pure, deterministic job doesn't make that job unsafe.

We should probably also consider `Once` and possibly even `Keep` to be unsafe, since they both also mark various degrees of intentionally forcing Jobs to be rerun even when their (reified) inputs have not changed.

Also, to make a lesser, side point: I _do_ think `localRunner` is intrinsically unsafe, not due to `ReRun` but because asking for the `localRunner` means that you are explicitly asking for access to more inputs that the sandboxed runner does not have access to, and pretty much all of those additional inputs are either nondeterministic or preclude shared caching.

> Also I don't consider it safe to paper over non-determinism with with memoization/caching; that's a recipe for a very fragile build. I explicitly want us to never do that. You could ask if a wake program behaves the same with or without job memoization in order to justify job memoization as non-fragile.

I totally agree with this.


> I'd still rather have a good model of what is safe/unsafe however and break the rules of the model occasionally personally.

I agree with this in general, but I feel like the current model is not capable of expressing notions of safety across multiple runs, which I do think is important to capture. Unless you disagree that `setPlanPersistence Once` and `setPlanPersistence Keep` are unsafe, then we need a model that can explain why they are unsafe, and I feel that framing it in terms of reproducibility/purity is more accurate than expressing it as a set of code transformations (the laws you described earlier), which only captures the set of transformations that wake assumes is safe to make, which is really just for performance reasons.

JakeSiFive:
Ok ready for another round of review. I kept unsafeBind but I also have unsafeFlatMap as its opposite. I found that the ordering of arguments of bind was more often what was needed but occasionally unsafeFlatMap's ordering was needed. In Scala the ordering is that of Bind so things tend to work better and the use case of pipes for the opposite ordering isn't done because in Scala you'd just make a method call. I want to provide people with both orderings since both are useful in different contexts. I think `x.unsafeBind (\x ...)` looks good to me but when unsafeFlatMap is used outside of a pipe chain it's harder to read. I'm open to suggestions for new names for both of these.

I didn't change the name to `umap` for a few reasons 1) that only works for map and not for anything else so I'd have to go back to prepending "unsafe" anyway 2) I think this code breaking with wake naming norms is a feature not a bug. It should be apparent.

I added proper documentation that explains the uses and use cases.

V-FEXrt (approved):
The function  names still feel slightly off compared to the standard set by result.wake and option.wake but not enough that I feel the need to block.

For completeness, I'm  mainly talking `omap`/`rmap/`ofilter` type functions which suggest `unsafeMap` should be named `umap`

share/wake/lib/core/unsafe.wake
```diff
@@ -0,0 +1,142 @@
...
+#   unsafeFlatMap (\runner
+#     makePlan "unsafe" Nil "echo foo > foo.txt; echo bar > bar.txt"
+#     | setPlanFnOutputs runner
+#   ) unsafeRunner
+# ```
+export def unsafeFlatMap (f: a => Unsafe b) (x: Unsafe a): Unsafe b =
+    unsafeBind x f
+
+# Takes an unsafe value and pipes it through a function returning an
+# an unsafe value. This is equivlient to unsafeFlatMap with its
```
richardxia, line 100:
```suggestion
# an unsafe value. This is equivalent to unsafeFlatMap with its
```

share/wake/lib/core/unsafe.wake
```diff
@@ -0,0 +1,142 @@
...
+# arguments reversed. This can be nice to have over unsafeFlatMap
+# for syntatic reasons. You can think of it as "binding" the unsafe
+# value to the argument of the function. This is particularlly
+# nice when a lambda is used for the function argument.
+#
+# Example:
+# ```
+#   x.unsafeBind (\x y.unsafeBind (\y f x y))
+# ```
+export def unsafeBind (x: Unsafe a) (f: a => Unsafe b): Unsafe b =
```
richardxia, line 110: I know you said in your latest PR comment that you believe that both the `unsafeFlatMap` and the `unsafeBind` argument orderings are useful, but I'm still a bit concerned about having two, more or less equivalent, APIs, since it's one extra thing for a user to have to know about and then have to remember which is which. In general, I feel that idiomatic wake tends to use more of `unsafeFlatMap`'s style. I haven't pored through the codebase for this, but my gut feeling is that we tend to use `|` for functions of arity greater than 1 whenever the function is like a "method". This allows us to mimic a fluent style that most other languages get with `.` chaining, but we use `|` instead. I think we tend to only use `.` for 1-arity functions, where it looks like an "attribute access".

I think your example in the comment above would look pretty similar with `| unsafeFlatMap`:

```wake
x | unsafeFlatMap (\x y | unsafeFlatMap (\y f x y))
```

I acknowledge that this is a subjective, aesthetic opinion of mine, so I don't think it's worth blocking over this if you feel strongly about keeping it.

richardxia (approved):

share/wake/lib/core/unsafe.wake
JakeSiFive: I'm happy with just unsafeFlatMap, I can reorder things using `|` as suggested

ag-eitilt:
I'd still like to see an extended use example for this: are `|`, `$`, and judicious use of `def` enough to make up for the lack of `do` notation ergonomics?

JakeSiFive:
They are not sufficient, and the resulting code is generally not too fun to write. I added unsafeSubsume to handle the most tricky cases however.

For unrelated reasons, I've been considering adding do notation but instead of using type classes, it would take a pair of functions (one for flatmap and one for pure) and these would be assigned to let bindings (and thus would be polymorphic). The error messages in the current type checker would be awful however. GHC has an extension for a feature like this but I don't remember its name.

The big advantage of such a feature is that aside from giving us something like do notation for things like unsafe, result, option, etc... it also gives us list comprehensions which is separately a feature that I was hoping to add. Anyway I'm not going to do this anytime soon but it would be nice to dovetail things like that.

JakeSiFive:
I think its ok that things aren't super nice to write. We don't expect people to be writing extensive code like this. The pattern is roughly

1) subsume/flatmap your unsafe things you need to use
2) put your body in the middle
3) potentially use `| makeUnsafe` at the end

This kind of code tends not to format very nicely however.

ag-eitilt:
Sounds good, and since we need this code to be able to write examples using this code, I'm fine merging this as effectively-dead code and seeing it in practice later.  Most of my hesitance doesn't apply to this here so much as it is discussion of the future with no better place for that discussion to happen yet.  So on that note:

> files: returns different results for the same invocation at different points in the program
> which: same thing as files

I agree these definitely are unsafe, but I'm surprised that they can return different things, especially `which` -- we should probably describe what those different results/points are in their docs.

> claim: claim is up for grabs. Assuming no outside forces are modifying files in your workspace then claim violates no safety laws
> claimFileAsPath: Also up for grabs but with a weaker claim since we have to assume that nothing modifies this file while a build accessing this file is using it

Doesn't `claimFileAsPath` fully copy the file into the workspace?  The only time this should be non-idempotent is if it's called on the same file twice, with different output paths, when that file is edited in the mean time; otherwise wouldn't it be cached and any subsequent calls on the same input+output just use the initial `Path` (if in the same invocation) or mark anything downstream as stale like any updated output (if in different ones)?  Yeah, that's a bit of a hole, but I'm not sure I'd call treating two separate `Path`s as coincidentally identical as enough of a hole to warrant `Unsafe`.

The more likely argument to convince me is that we detect when an output file has been modified and would be clobbered when rerunning, but since we don't (and can't) do that for `source` I'm still not sure why we're holding `claim` to a higher standard.

> localRunner: This lets you do literally anything and can implement all of the unsafe functions
> setPlanFnOutputs: This violates none of the safety laws or input laws but I want to catch it in review
> editPlanFnOutputs: ditto

Yeah, these are all definitely bad ones.  `*PlanFnOutputs` are actually *very* unsafe unless we've patched them in the past month -- they at least previously didn't check what `String` paths you gave them, and were one of the (luckily only theoretical so far) means of producing a `Path` to outside the workspace.

> getJobId: as you run wake many times with the same database this will change for the same job

I wanted to say "only if you set `ReRun` or have modified inputs somewhere upstream", but since old jobs will be evicted if something else produces the same files (among other things?) it's definitely possible to have different job IDs for the same set of inputs -- I have that issue with some opt-in `claim` calls causing the original producer to rerun after opting back out -- so I agree this is `Unsafe`, if not easily so.

> wakePath: this will change depending on what version of wake you use or where your wake binary came from

What's the scope for `Unsafe`, again?  If we acknowledge that we don't guarantee safety between different repo states (which is a required admission unless we make `source` unsafe) then that *seems* like we can extend that to "the same checkout, with the same `wake`".

This one in particular is why I'm being very picky about how we define our terms -- there's very fine lines between some of the things which we're assigning to different sides of the monad, and even though this particular PR doesn't migrate anything to use it, I want us to be very deliberate about things once we do start that second stage.

> workspace: this will be different even if both workspaces have the same checkout of a repo
> cwd: this will vary based on where wake is run from
> getenv: different users will have different environments but we might consider some special WAKE prefixed variables to be inputs to the build and thus its ok that this changes
> sysname: We now only support Linux so maybe this is safe but we have yet to enforce that
> machine: We do support Arm and x86 and I'm sure we could support Risc-v without too much effort. I don't think anyone is going to use this so I think the point is moot.

These are all variants on that same scope question: unless someone is sharing a workspace between two different machines, `sysname`, `machine`, and even `workspace` are all going to be even more stable than the `source` files; is it just that we don't think about them for most code/documentation/guarantees that they're `Unsafe`?

JakeSiFive:
Here's an example of using an unsafe verison of localRunner

```
export def unsafeFiles (dir: String): Unsafe (Result (List String) Error) =
    def unsafeLocalRunner = makeUnsafe localRunner

    unsafeLocalRunner
    | unsafeFlatMap (
        makePlan "ls" Nil "ls {dir}"
        | runJobWith _
        | getJobStdout
        | rmap (tokenize `\n` _) # Note I could not use require
        | makeUnsafe
    )
```

This took me several iterations to work out and there were a lot of pains in getting here. Issues:

1) At first I tried using an lambda with a block inside of it and a require. But then I remembered that its not possible to use require in unsafe code! That means that I have to use rmap. With many requires this would get *really* verbose so you would not want to keep going like this. In this particular case however since I have a `makeUnsafe` at the end of my code I could have written this with unsafeMap which would have allowed me to use require!

2) The lambda syntax was hard to get right but eventually I realized that I could just use a `_` because I happened to have no parens. I think most people would opt for defining a function for any non-trivial flat map that they had to do.


Alternates:
```
export def unsafeFiles (dir: String): Unsafe (Result (List String) Error) =
    def unsafeLocalRunner = makeUnsafe localRunner

    unsafeLocalRunner
    | unsafeMap (
        makePlan "ls" Nil "ls {dir}"
        | runJobWith _
        | getJobStdout
        | rmap (tokenize `\n` _) # Note I could not use require
    )
```

```
export def unsafeFiles (dir: String): Unsafe (Result (List String) Error) =
    def unsafeLocalRunner = makeUnsafe localRunner
    def myRunWith runner =
        require Pass output =
            makePlan "ls" Nil "ls {dir}"
            | runJobWith runner
            | getJobStdout

        output
        | tokenize `\n`
        | Pass

    unsafeLocalRunner
    | unsafeMap myRunWith
```

The last one is probably what the average person would write? I'm not sure. I don't expect non-experts to be writing unsafe code without heavy hand holding personally.

ag-eitilt:
Ah, thanks, that's much better than I was afraid of!  I'm still a bit concerned about heavily-nested definitions, but all value manipulation being inside `unsafeMap` rather than at the same level as `unsafeLocalRunner` avoids most of the contortions I thought we'd have to be dealing with.

JakeSiFive:
`claimFileAsPath` is unsafe because its reading from an external mutable source. So if you call it one time, and then call it a second time with the same inputs, the output can be different. That's the definition of unsafe. Still we might intentionally choose not to make this unsafe but considering people use it to pull from source we'd rather them not pull from I think its good to consider making this unsafe. It might be too much of a head ache to port however. But considering that claimFileAsPath lets you read litterally anything on the filesystem, I'd say its unsafe.

Anyone depending `wakePath` or `workspace` could write things that differ from CI run to CI run which we 100% do not want so it should be unsafe. This is part of what my model is lacking but you can imagine that there are some effectual functions like "forget the database" or "forget targets" or "move from an unmodified checkout, to a seperate unmodified checkout on the same commit" that we'd like to be invariant over.

A lot of these come down to "what do we consider a non-invariant input to the build?" and that is an as yet unanswered question that we will have to resolve over time.

PR #1451, opened 2023-10-30
Add toUpper and toLower

JakeSiFive:
This PR just adds primitives for converting to all upper case and all lower case, we I recently needed but didn't have access to.

### DIFF
--- a/share/wake/lib/core/string.wake
+++ b/share/wake/lib/core/string.wake
@@ -223,7 +223,7 @@ export def unicodeIdentifier (str: String): String =
 
     p str
 
-# unicodeLowercase: converts upper case codepoints to thier lower case counterparts
+# unicodeLowercase: converts upper case codepoints to their lower case counterparts
 export def unicodeLowercase (str: String): String =
     def p s = prim "tolower"
 
--- a/share/wake/lib/core/string.wake
+++ b/share/wake/lib/core/string.wake
@@ -233,6 +233,8 @@ export def unicodeLowercase (str: String): String =
 export def unicodeUppercase (str: String): String =
     def p s = prim "toupper"
 
+    p str
+
 # sortStrings: sort a list of strings as a human would judge them.
 export def sortStrings (list: List String): List String =
     sortBy (_ <=>~ _) list
--- a/share/wake/lib/core/string.wake
+++ b/share/wake/lib/core/string.wake
@@ -342,11 +342,15 @@ export def hashString (x: String): String =
 
     hashStr x
 
+# Returns the string given but with all upper case letters converted
+# to lower case
 export def toLower (x: String): String =
     def lower x = prim "tolower"
 
     lower x
 
+# Returns the string given but with all lower case letters converted
+# to upper case
 export def toUpper (x: String): String =
     def upper x = prim "toupper"
 
--- a/share/wake/lib/core/string.wake
+++ b/share/wake/lib/core/string.wake
@@ -223,11 +223,15 @@ export def unicodeIdentifier (str: String): String =
 
     p str
 
-# unicodeLowercase: eliminate case distinctions; C => c
+# unicodeLowercase: converts upper case codepoints to thier lower case counterparts
 export def unicodeLowercase (str: String): String =
-    def p s = prim "scaseNFKC"
+    def p s = prim "tolower"
 
     p str
+
+# unicodeUppercase: converts lower case codepoints to their upper case counterparts
+export def unicodeUppercase (str: String): String =
+    def p s = prim "toupper"
 
 # sortStrings: sort a list of strings as a human would judge them.
 export def sortStrings (list: List String): List String =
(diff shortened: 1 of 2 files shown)


share/wake/lib/core/string.wake
```diff
@@ -341,3 +341,13 @@ export def hashString (x: String): String =
     def hashStr x = prim "hash_str"
 
     hashStr x
+
+export def toLower (x: String): String =
```
V-FEXrt, line 345: These need doc comments

src/runtime/string.cpp
```diff
@@ -680,6 +680,56 @@ static PRIMFN(prim_hash_str) {
...
+  const char *in = str->c_str();
+  ssize_t in_size = str->length;
+  std::string out;
+  while (in_size > 0) {
+    int32_t codepoint = 0;
+    ssize_t size = utf8proc_iterate(reinterpret_cast<const uint8_t *>(in), in_size, &codepoint);
+    in += size;
+    in_size -= size;
+    int32_t upper = utf8proc_toupper(codepoint);
+    char buf[4] = {0};
```
V-FEXrt, line 700: maybe I'm misunderstanding the API, but `4` here doesn't feel safe

src/runtime/string.cpp
```diff
@@ -680,6 +680,56 @@ static PRIMFN(prim_hash_str) {
...
+    ssize_t bytes = utf8proc_encode_char(upper, reinterpret_cast<uint8_t *>(buf));
+    out.insert(out.end(), buf, buf + bytes);
+  }
+
+  RETURN(String::alloc(runtime.heap, std::move(out)));
+}
+
+static PRIMTYPE(type_to_lower) {
```
V-FEXrt, line 708: we 100% shouldn't do this but I just realized we could probably have a `static PRIMTYPE(type_str_to_str) { ... }`  that we share between all `String => String`  prims

V-FEXrt (approved):

V-FEXrt (changes requested):
oops wrong  button

src/runtime/string.cpp
JakeSiFive: encode encodes a single code point and returns the number of bytes it took up. The API specifically wants 4 bytes here.

src/runtime/string.cpp
JakeSiFive: Yeah I always wonder if I should do that and then never do it lol.

share/wake/lib/core/string.wake
JakeSiFive: Added. I didn't feel like adding super detailed docs since I felt like it was mostly self explanatory.

src/runtime/string.cpp
```diff
@@ -680,6 +680,56 @@ static PRIMFN(prim_hash_str) {
...
+  return args.size() == 1 && args[0]->unify(Data::typeString) && out->unify(Data::typeString);
+}
+
+static PRIMFN(prim_to_lower) {
```
richardxia, line 712: What's the difference between this new prim and the existing [`scaseNFKC`](https://github.com/sifive/wake/blob/31c5cda2c8b0864ee65e9ee29e9d95c8acc8bcb2/src/runtime/string.cpp#L520) prim, which is used by string.wake's [`unicodeLowercase`](https://github.com/sifive/wake/blob/31c5cda2c8b0864ee65e9ee29e9d95c8acc8bcb2/share/wake/lib/core/string.wake#L226-L230) function? I'm especially curious about how the existing function was basically able to delegate the entire thing to the library, without having to do any iteration on its own.

JakeSiFive: I actually couldn't find that! I was searching for the upper case conversion which I couldn't find because it wasn't implemented. On further inspection, that primitive is kind of abusing case folding. Case folding happens in practice to lower case everything but all it says it does is that it canonicalizes the cases. There's no way to repeat that for the to upper case.

I think we should get rid of the previous primitive and I should just add `unicodeUppercase` to be consistent with naming.

richardxia: Yeah, I figured something weird was going on with it, since it wasn't obvious to me how the `UTF8PROC_CASEFOLD` flag implied lowercase. 👍 on removing it and replacing the previous implementation of `unicodeLowercase` with your new function.

JakeSiFive: to answer directly:
the old primitive used utf8proc_map which just has a list of transformations that can be applied. One of them is called "case folding" which has unspecified semantics but maps everything into a unique form with respect to case. In practice this is lower case. Additionally you have to have pre-allocated a large enough buffer for that to be used and I see now actual assurance that such a thing is generally true for lower and upper case. The map function is allowed to return an error if the output is larger than the input (although that might never happen with upper/lower cases).

The new iterative implementation makes no such assumptions and cannot fail unless memory allocation fails.

I got rid of the old scase primitive meant for canonicalization and replaced it with the iterative version. If there was a "toupper" version of scaseNFKC I would use that instead but it doesn't appear to exist in our unicode library. In the intrest of having tolower and toupper function the same way, and as a bonus fixing those API holes we just found, think we should stick to the iterative version.

I did however switch to using the previous naming scheme

share/wake/lib/core/string.wake
```diff
@@ -223,9 +223,15 @@ export def unicodeIdentifier (str: String): String =
 
     p str
 
-# unicodeLowercase: eliminate case distinctions; C => c
+# unicodeLowercase: converts upper case codepoints to thier lower case counterparts
```
V-FEXrt, line 226: typo: `thier` -> `their`

V-FEXrt (approved):

PR #1453, opened 2023-10-31
Identify `write` job with mode as octal, not decimal

ag-eitilt:
In trying to debug an odd `write` job, @maniltongya and I couldn't find where it was being called despite what seemed like a very unique `420` mode being printed in the job metadata.  I eventually realized that was being printed as *decimal* instead, and was the much more typical `0644`.  This PR still doesn't help us find where the offending `write` is happening, but it will at least avoid the same confusion over what's being set.

This and `mkdirImp` seem to be the only jobs which specify the mode, and that one already uses the proper base.  There shouldn't be any issue with updating this (beyond all `write` jobs having to be rerun) as the actual `prim` call never interacts with the label/cmdline, and a very simple `wake -x 'write "tmp.txt" "Test"'` shows everything being correct.

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -68,8 +68,8 @@ target writeImp inputs mode path content =
 
     # If all those checks pass we go ahead and perform the write. The write will
     # overwrite single files but it will not overwrite a whole directory with a file.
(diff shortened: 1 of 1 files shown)


JakeSiFive (approved):
Yep. 100% a good change

V-FEXrt (approved):

PR #1454, opened 2023-11-01
Implement tag filters for database inspection

V-FEXrt:
Implements database inspection for filtering on tags. Using the same semantics as all other filters except that two values are provided with a `=` between them. `--tag a=b` filters for all tags where the URI is `a` and the content is `b`. `*` and `?` work as expected.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -435,8 +435,6 @@ std::string Database::open(bool wait, bool memory, bool tty) {
       " from jobs j left join stats s on j.stat_id=s.stat_id join runs r on j.run_id=r.run_id "
       "inner join tags t on j.job_id = t.job_id"
       " where t.uri like ? AND t.content like ? order by j.job_id";
-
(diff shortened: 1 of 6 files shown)


src/runtime/database.cpp
```diff
@@ -1289,13 +1303,13 @@ std::vector<JobReflection> Database::last_exe() { return find_all(this, imp->fin
 
 std::vector<JobReflection> Database::last_use() { return find_all(this, imp->find_last_use); }
 
-std::vector<JobReflection> Database::labels_matching(const std::string glob) {
+std::vector<JobReflection> Database::labels_matching(const std::string &glob) {
```
JakeSiFive, line 1306: 💀 how did we let that happen whoops

JakeSiFive (approved):

src/runtime/database.cpp
V-FEXrt: I had the exact same thought

PR #1455, opened 2023-11-03
TTL eviction policy

JakeSiFive:
This PR is 95% of what needs to be done for adding a TTL eviction policy. Specifically it does the following already

1) It implements a TTL eviction policy
2) It implements plumbing of eviction from the config system, to the wake tool, to the job-cache client, across the process boundary to daemon, and across another boundary to the eviction process (which could probably be made a thread at this point?)
3) Updates almost all tests to work again

TODO:
1) I have to fix the basic-lru test. For some reason I broke it
2) I have to write a "basic-ttl" test
3) I want to have fuzz testing for both ttl and lru, not just LRU

Once those TODOs are TODONE we'll have a better eviction since LRU is broken (doesn't mean we should let the still valid tests we currently have rot)


src/job_cache/eviction_policy.cpp
```diff
@@ -42,6 +42,62 @@
...
+      auto reset1 = wcl::make_defer([this]() { jobs_older_than.reset(); });
+      jobs_older_than.bind_integer(1, deadline);
+      // First find the use time that we want to remove from
+      while (jobs_older_than.step() == SQLITE_ROW) {
+        int64_t job_id = jobs_older_than.read_integer(0);
+        std::string cmd = jobs_older_than.read_string(1);
+        for (auto& ch : cmd) {
+          if (ch == '\0') ch = ' ';
+        }
+        jobs_to_remove.emplace_back(job_id, std::move(cmd));
```
V-FEXrt, line 79: why do we care about the command here? Isn't the job_id sufficient?

src/job_cache/eviction_policy.cpp
```diff
@@ -42,6 +42,62 @@
...
+      }
+
+      // If there are jobs to remove then we go about actully removing them here
+      if (!jobs_to_remove.empty()) {
+        auto reset2 = wcl::make_defer([this]() { remove_jobs_older_than.reset(); });
```
V-FEXrt, line 84: nit: I think `reset1` and `reset2` can just be `reset` since they are both in separate scopes

src/job_cache/eviction_policy.cpp
```diff
@@ -364,6 +420,32 @@ LRUEvictionPolicy::LRUEvictionPolicy(uint64_t max, uint64_t low)
...
+  gc_thread = std::thread(garbage_collect_orphan_folders, db);
+}
+
+void TTLEvictionPolicy::read(int job_id) {}
+
+void TTLEvictionPolicy::write(int job_id) {
+    static time_t time_of_last_cleanup = 0;
+    time_t current = time(nullptr);
+    // We spend a bit of time seeing if we even need to clean anything up
+    if (current - time_of_last_cleanup > 5) {
```
V-FEXrt, line 439: The `5` here should probably be a `resolution` variable in `EvictionConfig`

tools/job-cache/main.cpp
```diff
@@ -23,26 +23,144 @@
...
+      std::cerr << "Encountered '" << arg << "' which is not a recognized option" << std::endl;
+    }
+  }
+};
+
 // argv[0] = program name
 // argv[1] = cache dir
 // argv[2] = bulk cache dir
 // argv[3] = low cache size
 // argv[4] = max cache size
```
V-FEXrt, line 90: This is not longer an accurate comment

V-FEXrt (commented):
All in all looks like the right direction. Waiting to approve for the final TODOs

src/job_cache/eviction_policy.cpp
JakeSiFive: We want to log the command of the job that we're evicting

src/job_cache/eviction_policy.cpp
JakeSiFive: It's a lot of extra plumbing. Do you mind if we skip this for now?

V-FEXrt: yeah thats fine

src/job_cache/daemon_cache.cpp
```diff
@@ -1140,8 +1139,18 @@ void DaemonCache::launch_evict_loop() {
 
     // Finally enter the eviction loop, if it exits cleanly
     // go ahead and exit with its result.
-    int result =
-        eviction_loop(".", std::make_unique<LRUEvictionPolicy>(max_cache_size, low_cache_size));
+    std::unique_ptr<EvictionPolicy> policy;
+    switch (config.type) {
+      case EvictionPolicyType::TTL:
+        wcl::log::info("Using TTL eviction policy")();
```
colinschmidt, line 1145: I feel like theres no reason not to include the top-level details of these policies in the log (e.g. second to live for TTL and min and max size for LUR)

tools/wake-unit/fuzz_test_job_cache.cpp
```diff
@@ -520,7 +525,8 @@ TEST(job_cache_basic_fuzz) {
   config.number_of_steps = 10000;
   config.cache_dir = ".job_cache_test";
   config.dir = "job_cache_test";
-  TEST_FUNC_CALL(fuzz_loop, config, std::move(gen));
+  TEST_FUNC_CALL(fuzz_loop, config, true, gen);
+  TEST_FUNC_CALL(fuzz_loop, config, false, std::move(gen));
```
V-FEXrt, line 529: why does only one of these have `std::move`?

V-FEXrt: Making a copy for the first call and moving on the second I presume?

tests/job-cache/basic-ttl/pass.sh
```diff
@@ -0,0 +1,112 @@
+#! /bin/sh
+
+if [ $(uname) != Linux ] ; then
+  exit 0
+fi
```
V-FEXrt, line 5: nit: I don't think this is needed anymore

src/job_cache/eviction_policy.cpp
```diff
@@ -42,6 +42,64 @@
...
+      while (jobs_older_than.step() == SQLITE_ROW) {
+        int64_t job_id = jobs_older_than.read_integer(0);
+        std::string cmd = jobs_older_than.read_string(1);
+        for (auto& ch : cmd) {
+          if (ch == '\0') ch = ' ';
+        }
+        jobs_to_remove.emplace_back(job_id, std::move(cmd));
+      }
+
+      // If there are jobs to remove then we go about actully removing them here
```
V-FEXrt, line 84: nit: `actully` -> `actually`

V-FEXrt (approved):

src/job_cache/daemon_cache.cpp
```diff
@@ -1140,8 +1139,20 @@ void DaemonCache::launch_evict_loop() {
...
-        eviction_loop(".", std::make_unique<LRUEvictionPolicy>(max_cache_size, low_cache_size));
+    std::unique_ptr<EvictionPolicy> policy;
+    switch (config.type) {
+      case EvictionPolicyType::TTL:
+        wcl::log::info("Using TTL eviction policy, seconds_to_live = %lu",
+                       config.ttl.seconds_to_live)();
+        policy = std::make_unique<TTLEvictionPolicy>(config.ttl.seconds_to_live);
+        break;
+      case EvictionPolicyType::LRU:
+        wcl::log::info("Using LRU eviction policy, max = %lu, min = %lu", config.lru.low_size,
```
colinschmidt, line 1150: It looks like the debug print has the max and min switched.

JakeSiFive: It does indeed, I'll fix it in a follow up before we release

PR #1456, opened 2023-11-03
Add file streams for teeing Job stdout/err to disc

ag-eitilt:
For logs, etc., it's desirable to write Job outputs to known files in addition to the existing streams mechanisms -- for example, directing specific Jobs to different locations, and writing those logs even if Wake is killed partway through.

### DIFF
--- a/src/runtime/job.cpp
+++ b/src/runtime/job.cpp
@@ -678,22 +678,17 @@ JobTable::~JobTable() {
   }
 }
 
-static char **split_null(std::string &str) {
-  int nulls = 0;
-  for (char c : str) nulls += (c == 0);
-  char **out = new char *[nulls + 1];
-
-  nulls = 0;
-  out[0] = const_cast<char *>(str.c_str());
-  char *end = out[0] + str.size();
-  for (char *scan = out[0]; scan != end; ++scan) {
-    if (*scan == 0) {
-      ++nulls;
-      out[nulls] = scan + 1;
+static std::vector<std::string> split_null(std::string &str) {
+  std::vector<std::string> out;
+  std::string buf;
+  for (char c : str) {
+    if (c == '\0') {
+      out.emplace_back(std::move(buf));
+    } else {
+      buf += c;
     }
   }
-  out[nulls] = 0;
-
+  out.emplace_back(std::move(buf));
   return out;
 }
 
@@ -835,24 +830,22 @@ static void launch(JobTable *jobtable) {
       err = std::make_unique<NullBuf>();
     }
 
-    auto stdout_teefiles = split_null(task.job->stdout_teefiles);
+    const auto stdout_teefiles = split_null(task.job->stdout_teefiles);
     auto stdout_tee_names = std::vector<std::string>();
-    for (int i = 0; stdout_teefiles[i]; ++i) {
-      stdout_tee_names.push_back(stdout_teefiles[i]);
-      if (jobtable->imp->teefiles.find(stdout_teefiles[i]) == jobtable->imp->teefiles.end()) {
-        jobtable->imp->teefiles[stdout_teefiles[i]] = std::make_unique<std::ofstream>(stdout_teefiles[i], std::ios::out | std::ios::trunc);
+    for (const auto &file : stdout_teefiles) {
+      stdout_tee_names.push_back(file);
+      if (jobtable->imp->teefiles.find(file) == jobtable->imp->teefiles.end()) {
+        jobtable->imp->teefiles[file] = std::make_unique<std::ofstream>(file, std::ios::out | std::ios::trunc);
       }
     }
-    auto stderr_teefiles = split_null(task.job->stderr_teefiles);
+    const auto stderr_teefiles = split_null(task.job->stderr_teefiles);
     auto stderr_tee_names = std::vector<std::string>();
-    for (int i = 0; stderr_teefiles[i]; ++i) {
-      stderr_tee_names.push_back(stderr_teefiles[i]);
-      if (jobtable->imp->teefiles.find(stderr_teefiles[i]) == jobtable->imp->teefiles.end()) {
-        jobtable->imp->teefiles[stderr_teefiles[i]] = std::make_unique<std::ofstream>(stderr_teefiles[i], std::ios::out | std::ios::trunc);
+    for (const auto &file : stderr_teefiles) {
+      stderr_tee_names.push_back(file);
+      if (jobtable->imp->teefiles.find(file) == jobtable->imp->teefiles.end()) {
+        jobtable->imp->teefiles[file] = std::make_unique<std::ofstream>(file, std::ios::out | std::ios::trunc);
       }
     }
-    delete[] stdout_teefiles;
-    delete[] stderr_teefiles;
 
     std::shared_ptr<JobEntry> entry = std::make_shared<JobEntry>(
         jobtable->imp.get(), std::move(task.job), std::move(out), std::move(err), std::move(stdout_tee_names), std::move(stderr_tee_names));
@@ -893,8 +886,6 @@ static void launch(JobTable *jobtable) {
     pid_t pid = wake_spawn(cmdline[0], cmdline, environ);
     sigprocmask(SIG_BLOCK, &set, 0);
 
-    delete[] cmdline;
-    delete[] environ;
     ++jobtable->imp->num_running;
     jobtable->imp->pidmap[pid] = entry;
     entry->job->pid = entry->pid = pid;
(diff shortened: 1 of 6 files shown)


src/runtime/job.cpp
```diff
@@ -105,7 +106,9 @@ struct Job final : public GCObject<Job, Value> {
   bool keep;
   std::string echo;
   std::string stream_out;
+  std::vector<std::string> tee_out;
   std::string stream_err;
+  std::vector<std::string> tee_err;
```
JakeSiFive, line 111: Something I realized is that we need to make sure that virtual jobs write to these files.

Also also, I just realized that we can make runners treat these files as outputs of some random virtual job so that they still get hashed and can be overwritten or otherwise read in.

src/runtime/job.cpp
```diff
@@ -824,8 +830,18 @@ static void launch(JobTable *jobtable) {
     } else {
       err = std::make_unique<NullBuf>();
     }
+
+    auto stdout_teefiles = std::vector<std::unique_ptr<std::ofstream>>();
+    for (auto tee_path : task.job->tee_out) {
+      stdout_teefiles.push_back(std::make_unique<std::ofstream>(tee_path, std::ios::out | std::ios::app));
```
JakeSiFive, line 836: hmm, we should talk about if these should be append only or not. I think I'm inclined to say they should not be appended to.

src/runtime/job.cpp
```diff
@@ -1286,18 +1316,35 @@ static PRIMFN(prim_job_virtual) {
...
          args[3]->unify(Data::typeString) && args[4]->unify(Data::typeString) &&
          args[5]->unify(Data::typeInteger) && args[6]->unify(Data::typeString) &&
          args[7]->unify(Data::typeInteger) && args[8]->unify(Data::typeString) &&
          args[9]->unify(Data::typeString) && args[10]->unify(Data::typeString) &&
-         args[11]->unify(Data::typeInteger) && out->unify(Data::typeJob);
+         args[11]->unify(Data::typeString) && args[12]->unify(Data::typeString) &&
+         args[13]->unify(Data::typeInteger) && out->unify(Data::typeJob);
+}
+
+std::vector<std::string> null_sep_list(const std::string &str) {
```
JakeSiFive, line 1329: A function like this already exists further into the job.cpp code that is used on cmd and env that you can use instead.

src/runtime/job.cpp
```diff
@@ -1396,7 +1448,7 @@ static PRIMFN(prim_job_cache) {
   Value *joblist;
   if (reuse.found && !jobtable->imp->check) {
     Job *jobp = Job::claim(runtime.heap, jobtable->imp->db, dir, dir, stdin_file, env, cmd, true,
-                           STREAM_ECHO, STREAM_INFO, STREAM_WARNING);
+                           STREAM_ECHO, STREAM_INFO, std::vector<std::string>(), STREAM_WARNING, std::vector<std::string>());
```
JakeSiFive, line 1451: You can use `{}` for these cases, but per my previous comment, I'd recommend using `""` and passing a string here to delay the parsing.

src/runtime/job.cpp
ag-eitilt: Over-the-wall discussion about storing these all to the JobTable has resolved this.

src/runtime/job.cpp
ag-eitilt: Good hit on testing with virtual jobs.  I think we *don't* want to hash these (yet), though -- you mentioned "incidental files" during the brainstorming in a way which made me think they'll allow things like `wake -o` without full `Path` citizenship, which I think will be ideal for these logs once we get that implemented, but for now I think it's better to just drop them from FnOutputs entirely.

src/runtime/job.cpp
ag-eitilt: Ooh, there is?  Great, thanks!  (I thought I had searched for it, but apparently I did so particularly poorly.)

src/runtime/job.cpp
ag-eitilt: Ah, yeah, that makes sense.  Let me make sure it doesn't conflict with anywhere we already need to have the `vector`, but otherwise that'll make it cleaner.

src/runtime/job.cpp
JakeSiFive: Yeah we can't hash since you pointed out that we want to tee things out to the same file. The incidental files can be used but it won't panic if someone overwrites a tee file. I think we can add a check that incidental files are not hashed though and then wake --clean will still work.

mwachs5:
does anything in the `logging.md` need to get updated for this change?

JakeSiFive:
> does anything in the `logging.md` need to get updated for this change?

It shouldn't invalidate anything but it will add a new feature. If Sam want's to take that, I'm ok with that, otherwise I'll do it in a follow up PR

ag-eitilt:
I can definitely update that as part of this, once I get the implementation side of things done to know what I'll be updating it with.  Thanks for pointing it out!

src/runtime/job.cpp
```diff
@@ -824,8 +834,28 @@ static void launch(JobTable *jobtable) {
...
+    for (int i = 0; stderr_teefiles[i]; ++i) {
+      stderr_tee_names.push_back(stderr_teefiles[i]);
+      if (jobtable->imp->teefiles.find(stderr_teefiles[i]) == jobtable->imp->teefiles.end()) {
+        jobtable->imp->teefiles[stderr_teefiles[i]] = std::make_unique<std::ofstream>(stderr_teefiles[i], std::ios::out | std::ios::trunc);
+      }
+    }
+    delete[] stdout_teefiles;
+    delete[] stderr_teefiles;
+
     std::shared_ptr<JobEntry> entry = std::make_shared<JobEntry>(
```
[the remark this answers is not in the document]
JakeSiFive:
```
    auto stdout_tee_names = std::vector<std::ostream*>();
    for (const auto& file : stdout_teefiles) {
      auto iter = jobtable...->teefiles.find(file);
      if (iter == ... .end()) {
          auto out = std::make_unique<std::ofstream>(stdout_teefiles[i], std::ios::out);
          iter = jobtable->imp->teefiles[stdout_teefiles[i]].emplace(file, std::move(out)).first
      }
      stdout_tee_names.push_back(iter->second.get());
    }
 ```

File: src/runtime/job.cpp [hunk 6]
```diff
@@ -824,8 +834,28 @@ static void launch(JobTable *jobtable) {
     } else {
       err = std::make_unique<NullBuf>();
     }
+
+    auto stdout_teefiles = split_null(task.job->stdout_teefiles);
+    auto stdout_tee_names = std::vector<std::string>();
+    for (int i = 0; stdout_teefiles[i]; ++i) {
+      stdout_tee_names.push_back(stdout_teefiles[i]);
+      if (jobtable->imp->teefiles.find(stdout_teefiles[i]) == jobtable->imp->teefiles.end()) {
+        jobtable->imp->teefiles[stdout_teefiles[i]] = std::make_unique<std::ofstream>(stdout_teefiles[i], std::ios::out | std::ios::trunc);
```
JakeSiFive, line 843: We should do an unlink after this so that we get a new inode instead of modifying the existing file.

JakeSiFive (commented):
Everything is looking good!

src/runtime/job.cpp
```diff
@@ -339,6 +347,9 @@ struct JobTable::detail {
     for (auto &entry : term_bufs) {
       entry.second.release();
     }
+    for (auto &entry : teefiles) {
+      entry.second->close();
+    }
```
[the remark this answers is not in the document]
ag-eitilt: Good to know!

src/runtime/job.cpp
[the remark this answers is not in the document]
ag-eitilt: That's a nice little tweak that makes me glad I don't write C++ on the regular...

src/runtime/job.cpp
ag-eitilt: I'm not seeing that function in the C++ reference I'm using, though I do recognize what it means in context.  Mind giving me a bit more guidance what that would look like in code?

JakeSiFive: unlink is a posix function: https://man7.org/linux/man-pages/man2/unlink.2.html

ag-eitilt: Apparently it's a bit less complete a reference than I expected, thanks!  (They could at least have included it and marked it as system dependent somehow...)

JakeSiFive:
```
unlink(file.c_str());
```

Is all you would need but its probably best to only ignore the error if ENOENT is returned. Everything else should be logged as a warning


```
if (unlink(file.c_str()) < 0 && errno != ENOENT) {
    wcl::log::warning("Could not unlink %s while creating a new file for tee-ing: %s", file.c_str(), strerror(errno))();
}
```

PR #1466, opened 2023-11-16
Optimize ORM insert queries

V-FEXrt:
Previously we were making a query per file entry in add_job. Now we always make a constant 5 queries on add_job

### DIFF
--- a/rust/rsc/src/rsc/add_job.rs
+++ b/rust/rsc/src/rsc/add_job.rs
@@ -1,6 +1,9 @@
 use axum::{http::StatusCode, Json};
+use entity::prelude::{OutputDir, OutputFile, OutputSymlink, VisibleFile};
 use entity::{job, output_dir, output_file, output_symlink, visible_file};
-use sea_orm::{ActiveModelTrait, ActiveValue, DatabaseConnection, DbErr, TransactionTrait};
+use sea_orm::{
+    ActiveModelTrait, ActiveValue, DatabaseConnection, DbErr, EntityTrait, TransactionTrait,
+};
 use std::sync::Arc;
 use tracing;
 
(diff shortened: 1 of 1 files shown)


rust/rsc/src/rsc/add_job.rs
```diff
@@ -44,49 +47,67 @@ pub async fn add_job(
                 let job = insert_job.save(txn).await?;
                 let job_id = job.id.unwrap();
 
-                // TODO: Convert these loops to singular add_many calls
+                let mut visible_files = Vec::new();
                 for vis_file in vis {
```
JakeSiFive, line 51: let visible_files = vis.iter().map(|visible_file| {
    visible_file::ActiveModel {
                        id: ActiveValue::NotSet,
                        path: ActiveValue::Set(vis_file.path),
                        hash: ActiveValue::Set(vis_file.hash.into()),
                        job_id: ActiveValue::Set(job_id)
                        }
}).collect();

JakeSiFive: Also lol, apparently I had a TODO for this already

JakeSiFive (approved):
Approved with a nit about not introducing mutable things by using map

rust/rsc/src/rsc/add_job.rs
```diff
@@ -47,61 +47,61 @@ pub async fn add_job(
                 let job = insert_job.save(txn).await?;
                 let job_id = job.id.unwrap();
 
-                let mut visible_files = Vec::new();
-                for vis_file in vis {
-                    visible_files.push(visible_file::ActiveModel {
+                let visible_files: Vec<visible_file::ActiveModel> = vis
```
JakeSiFive, line 50: BTW, you can use `Vec<_>` here. In fact I think you might be able to get away with no type annotation because `insert_many` expects a Vec and your pre-collector iterator has the right item type

JakeSiFive (approved):
LGTM

rust/rsc/src/rsc/add_job.rs
V-FEXrt: `Vec<_>` worked but completely removing didn't

JakeSiFive: Interesting, ok. I'm still not 100% sure how Rust type inference works. Maybe insert_many has a more general signature than last time I looked at it.

JakeSiFive: Ah yeah, they updated it to use an iterator! You can actually avoid the collect and pass the iterator directly in and you'll be even *more* efficient. I remember the signature of that being a Vec in the previous version.

PR #1468, opened 2023-11-17
rsc: Add created_at column to all tables

V-FEXrt:
Adds an automatically filled `created_at` column to all tables.


rust/migration/src/m20231117_162713_add_created_at.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        Ok(())
+    }
+}
+
+#[derive(DeriveIden)]
+pub enum Job {
+    Table,
+    CreatedAt,
+}
```
JakeSiFive, line 76: I think you can just add the columns to the existing Idens for these tables

rust/migration/src/m20231117_162713_add_created_at.rs
```diff
@@ -0,0 +1,112 @@
...
+        } as (t, c) in {
+            manager
+                .alter_table(
+                    Table::alter()
+                        .table(t)
+                        .add_column(
+                            ColumnDef::new(c)
+                                .timestamp()
+                                .not_null()
+                                .default(SimpleExpr::Keyword(Keyword::CurrentTimestamp))
```
JakeSiFive, line 40: It took me seeing the NotSet to realize that how nice this was. I thought it was setting the infill for the migration but it makes the other insertions automatically set it to?

Something is going on with SQL there that I don't understand lol

V-FEXrt: > but it makes the other insertions automatically set it to

Correct, that column is automatically filled by the db engine

rust/migration/src/m20231117_162713_add_created_at.rs
V-FEXrt: I would have to import them from the other migration if I did that. And it would also be confusing that the old migration has the `CreatedAt` column but doesn't use it. I think its better to keep migrations self contained

rust/migration/src/m20231117_162713_add_created_at.rs
JakeSiFive: This is just me being curious about SQL: What happens if you do `2 * CurrentTimestamp`? Is the default allowed to be a *function* of the current time?! Or is there just a special keyword in SQL that I'm not familiar with and something will error out if you try to use a function in that context?

V-FEXrt: its a function but yeah https://www.postgresqltutorial.com/postgresql-date-functions/postgresql-current_timestamp/

JakeSiFive: ok this clears things up for me: https://www.postgresql.org/docs/current/ddl-default.html

> The default value can be an expression, which will be evaluated whenever the default value is inserted (not when the table is created). A common example is for a timestamp column to have a default of CURRENT_TIMESTAMP, so that it gets set to the time of row insertion. Another common example is generating a “serial number” for each row. In PostgreSQL this is typically done by something like:

JakeSiFive: I didn't know this was possible, TIL!

rust/migration/src/m20231117_162713_add_created_at.rs
V-FEXrt: My strongest counter here is, what should we do if we (over the series of months) 1) create table Foo 2) delete table Foo 3) create a _new_ table Foo with no association to the original?

V-FEXrt: if that doesn't convenience ya then happy to move forward with that plan

JakeSiFive: That would be a new table that just happens to have the same name but it would be reflected in the fact that it was a different type.

JakeSiFive (approved):
Spoke over chat about this, approved either way.

PR #1472, opened 2023-11-28
rsc: Paydown JobUse tech debt

V-FEXrt:
Last bit of tech debt cleanup before digging in to new feature work.

- Removes `time` from `JobUses` now that its covered by `CreatedAt`
- Renames `JobUses` to `JobUse` to match all other models.

### DIFF
--- a/rust/migration/src/m20231127_232833_drop_use_time.rs
+++ b/rust/migration/src/m20231127_232833_drop_use_time.rs
@@ -1,0 +1,40 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(JobUses::Table)
+                    .drop_column(JobUses::Time)
+                    .to_owned(),
+            )
+            .await
+    }
+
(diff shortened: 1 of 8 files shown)


rust/migration/src/m20231127_232833_drop_use_time.rs
```diff
@@ -0,0 +1,40 @@
...
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(JobUses::Table)
+                    .add_column(
+                        ColumnDef::new(JobUses::Time)
+                            .timestamp()
+                            .not_null()
+                            .default(SimpleExpr::Keyword(Keyword::CurrentTimestamp)),
```
JakeSiFive, line 28: This should be replaced with CreateTime not the current time stamp

V-FEXrt: lol how do you do that?

JakeSiFive: I don't know lol

JakeSiFive (commented):
We spoke off line. You need to write a query to do this and that's too much work right now for the gain but the pattern would look like this

1) Add the alter table thing but allow the new field to be nullable
2) where these values are null, update the value as desired using an update query
3) Modify the column using a new column def identical to the last but where not_null is set (Which is now true thanks for the above update query).

We could have avoided this by using `rename_column` for this use case on the previous migration but in this particular downward migrartion we're migrating from having a single time column to having two time columns. In the future maybe we'll have more foresight.

JakeSiFive (approved):

PR #1475, opened 2023-11-28
rsc: Add basic integraiton tests

V-FEXrt:
As we start spinning up larger development churn it would be good to ensure things don't regress. I've added a basic integration test for the core flow to help with that.


JakeSiFive (approved):
Awesome!

PR #1476, opened 2023-11-29
don't join threads in transactions (or block in anyway)

JakeSiFive:
So we're not quite sure *why* this thread was hanging but it wound up hanging somehow, possibly just long enough to timeout another transaction. So anyway, a transaction in another process was timed out due to this thread joining and we had to avoid that. It turns out its safe for multiple instances of remove_backing_files to run at the same time so we just allow that and detach the thread so that it can run independently. This ensures that we don't do any blocking operations inside the transaction and that the eviction process keeps responding.

### DIFF
--- a/src/job_cache/eviction_policy.cpp
+++ b/src/job_cache/eviction_policy.cpp
@@ -68,6 +68,12 @@ struct TTLEvictionPolicyImpl {
 
     wcl::log::info("Performing a cleanup!")();
 
+    // We're about to assign to this thread but we don't want to
+    // wait for it to exit so we detach the thread before making a new one.
+    // The function being run is thread safe so many instances can be run at
+    // once safely.
+    if (cleaning_thread.joinable()) cleaning_thread.detach();
+
     transact.run([this, &jobs_to_remove, deadline]() {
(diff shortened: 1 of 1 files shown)


src/job_cache/eviction_policy.cpp
```diff
@@ -68,6 +68,12 @@ struct TTLEvictionPolicyImpl {
 
     wcl::log::info("Performing a cleanup!")();
 
+    // We're about to assign to this thread but we don't want to
+    // wait for it to exit so we detach the thread before making a new one.
+    // The function being run is thread safe so many instances can be run at
+    // once safely.
+    if (cleaning_thread.joinable()) cleaning_thread.detach();
```
colinschmidt, line 75: Do we need to keep track of the detached threads in anyway? My reading of the documentation is that all resources will be freed once the detached thread exits, so we only need to be sure that `remove_backing_files` eventually completes, which seems reasonable.

JakeSiFive: Correct

V-FEXrt (approved):

PR #1478, opened 2023-11-30
rsc: Implement basic TTL eviction for jobs

V-FEXrt:
Adds a TTL eviction loop that runs every 10 minutes to delete expired jobs.

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -64,13 +64,13 @@ fn make_app(state: Arc<sea_orm::DatabaseConnection>) -> Router {
         )
 }
 
-fn launch_eviction(state: Arc<sea_orm::DatabaseConnection>) {
+fn launch_eviction(state: Arc<sea_orm::DatabaseConnection>, tick_interval: u64, deadline: i64) {
     tokio::spawn(async move {
-        let mut interval = tokio::time::interval(std::time::Duration::from_secs(60 * 10));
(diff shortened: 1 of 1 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -60,16 +64,35 @@ fn make_app(state: Arc<sea_orm::DatabaseConnection>) -> Router {
         )
 }
 
+fn launch_eviction(state: Arc<sea_orm::DatabaseConnection>) {
+    tokio::spawn(async move {
+        let mut interval = tokio::time::interval(std::time::Duration::from_secs(60 * 10));
```
JakeSiFive, line 69: lessons learned, this should be configurable lol

rust/rsc/src/rsc/main.rs
```diff
@@ -60,16 +64,35 @@ fn make_app(state: Arc<sea_orm::DatabaseConnection>) -> Router {
...
+        loop {
+            interval.tick().await;
+
+            let deadline = (Utc::now() - Duration::days(7)).naive_utc();
```
JakeSiFive, line 73: ditto

JakeSiFive (commented):
needs a test now that we're doing that lol

JakeSiFive (approved):
This is landable as is with my approval. Either in this PR or in a follow up one of us should add TTL config to the config type however.

JakeSiFive (approved):
I was thinking about this, lets add this to the config in a follow up and we can add config tests before/during that change. Feels like a better way to break things up.

PR #1480, opened 2023-12-01
Make `stat` actually recognize symlinks

ag-eitilt:
Whose bright idea was it to make the syscall you use to check the type of a path, *not* actually check the type of all paths, and instead hide the uniform behaviour behind a one-letter difference?

Regardless, the `stat` function was previously returning `PathTypeRegularFile` (and permissions/size of the target file) when called on symlinks.  It now returns `PathTypeSymlink` (and less helpful but more accurate permissions/size).


JakeSiFive (approved):
Dope 🤦 

Thanks for fixing this @ag-eitilt

PR #1481, opened 2023-12-01
Export `Stat` tuple accessors

ag-eitilt:
These had always been intended to be exported, but Wake's tuple visibility marking is a bit confusing.


JakeSiFive (approved):
Thanks for fixing this too!

PR #1484, opened 2023-12-07
rsc: Add support for standalone rsc

V-FEXrt:
Running rsc with `--standalone` now launches the cache backed by a in memory sqlite3 database. Useful for development without spinning up a postgres server. 

This feature should only be used for development and not for actual deployment of the cache.

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -5,7 +5,7 @@ use std::io::{Error, ErrorKind};
 use std::sync::Arc;
 use tracing;
 
-use sea_orm::{ColumnTrait, DeleteResult, EntityTrait, QueryFilter};
+use sea_orm::{ActiveModelTrait, ActiveValue::*, ColumnTrait, Database, DatabaseConnection, DeleteResult, EntityTrait, QueryFilter};
 
 use chrono::{Duration, Utc};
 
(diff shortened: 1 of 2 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -38,9 +38,15 @@ struct ServerOptions {
         long
     )]
     database_url: Option<String>,
+
+    #[arg(
+        help = "Launches the cache without an external database",
+        long
+    )]
+    standalone: bool,
```
JakeSiFive, line 46: I don't think this is needed, you could have just specified a sqlite database via the database_url instead

edit: the insecure key part does stand out and it does seem prudent to not generate insecure keys in a way that persists which implies we should tie it to in memory databases.

rust/rsc/src/rsc/main.rs
```diff
@@ -64,7 +70,59 @@ fn make_app(state: Arc<sea_orm::DatabaseConnection>) -> Router {
...
+        );
+        tracing::error! {%err, "unperformed migrations, please apply these migrations before starting rsc"};
+        Err(err)?
+    }
+
+    Ok(connection)
+}
+
+async fn create_insecure_api_key(db: &DatabaseConnection) -> Result<String, Box<dyn std::error::Error>> {
+    let key = data_encoding::BASE64.encode(b"Insecure Key");
```
JakeSiFive, line 98: Why not just make the key "insecurekey" instead? That way if you're writing a curl command you know what to type every time without memorizing some magic string

JakeSiFive (approved):
LGTM, I had some thoughts but this can land as is

rust/rsc/src/rsc/main.rs
V-FEXrt: oh that's just clearly better. Done

JakeSiFive (approved):

PR #1486, opened 2023-12-08
Add basic implementation of blob upload

V-FEXrt:
This change introduces two new routes.

1) `GET /blob` which returns the url that blobs should be POSTed to
2) `POST /blob` which right now accepts a blob and writes it to the location specified by `config.local_store`. 

The POST route isn't complete since the blobs should be tracked in a blob database and a detailed response should be given to the client however that is a sizable chunk of work and so I wanted to split  the review into two parts. This one that handles file upload, and the next one that handle blob tracking.

Future work will be to support multiple backing stores so the routes are being implemented with that in mind.

### DIFF
--- a/rust/rsc/src/rsc/blob.rs
+++ b/rust/rsc/src/rsc/blob.rs
@@ -1,8 +1,9 @@
 use crate::config::RSCConfig;
 use crate::types::GetUploadUrlResponse;
+use async_trait::async_trait;
 use axum::{body::Bytes, extract::Multipart, http::StatusCode, BoxError, Json};
 use data_encoding::BASE64URL;
-use futures::{Stream, TryStreamExt};
+use futures::{Stream, TryStream, TryStreamExt};
 use rand_core::{OsRng, RngCore};
 use sea_orm::DatabaseConnection;
 use std::sync::Arc;
@@ -11,37 +12,40 @@ use tokio::io::BufWriter;
 use tokio_util::io::StreamReader;
 use tracing;
 
+#[async_trait]
+pub trait BlobStore {
+    async fn stream<A, B>(&self, reader: StreamReader<A, B>) -> Result<String, std::io::Error>
+    where
+        StreamReader<A, B>: tokio::io::AsyncRead + std::marker::Send;
+}
+
+pub struct LocalBlobStore {
+    pub root: String,
+}
+
+#[async_trait]
+impl BlobStore for LocalBlobStore {
+    async fn stream<A, B>(&self, reader: StreamReader<A, B>) -> Result<String, std::io::Error>
+    where
+        StreamReader<A, B>: tokio::io::AsyncRead + std::marker::Send,
+    {
+        futures::pin_mut!(reader);
+
+        let name = create_temp_filename();
+        let path = std::path::Path::new(&self.root).join(name.clone());
+        let mut file = BufWriter::new(File::create(path).await?);
+
+        tokio::io::copy(&mut reader, &mut file).await?;
+
+        Ok(name)
+    }
+}
+
 fn create_temp_filename() -> String {
     let mut key = [0u8; 16];
     OsRng.fill_bytes(&mut key);
     // URL must be used as files can't contain /
     BASE64URL.encode(&key)
-}
-
-async fn stream_to_file<S, E>(parent: &str, stream: S) -> Result<String, (StatusCode, String)>
-where
-    S: Stream<Item = Result<Bytes, E>>,
-    E: Into<BoxError>,
-{
-    async {
-        // Convert the stream into an `AsyncRead`.
-        let body_with_io_error =
-            stream.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err));
-        let body_reader = StreamReader::new(body_with_io_error);
-        futures::pin_mut!(body_reader);
-
-        // Create the file. `File` implements `AsyncWrite`.
-        let name = create_temp_filename();
-        let path = std::path::Path::new(parent).join(name.clone());
-        let mut file = BufWriter::new(File::create(path).await?);
-
-        // Copy the body into the file.
-        tokio::io::copy(&mut body_reader, &mut file).await?;
-
-        Ok::<String, std::io::Error>(name)
-    }
-    .await
-    .map_err(|err| (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()))
 }
 
 #[tracing::instrument]
@@ -56,16 +60,20 @@ pub async fn create_blob(
     _conn: Arc<DatabaseConnection>,
     config: Arc<RSCConfig>,
 ) -> (StatusCode, String) {
-    while let Ok(Some(field)) = multipart.next_field().await {
-        let Some(ref local_store) = config.local_store else {
-            return (
-                StatusCode::INTERNAL_SERVER_ERROR,
-                "Invalid configuration".into(),
-            );
-        };
+    let Some(local_store) = config.local_store.clone() else {
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "Invalid configuration".into(),
+        );
+    };
 
-        if let Err(inner) = stream_to_file(local_store, field).await {
-            return inner;
+    let store = LocalBlobStore { root: local_store };
+
+    while let Ok(Some(field)) = multipart.next_field().await {
+        let body = field.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err));
+        match store.stream(StreamReader::new(body)).await {
+            Ok(key) => println!("{:?}", key),
+            Err(msg) => return (StatusCode::INTERNAL_SERVER_ERROR, msg.to_string()),
         }
     }
 
--- a/rust/rsc/src/rsc/blob.rs
+++ b/rust/rsc/src/rsc/blob.rs
@@ -1,9 +1,9 @@
 use crate::config::RSCConfig;
 use crate::types::GetUploadUrlResponse;
 use async_trait::async_trait;
-use axum::{body::Bytes, extract::Multipart, http::StatusCode, BoxError, Json};
+use axum::{extract::Multipart, http::StatusCode, Json};
 use data_encoding::BASE64URL;
-use futures::{Stream, TryStream, TryStreamExt};
+use futures::TryStreamExt;
 use rand_core::{OsRng, RngCore};
 use sea_orm::DatabaseConnection;
 use std::sync::Arc;
--- a/rust/rsc/src/rsc/blob.rs
+++ b/rust/rsc/src/rsc/blob.rs
@@ -1,43 +1,28 @@
 use crate::config::RSCConfig;
 use crate::types::GetUploadUrlResponse;
 use axum::{body::Bytes, extract::Multipart, http::StatusCode, BoxError, Json};
+use data_encoding::BASE64URL;
+use futures::{Stream, TryStreamExt};
+use rand_core::{OsRng, RngCore};
 use sea_orm::DatabaseConnection;
 use std::sync::Arc;
-use tracing;
-
-use futures::{Stream, TryStreamExt};
 use tokio::fs::File;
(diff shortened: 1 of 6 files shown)


rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,71 @@
...
+#[tracing::instrument]
+pub async fn get_upload_url(
+    _conn: Arc<DatabaseConnection>,
+    config: Arc<RSCConfig>,
+) -> Json<GetUploadUrlResponse> {
+    let url = config.server_addr.clone() + "/blob";
+    Json(GetUploadUrlResponse { url })
+}
+
+pub async fn create_blob(mut multipart: Multipart, config: Arc<RSCConfig>) -> hyper::StatusCode {
```
JakeSiFive, line 17: Instead of passing the whole config, just pass the part that matters

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,71 @@
+use crate::config::RSCConfig;
+use crate::types::GetUploadUrlResponse;
+use axum::{extract::Multipart, Json};
+use sea_orm::DatabaseConnection;
+use std::sync::Arc;
+use tracing;
+
+#[tracing::instrument]
+pub async fn get_upload_url(
+    _conn: Arc<DatabaseConnection>,
```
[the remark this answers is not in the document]
JakeSiFive: Also instead of passing the whole config, just pass the part that matters

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+        let name = field.name().unwrap().to_string();
+        let Ok(data) = field.bytes().await else {
+            break false;
+        };
+
+        if name == "hash" {
+            let key = String::from_utf8(data.to_ascii_lowercase()).unwrap();
+            hash = Some(key);
+        }
```
JakeSiFive, line 42: I think the server should be the one doing the hashing, not the client? I can't remember what we decided but my current thinking is that the server should be the one hashing the blob

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        if name == "file" {
+            contents = Some(data);
+        }
```
JakeSiFive, line 46: We can't fit all the data into memory so we need a data stream of some kind. We should do something like this: https://github.com/tokio-rs/axum/blob/main/examples/stream-to-file/src/main.rs

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,71 @@
...
+    };
+
+    let Some(local_store) = config.local_store.clone() else {
+        return hyper::StatusCode::INTERNAL_SERVER_ERROR;
+    };
+
+    let dest = local_store + &hash;
+    let Ok(_) = std::fs::write(dest, contents) else {
+        return hyper::StatusCode::INTERNAL_SERVER_ERROR;
+    };
```
JakeSiFive, line 68: Can we go ahead and make a trait for the storage backend that accepts a stream and then pass a Arc<dyn BlobStore> to the router handler?

rust/rsc/src/rsc/blob.rs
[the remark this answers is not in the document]
V-FEXrt: Next PR will need the connection, and when we move to multiple backing implementations we'd need the whole config anyways. Trying to reduce churn

rust/rsc/src/rsc/blob.rs
V-FEXrt: Sent in a DM yesterday :)

For the MVP I was thinking we should just have the client hash because 1) it already knows the hash 2) we trust all the uploaders 3) I think server rehashing messes with streaming

rust/rsc/src/rsc/blob.rs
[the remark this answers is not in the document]
JakeSiFive: Why will those things be needed?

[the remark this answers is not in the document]
V-FEXrt: 1) we have to write blob info into the database
2) we need the config to know which backing store to even use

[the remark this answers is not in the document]
JakeSiFive: Just pass the backing store object that was constructed from the config

rust/rsc/src/rsc/blob.rs
V-FEXrt: pushing streaming off to later PR per DMs

rust/rsc/src/rsc/blob.rs
V-FEXrt: this thread was mixed up with the other one. For create_blob we actually do need a DB connection and more context about the config so I'm going to leave this in for now. Can circle back to trim it down after we have blob tracking

rust/rsc/src/rsc/blob.rs
```diff
@@ -1,71 +1,96 @@
...
+use futures::{Stream, TryStreamExt};
+use tokio::fs::File;
+use tokio::io::BufWriter;
+use tokio_util::io::StreamReader;
 
-pub async fn create_blob(mut multipart: Multipart, config: Arc<RSCConfig>) -> hyper::StatusCode {
-    let mut hash: Option<String> = None;
-    let mut contents: Option<axum::body::Bytes> = None;
+// to prevent directory traversal attacks we ensure the path consists of exactly one normal
+// component
```
JakeSiFive, line 14: Can you explain this attack to me?

edit: Ah, I see you currently allow the user to name the file. That should not be a part of the protocol at all, the file should be named by the hash and a temporary should be used.

rust/rsc/src/rsc/blob.rs
```diff
@@ -1,71 +1,96 @@
...
+                "Invalid configuration".into(),
+            );
+        };
 
-    let dest = local_store + &hash;
-    let Ok(_) = std::fs::write(dest, contents) else {
-        return hyper::StatusCode::INTERNAL_SERVER_ERROR;
-    };
+        // TODO: instead of writing the user provided field name to disk we should
+        // generate a random temporary name.
```
JakeSiFive, line 89:
```
use rand_core::{RngCore, OsRng};
use data_encoding::BASE64;

let mut key = [0u8; 16];
OsRng.fill_bytes(&mut key);
let dst = local_store + "/" + BASE64.encode(&key);
```

The online hashing part would still need to be done as well of course but I'm happy delaying that. I just don't think we should commit code that uses a user specified location as the destination.

JakeSiFive: I think my priorities are

1. Do not commit code with security flaws in it at all
2. Try to make each PR a "functional" member (happy to slip on this one)
3. Be incremental to keep PR size down

I'm very happy to give ground on 2 to make 1 hold

V-FEXrt: re: 1

Out of self preservation I'll call out that it doesn't actually have a security flaw. It has a _potential_ for a security flaw that is avoided by erroring out if the user gives bad input.

That said, there isn't any reason to just use a random name from the start so I'll go ahead and do that

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,81 @@
...
+use rand_core::{OsRng, RngCore};
+use sea_orm::DatabaseConnection;
+use std::sync::Arc;
+use tokio::fs::File;
+use tokio::io::BufWriter;
+use tokio_util::io::StreamReader;
+use tracing;
+
+#[async_trait]
+pub trait BlobStore {
```
JakeSiFive, line 16: Can you add a TODO for the "returns a URL" task here?

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,81 @@
...
+    Json(GetUploadUrlResponse { url })
+}
+
+#[tracing::instrument]
+pub async fn create_blob(
+    mut multipart: Multipart,
+    _conn: Arc<DatabaseConnection>,
+    config: Arc<RSCConfig>,
+) -> (StatusCode, String) {
+    let Some(local_store) = config.local_store.clone() else {
```
JakeSiFive, line 63: I think you can use `let Some(local_store) = &config.local_store` instead. This will change the type of local_store from an owned string to an &String

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,81 @@
...
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "Invalid configuration".into(),
+        );
+    };
+
+    let store = LocalBlobStore { root: local_store };
```
JakeSiFive, line 70: ...but I guess it doesn't matter because you'd have to clone it here anyways 🤷

rust/rsc/src/rsc/blob.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    while let Ok(Some(field)) = multipart.next_field().await {
+        let body = field.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err));
+        match store.stream(StreamReader::new(body)).await {
+            Ok(key) => println!("{:?}", key),
```
JakeSiFive, line 75: Add a TODO here for something about adding it to the database

rust/rsc/src/rsc/main.rs
```diff
@@ -44,30 +49,51 @@ struct ServerOptions {
...
+            get({
+                let config = config.clone();
+                move || blob::get_upload_url(config.server_addr.clone())
             }),
         )
+        .route(
+            "/blob",
+            post({
+                let conn = conn.clone();
+                let config = config.clone();
```
JakeSiFive, line 92: You don't need to pass the config here, just the blob store

rust/rsc/src/rsc/main.rs
```diff
@@ -44,30 +49,51 @@ struct ServerOptions {
...
         .route(
             "/job/matching",
             post({
-                let shared_state = state.clone();
-                move |body| read_job::read_job(body, shared_state)
+                let conn = conn.clone();
+                move |body| read_job::read_job(body, conn)
+            }),
+        )
+        .route(
```
JakeSiFive, line 81: Add a todo here about using the blob store for the url retrieval.

JakeSiFive (commented):
I'm very happy with this now, I have a few TODO comments and one last comment about constructing the blob store when creating the route from the config

rust/rsc/src/rsc/blob.rs
V-FEXrt: yep, I did the ref thing originally :)

rust/rsc/src/rsc/main.rs
JakeSiFive: @V-FEXrt This is my only blocking comment and its not a hard blocker. I just want use to construct a *single* blob store that gets passed to multiple routes on demand (using an Arc and cloning)

V-FEXrt: How should I return an http error when the configuration is wrong?

JakeSiFive (approved):
ok approved with TODOs added. We spoke offline and its a bit tricky to get the lifetimes and traits all happy for everything right now for the backend. We can play with the interface of BlobStore later to make it object safe or find some other mechnism for dynamic dispatch.

rust/rsc/src/rsc/main.rs
V-FEXrt: nvm I have a solution I'm happy with

rust/rsc/src/rsc/main.rs
V-FEXrt: The upload url is not really tied to the blob store. At least for as long as we do the double upload. We mostly kept this route as a safety net for ourselves but tbh I think each blob store will be so different that it won't ever make since to change this.

V-FEXrt:
Going to go ahead and merge this now, feel free to look at the latest commit and suggest changes I can grab in a follow up

PR #1487, opened 2023-12-11
[LSC] Aquire daemon lock before opening db

JakeSiFive:
Previously we (unintentionally) would open the database BEFORE inquiring the cache lock. The vast majority of the time this is harmless because we connect to the database with randomized exponential backoff and sqlite3 does a good enough job at this with this back off strategy. Later once the cache fails to aquire the cache lock the connection is closed and no harm was done. However it can get overwhelmed and cause issues. The solution is simple, just move the lock acquire to be first like it should have always been.


colinschmidt (approved):
LGTM

PR #1489, opened 2023-12-11
Poll shared cache connect

JakeSiFive:
I'll start out by saying, I'm not a fan of this PR but I was just noodling on what we could do to help with this problem if it occurs in the future. Ideally we'd having a polling mechanism here but at this moment I'm not at all sure how to do that. It appears to me that Tokio and mio have solved this problem somehow so I think there's a better solution out there. This change does not eliminate the core issue like the previous PR did and because it at best reduces spurious daemon launches by 2x it doesn't eliminate DoS-ing yourself by spamming wake invocations, it just reduces the impact a little bit...I'm gonna put this up and see what people think but I'm rather on the fence about landing it.

### DIFF
--- a/src/job_cache/job_cache.cpp
+++ b/src/job_cache/job_cache.cpp
@@ -134,7 +134,7 @@ static bool daemonize(std::string dir) {
   return true;
 }
 
-wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
+wcl::optional<wcl::unique_fd> try_connect(std::string dir, bool& addr_bound) {
   wcl::unique_fd socket_fd;
   {
     int local_socket_fd = socket(AF_UNIX, SOCK_STREAM | SOCK_NONBLOCK | SOCK_CLOEXEC, 0);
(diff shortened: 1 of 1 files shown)


src/job_cache/job_cache.cpp
```diff
@@ -167,9 +167,15 @@ wcl::optional<wcl::unique_fd> try_connect(std::string dir) {
...
   addr.sun_path[0] = '\0';
 
+  // Among the failures that can occur here, we could get EAGAIN. We can't use
+  // epoll to avoid this however if we get EAGAIN it means that we don't have
+  // to launch another process because the address is already bound.
   wcl::log::info("key = %s, sizeof(key) = %lu", key, sizeof(key))();
   memcpy(addr.sun_path + 1, key, sizeof(key));
   if (connect(socket_fd.get(), reinterpret_cast<const sockaddr *>(&addr), sizeof(key)) == -1) {
+    if (errno == EAGAIN) {
+      addr_bound = true;
```
V-FEXrt, line 177: The logic around the `bool&` here is tricky and hard to following. You should just change the return type to `wcl::result` with an error enum of `enum { Generic, AddrBound }` which will make this function clearer but more importantly will make the callsite several times better

JakeSiFive: Yeah that does seem better. I was thinking of doing something like that but I got lazy lol

JakeSiFive:
> I don't hate the overall structure and it does seem reasonable to me to limit daemon launches if we already know its launched. My vote would be to land if we can clean up the logic a little bit

Yeah that makes sense, let me investigate how tokio and mio solve this first before going this route. If we do go this route then using a result to clean up the logic is preferable no doubt.

JakeSiFive:
I was thinking about this and my initial logic on this was wrong. I *thought* it would flip back and forth between launching and not launching but it actually should re-test for the connection being valid every time so it should keep not re-launching by continuously reconfirming that the other end of the socket is still bound. So this is actually quite a good strategy for reducing log churn, process churn, and should help reduce contention so I'm on board with this. I'll clean this up on Wednesday and we can land it.

PR #1490, opened 2023-12-13
rsc: Create tables for tracking Blobs

V-FEXrt:
**Note:** Due to sqlite3 limitations we had to rewrite migration history to pass our tests. After initial deployment this becomes much more difficult. We are researching if we can avoid sqlite3 for our test flow.

Creates the tables and relationships required for tracking uploaded blobs. Minimally updates routes and tests so they will pass. Updating the routes to work as expected will come in a future PR


rust/migration/src/m20231213_194935_create_blob_table.rs
```diff
@@ -0,0 +1,141 @@
...
+            .to_owned();
+
+        manager.exec_stmt(insert).await?;
+
+        manager
+            .create_table(
+                Table::create()
+                    .table(Blob::Table)
+                    .col(ColumnDef::new(Blob::Id).string().not_null().primary_key())
+                    .col(ColumnDef::new(Blob::Store).integer().not_null())
```
JakeSiFive, line 43: I think you probably want a blob url here, not just the store type. Just knowing the store type doesn't help. You need to know the store type and configuration details of the store to know anything useful. I'd argue we should either drop this or make it a full URL

rust/migration/src/m20231213_194935_create_blob_table.rs
```diff
@@ -0,0 +1,141 @@
...
+                    .col(ColumnDef::new(Blob::WorkspacePath).string().not_null())
```
JakeSiFive, line 44: What is a WorkspacePath?

rust/migration/src/m20231213_194935_create_blob_table.rs
```diff
@@ -0,0 +1,141 @@
...
+                            .to_col(BlobStore::Id)
+                            // All blobs for a given store must be manually
+                            // deleted before the blob store may be deleted.
+                            .on_delete(ForeignKeyAction::Restrict),
+                    )
+                    .to_owned(),
+            )
+            .await?;
+
+        manager
```
JakeSiFive, line 60: You're creating a new table here but I don't think that's what you want, we already have a table that stores output files for this purpose.

rust/migration/src/m20220101_000001_create_blob_tables.rs
```diff
@@ -0,0 +1,128 @@
...
+                    .foreign_key(
+                        ForeignKeyCreateStatement::new()
+                            .name("fk-store_id-blob_store")
+                            .from_tbl(Blob::Table)
+                            .from_col(Blob::StoreId)
+                            .to_tbl(BlobStore::Table)
+                            .to_col(BlobStore::Id)
+                            // All blobs for a given store must be manually
+                            // deleted before the blob store may be deleted.
+                            .on_delete(ForeignKeyAction::Restrict),
```
JakeSiFive, line 56: What are your thoughts on restrict vs cascade? What about specifically for blob store?

rust/migration/src/m20220101_000002_create_table.rs
```diff
@@ -36,14 +36,36 @@ impl MigrationTrait for Migration {
...
+                    .foreign_key(
+                        ForeignKeyCreateStatement::new()
+                            .name("fk-stdout_id-blob")
+                            .from_tbl(Job::Table)
+                            .from_col(Job::StdoutId)
+                            .to_tbl(Blob::Table)
+                            .to_col(Blob::Id)
+                            // A blob may not be deleted if there exists a
+                            // job that depends on it.
+                            .on_delete(ForeignKeyAction::Restrict),
```
JakeSiFive, line 56: These in particular seem like they might be too strong? Sometimes a blob will go missing from storage and if that happens we want to delete the blob from the database, but also delete any jobs that depend on it. I think deleting stores is often not needed so I'm less concerned with that but deleting blobs seems to be more important and common.

rust/migration/src/m20220101_000002_create_table.rs
```diff
@@ -101,6 +123,17 @@ impl MigrationTrait for Migration {
...
+                    .foreign_key(
+                        ForeignKeyCreateStatement::new()
+                            .name("fk-blob_id-blob")
+                            .from_tbl(OutputFile::Table)
+                            .from_col(OutputFile::BlobId)
+                            .to_tbl(Blob::Table)
+                            .to_col(Blob::Id)
+                            // A blob may not be deleted if there exists an
+                            // output file that depends on it.
+                            .on_delete(ForeignKeyAction::Restrict),
```
JakeSiFive, line 135: ditto, for instance when we delete a job, we cascade the deletion to the output file

rust/rsc/src/rsc/add_job.rs
```diff
@@ -29,8 +29,8 @@ pub async fn add_job(
         stdin: Set(payload.stdin),
         is_atty: Set(payload.is_atty),
         hidden_info: Set(payload.hidden_info),
-        stdout: Set(payload.stdout),
-        stderr: Set(payload.stderr),
+        stdout_id: Set(payload.stdout_id),
+        stderr_id: Set(payload.stderr_id),
```
JakeSiFive, line 33: minor nit: Both here and above in the DB, I think "stdfoo_blob_id" or "stdfoo_blob_key" or something like that would be better. Just seeing "stdout_id" doesn't really tell me whats happening here.

rust/rsc/src/rsc/main.rs
```diff
@@ -389,8 +403,8 @@ mod tests {
                 "output_dirs": [],
                 "output_symlinks": [],
                 "output_files":[],
-                "stdout":[116, 104, 105, 115, 32, 105, 115, 32, 97, 32, 116, 101, 115, 116],
-                "stderr":[116, 104, 105, 115, 32, 105, 115, 32, 97, 32, 118, 101, 114, 121, 32, 108, 111, 110, 103, 32, 115, 116, 114, 105, 110, 103, 32, 102, 111, 114, 32, 97, 32, 116, 101, 115, 116],
+                "stdout_id": blob_id,
+                "stderr_id": blob_id,
```
JakeSiFive, line 407: oh neat upshot!

JakeSiFive (approved):
Approved. I suspect some of the on delete resitrcts should be cascade instead most likely

rust/migration/src/m20220101_000001_create_blob_tables.rs
V-FEXrt: In the comments I gave my core justification. But in general cascade is pretty dangerous.

In this case, if you deleted a BlobStore, it would cascade to the Blobs, which would cascade to the OutputFiles, which would cascade to the Jobs.

To me that really isn't acceptable especially since we have a clear workflow for deleting all those things 

1) Job eviction deletes a Job which cascades to its OutputFiles 
2) Blob eviction later sees those blobs with no references so it triggers proper deletion using the backing blob store API then deletes the entry in the blob table
3) We can then manually remove the decommissioned BlobStore

If we make it cascade we will be guaranteeing that blobs get orphaned in the remote store without really gaining anythign in the API

rust/migration/src/m20220101_000002_create_table.rs
V-FEXrt: If a blob goes missing from storage you should just delete the whole job. Then you can either delete the blob manually or let blob eviction clean up the entry

V-FEXrt: In fact, I claim you should _always_ delete the job first. That way you stop the bleeding with clients. Deleting the blob first makes the bleeding worse for a short amount of time

JakeSiFive: I agree you should delete the job but there may be many jobs you need to delete not just one

V-FEXrt: If this is a case we see happening a lot we should write code to handle the cleanup (we can use rsc_tool or the http api to delete all jobs that depend on blob x) instead of having the database work against the guarantees we want

V-FEXrt: I cascade wouldn't be as bad if we have some way to ensure we cleanup on the remote stores with a blob is deleted. I'm not sure how we'd do that but if you have any ideas I'm open to hearing them

JakeSiFive: The way I've been doing it in the local shared cache is that you scan both ways because you shouldn't trust either side. One scan checks for non-existant blobs that are referenced in the DB, the other scan checks for blobs that exist but are orphaned. This appears to be a required step because you can't atomically delete the blob on one side and the blob in the database. One must be deleted before the other so one of the scan directions is required. In practice I've found that both are helpful.

JakeSiFive: Side note: We have been using on delete cascade, if you want to change that lets change it everywhere you think it makes sense while we're changing the DB.

I'm fine with it either way, using cascade only makes the delete *Easier* it doesn't actully buy us any kind of assurance so I'm quite happy to keep restrict over cascade. But I think the assurances it buys us are week (but certainly greater than the absence of assurance we get from cascade).

I will warn against thinking that this protects you from the blob store and the database falling out of sync, I do not think it accomplishes that. Distributed systems are hard.

JakeSiFive: Yeah ok, I think I've landed on liking restrict with the following reasoning:

1) It corrects for some mistakes
2) We still need at least one direction of out of sync checks if not both but that's ok (its not a + to either side)
3) We can always do the cascade deletes manually if we need to

JakeSiFive (approved):

PR #1491, opened 2023-12-14
Use Arc dyn BlobStore insteadl of impl

JakeSiFive:
This changes the BlobStore trait to be object safe while maintaining most of the generality of the previous implementation.

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -61,7 +61,9 @@ struct ServerOptions {
 fn create_router(conn: Arc<DatabaseConnection>, config: Arc<config::RSCConfig>) -> Router {
     // If we can't create a store, just exit. The config is wrong and must be rectified.
     let root = config.local_store.clone().unwrap();
-    let store = blob::LocalBlobStore { root };
+    //let store = blob::LocalBlobStore { root };
+    let store: Arc<dyn blob::DebugBlobStore + Send + Sync> =
+        Arc::new(blob::LocalBlobStore { root });
     Router::new()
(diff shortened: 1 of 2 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -61,7 +61,9 @@ struct ServerOptions {
 fn create_router(conn: Arc<DatabaseConnection>, config: Arc<config::RSCConfig>) -> Router {
     // If we can't create a store, just exit. The config is wrong and must be rectified.
     let root = config.local_store.clone().unwrap();
-    let store = blob::LocalBlobStore { root };
+    //let store = blob::LocalBlobStore { root };
```
V-FEXrt, line 64: commented code should be removed

V-FEXrt (approved):
Approving with the musing that this feels a lot like using Rust magic to go out of the way to avoid a copy when a copy would have been just fine

JakeSiFive:
> Approving with the musing that this feels a lot like using Rust magic to go out of the way to avoid a copy when a copy would have been just fine

copies don't share state and thus a) can mutate when shared across threads and b) can't update when they need to. The point is NOT to avoid the copy, this is more expensive than just doing a copy, the point is to have a single source of truth.

JakeSiFive:
Also the core core core issue that you seem to keep missing is that you CANNOT dynmaically dispatch based on the config across an `impl` so if you'd have to either duplicate the routing code or have other config code that uses pattern matching in order to call the correct monomorphization of the router creation code. This avoids having the issue of dispatching to different monomorphizations

PR #1493, opened 2023-12-16
rsc: Basic attempt at uuid as id

V-FEXrt:
This change is updates all models/tables to use UUIDs. This is important for concurrently/db destruction/multi-client as we can't guarentee that two different clients won't ever get the same integer id representing 2 different jobs. With this change, even if bad db stuff happens we ensure running clients don't confuse jobs as every new job will always get a unique id regardless of the previous state of the db. 

It properly generates new UUIDs by default but it takes a hard dependence on Postgres. To land this PR we would have to drop sqlite3 support which has now been completed.

This change rewrites history but we just did that in the last PR so might as well get them all in at once

The standalone tests have been updated and now properly use a sandboxed postgres

### DIFF
--- a/rust/migration/src/m20220101_000001_create_blob_tables.rs
+++ b/rust/migration/src/m20220101_000001_create_blob_tables.rs
@@ -12,10 +12,10 @@ impl MigrationTrait for Migration {
                     .table(BlobStore::Table)
                     .col(
                         ColumnDef::new(BlobStore::Id)
-                            .integer()
+                            .uuid()
                             .not_null()
-                            .auto_increment()
-                            .primary_key(),
+                            .primary_key()
+                            .default(SimpleExpr::FunctionCall(PgFunc::gen_random_uuid())),
                     )
                     .col(ColumnDef::new(BlobStore::Type).string().not_null())
                     .to_owned(),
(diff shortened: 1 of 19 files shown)


rust/migration/src/m20220101_000001_create_blob_tables.rs
```diff
@@ -65,7 +65,7 @@ impl MigrationTrait for Migration {
                     .table(LocalBlobStore::Table)
                     .col(
                         ColumnDef::new(LocalBlobStore::Id)
-                            .integer()
+                            .uuid()
                             .not_null()
                             .primary_key(),
```
colinschmidt, line 70: Does this need the default call pointing to `gen_random_uuid`?

colinschmidt (commented):
By drop sqlite support you mean "drop sqlite support for rsc" right?
If so I think this seems like a clean change, modulo my comment.

rust/migration/src/m20220101_000001_create_blob_tables.rs
V-FEXrt: Nice call out, in this case no but only for a sneaky reason. 

LocalBlobStore is a "dependent table" of the BlobStore (kind of like a subclass) and its ID is a foreign key to BlobStore's ID. That means the ID must always be manually set to and already existing BlobStore ID

V-FEXrt:
> By drop sqlite support you mean "drop sqlite support for rsc" right?

Yeah lol, changing that for wake would be a very large project

colinschmidt (approved):
LGTM

PR #1494, opened 2023-12-18
Do not use cache after the failure cap is reached

JakeSiFive:
This change just makes it so that no attempt is made whatsoever after the failure cap is reached, this includes add.

As a toss in I also fixed add not using the config.

### DIFF
--- a/src/job_cache/job_cache.cpp
+++ b/src/job_cache/job_cache.cpp
@@ -324,7 +324,7 @@ static int misses_from_failure = 0;
 
 FindJobResponse Cache::read(const FindJobRequest &find_request) {
   if (misses_from_failure > timeout_config.max_misses_from_failure) {
-    return FindJobResponse(wcl::optional<MatchingJob>{});  
+    return FindJobResponse(wcl::optional<MatchingJob>{});
   }
 
   wcl::log::info("Cache::read enter")();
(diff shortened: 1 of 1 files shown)


src/job_cache/job_cache.cpp
```diff
@@ -320,10 +320,15 @@ wcl::result<FindJobResponse, FindJobError> Cache::read_impl(const FindJobRequest
   return wcl::result_value<FindJobError>(FindJobResponse(json));
 }
 
+static int misses_from_failure = 0;
+
 FindJobResponse Cache::read(const FindJobRequest &find_request) {
+  if (misses_from_failure > timeout_config.max_misses_from_failure) {
```
colinschmidt, line 326: Are the two new guards supposed to also include `miss_on_failure`? I guess if `miss_on_failure` is not set we wouldn't expect `misses_from_failure` to be non-zero and therefore these guards should never trigger, so its just a style question not a functionality quesiton.

colinschmidt (approved):
LGTM

src/job_cache/job_cache.cpp
JakeSiFive: Yeah its never going to be reached. We only use `miss_on_failure` to determine if the miss is a failure or not

PR #1498, opened 2023-12-27
Allow remote shared cache testing from a network isolated sandbox

JakeSiFive:
This is still very much a draft but it displays the basic components needed to get this idea to work.

Issues:
1) Each test has to use a distinct database name manually right now, it would be nice if these were auto selected
2) database names must be created prior to executing, it would be nice if the databases were created on the fly somehow
3) Managing the workflow with vendoring is a bit odd right now, it would be nice if vendored dep settings could be bind mounted into place instead
4) A non-rustup version of cargo has to be used to build this. I managed to get this working locally via a local install of a non-rustup toolchain. This might get a bit annoying if we don't have a way to download the version of cargo/rust that you're meant to build with
5) Right now things are just in the state I first got them working in, it would be nice if we could have a generic mechanism for running RSC commands in these environments


postgres-cmd-spec.json
```diff
@@ -0,0 +1,29 @@
...
+  "group-id": 0,
+  "visible": [],
+  "directory": ".",
+  "stdin": "",
+  "resources": [],
+  "isolate-network": true,
+  "isolate-pids": true,
+  "mount-ops": [
+    {"type": "tmpfs", "destination": "/tmp"},
+    {"type": "tmpfs", "destination": "/root"}
```
JakeSiFive, line 19: I don't think I need this anymore

postgres-cmd-spec.json
```diff
@@ -0,0 +1,29 @@
+{
+  "label": "postgres command",
+  "environment": [
+    "PATH=/usr/lib64/ccache:/usr/bin:/usr/sbin:/bin:/sifive/tools/rust/stable/1.72.0/bin",
```
JakeSiFive, line 4: This needs to be added somehow. I think a good way to go about this would be to make a new kind of wakebox runner so we can run arbitrary commands with whatever environment we pick.

postgres-cmd.sh
```diff
@@ -0,0 +1,15 @@
+set -ex
+echo $SHELL
+echo $USER
+mkdir $DATA_DIR
```
JakeSiFive, line 4: DATA_DIR can be fixed to `/tmp/pg_data` here since `/tmp` is a fresh tmpfs in the sandbox's mount namespace

rust/rsc/src/rsc/main.rs
```diff
@@ -101,8 +101,8 @@ fn create_router(conn: Arc<DatabaseConnection>, config: Arc<config::RSCConfig>)
         )
 }
 
-async fn create_standalone_db() -> Result<DatabaseConnection, sea_orm::DbErr> {
-    let db = Database::connect("sqlite::memory:").await?;
+async fn create_standalone_db(db: &str) -> Result<DatabaseConnection, sea_orm::DbErr> {
```
JakeSiFive, line 104: It would be nicer to generate unique database names on the fly, this will also require creating those databases however but I think that should be manageable from SeaORM. I think we can get away with just a single database existing that we can connect to, use to make a new database, disconnect, and reconnect to our generated database. If that database name is a UUID it should be unique.

build.wake
```diff
@@ -17,6 +17,38 @@ package build_wake
 
 from wake import _
 
+
+export def testPostgres postgresDir cmd =
```
JakeSiFive, line 21: TODO: Encapsulate some of the wierd logic here as a runner

postgres-cmd-spec.json
JakeSiFive: It turns out that I do in fact need an extra temp home directory somewhere because some tools like ccache will add things to it.

scripts/which_wake_files
```diff
@@ -16,7 +16,7 @@ if [[ ${1} == "all" ]]; then
 fi
 
 if [[ ${1} == "changed" ]]; then
-  CHANGED_FILES=$(git --git-dir ${WAKE_ROOT}.git diff --name-only | grep '\.wake' | grep -v 'tests/' | grep -v "manifest.wake")
+  CHANGED_FILES=$(git --git-dir ${WAKE_ROOT}.git diff --name-only | grep '\.wake$' | grep -v 'tests/' | grep -v "manifest.wake")
   STAGED_FILES=$(git --git-dir ${WAKE_ROOT}.git diff --name-only --cached | grep '\.wake' | grep -v 'tests/' | grep -v "manifest.wake")
```
V-FEXrt, line 20: `STAGED_FILES` also needs this change

JakeSiFive:
ok almost done, I'm gonna get these tests running in CI as the last step and then I'll take this off of draft. Go ahead and review the current form of it now because I only plan on adding code, not making modifications outside of review requests now.

rust/rsc/postgres-tests.wake
```diff
@@ -0,0 +1,89 @@
...
+        Some dir -> Pass "{dir}/bin"
+        None ->
+            require Some osPath = getenv "PATH"
+            else failWithError "could not find postgres"
+
+            def postgresPath = whichIn osPath "postgres"
+
+            if postgresPath ==* "postgres" then
+                failWithError "could not find postgres on path"
+            else
```
V-FEXrt, line 36: You should use `whichInEnvPath` here

rust/rsc/src/rsc/main.rs
```diff
@@ -102,7 +104,16 @@ fn create_router(conn: Arc<DatabaseConnection>, config: Arc<config::RSCConfig>)
...
-    let db = Database::connect("sqlite::memory:").await?;
+    let shim_db = Database::connect("postgres://127.0.0.1/shim").await?;
+    let mut buf = [0u8; 24];
+    OsRng.fill_bytes(&mut buf);
+    let rng_str = HEXLOWER.encode(&buf);
+    let db = format!("db_{}", rng_str);
+    shim_db
+        .execute_unprepared(&format!("CREATE DATABASE {}", db))
+        .await?;
+    drop(shim_db);
```
V-FEXrt, line 115: Can you explain what `drop` is doing here/why it is needed?

JakeSiFive: I don't think its needed but it causes the shim_db to be droped before we create a new one so that we don't have two connections at once.

It also might help enforce that the follow up connection uses the same view version of the databases as the one we just created (e.g. drop might wait for transactions to be committed but I'm not too sure about that and it shouldn't be needed strictly here, I think the await is enough to know that the commit has happened and any new connections will see the new database.

rust/rsc/postgres-tests.wake
```diff
@@ -0,0 +1,84 @@
...
+export target testPostgres Unit =
+    # First we define a few directories, using various
+    # environment variables
+    def dataDir = "/tmp/pg_data"
+
+    require Pass postgresDir = match (getenv "POSTGRES_DIR")
+        Some dir -> Pass "{dir}/bin"
+        None ->
+            require Some postgresPath = whichInEnvPath "postgres"
+            else failWithError "could not find postges on path"
```
V-FEXrt, line 30: typo: `postges` -> `postgres`

V-FEXrt (approved):

Makefile
```diff
@@ -80,6 +80,9 @@ test:		wake.db
 unittest:	all
 	$(WAKE_ENV) ./bin/wake --in test_wake runUnitTests
 
+remoteCacheTests:	all
+	$(WAKE_ENV) ./bin/wake -d -x 'testPostgres Unit'
```
V-FEXrt, line 84: nit: Do we need `-d` here? Also should `testPostgres` be in `test_wake` like the other tests?

JakeSiFive: -d It makes it a million times

Also I'm not going to mess with `test_wake` stuff right now since thats its own bag of worms

.github/workflows/build.yaml
```diff
@@ -18,8 +18,17 @@ jobs:
...
+
+        - name: Build postgres
+          run: cd /var/tmp/postgres && ./configure --prefix /var/tmp/pg-server && make install
+
         - name: Install Deps
-          run: sudo apt-get update && sudo apt-get install -y build-essential fuse libfuse-dev libsqlite3-dev libgmp-dev libncurses5-dev pkg-config git g++ gcc libre2-dev python3-sphinx clang-format-12
+          run: sudo apt-get update && sudo apt-get install -y build-essential fuse libfuse-dev libsqlite3-dev libgmp-dev libncurses5-dev pkg-config git g++ gcc libre2-dev python3-sphinx clang-format-12 postgresql
+
+        - name: Run Remote Cache Tests
+          run: POSTGRES_DIR=/var/tmp/pg-server make remoteCacheTests
```
V-FEXrt, line 31: doesn't `cargo vendor` need to happen before this?

JakeSiFive: yes

.github/workflows/dockerfiles/alpine
```diff
@@ -1,6 +1,6 @@
 FROM alpine:3.14.0
 
-RUN apk add m4 g++ make pkgconf git tar xz gmp-dev re2-dev sqlite-dev fuse-dev ncurses-dev dash sqlite-static ncurses-static linux-headers jq
+RUN apk add m4 g++ make pkgconf git tar xz gmp-dev re2-dev sqlite-dev fuse-dev ncurses-dev dash sqlite-static ncurses-static linux-headers jq postgresql
```
V-FEXrt, line 3: Since we are building manually all these should be deleted

rust/rsc/postgres-cmd-spec.json
```diff
@@ -0,0 +1,43 @@
+{
```
V-FEXrt, line 1: Can this be deleted if we are generated the spec file?

.github/workflows/dockerfiles/alpine
JakeSiFive: I think a lot of them are needed, just not the postgres dep

rust/rsc/postgres-cmd-spec.json
JakeSiFive: yep, I just haven't gotten around to it, hold your horses on clean up comments.

.github/workflows/dockerfiles/alpine
V-FEXrt: oh that is what I meant lol

V-FEXrt: "all these" -> postgres from all these different files not the entire dep list

.github/workflows/build.yaml
```diff
@@ -33,27 +48,9 @@ jobs:
         - name: Build postgres
           run: pushd /var/tmp/postgres && ./configure --prefix /var/tmp/pg-server && make install && popd
 
-        - name: Install Deps
-          run: sudo apt-get update && sudo apt-get install -y build-essential fuse libfuse-dev libsqlite3-dev libgmp-dev libncurses5-dev pkg-config git g++ gcc libre2-dev python3-sphinx clang-format-12 postgresql
-
-        - name: Vendor remote cache deps
-          run: cd rust/rsc && cargo vendor --locked
```
V-FEXrt: looks like you deleted too much

PR #1502, opened 2024-01-16
rsc: Resolve Blobs on request

V-FEXrt:
Blobs returned to the client via a matching jobs now resolve and return the URL where that blob can be downloaded from.

Example:
```json
{
    "cputime": 1.0,
    "ibytes": 100000,
    "memory": 1000,
    "obytes": 1000,
    "output_dirs": [],
    "output_files": [
        {
            "blob": {
                "id": "4ad74b40-4a3f-4464-a488-77f6297c8121",
                "url": "file:///foo/bar/remote_cache_local_store/57KcxXl4PxMyjPcZIRJvDw=="
            },
            "mode": 0,
            "path": "foo"
        }
    ],
    "output_symlinks": [],
    "runtime": 1.0,
    "status": 0,
    "stderr_blob": {
        "id": "5bfb8a97-c244-4241-a091-5aa3388c20eb",
        "url": "file:///foo/bar/remote_cache_local_store/X1MS-Semg55oqRF3U9LAYA=="
    },
    "stdout_blob": {
        "id": "4ad74b40-4a3f-4464-a488-77f6297c8121",
        "url": "file:///foo/bar/remote_cache_local_store/57KcxXl4PxMyjPcZIRJvDw=="
    },
    "type": "Match"
}
```

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -7,13 +7,14 @@ use clap::Parser;
 use data_encoding::HEXLOWER;
 use migration::{Migrator, MigratorTrait};
 use rand_core::{OsRng, RngCore};
+use std::collections::HashMap;
 use std::io::{Error, ErrorKind};
 use std::sync::Arc;
 use tracing;
 
 use sea_orm::{
-    ActiveModelTrait, ActiveValue::*, ColumnTrait, ConnectionTrait, Database, DatabaseConnection,
-    DeleteResult, EntityTrait, QueryFilter,
+    prelude::Uuid, ActiveModelTrait, ActiveValue::*, ColumnTrait, ConnectionTrait, Database,
+    DatabaseConnection, DeleteResult, EntityTrait, QueryFilter,
 };
 
 use chrono::{Duration, Utc};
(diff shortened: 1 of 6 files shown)


rust/rsc/src/rsc/read_job.rs
```diff
@@ -44,12 +75,29 @@ pub async fn read_job(
...
+                    });
+
+                let output_files: Result<Vec<ResolvedBlobFile>, String> =
+                    futures::future::join_all(output_files)
+                        .await
+                        .into_iter()
+                        .collect();
+
+                let output_files = match output_files {
+                    Err(_) => return Ok((None, ReadJobResponse::NoMatch)),
```
colinschmidt, line 98: Do we really want to have the same response for a failure to find blobs of a known job as we do for not finding a job?
I think from the client perspective it makes sense that in neither of these cases can we use the RSC for this job, but in the bad blob case there is no trail on the client that anything went wrong even though any number of things could have gone wrong AFAICT.

rust/rsc/src/rsc/read_job.rs
```diff
@@ -79,8 +127,12 @@ pub async fn read_job(
...
                         output_dirs,
                         output_files,
-                        stdout_blob_id: matching_job.stdout_blob_id,
-                        stderr_blob_id: matching_job.stderr_blob_id,
+                        stdout_blob: resolve_blob(matching_job.stdout_blob_id, txn, &blob_stores)
+                            .await
+                            .unwrap(),
+                        stderr_blob: resolve_blob(matching_job.stderr_blob_id, txn, &blob_stores)
+                            .await
+                            .unwrap(),
```
colinschmidt, line 135: What happens if these blobs aren't resolvable? It seems like we will still return a match even though we don't have all the data needed.

rust/rsc/src/rsc/main.rs
```diff
@@ -52,20 +53,49 @@ struct ServerOptions {
...
+        Err(_) => Vec::new(),
+    };
+
+    let mut active_stores: HashMap<Uuid, Arc<dyn blob::DebugBlobStore + Sync + Send>> =
+        HashMap::new();
+
+    for store in stores.into_iter() {
+        active_stores.insert(
+            store.id,
+            Arc::new(blob::LocalBlobStore { root: store.root }),
```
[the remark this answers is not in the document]
V-FEXrt: LocalBloStore is a different type/"service provider"

The LocalBlobStore in the database is just tracking data needed to construct the "runtime" LocalBlobStore which actually implements the api/trait to upload/resolve/delete a blob

rust/rsc/src/rsc/main.rs
```diff
@@ -52,20 +53,49 @@ struct ServerOptions {
...
 
-    #[arg(
-        help = "Specify an absolute path for a local directory based store",
-        value_name = "LOCAL_STORE",
-        long
-    )]
-    local_store: Option<String>,
+// Maps databse store uuid -> dyn blob::DebugBlobStore
+// that represent said store.
+async fn activate_stores(
```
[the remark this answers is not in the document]
V-FEXrt: This kind of ties in to the question below. Its not quite just collecting data, it also transforms the database model into the object responsible for doing things. The reason for "activate" was because we had to upgrade the data to a service to make it usable. Otherwise it would've just been called fetch or get.

One thing that may be important to note, this function will also be where other non-local stores are constructed/upgraded in the future

rust/rsc/src/rsc/read_job.rs
V-FEXrt: I agree this response isn't super great. This was me just not wanting to massage rust to be happy here. I'll see what I can do. Mainly I still want to return NoMatch, but maybe I can set a status code or something so that we get an idea that something went wrong

rust/rsc/src/rsc/read_job.rs
V-FEXrt: ah lol no, its actually much worse. `unwrap()` crashes the server on `Err` honestly we should probably just have a "no unwrap ever" policy in the RSC. I was just ignoring the error here, don't remember why, probably because its more code to handle it. I'll fix it

rust/rsc/src/rsc/read_job.rs
colinschmidt: In scenarios where we are debugging an RSC read failure, will there be any log/record of the `Err` that is hidden on this line?

V-FEXrt: I can add a log super easily

rust/rsc/src/rsc/read_job.rs
```diff
@@ -121,18 +123,30 @@ pub async fn read_job(
...
+                    Err(err) => {
+                        tracing::error! {%err, "Failed to resolve stdout blob. Resoving job as a cache miss."};
+                        return Ok((None, ReadJobResponse::NoMatch))
+                    },
+                    Ok(blob) => blob,
+                };
+
+                let stderr_blob = match resolve_blob(matching_job.stderr_blob_id, txn, &blob_stores).await {
+                    Err(err) => {
+                        tracing::error! {%err, "Failed to resolve stderr blob. Resoving job as a cache miss."};
```
colinschmidt, line 136: FYI: You made a typo in "Resolving" in all of these error messages.

colinschmidt (approved):
LGTM

PR #1503, opened 2024-01-18
rsc: Implement blob eviction

V-FEXrt:
Implements blob eviction loop for RSC. Depends on blob resolution PR landing first.

### DIFF
--- a/rust/rsc/src/rsc/blob_store_service.rs
+++ b/rust/rsc/src/rsc/blob_store_service.rs
@@ -26,9 +26,15 @@ pub async fn fetch_unreferenced_blobs(
     db: &DatabaseConnection,
     ttl: NaiveDateTime,
 ) -> Result<Vec<blob::Model>, DbErr> {
-    // TODO: this can probably be written ala
+    // TODO: this can probably be refactored as a ORM query with something like
     // Blob::find()
-    //     .filter(blob::Column::Id.not_in_subquery(sea_orm::query::Query::select().columns([OutputFile::BlobId]).from(OutputFile::Table));
+    //     .filter(
+    //         blob::Column::Id.not_in_subquery(
+    //             sea_orm::query::Query::select()
+    //                 .column(OutputFile::BlobId)
+    //                 .from(OutputFile::Table
+    //                 .union( *OTHER STUFF* ));
+    //                 .take().
 
     Blob::find()
         .from_raw_sql(Statement::from_sql_and_values(
@@ -36,9 +42,9 @@ pub async fn fetch_unreferenced_blobs(
             r#"
             SELECT * FROM blob
             WHERE created_at <= $1
-            AND id NOT IN 
+            AND id NOT IN
             (
-                SELECT blob_id FROM output_file 
+                SELECT blob_id FROM output_file
                 UNION SELECT stdout_blob_id FROM job
                 UNION SELECT stderr_blob_id FROM job
             )
--- a/rust/rsc/src/rsc/blob_store_service.rs
+++ b/rust/rsc/src/rsc/blob_store_service.rs
@@ -1,7 +1,8 @@
-use entity::local_blob_store;
-use entity::prelude::LocalBlobStore;
-use sea_orm::DbErr;
-use sea_orm::{prelude::Uuid, DatabaseConnection, EntityTrait};
+use chrono::NaiveDateTime;
+use entity::prelude::{Blob, LocalBlobStore};
+use entity::{blob, local_blob_store};
+use sea_orm::{prelude::Uuid, ColumnTrait, DatabaseConnection, EntityTrait, QueryFilter};
+use sea_orm::{DbBackend, DbErr, DeleteResult, Statement};
 
 pub async fn fetch_local_blob_store(
     db: &DatabaseConnection,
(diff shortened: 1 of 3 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -213,7 +217,55 @@ fn launch_eviction(conn: Arc<DatabaseConnection>, tick_interval: u64, deadline:
...
+            let blobs = blob_store_service::fetch_unreferenced_blobs(conn.as_ref(), deadline)
+                .await
+                .unwrap();
+            let blob_ids: Vec<Uuid> = blobs.iter().map(|blob| blob.id).collect();
+            let mut affected: u64 = blob_ids.len().try_into().unwrap();
+
+            if affected == 0 {
+                tracing::info!(%affected, "Performed Blob TTL eviction tick.");
+                continue;
+            }
```
colinschmidt, line 254: If we remove the early "return" do the remaining functions handle the empty Vec correctly? Not sure its a good tradeoff here but the simplicity of this function just being straight line code might be nice.

rust/rsc/src/rsc/main.rs
```diff
@@ -213,7 +217,55 @@ fn launch_eviction(conn: Arc<DatabaseConnection>, tick_interval: u64, deadline:
...
+            // Delete blobs from database
+            let res = blob_store_service::delete_blobs_by_ids(conn.as_ref(), blob_ids)
+                .await
+                .unwrap();
+            affected = res.rows_affected;
+
+            // Delete blobs from blob store
+            for blob in blobs {
+                let store = blob_stores.get(&blob.store_id).unwrap();
+                store.delete_key(blob.key).await.unwrap();
```
colinschmidt, line 265: What happens if the database removal succeeds but the store removal fails? Do we need some recovery mechanism? or should we have some kind of critical block here to ensure both things always happen?

rust/rsc/src/rsc/main.rs
```diff
@@ -213,7 +217,55 @@ fn launch_eviction(conn: Arc<DatabaseConnection>, tick_interval: u64, deadline:
...
+            if affected == 0 {
+                tracing::info!(%affected, "Performed Blob TTL eviction tick.");
+                continue;
+            }
+
+            // Delete blobs from database
+            let res = blob_store_service::delete_blobs_by_ids(conn.as_ref(), blob_ids)
+                .await
+                .unwrap();
+            affected = res.rows_affected;
```
colinschmidt, line 260: We want this to be the same variable rather than a `deleted` non mutable variable because of the way the `tracing::info!` macro works?

colinschmidt: It seems like if those two numbers are deleted we would want to record that actually. It would mean that we had some mismatch in the queries or a partial failure in the deletion.

rust/rsc/src/rsc/blob_store_service.rs
```diff
@@ -20,3 +21,47 @@ pub async fn fetch_local_blob_stores(
 ) -> Result<Vec<local_blob_store::Model>, DbErr> {
     LocalBlobStore::find().all(db).await
 }
+
+pub async fn fetch_unreferenced_blobs(
```
colinschmidt, line 25: Can we get a comment describing what this does because the TTL aspect isn't clear from the function name.

rust/rsc/src/rsc/main.rs
V-FEXrt: not currently no. `delete_blobs_by_ids` calls SeaORMs `delete_many` which panics if you give it 0 things to delete.

I could probably add a case to that function to skip on 0 ids which would be cleaner

rust/rsc/src/rsc/main.rs
V-FEXrt: This _will_ happen and there is pretty much nothing we can do about it here.

The store removal can be a remove server taking 100s of times longer to respond than we are willing to wait. (in fact, I should probably spawn a thread for the call to delete_key). 

This outcome is better than the reverse imo. We get to choose from
1) exists in the database but not the blob store
2) exists in the blob store but not the database

One thing we can do in the long term to mitigate #2 is to have a periodic task for "go collect all blobs on a store, figure out which ones aren't referenced, try to delete them again, make a report for human review". #1 doesn't have as much headache in later cleanup but it means that there is a much higher chance that a client will be given a reference to a blob that doesn't exist (which to be fair, we have to be tolerant of this regardless)

rust/rsc/src/rsc/main.rs
V-FEXrt: I think it'll be clearer if I just have two `tracing::info!`s here. Once for "found this many to delete" once for "actually deleted this many"

rust/rsc/src/rsc/main.rs
```diff
@@ -199,9 +201,11 @@ async fn connect_to_database(
     create_remote_db(config).await
 }
 
-fn launch_eviction(conn: Arc<DatabaseConnection>, tick_interval: u64, deadline: i64) {
+fn launch_eviction(conn: Arc<DatabaseConnection>, tick_interval: i64, deadline: i64) {
```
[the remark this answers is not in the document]
V-FEXrt: They are now both u64

The specific/silly API reasons that I needed u64 for one and i64 for the other and figured it made since to just hide the conversion inside of the function so that consumers of `launch_eviction` didn't have to think about it but I think I managed to sidestep those headaches.

colinschmidt (approved):
LGTM.
Thanks for adding all the tracing messages, hopefully this will make debugging such a system slightly more tolerable.

PR #1505, opened 2024-01-22
wake: Add basic http request API

V-FEXrt:
_NOTE:_ The diff on `.wakemanifest` is because it needed to be sorted the only real change to it was adding `http.wake`

This PR adds package `http` which provides a simple structured API for making web requests. This functionality is needed for the remote cache to access the RSC server however its general enough and useful enough to be elevated to an stdlib package.

The main concern is that we don't want to encourage users to make web requests as the web is a source of non-determinism in the build. In reality though, if someone needs to make a web request then they are going to do it regardless so its better that we implement it since we can more easily audit web requests and it makes the code simpler/cleaner/more abstract when it does come up.

Core API looks something like 

```
def foo _ = 
    require Pass (HttpResponse statusCode _ body) =
        buildHttpRequest "localhost"
        | addMethod HttpMethodPost
        | addHeader "Content-Type" "application/json"
        | addBody "{}"
        | makeRequest
        
    def _ = println "{statusCode | str}: {body}"

    Pass statusCode
```


share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+# The description of a http request to send to a web server.
+tuple HttpRequest =
+    # The url of the function to be called
+    Url: String
+    # The method to be called on the destination
+    Method: HttpMethod
+    # The headers to be set on the request
+    Headers: List HttpHeader
+    # The body of the request
+    Body: String
```
ag-eitilt, line 33: ...on that note, why are these private to begin with?  There's not any validation nor transformation being done in the helper functions, and together they expose the entire surface area of the type, so even if we don't recommend people mess with this directly, neither do we gain anything from keeping it hidden.

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+
+export tuple HttpResponse =
+    # The status code returned from the web server
+    StatusCode: Integer
+    # The headers returned from the web server
+    Headers: List HttpHeader
+    # The body returned from the web server
+    Body: String
```
ag-eitilt, line 41: Since you're exporting the constructor/destructor, I expect this is just the standard confusion over tuple syntax:
```suggestion
export tuple HttpResponse =
    # The status code returned from the web server
    export StatusCode: Integer
    # The headers returned from the web server
    export Headers: List HttpHeader
    # The body returned from the web server
    export Body: String
```

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+        "--request", method,
+        "--data", request.getHttpRequestBody,
+        "--url", request.getHttpRequestUrl,
+        headers
+
+    require Pass stdout =
+        makeExecPlan cmd Nil
+        | setPlanLabel "http: {method} {request.getHttpRequestUrl}"
+        # Curl dumps download stats to stderr
+        | setPlanStderr logNever
```
ag-eitilt, line 192: Might still be helpful sometimes?  Or does our virtual TTY throw curl's formatting off?
```suggestion
        | setPlanStderr logInfo
```

V-FEXrt:
Looks like I have a couple use cases for the RSC that I haven't fully captured in this API yet. Will get those first then come back around to the comments. Thanks @ag-eitilt

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+
+        headers
+        | map loop
+        | findFail
+
+
+    Pass (HttpResponse status headers body)
+
+
+# Makes the request specified by request and parses into a HttpResponse
```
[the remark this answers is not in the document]
V-FEXrt: Weird it should be setup for that. Not sure why it passed. I'll have to poke around. Regardless, formatted

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+        "--url", request.getHttpRequestUrl,
+        headers
+
+    require Pass stdout =
+        makeExecPlan cmd Nil
+        | setPlanLabel "http: {method} {request.getHttpRequestUrl}"
+        # Curl dumps download stats to stderr
+        | setPlanStderr logNever
+        # Web is unreliable and should never be reused
+        | setPlanPersistence ReRun
```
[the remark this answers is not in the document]
V-FEXrt: Maybe we can add another parameter to the request, but I'd like to keep the fact that this is a 1) a job and 2) backed by curl a "secret". Basically I want to reserve the right to change the backing job or even just make this a runtime prim

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+        | runJobWith localRunner
+        | getJobStdout
+
+    parseCurlResponse stdout
```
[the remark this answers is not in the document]
V-FEXrt: I'm personally not a big fan of that. `isHttpResponseOk` would have to span 200-399 which hides a _lot_ of info. I'm not opposed to something like `isStatusCode (x: Integer): Boolean` though. Or even `isStatusCodeFn (x: Integer => Boolean): Boolean` for working on ranges but idk if that's any better than just reading the code and acting accordingly

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+export def addHeader (name: String) (value: String): HttpRequest => HttpRequest =
+    editHttpRequestHeaders ((HttpHeader name value), _)
+
+# Adds the Content-Type header to the http request.
+export def addContentTypeHeader (value: String): HttpRequest => HttpRequest =
+    addHeader "Content-Type" value
+
+# Adds the Content-Type header to the http request and sets it to application/json
+export def addContentTypeJsonHeader: HttpRequest => HttpRequest =
+    addHeader "Content-Type" "application/json"
```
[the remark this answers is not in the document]
V-FEXrt: Just forgot to use it lol

For `Accept`, I'm not against adding a function for it, but its not sufficient as I'm sure some server will just ignore it. Instead the client should ensure the response header has `Content-Type: application/json` set

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,199 @@
...
+#
+# buildHttpRequest defaults the method to GET and leave the headers and body empty.
+#
+# NOTE: Relying on the internet for your build is inherently unreproducible and should be avoided.
+export def buildHttpRequest (url: String): HttpRequest =
+    HttpRequest url HttpMethodGet Nil ""
+
+# Adds an http method to the http request.
+export def addMethod (method: HttpMethod): HttpRequest => HttpRequest =
+    setHttpRequestMethod method
```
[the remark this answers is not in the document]
V-FEXrt: My thought here is that the client side of the API is wonky if only one function has the full `setHttpRequestFoo` name

Something like
```
        buildHttpRequest "localhost"
        | setHttpRequestMethod HttpMethodPost
        | addHeader "Content-Type" "application/json"
        | addBody "{}"
        | makeRequest
```
looks weird to me. Happy to just change it to `setMethod` however

share/wake/lib/system/http.wake
V-FEXrt: Latest commit makes things more wonky in the type now, worth taking a look and seeing if you still agree.

Mainly my thought is just that the API has a very specific public surface and I don't want people using anything else because I want to be able to manipulate the internals without a major refactor for folks who reached in to the data type

Something something "all observable effects of your api will eventually be depended on"

share/wake/lib/system/http.wake
V-FEXrt: yep, had already fixed that locally, thanks

share/wake/lib/system/http.wake
V-FEXrt: I don't feel strongly about this, main argument against imo is that we should be hitting this _a lot_ and logging this info doesn't see valuable enough for the extra time it takes to write it to the stream since it'll be in the database anyways

V-FEXrt:
Alright! Ready for another look!

share/wake/lib/system/http.wake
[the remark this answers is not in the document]
ag-eitilt: My main concern is mostly just people being lazy and seeing if their response is okay with a simple `(_ == 200)`; I definitely share your hesitation over the broad span that would get collapsed.  Then again, the response codes offer a relatively fine granularity for a reason, and if someone decides to forgo that then there's only so much we can do (especially with this entire API being recommended *against*.)

share/wake/lib/system/http.wake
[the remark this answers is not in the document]
ag-eitilt: Yeah, if we're reserving the right to hide internals then `setMethod` is probably best.

share/wake/lib/system/http.wake
ag-eitilt: `logInfo` should only be visible if people run with `-v`, right?  Does the performance get hit even if it's not shown?  But, yeah, I was forgetting that people should probably just be looking it up in the database.

share/wake/lib/system/http.wake
ag-eitilt: Yeah, that's probably getting complex enough to warrant the indirection.  And if we don't see the structure which will eventually be merged as the final representation, then I do definitely agree that it's right to hide it even if it stayed simple for the MVP.

share/wake/lib/system/http.wake
```diff
@@ -29,36 +37,40 @@ tuple HttpRequest =
...
+    Body: Option String
+    # Multipart form data. May be a list of strings or files. Mutually exclusive with body
+    FormData: Option (List HttpFormData)
 
 export tuple HttpResponse =
-    # The status code returned from the web server
-    StatusCode: Integer
+    # The status code returned from the web server. 
+    #
+    # Not all response have a status code so its not guarenteed to be set.
```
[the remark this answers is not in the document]
V-FEXrt: haha I actually only added this because I hit it on some code I'm working on if you use curl to download a file on the local machine ala `curl -i file://foo/bar` you don't get the `HTTP/1.1 200` line

share/wake/lib/system/http.wake
```diff
@@ -29,36 +37,40 @@ tuple HttpRequest =
     Method: HttpMethod
     # The headers to be set on the request
     Headers: List HttpHeader
-    # The body of the request
-    Body: String
+    # A string of the contents of the request. Mutually exclusive with form data
+    Body: Option String
+    # Multipart form data. May be a list of strings or files. Mutually exclusive with body
+    FormData: Option (List HttpFormData)
```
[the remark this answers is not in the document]
V-FEXrt: Yeah, this would be covered in the type system if we used `Result` but the type name would certainly be confusing.

share/wake/lib/system/http.wake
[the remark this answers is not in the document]
ag-eitilt: Ah, right, not everything is HTTP!  Makes sense.

V-FEXrt:
Reminder to self: Add a tag to all http jobs

share/wake/lib/core/option.wake
```diff
@@ -98,6 +98,15 @@ export def omapPartial (f: a => Option b): Option a => Option b = match _
     Some a -> f a
     None -> None
 
+# omapFailable: A variant of omap that calls a function on a non-empty option that returns a Result
+# type. The Result-type-ness is then hoist outside the resulting option, making it easier to write
+# `require Pass` statements on code that maps over `Option` types.
+export def omapFailable (fn: a => Result b fail): Option a => Result (Option b) fail = match _
```
ag-eitilt, line 104: Give us typeclasses and we can have a nice, unguessable `traverse`! 😜  I think the existing naming pattern would have this be `omapPass`, but I don't like the existing naming pattern so I'm happy with this.  It's definitely a function that I can see getting a lot more use out of now that it's in the main library (and now that I'm reminded it exists).

share/wake/lib/system/http.wake
```diff
@@ -21,13 +21,13 @@ tuple HttpHeader =
...
-    HttpFormDataFile Path
-
+# A form data entry. It consists of a user defined named and a file on disk to upload.
+#
+# Note: while raw strings are generally accepted as http form data the actual encoding of the
+# string restults in shell escape headaches. Strings should be written to a file before upload.
 tuple HttpFormData =
     Name: String
-    Value: HttpFormDataValue
+    File: String
```
ag-eitilt: I know there's cases where using a file would be overkill, but it seems like the curl argument you generate always *expects* it to be a file, so can we use `Path` here instead of `String`?  If you want to accept non-`Path` data to `addFormData`, just have that write the input to a temporary file.

share/wake/lib/system/http.wake
```diff
@@ -175,53 +165,106 @@ def parseCurlResponse (str: String): Result HttpResponse Error =
...
     parseCurlResponse stdout
+
+# Makes the request specified by request and writes body to a file.
+#
+# The backup method to make a request. It must be used when the response is not string encodable.
+# WARNING: This function will likely break the sandbox and make your build unreliable
+export target makeBinaryRequest (request: HttpRequest): Result Path Error =
+    # We can't create a unique file name ahead of time by hashing the contents so we hash the
+    # request instead. This won't create 'file output by many jobs' in a single invocation since
+    # the function is a target, and across multiple invocations its fine as the job is set to
```
ag-eitilt, line 242: The other consideration with making the HTTP stuff general is that someone might not give this a GET `HttpRequest` -- by making it a `target`, if we have something like a POST (which is explicitly non-idempotent, as opposed to PUT) then someone might *expect* to be submitting multiple identical requests to their server but the server will only see the first.  I'd question why they're wanting their build system to be posting keepalives or whatever, but I can foresee someone at some point in some project adding visualization tooling which is poorly-enough written to not be able to handle being left alone even while nothing is changing.

ag-eitilt (approved):
The complexity and edge cases of this all almost make me wish we could go back to sneakernets and installation discs, to know that everything we rely on is *immediately* local to the machine.  And that's even without touching any JavaScript!  Still, code looks good.

share/wake/lib/system/http.wake
V-FEXrt: It _should_ be a path but I'm fighting with the Runner api for RSC. The returned outputs are returned as a string not a path as they haven't been hashed yet and afaik I can't upgrade them to a Path without either causing 'multiple jobs output file' or breaking something downstream.

as for why its a always a file, that comes down to really long strings messing with your shell/curl so its better to always be a file on disk

share/wake/lib/core/option.wake
V-FEXrt: I just copied over from the `utils` dir but happy to rename if you want

share/wake/lib/system/http.wake
V-FEXrt: hmmm yeah. That is a concern. I'd really prefer to not specialize the API but this _will_ confuse someone eventually.

On one hand any non-protobuf POST request should never be returning binary data (and a protobuf POST would need a whole other library anyways) but on the other this will come up.

Personally I'm tempted to just do something like 

```
def _ = 
    if method !=~ "GET" then
        printlnLogLevel "makeBinaryRequest is idenpoent and using non-GET verbs will act unexpectedly" logWarning
    else Unit
```

since very few people will use the not recommended sub api in the not recommended api

share/wake/lib/system/http.wake
ag-eitilt: Ah, yeah, that would be a problem.  Maybe add a comment that they are deliberately left unhashed to avoid the "multiple jobs" error?  Since this is an internal tuple we can enforce that invariant, but only if we know what we're enforcing, and if we don't forget and export this later.

share/wake/lib/system/http.wake
ag-eitilt: Yeah, I think a warning is probably sufficient here.  Anybody using this (should) already know they have to be particularly careful, so some degree of sub-par behaviour in particular edge cases should be fine.

Thinking about it, it does mean the `ReRun` we *say* the job is using actually functions like `Once`, but I don't think that's a distinction we have to make in the code.

share/wake/lib/core/option.wake
ag-eitilt: We probably need to go through the library as a whole and make sure we're using consistent language (and specifically the consistent language we actually want) at some point, so I'd say leave it and we'll update later as necessary.  Besides, this still isn't *exactly* the same semantics as `rmapPass` (which returns a `Fail` if the original value was a `Fail`, while this would make it a `Pass None`).

share/wake/lib/system/http.wake
V-FEXrt: Yep SG. In the longer term (I'll leave a todo) it would be good to figure out how to use a path here without breaking rsc

share/wake/lib/system/http.wake
V-FEXrt: yeah the target does actually just bottom out in Once (which also uses a more precise target). I'll just make that change

share/wake/lib/system/http.wake
```diff
@@ -247,19 +262,34 @@ export target makeBinaryRequest (request: HttpRequest): Result Path Error =
...
     def job =
         makeExecPlan cmd Nil
-        | setPlanLabel "http: {method | methodToString} {url}"
+        | setPlanLabel "http: {methodStr} {url}"
         | setPlanStdout logNever
         # Curl dumps download stats to stderr
         | setPlanStderr logInfo
-        # Web is unreliable and should never be reused
-        | setPlanPersistence ReRun
+        # Ideally we'd rerun on every invocation of the request. Unfortunately the desintation
```
ag-eitilt, line 285:
```suggestion
        # Ideally we'd rerun on every invocation of the request. Unfortunately the destination
```

share/wake/lib/system/http.wake
```diff
@@ -207,7 +216,9 @@ def makeCurlCmd ((HttpRequest url method headers body formData): HttpRequest) (e
 
 # Makes the request specified by request and parses into a HttpResponse.
 #
-# The default method to make a request. It should be used when the response is string encodable.
+# The standard method to make a request. When the response is string encodabled it must be used.
```
ag-eitilt, line 219: `encodeabled` -> `encodeable`.  Also, the reversal of the clauses in the sentence make it harder to figure out "it" means "this function".

share/wake/lib/system/http.wake
```diff
@@ -97,8 +100,14 @@ export def addAuthorizationHeader (value: String): HttpRequest => HttpRequest =
 export def setBody (body: String): HttpRequest => HttpRequest =
     setHttpRequestBody (Some body)
 
-# Adds a named file to the requests form data
-export def addFormData (name: String) (value: String): HttpRequest => HttpRequest =
+# Adds a named file to the requests form data.
+export def addFormData (name: String) (value: Path): HttpRequest => HttpRequest =
+    unsafe_addFormData name value.getPathName
+
+# Adds a named file to the requests form data from an unhashed file.
```
ag-eitilt, line 107:
```suggestion
# Adds a named file to the request's form data from an unhashed file.
```

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,300 @@
...
+    HttpMethodPatch
+
+# The headers that may be assocated with an http request.
+#
+# These are often used to specify how to interpret the body or to provide authorization.
+tuple HttpHeader =
+    Name: String
+    Value: String
+
+# A form data entry. It consists of a user defined named and a file on disk to upload.
```
ag-eitilt, line 24:
```suggestion
# A form data entry. It consists of a user defined name and a file on disk to upload.
```

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,300 @@
...
+# Creates a basic HttpRequest to be used as part of a builder.
+#
+# This function should be used to start the builder chain of a desired http web request.
+# Ex:
+# def foo =
+#     buildHttpRequest "localhost"
+#     | setMethod HttpMethodGet
+#     | addHeader "Content-Type" "application/json"
+#     | setBody "{}"
+#     | makeRequest
```
ag-eitilt, line 68:
```suggestion
#
# Ex:
# ```
# def foo =
#     buildHttpRequest "localhost"
#     | setMethod HttpMethodGet
#     | addHeader "Content-Type" "application/json"
#     | setBody "{}"
#     | makeRequest
# ```
```

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,300 @@
...
+#
+# buildHttpRequest defaults the method to GET and leave the headers and body empty.
```
ag-eitilt, line 70:
```suggestion
# buildHttpRequest defaults the method to GET and leaves the headers and body empty.
```

share/wake/lib/system/http.wake
```diff
@@ -0,0 +1,300 @@
...
+# helper function for building up the curl cmd representing a request
+def makeCurlCmd ((HttpRequest url method headers body formData): HttpRequest) (extraFlags: List String): Result (List String) Error =
+    def headerToCurlFlag (HttpHeader name value) = "--header", "{name}:{value}", Nil
+    def bodyToCurlFlag body = "--data", body, Nil
+    def formDataToCurlFlag (HttpFormData name file) = "--form", "{name}=@{file}", Nil
+
+    require False = body.isSome && formData.isSome
+    else failWithError "Request Body and FormData are both set. This is not allowed."
+
+    # Curl does a preflight check bofore sending the request if it is large. In core use cases
```
ag-eitilt, line 198:
```suggestion
    # Curl does a preflight check before sending the request if it is large. In core use cases
```

ag-eitilt (approved):
I like the code!  This is basically just a proofreading pass at this point.

PR #1508, opened 2024-02-06
Allow spec to disable fuse caching

V-FEXrt:
A new key `disable-fuse-cache` may be set in the wakebox spec json to turn of caching in the fuse daemon. This will negatively effect performance but when enabled will avoid fuse caching bugs when reading from a hardlink that was written to via a different path name.


colinschmidt (commented):
Do we need to add the tests to some list, or do all tests in that folder get run automatically? Otherwise looks good to me.

V-FEXrt:
These are automatically picked up. 

The integration tests just get picked up but the unit tests have to manually be added

colinschmidt (approved):
LGTM

PR #1510, opened 2024-02-07
Revert "Update wake-format's search libdir to match wake's libdir"

V-FEXrt:
Reverts sifive/wake#1504

While this PR is correct, it reveals a previously hidden bug. We should circle around to fixing that but other key bug fixes need to go in asap


colinschmidt (approved):
Looks like a clean revert to me.

PR #1513, opened 2024-02-08
Only disable fuse caching for verifed hardlinks

V-FEXrt:
The previous fix in #1508 was overly broad and resulted in unintended side effects. Instead of turning direct_io on for an entire job, instead we only enable it for verified hardlinks.


tools/fuse-waked/main.cpp
```diff
@@ -1009,6 +1015,7 @@ static int wakefuse_open(const char *path, struct fuse_file_info *fi) {
 static int wakefuse_open_trace(const char *path, struct fuse_file_info *fi) {
   int out = wakefuse_open(path, fi);
   fprintf(stderr, "open(%s) = %s\n", path, trace_out(out));
+  fprintf(stderr, "open(%s): direct_io = %d\n", path, fi->direct_io);
```
colinschmidt, line 1018: Why not combine these two prints?

colinschmidt (approved):
LGTM except for a nit in the trace print.

To clarify the previous fail test now always works so became the pass test, and the previous pass test is unneeded because there is no way to configure this differently.

V-FEXrt:
> To clarify the previous fail test now always works so became the pass test, and the previous pass test is unneeded because there is no way to configure this differently.

yeah that's the idea accept the previous pass is still the current pass just renamed from `hardlink-pass` to `hardlink` since there is no `hardlink-fail`

PR #1514, opened 2024-02-08
bsp: restore --tag functionality for bsp-wake

V-FEXrt:
`--tag` was changed to be more flexible and filterable in #1454 however bsp-wake depended on the previous functionality.

This PR adds a new flag `--tag-uri` that acts like `--tag` used to and updates bsp-wake to use it


colinschmidt (approved):
LGTM

PR #1518, opened 2024-02-13
stdlib: Add unsafe_stat and writeTempFile to io.wake

V-FEXrt:
`unsafe_stat`: `stat` but doesn't require path. Needed by the cacheRunner to stat a file before it can be elevated into a Path

`writeTempFile`: writes a file from a string. Useful for jobs/tools that require a file as input. Filenames are generated from the hash of the content and duplicate content is de-duplicated therefore guarantees to never raise the `file output by multiple jobs` error

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -195,11 +195,11 @@ export def write (path: String) (content: String): Result Path Error =
 
 # Writes a string to an intermediate file that will not conflict with other files.
 #
-# Idential prefix+files will be automatically deduplicated
+# Identical prefix+files will be automatically deduplicated
 export def writeTempFile (prefix: String) (content: String): Result Path Error =
     def name = "{prefix}.{hashString content}"
 
-    require Pass dir = mkdir "build/.temp"
+    require Pass dir = mkdir ".build/tmp/writes"
 
     writeIn dir 0644 name content
 
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -106,10 +106,16 @@ def statTypeNumberToPathType (inodeType: Integer): PathType = match inodeType
 
 # Returns the system Stat for a given path
 export target stat (path: Path): Result Stat Error =
+    unsafe_stat path.getPathName
+
+# Returns the system Stat for a given path string. Prefer stat instead
+#
+# Should only be used in rare cases where wake invariants are upheld by the caller
+export target unsafe_stat (path: String): Result Stat Error =
     def imp p = prim "stat"
 
     require Pass (Pair size (Pair inodeType mode)) =
-        imp path.getPathName
+        imp path
         | rmapError makeError
 
     Stat (statTypeNumberToPathType inodeType) (bits2mode mode) size
@@ -186,6 +192,16 @@ export def write (path: String) (content: String): Result Path Error =
         | mkdir
 
     writeImp (parent, Nil) 0644 spath content
+
+# Writes a string to an intermediate file that will not conflict with other files.
+#
+# Idential prefix+files will be automatically deduplicated
+export def writeTempFile (prefix: String) (content: String): Result Path Error =
+    def name = "{prefix}.{hashString content}"
+
+    require Pass dir = mkdir "build/.temp"
+
+    writeIn dir 0644 name content
 
 # Create a file with the given mode in the specified directory
 export def writeIn (parent: Path) (mode: Integer) (name: String) (content: String): Result Path Error =


share/wake/lib/system/io.wake
```diff
@@ -187,6 +193,16 @@ export def write (path: String) (content: String): Result Path Error =
 
     writeImp (parent, Nil) 0644 spath content
 
+# Writes a string to an intermediate file that will not conflict with other files.
+#
+# Idential prefix+files will be automatically deduplicated
+export def writeTempFile (prefix: String) (content: String): Result Path Error =
+    def name = "{prefix}.{hashString content}"
+
+    require Pass dir = mkdir "build/.temp"
```
V-FEXrt, line 202: This should be `.build`

share/wake/lib/system/io.wake
```diff
@@ -187,6 +193,16 @@ export def write (path: String) (content: String): Result Path Error =
 
     writeImp (parent, Nil) 0644 spath content
 
+# Writes a string to an intermediate file that will not conflict with other files.
+#
+# Idential prefix+files will be automatically deduplicated
```
colinschmidt, line 198:
```suggestion
# Identical prefix+files will be automatically deduplicated
```

share/wake/lib/system/io.wake
```diff
@@ -187,6 +193,16 @@ export def write (path: String) (content: String): Result Path Error =
...
+export def writeTempFile (prefix: String) (content: String): Result Path Error =
```
colinschmidt, line 199: For my understanding: If this were a `target` would the deduplication happen faster? Would it be a stronger deduplication guarantee (exact input equality vs hash(input) equality)?

share/wake/lib/system/io.wake
colinschmidt: I feel like `tmp` is a more common name for these directories but that obviously doesn't matter.

share/wake/lib/system/io.wake
```diff
@@ -106,10 +106,16 @@ def statTypeNumberToPathType (inodeType: Integer): PathType = match inodeType
 
 # Returns the system Stat for a given path
 export target stat (path: Path): Result Stat Error =
+    unsafe_stat path.getPathName
+
+# Returns the system Stat for a given path string. Prefer stat instead
+#
+# Should only be used in rare cases where wake invariants are upheld by the caller
```
ag-eitilt, line 113: Might be nice to describe the invariants which are at risk, but since we're expecting people to know what they're doing if they use this, it's probably not required.

share/wake/lib/system/io.wake
ag-eitilt: To avoid nesting dotfiles, maybe `.build/tmp`?

ag-eitilt (approved):
Beyond the known directory naming issue this looks good!

share/wake/lib/system/io.wake
ag-eitilt: Thinking about it further, though, `tmp` seems like it's a bit underspecified because it *is* such a common name; `.build/temp-writes`?  That doesn't really flow nicely.

V-FEXrt: `.build/tmp/writes`?

share/wake/lib/system/io.wake
V-FEXrt: `writeIn` bottoms out in a `target` so there is a couple tradeoffs being made here.

making `writeTempFile` a target will 
 - skip the `hashStr` hash
 - hash prefix & content (the runtime needs to hash them to look it up in the memo table)
 - have two hash tables with nearly duplicate data (`writeIn` would have an entry and `writeTempFile` would have an entry)
making `writeTempFile` a def will
  - calculate the `hashStr` hash every time 
  - not hash prefix & content params
  - only have one entry in the `writeIn` memo table
  
  I'm happy either way but to me a `def` seems like slightly less overhead

colinschmidt: yeah the current implementation sounds like the right tradeoff to me. Thanks for the education on how this bottoms out.

share/wake/lib/system/io.wake
ag-eitilt: Yeah, that seems good.  Allows further temporary namespacing if anything comes up in the future.

share/wake/lib/system/io.wake
V-FEXrt: Honestly this probably isn't all that unsafe of a function. Stat doesn't modify the file so its not a huge risk there. The main issue is that you can extract data about a non-path file which isn't ideal but not the end of the world either

PR #1520, opened 2024-02-16
lsc: Fix incomplete capture of job output

V-FEXrt:
The jobCacheRunner was caching the filtered output when writing to the cache, it should instead be caching the raw output. This PR fixes that.


colinschmidt (approved):
LGTM

PR #1521, opened 2024-02-16
rsc: Update api types to align with client expectations

V-FEXrt:
- Changes job's `cmd` from a string to a vector of bytes as multipart commands are delimited with `\0` which is not valid in rust strings
- Changes visible_file's `hash` from a vector of bytes to a string as that is a much simpler representation for the client to construct.

### DIFF
--- a/rust/entity/src/job.rs
+++ b/rust/entity/src/job.rs
@@ -7,8 +7,8 @@ use sea_orm::entity::prelude::*;
 pub struct Model {
     #[sea_orm(primary_key, auto_increment = false)]
     pub id: Uuid,
-    #[sea_orm(column_type = "Binary(BlobSize::Blob(None))", unique)]
-    pub hash: Vec<u8>,
+    #[sea_orm(unique)]
+    pub hash: String,
     #[sea_orm(column_type = "Binary(BlobSize::Blob(None))")]
     pub cmd: Vec<u8>,
     #[sea_orm(column_type = "Binary(BlobSize::Blob(None))")]
--- a/rust/entity/src/job.rs
+++ b/rust/entity/src/job.rs
@@ -9,7 +9,8 @@ pub struct Model {
     pub id: Uuid,
     #[sea_orm(column_type = "Binary(BlobSize::Blob(None))", unique)]
     pub hash: Vec<u8>,
-    pub cmd: String,
+    #[sea_orm(column_type = "Binary(BlobSize::Blob(None))")]
+    pub cmd: Vec<u8>,
     #[sea_orm(column_type = "Binary(BlobSize::Blob(None))")]
     pub env: Vec<u8>,
     pub cwd: String,
--- a/rust/migration/src/m20220101_000002_create_table.rs
+++ b/rust/migration/src/m20220101_000002_create_table.rs
@@ -30,7 +30,7 @@ impl MigrationTrait for Migration {
                             .default(SimpleExpr::FunctionCall(PgFunc::gen_random_uuid())),
                     )
                     .col(ColumnDef::new(Job::Hash).ezblob().unique_key())
-                    .col(ColumnDef::new(Job::Cmd).string().not_null())
+                    .col(ColumnDef::new(Job::Cmd).ezblob().not_null())
                     .col(ColumnDef::new(Job::Env).ezblob())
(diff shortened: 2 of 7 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -482,8 +482,8 @@ mod tests {
                     .header("Authorization", api_key)
                     .body(Body::from(
                         serde_json::to_vec(&json!({
-                            "cmd": "blarg",
-                            "env":"PATH=/usr/bin",
+                            "cmd": [98, 108, 97, 114, 103], // blarg
```
colinschmidt, line 485: I think we should add/use some helper here so test writers don't have to manually to str.to_bytes.

rust/entity/src/job.rs
```diff
@@ -9,7 +9,8 @@ pub struct Model {
     pub id: Uuid,
     #[sea_orm(column_type = "Binary(BlobSize::Blob(None))", unique)]
     pub hash: Vec<u8>,
-    pub cmd: String,
+    #[sea_orm(column_type = "Binary(BlobSize::Blob(None))")]
+    pub cmd: Vec<u8>,
```
colinschmidt, line 13: Can `cmd`'s contain unicode? I assume rust's `String` could but not sure if we actually used that? Not enough of a unicode person to know whether a vec of bytes is an appropriate representation but wanted to double check, since it seems potentially problematic.

rust/migration/src/m20220101_000002_create_table.rs
```diff
@@ -84,7 +84,7 @@ impl MigrationTrait for Migration {
                             .default(SimpleExpr::FunctionCall(PgFunc::gen_random_uuid())),
                     )
                     .col(ColumnDef::new(VisibleFile::Path).string().not_null())
-                    .col(ColumnDef::new(VisibleFile::Hash).ezblob())
+                    .col(ColumnDef::new(VisibleFile::Hash).string())
```
colinschmidt, line 87: Do we think all `Hash` types should become strings? I noticed `Job::Hash` above is still using `ezblob`

rust/rsc/src/rsc/add_job.rs
```diff
@@ -24,7 +24,7 @@ pub async fn add_job(
         created_at: NotSet,
         hash: Set(hash.clone().into()),
         cmd: Set(payload.cmd),
-        env: Set(payload.env.as_bytes().into()),
```
colinschmidt, line 27: This change is strange to me since we didn't change how `env` is represented but how `cmd` is represented but here we are changing `env`
Edit: I see that we did in fact change the payload type of `env` to match the entity type. So this makes sense.

colinschmidt (approved):
LGTM except for a nit about not using byte vecs directly in the testing.

rust/entity/src/job.rs
V-FEXrt: > Can `cmd`'s contain unicode?

Not sure but we should assume yes since wake strings can

> since it seems potentially problematic

We have to use a Vec<u8> here because we stuff null bytes into the `cmd` string where they don't belong. Which causes Rust's string type to (correctly) fail. (Can't remember if it panics or just fails to parse). There is no harm in treating unicode strings (even if they are multi-byte code points) as a vector of bytes as long as we never try to use that vector as a string later. In reality all we do with this vec of bytes is hash & store it for matching. We don't even return it to the client on a job match because the client already know the cmd in that case

rust/migration/src/m20220101_000002_create_table.rs
V-FEXrt: I believe `Job::Hash` is never exposed via the API so using `ezblob` is allowed. That said probably still makes more sense to have them be uniform

rust/rsc/src/rsc/main.rs
V-FEXrt: that was way easier than I had thought it would be

rust/entity/src/job.rs
colinschmidt: Gotcha since it just becomes data that makes sense. I guess we just want to make sure people looking at the DB don't expect to be able to copy the command out of this field, but hopefully that is an easy thing to do given the number of people looking at the DB :)

PR #1522, opened 2024-02-16
rsc: Support database backed blobs

V-FEXrt:
For small enough blobs, the overhead of writing out to disk/a second provider isn't worth it. This PR allows the client to opt in to storing certain blobs directly in the database. This is primarily useful for the empty blob but will also be useful for other small blobs.

Clients can opt in by setting the content type of the mutipart part to `"blob/small"`. Right now only blobs up to 100 bytes are allowed but there isn't a strong reason (other than de-duplication) to let the client arbitrarily defined what small is.

When a small blob is resolved by the server, its contents are directly returned as the blob url using standard `%BYTE%BYTE` url encoding. the schema `db://` can be used to detect this.

### DIFF
--- a/rust/rsc/src/rsc/blob_store_impls.rs
+++ b/rust/rsc/src/rsc/blob_store_impls.rs
@@ -5,6 +5,7 @@ use futures::stream::BoxStream;
 use rand_core::{OsRng, RngCore};
 use sea_orm::prelude::Uuid;
 use tokio::fs::File;
+use tokio::io::AsyncReadExt;
 use tokio::io::BufWriter;
 use tokio_util::bytes::Bytes;
 use tokio_util::io::StreamReader;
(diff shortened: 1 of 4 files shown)


rust/rsc/src/rsc/blob_store_impls.rs
```diff
@@ -75,3 +85,56 @@ impl BlobStore for TestBlobStore {
...
+        stream: BoxStream<'a, Result<Bytes, std::io::Error>>,
+    ) -> Result<String, std::io::Error> {
+        let reader = StreamReader::new(stream);
+        futures::pin_mut!(reader);
+        let mut buffer = [0u8; 101];
+        let mut count = 0;
+
+        loop {
+            // If we have filled the buffer then the request has exceeded the max.
+            if count == 101 {
```
colinschmidt, line 111: Maybe this isn't a problem with your invocation of `reader.read` but I would expect this to be `>=` in case we read say 200 bytes in the first go.

colinschmidt (approved):
Just one bounds check that might need updating, otherwise LGTM.

rust/rsc/src/rsc/blob_store_impls.rs
V-FEXrt: Happy to change it but `reader.read` uses the type to ensure it never overflows the buffer so the count can never go past `101`

colinschmidt: Fancy. No need to change it then.

PR #1523, opened 2024-02-17
rsc: Create configurable no-op rsc runner

V-FEXrt:
Creates a rsc runner that just falls through to the underlying runner. The RSC runner can be configured with the ENV var `WAKE_REMOTE_CACHE` with the format `domain:port:auth` or `domain:port:` for no auth.

Example: `WAKE_REMOTE_CACHE=localhost:3001:MyKey ./bin/wake -x Unit`

This PR will race with https://github.com/sifive/wake/pull/1519

### DIFF
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -148,17 +148,34 @@ export def fuseRunner: Runner =
     | makeJSONRunner
 
 export def defaultRunner: Runner =
-    require Some _ = getenv "WAKE_LOCAL_JOB_CACHE"
-    else fuseRunner
+    require Some config = getenv "WAKE_REMOTE_CACHE"
+    else
+        require Some _ = getenv "WAKE_LOCAL_JOB_CACHE"
+        else fuseRunner
 
-    # The fuseRunner does not actully mount over / and instead uses the
-    # the host root as the sandbox root. This means that wakeroot will
-    # change from depending on where wake is being run from. As a hack
-    # to work around this we abuse the fact that the fuseRunner only
-    # works in relative paths to make different runs consistent. Ideally
-    # you'd have a more complex sandbox that kept the wakeroot at a
-    # consistent place across runs.
-    mkJobCacheRunner (\_ Pass "") "/workspace" fuseRunner
+        # The fuseRunner does not actully mount over / and instead uses the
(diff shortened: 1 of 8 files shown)


share/wake/lib/system/runner.wake
```diff
@@ -148,17 +148,34 @@ export def fuseRunner: Runner =
...
+        else fuseRunner
+
+        # The fuseRunner does not actully mount over / and instead uses the
+        # the host root as the sandbox root. This means that wakeroot will
+        # change from depending on where wake is being run from. As a hack
+        # to work around this we abuse the fact that the fuseRunner only
+        # works in relative paths to make different runs consistent. Ideally
+        # you'd have a more complex sandbox that kept the wakeroot at a
+        # consistent place across runs.
+        mkJobCacheRunner (\_ Pass "") "/workspace" fuseRunner
```
colinschmidt, line 163: Nit, but sometimes these nested requires are easier to read as a match.

share/wake/lib/system/runner.wake
```diff
@@ -148,17 +148,34 @@ export def fuseRunner: Runner =
...
+        panic "Remote cache config was set with incorrect format. Saw: '{config}'. Expected 'domain:port:auth' (auth may be omitted)"
+
+    require Some (port) =
+        portStr
+        | int
+    else panic "Remote cache config was set with non-integer port. Saw: {portStr}"
+
+    def auth = if authStr ==* "" then None else Some authStr
+
+    rscRunner (RemoteCacheApi domain port auth)
```
colinschmidt, line 178: It feels like this should maybe be another function that lives in remote_cache_api.wake
`def makeRemoteCacheAPI configKey: RemoteCacheAPI`

share/wake/lib/system/runner.wake
V-FEXrt: This is another example where `when` would be _so much_ nicer

```
def defaultRunner =
    when Some config = getenv "WAKE_REMOTE_CACHE"
        mkRscRunner ...
    when Some _ = getenv "WAKE_LOCAL_JOB_CACHE"
        mkJobCacheRunner ...
    fuseRunner
```

I'm still partial to the decreased nesting of `require` over `match` here though

colinschmidt (approved):
LGTM

PR #1524, opened 2024-02-28
stdlib: Add |> and |< operators for result pipelines

V-FEXrt:
Result pipelines allow for much nicer user APIs (when designed for). This PR adds operators to allow for such pipelines

### DIFF
--- a/share/wake/lib/core/result.wake
+++ b/share/wake/lib/core/result.wake
@@ -240,7 +240,7 @@ export def addErrorContext (prefix: String): Result a Error => Result a Error = 
 export def failWithError (cause: String): Result a Error =
     Fail (makeError cause)
 
-# |>: Result compatible pipeline function. Used in the same way as '|'.
+# |>: Result compatible pipeline function. Used in the same way as '|' when both the input and the following function return a Result.
 #
 # Example:
 # ```
--- a/share/wake/lib/core/result.wake
+++ b/share/wake/lib/core/result.wake
@@ -239,3 +239,25 @@ export def addErrorContext (prefix: String): Result a Error => Result a Error = 
 #   else failWithError "Could not parse {str} as an Integer"
 export def failWithError (cause: String): Result a Error =
     Fail (makeError cause)
+
+# |>: Result compatible pipeline function. Used in the same way as '|'.
+#
+# Example:
+# ```
+# Pass "foo.json" 
+# |> source 
+# |> parseJSONFile 
+# ```
+export def (argument: Result a b) |> (pipeFn: a => Result c b): Result c b =
+    rmapPass (\a pipeFn a) argument
+
+# |<: Dual to |>. Used to inject a function that doesn't return a result into the result pipeline.
+# Example:
+# ```
+# Pass 1 
+# |< str
+# |> source 
+# |> parseJSONFile 
+# ```
+export def (argument: Result a b) |< (pipeFn: a => c): Result c b =
+    rmap (\a pipeFn a) argument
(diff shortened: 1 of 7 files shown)


tests/wake-format/basic/stdout
```diff
@@ -1783,3 +1783,20 @@ def a =
...
+    |< b
+    |< c
+
+def a =
+    a |> b
+    |< c
+
+def a =
+    a |< b
+    |> c
```
V-FEXrt, line 1802: need to fix the formatting here

share/wake/lib/core/result.wake
```diff
@@ -239,3 +239,25 @@ export def addErrorContext (prefix: String): Result a Error => Result a Error =
 #   else failWithError "Could not parse {str} as an Integer"
 export def failWithError (cause: String): Result a Error =
     Fail (makeError cause)
+
+# |>: Result compatible pipeline function. Used in the same way as '|'.
```
ag-eitilt, line 243: Maybe?
```suggestion
# |>: Result compatible pipeline function. Used in the same way as '|' when both the input and the following function return a Result.
```

tests/wake-format/basic/stdout
ag-eitilt: It's an interesting formatting bug, definitely!  Nice to have it exercising whatever odd edge case is in the code here.

ag-eitilt (approved):
Yeah, I like how this looks.

tests/wake-format/basic/stdout
V-FEXrt: Fixed!

tools/wake-format/emitter.cpp
```diff
@@ -409,9 +431,25 @@ static std::vector<CSTElement> collect_right_binary(CSTElement collect_over, CST
...
+  // Ex: OP_OR("|") != OP_OR("||")
+  // Ex: OP_OR("||") == OP_OR("||")
+  bool matches = false;
+  for (const auto& i : collect_over) {
+    if (op.firstChildElement().id() == i.first &&
+        op.firstChildElement().fragment().segment().str() == i.second) {
+      matches = true;
+      break;
+    }
+  }
```
ag-eitilt, line 450: Is this significant enough to DRY out?  I don't see any differences from the other block at a glance.

tools/wake-format/emitter.cpp
```diff
@@ -1361,9 +1399,14 @@ wcl::doc Emitter::walk_binary(ctx_t ctx, CSTElement node) {
 
   std::vector<CSTElement> parts = {};
   if (is_op_left_assoc(op_token)) {
-    parts = collect_left_binary(op_token, node);
+    if (is_binop_matching_str(op_token, TOKEN_OP_OR, "|>") ||
+        is_binop_matching_str(op_token, TOKEN_OP_OR, "|<")) {
+      parts = collect_left_binary({{TOKEN_OP_OR, "|>"}, {TOKEN_OP_OR, "|<"}}, node);
```
ag-eitilt, line 1404: This is the sort of thing we'll need to fix `,` and `++`, right?

ag-eitilt (approved):
Still looks good!

tools/wake-format/emitter.cpp
V-FEXrt: Yeah its the same, I just didn't have a good enough name for it

tools/wake-format/emitter.cpp
V-FEXrt: I think so, not 100% sure. I think it might be bit more subtle. Its at least worth trying

PR #1525, opened 2024-03-01
test: Add diffs and simplify flows

V-FEXrt:
The wake test system has had hacky support for `wake-unit` for quite some time. This lead to basically having two nearly but not quite identical test subsystems. In preparing to add the  lsp tests to CI I realized it was closer to `wake-unit` than the other integration tests so I either needed to further ingrain the hack or fully remove the hack. 

I went ahead and removed the hack but this does require that more binaries (including test ones) be bootstrapped. This will somewhat increase the default install size but that can be worked around if it ever becomes a burden.

This PR also resolves a long standing TODO to print out a diff of the expected vs actual stdout/stderr when a failure occurs which will making resolving failures much easier.
Screenshot from 2024-03-01 14-03-44


ag-eitilt (approved):
I've not compared the removed code to its replacements in fine detail, but on the whole removing hacks is good progress.

PR #1526, opened 2024-03-01
CI Testing

V-FEXrt:
bonk


.github/workflows/build.yaml
```diff
@@ -101,80 +83,9 @@ jobs:
               extra_docker_build_args: ''
               extra_docker_run_args: ''
               build_cmd: make -C wake-* static
-              test_cmd: tar xvJf wake-*/wake-static_* && cd tests && ../wake-*/bin/wake runTests
+              test_cmd: tar xvJf wake-*/wake-static_* && cd tests && ../wake-*/bin/wake -v runTests
```
colinschmidt, line 86: Let's just disable running these tests for now since alpine seems to be increasingly borked.

PR #1530, opened 2024-03-05
ci: Don't test alpine builds

V-FEXrt:
This PR disables the now failing alpine test. For now we will continue to ensure that it builds.

This is not a change in policy as alpine has not been officially supported or released for over 1.5 years. See: https://github.com/sifive/wake/releases/tag/v0.28.5


colinschmidt (approved):
LGTM

PR #1532, opened 2024-03-11
Add --simple db inspection flag

V-FEXrt:
Just like the default db inspection rendering but without stdout/stderr


colinschmidt (approved):
LGTM

PR #1533, opened 2024-03-11
db: Add simple timeline rendering

V-FEXrt:
Adds a `--simple-timeline` flag similar to `--timeline` with  most of the data removed to generate a much smaller timeline file.

Keeps cmdline, label, job id, start time, and end time. All other info should be pulled from more targeted/filtered timeline files

### DIFF
--- a/share/wake/html/timeline_main.js
+++ b/share/wake/html/timeline_main.js
@@ -312,11 +312,6 @@ timeline.on('click', function (properties) {
     }
     let job = jobMap.get(parseInt(properties.item)).job;
     document.getElementById("job").innerHTML = job.job
-    if (job.usage.charAt(8) === '0') {
-        document.getElementById("job").style.color = "green";
-    } else {
-        document.getElementById("job").style.color = "red";
-    }
     document.getElementById("stale").innerHTML = job.stale ? "true" : "false";
     document.getElementById("label").innerHTML = job.label;
     document.getElementById("directory").innerHTML = job.directory;
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -1177,6 +1177,24 @@ std::string Time::as_string() const {
   return buf;
 }
 
+JAST JobReflection::to_simple_json() const {
+  JAST json(JSON_OBJECT);
+  json.add("job", job);
+  json.add("label", label.c_str());
+
+  std::stringstream commandline_stream;
+  for (const std::string &line : commandline) {
+    commandline_stream << line << " ";
+  }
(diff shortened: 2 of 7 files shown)


share/wake/html/timeline_main.js
```diff
@@ -312,11 +312,6 @@ timeline.on('click', function (properties) {
     }
     let job = jobMap.get(parseInt(properties.item)).job;
     document.getElementById("job").innerHTML = job.job
-    if (job.usage.charAt(8) === '0') {
-        document.getElementById("job").style.color = "green";
-    } else {
-        document.getElementById("job").style.color = "red";
-    }
```
V-FEXrt, line 319: Its unclear to me if anyone is depending on the coloring here but its not super color blind friendly and removing it simplifies the diff for this PR

tools/wake/describe.cpp
```diff
@@ -312,6 +312,42 @@ void describe_timeline(const std::vector<JobReflection> &jobs,
          "</html>\n";
 }
 
+void describe_simple_timeline(const std::vector<JobReflection> &jobs,
```
V-FEXrt, line 315: should probably make a shared function to DRY it up here but each rendering feels subtly different in such a way that following a DRY path here feels incorrect.

tools/wake/main.cpp
```diff
@@ -281,6 +285,7 @@ void print_help(const char *argv0) {
     << "    --failed   -f    Capture jobs which failed last build"                       << std::endl
     << "    --tag    KEY=VAL Capture jobs which are tagged, matching KEY and VAL globs"  << std::endl
     << "    --timeline       Report timeline of captured jobs as HTML"                   << std::endl
+    << "    --simple-timeline       Report timeline of captured jobs as HTML"            << std::endl
```
V-FEXrt, line 288: I need to fix the copy and spacing here. Will do that first thing tomorrow

src/runtime/database.cpp
```diff
@@ -1177,6 +1177,24 @@ std::string Time::as_string() const {
...
+JAST JobReflection::to_simple_json() const {
+  JAST json(JSON_OBJECT);
+  json.add("job", job);
+  json.add("label", label.c_str());
+
+  std::stringstream commandline_stream;
+  for (const std::string &line : commandline) {
+    commandline_stream << line << " ";
+  }
+  json.add("commandline", commandline_stream.str());
```
colinschmidt, line 1189: I think we should include `tags` to more closely mirror the information captured in `--simple`

colinschmidt (approved):
LGTM

PR #1534, opened 2024-03-12
db: Add option to capture canceled jobs

V-FEXrt:
Adds `--canceled` which captures jobs that were canceled in the last build (usually by ctrl+c). This works similarly to `--failed` in that it will forget the list of canceled jobs after the next wake invocation.


colinschmidt (approved):
LGTM

V-FEXrt:
Need to run a test to verify that wakebox timed out jobs show up as canceled then I can land this

tests/inspection/canceled/pass.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/sh
+
+set -e
+WAKE="${1:+$1/wake}"
+
+# Start from any empty db every time for stable job ids
+rm -f wake.db
+
+# Use || true to ignore the expected non-0 return from timeout
+timeout 1 ${WAKE} test || true
```
colinschmidt, line 10: We don't need `test Unit` here?

tools/wake/cli_options.h
```diff
@@ -139,9 +140,10 @@ struct CommandLineOptions {
       {'g', "globals", GOPT_ARGUMENT_FORBIDDEN},
       {'e', "exports", GOPT_ARGUMENT_FORBIDDEN},
       {0, "html", GOPT_ARGUMENT_FORBIDDEN},
-      {0, "timeline", GOPT_ARGUMENT_OPTIONAL},
-      {0, "simple-timeline", GOPT_ARGUMENT_OPTIONAL},
-      {0, "simple", GOPT_ARGUMENT_OPTIONAL},
+      {0, "timeline", GOPT_ARGUMENT_FORBIDDEN},
+      {0, "simple-timeline", GOPT_ARGUMENT_FORBIDDEN},
+      {0, "simple", GOPT_ARGUMENT_FORBIDDEN},
+      {0, "canceled", GOPT_ARGUMENT_FORBIDDEN},
```
colinschmidt, line 146: oops. thanks for fixing the other recent additions.

tools/wake/main.cpp
```diff
@@ -428,10 +434,11 @@ int main(int argc, char **argv) {
...
   }
 
-  bool is_db_inspection = !clo.job_ids.empty() || !clo.output_files.empty() ||
-                          !clo.input_files.empty() || !clo.labels.empty() || !clo.tags.empty() ||
-                          clo.last_use || clo.last_exe || clo.failed || clo.tagdag ||
-                          clo.timeline || clo.taguri || clo.simple || clo.simple_timeline;
+  bool is_db_inspection =
+      !clo.job_ids.empty() || !clo.output_files.empty() || !clo.input_files.empty() ||
+      !clo.labels.empty() || !clo.tags.empty() || clo.last_use || clo.last_exe || clo.failed ||
+      clo.tagdag || clo.timeline || clo.taguri || clo.simple || clo.simple_timeline || clo.canceled;
```
colinschmidt, line 440: Do we have a linter for C++?

tests/inspection/canceled/pass.sh
V-FEXrt: Nope. `test` is exported as a CLI function so it expects a `List String` and with nothing is specified then `Nil` is passed

tools/wake/main.cpp
V-FEXrt: Yep, just for formatting though. `clang-format` (version 12.0.X)

PR #1537, opened 2024-04-03
rsc: Fix large uploads and unescaped paths

V-FEXrt:
Fixes a few small rsc bugs:
- Large JSON body requests would fail to launch from the client due to linux's max character len for a command
- Large requests would fail to be accepted by the server due to max request size limits
- Filenames were not shell escaped and so things like `$` would be incorrectly resolved to shell values


colinschmidt (approved):
LGTM

PR #1543, opened 2024-04-16
rsc: Implement several bug fixes

V-FEXrt:
The following bugs are fixed in this PR

- Logging is temporarily disabled due to spammy logs creating a 2x slowdown
- Output files were previously being written to the wrong location when the Plan directory was changed
- Symlink content is now entirely defined by the content of the link, not a vague notion of the source path
- Symlinks would fail if a symlink was left on disk from a previous build
- The source path of the symlink was being returned as the output instead of the symlink itself
- Symlinks no longer wait for all files to download as it's not necessary

With these changes larger internal builds will complete successfully

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -333,8 +333,10 @@ fn launch_blob_eviction(
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // setup a subscriber for logging
-    let subscriber = tracing_subscriber::FmtSubscriber::new();
-    tracing::subscriber::set_global_default(subscriber)?;
+    // TODO: The logging is incredibly spammy right now and causes significant slow down.
+    //       for now, logging is disabled but this should be turned back on once logging is pruned.
+    // let subscriber = tracing_subscriber::FmtSubscriber::new();
+    // tracing::subscriber::set_global_default(subscriber)?;
 
     // Parse the arguments
     let args = ServerOptions::parse();
--- a/share/wake/lib/system/remote_cache_runner.wake
+++ b/share/wake/lib/system/remote_cache_runner.wake
@@ -162,40 +162,55 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: Result RunnerIn
                 else failWithError "rsc: Failed to make output directory"
 
                 def doDownload (CacheSearchOutputFile path mode blob) =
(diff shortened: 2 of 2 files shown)


ag-eitilt (approved):
Seems good so far as I can tell.

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -162,66 +162,79 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: Result RunnerIn
                 else failWithError "rsc: Failed to make output directory"
 
                 def doDownload (CacheSearchOutputFile path mode blob) =
-                    rscApiGetFileBlob blob "{input.getRunnerInputDirectory}/{path}" mode
+                    rscApiGetFileBlob blob path mode
```
colinschmidt, line 165: We might want to write down in the RunnerInput/Plan tuple documentation that only `Command` is relative to `Directory` both `Visible/Input` and `Output` are relative to `workspace`. It seems reasonable to assume either and I don't see it written down.

colinschmidt (approved):

rust/rsc/src/rsc/main.rs
```diff
@@ -333,8 +333,10 @@ fn launch_blob_eviction(
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // setup a subscriber for logging
-    let subscriber = tracing_subscriber::FmtSubscriber::new();
-    tracing::subscriber::set_global_default(subscriber)?;
+    // TODO: The logging is incredibly spammy right now and causes significant slow down.
+    //       for now, logging is disabled but this should be turned back on once logging is pruned.
```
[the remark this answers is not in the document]
V-FEXrt: Yeah, though I want to spend some time to get it right. i.e. a matrix of `where to log` X `what level to log at`. I feel like it'll be totally fine for "log errors to file" as the default. I just wanted to focus on making it work first before circling back around to that.

share/wake/lib/system/remote_cache_runner.wake
V-FEXrt: I made a comment/note on this line but it would probably be good for us to put together a high level document covering all these details/designs/limitations.

We have a bunch of smaller docs, but I don't think we have the top level doc covering all the content. Mainly because most of this was just iterating on the local cache learnings/design

colinschmidt (approved):
LGTM. Happy to defer logging config until we have a whole solution.

PR #1546, opened 2024-04-23
rsc: Update the tool to fully bootstrap a fresh db from one command

V-FEXrt:
`rsc_tool --bootstrap` now completes all tasks needed to bootstrap an empty but migrated postgres database. This includes supporting DbOnly Blob Store which was previously under supported. There is now only one place where the DbOnly Blob Store id is hard coded.

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -30,6 +30,8 @@ mod types;
 
 #[path = "../common/config.rs"]
 mod config;
+#[path = "../common/database.rs"]
+mod database;
 
 #[derive(Debug, Parser)]
 struct ServerOptions {
@@ -101,10 +103,18 @@ async fn activate_stores(
     }
 
     // ---    Activate DBOnly Store   ---
-    let dbonly_id = Uuid::parse_str("00000000-0000-0000-0000-000000000000").unwrap();
+    let dbonly_store = match database::read_dbonly_blob_store(&conn).await {
+        Ok(Some(store)) => store,
+        Ok(None) => {
+            panic!("Database is not configured with a DbOnly store. Please bootstrap the db")
+        }
+        Err(_) => panic!("Failed to load DbOnly store from database. Unable to continue"),
+    };
     active_stores.insert(
-        dbonly_id,
-        Arc::new(blob_store_impls::DbOnlyBlobStore { id: dbonly_id }),
+        dbonly_store.id,
+        Arc::new(blob_store_impls::DbOnlyBlobStore {
+            id: dbonly_store.id,
+        }),
     );
 
     return active_stores;
@@ -128,7 +138,7 @@ fn create_router(
         panic!("UUID for active store not in database");
     };
 
-    let dbonly_uuid = Uuid::parse_str("00000000-0000-0000-0000-000000000000").unwrap();
+    let dbonly_uuid = database::read_dbonly_blob_store_id();
     let Some(dbonly_store) = blob_stores.get(&dbonly_uuid).clone() else {
         panic!("UUID for db store not in database");
     };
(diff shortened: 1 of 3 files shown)


rust/rsc/src/rsc_tool/main.rs
```diff
@@ -166,6 +170,33 @@ async fn remove_all_jobs(db: &DatabaseConnection) -> Result<(), Box<dyn std::err
...
+    add_api_key(key, key_desc, db).await?;
+
+    // Create the local blob store
+    add_local_blob_store(local_store_root, db).await?;
+
+    // Create "DBOnly blob store"
+    database::create_dbonly_blob_store(db).await?;
+
+    println!("");
+    println!("Done! Use the store id to set the active store and share the key as appropiate.");
```
colinschmidt, line 196: Setting the active store is something that user does via another rsc_tool invocation?

rust/rsc/src/rsc_tool/main.rs
```diff
@@ -166,6 +170,33 @@ async fn remove_all_jobs(db: &DatabaseConnection) -> Result<(), Box<dyn std::err
...
+        Some("InsecureKey".into())
+    } else {
+        None
+    };
+    add_api_key(key, key_desc, db).await?;
+
+    // Create the local blob store
+    add_local_blob_store(local_store_root, db).await?;
+
+    // Create "DBOnly blob store"
```
colinschmidt, line 192: Not sure what the scare quotes are for here.

rust/rsc/src/rsc/main.rs
```diff
@@ -407,6 +417,8 @@ mod tests {
     async fn create_test_store(
         db: &DatabaseConnection,
     ) -> Result<Uuid, Box<dyn std::error::Error>> {
+        database::create_dbonly_blob_store(db).await?;
```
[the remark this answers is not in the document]
V-FEXrt: Adding the dbonly store does muddy the terminology a bit, but there is still only one "active" store.

The active store is just the store that, by default, a blob is pushed to. The dbonly store is always enabled but not the default.

Happy to use a different term if you have any suggestions

rust/rsc/src/rsc_tool/main.rs
V-FEXrt: me either lol, don't remember typing them

rust/rsc/src/rsc_tool/main.rs
V-FEXrt: Its a part of the launch parameters into the RSC. You aren't allowed to change the active store while the RSC is running since the "activation" of a store is a non trivial action on a nearly global hashmap

colinschmidt: Should probably add a README to `rsc_tool` to cover common use patterns that aren't covered by `--help`

rust/rsc/src/rsc/main.rs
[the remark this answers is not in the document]
colinschmidt: Do we have a test that this default dbonly blob store is working properly?

Regarding naming it seems like `active` store makes sense from the perspective of someone pushing blobs, but from someone pulling blobs they may get results from "inactive" stores. `default` also doesn't solve this confusion. It seems like "push" or "write" or some verb like that would need to be joined with the other adjective, so "default_push_store" or "active_write_store" or things like this but I'm not sure any of those combinations actually provide clarity, and may impair understanding.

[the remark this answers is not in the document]
V-FEXrt: > Do we have a test that this default dbonly blob store is working properly?

No not yet, I'll get one added in a followup. Testing the blob uploads is a bit more tricky since I need to stream files but it should be done

PR #1547, opened 2024-04-24
rsc: Limit results on unreferenced blobs

V-FEXrt:
Limits the unreferenced blob query to returning 16k results. This is another instance of the "no more that 2^16 parameters" panic.

Regardless of the panic this is something I was planning on doing as I noticed very large eviction steps would stall the eviction task for longer than desired. As long as we don't create more than 16k new unreferenced blobs between each tick the eviction algorithm will keep up. 

There may be periods of high activity that cause it to fall behind but I expect the periods of low activity such as late night to vastly outweigh it and provide the opportunity to catch up.


rust/rsc/src/rsc/blob_store_service.rs
```diff
@@ -51,6 +54,7 @@ pub async fn fetch_unreferenced_blobs(
                 UNION SELECT stdout_blob_id FROM job
                 UNION SELECT stderr_blob_id FROM job
             )
+            LIMIT 16000
```
colinschmidt, line 57: It seems like we should also limit the query in `delete_blobs_by_ids` in case we have a very long list of ids.

colinschmidt (commented):
Made a comment about the other query in this file that seems like it could also run into this limit.

rust/rsc/src/rsc/blob_store_service.rs
```diff
@@ -39,6 +39,9 @@ pub async fn fetch_unreferenced_blobs(
     //                 .union( *OTHER STUFF* ));
     //                 .take().
 
+    // Limit = 16k as the query is also subject to parameter max.
+    // Blob has 4 params so (2^16)/4 = 16384. Also generally best to chunk blob eviction
```
ag-eitilt, line 43: Is there any way to dynamically calculate this, or at least to explicitly list the params rather than using `SELECT *`?  I know we're quite unlikely to change the schema, but if we ever do for some reason we risk introducing a bug which might only be caught after rollout if the tests aren't big enough.

V-FEXrt: There isn't really a good way to do that no, We'd need some kind of language level inspection to determine how many parameters the struct has and I don't know if that is possible without some `macro!` magic that I don't really want to dive into.

TBH the better solution will be just just update SeaORM once they update their sqlx dependency. Newer version of sqlx allow for 2^32 parameter which we are waaay less likely to hit

rust/rsc/src/common/database.rs
```diff
@@ -157,7 +190,226 @@ pub async fn read_dbonly_blob_store(
...
+    db: &T,
+    visible_files: Vec<visible_file::ActiveModel>,
+) -> Result<(), DbErr> {
+    if visible_files.len() == 0 {
+        return Ok(());
+    }
+
+    let chunked: Vec<Vec<visible_file::ActiveModel>> = visible_files
+        .into_iter()
+        .chunks(MAX_SQLX_PARAMS / 5)
```
colinschmidt, line 213: Should we put a comment at the top of `types` to remind us to update these numbers if those ever change?
Or is there some other mechanism that will force us to remember to update this?
I don't think we want to go through the effort/complexity of creating a macro to calculate this even though thats possible.

rust/rsc/src/common/database.rs
```diff
@@ -157,7 +190,226 @@ pub async fn read_dbonly_blob_store(
...
+        .into_iter()
+        .map(|chunk| chunk.collect())
+        .collect();
+
+    for chunk in chunked {
+        VisibleFile::insert_many(chunk).exec(db).await?;
```
colinschmidt, line 219: Is there a benefit to awaiting after all chunks inserts are made?

rust/rsc/src/common/database.rs
```diff
@@ -157,7 +190,226 @@ pub async fn read_dbonly_blob_store(
...
+    //         blob::Column::Id.not_in_subquery(
+    //             sea_orm::query::Query::select()
+    //                 .column(OutputFile::BlobId)
+    //                 .from(OutputFile::Table
+    //                 .union( *OTHER STUFF* ));
+    //                 .take().
+
+    // Limit = 16k as the query is also subject to parameter max.
+    // Blob has 4 params so (2^16)/4 = 16384. Also generally best to chunk blob eviction
+    // to avoid large eviction stalls.
```
colinschmidt, line 360: How frequently does the eviction loop run? Given that this function only returns "some" of the unreferenced blobs if we are trying to evict more than 16k in between invocations we will run into problems. If eviction is running frequently enough I think there will be other more pertinent limiters but just want to check.

rust/rsc/src/common/database.rs
```diff
@@ -157,7 +190,226 @@ pub async fn read_dbonly_blob_store(
...
+            WHERE created_at <= $1
+            AND id NOT IN
+            (
+                SELECT blob_id FROM output_file
+                UNION SELECT stdout_blob_id FROM job
+                UNION SELECT stderr_blob_id FROM job
+            )
+            LIMIT 16000
+            "#,
+            [ttl.into()],
```
colinschmidt, line 375:
```suggestion
            LIMIT $2
            "#,
            [ttl.into(), MAX_SQLX_PARAMS / 5],
```

rust/rsc/src/rsc/main.rs
```diff
@@ -306,7 +303,7 @@ fn launch_blob_eviction(
             tracing::info!(%eligible, "Blobs eligible for eviction");
```
colinschmidt, line 303: This message is technically incorrect with the limit in place. This could be "At least N blobs eligible for eviction"

rust/rsc/src/common/database.rs
V-FEXrt: Yep, but I'd like to push that off into a separate "join all futures" PR

rust/rsc/src/common/database.rs
V-FEXrt: Yeah, I'll add  a reminder to `types.rs`, we can't put a comment above the actual types since those files are all generated

rust/rsc/src/common/database.rs
V-FEXrt: Its something we can configure but right now its set to tick every minute. So as long as we don't add more than 16k blobs per minute averaged across the whole day then it should keep up

colinschmidt (approved):
LGTM

PR #1548, opened 2024-04-24
rsc: Guard blob posting behind api key

V-FEXrt:
Posting blobs should only be allowed by authorized actors. This change guards that route behind the authorization middleware

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -134,6 +134,7 @@ fn create_router(
     };
 
     Router::new()
+        // Authorized Routes
         .route(
             "/job",
             post({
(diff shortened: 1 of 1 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -146,6 +147,22 @@ fn create_router(
...
+                let conn = conn.clone();
+                let active = active_store.clone();
+                let dbonly = dbonly_store.clone();
+
+                move |multipart: Multipart| blob::create_blob(multipart, conn, active, dbonly)
+            })
+            .layer(DefaultBodyLimit::disable())
+            .layer(axum::middleware::from_fn({
+                let conn = conn.clone();
+                move |req, next| api_key_check::api_key_check_middleware(req, next, conn.clone())
```
colinschmidt, line 162: Is the double `conn.clone()` correct/needed?

colinschmidt:
Oh also we should add a test that blobs can't be posted without auth.

rust/rsc/src/rsc/main.rs
V-FEXrt: Tried a couple variant but yeah looks like it is needed. Fortunately the first clone drops out of scope quite quickly

rust/rsc/src/rsc/main.rs
```diff
@@ -616,6 +618,56 @@ mod tests {
...
+        let res = router
+            .call(
+                Request::builder()
+                    .uri("/blob")
+                    .method(http::Method::POST)
+                    .header(
+                        CONTENT_TYPE,
+                        format!("multipart/form-data; boundary={}", BOUNDARY),
+                    )
+                    .header(AUTHORIZATION, api_key)
```
colinschmidt, line 663: Are you missing an `clone` on the `api_key`? Or do we not need one because this is the last request?

V-FEXrt: I don't need one here because its the last thing to take ownership of `api_key`. Rust wouldn't compile if a clone was needed here :)

PR #1549, opened 2024-04-24
rsc: Add route to check for valid api key

V-FEXrt:
Add the route `POST /auth/check` which returns `200` on valid auth key and `401` on invalid auth key. It will be used as an early warning to actors they they may have incorrectly specified their auth key when connecting to the rsc

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -171,6 +171,13 @@ fn create_router(
                 move |req, next| api_key_check::api_key_check_middleware(req, next, conn.clone())
             })),
         )
+        .route(
+            "/auth/check",
+            post(axum::http::StatusCode::OK).layer(axum::middleware::from_fn({
+                let conn = conn.clone();
+                move |req, next| api_key_check::api_key_check_middleware(req, next, conn.clone())
+            })),
+        )
         // Unauthorized Routes
         .route(
             "/job/matching",
@@ -507,6 +514,36 @@ mod tests {
             .unwrap();
 
         assert_eq!(res.status(), StatusCode::BAD_REQUEST);
+
+        // Authorization check with invalid auth should 401
+        let res = router
+            .call(
+                Request::builder()
+                    .uri("/auth/check")
+                    .method(http::Method::POST)
+                    .header("Authorization", "badauth")
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(res.status(), StatusCode::UNAUTHORIZED);
+
+        // Authorization check with valid auth should 200
+        let res = router
+            .call(
+                Request::builder()
+                    .uri("/auth/check")
+                    .method(http::Method::POST)
+                    .header("Authorization", api_key.clone())
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(res.status(), StatusCode::OK);
 
         // Correctly inserting a job should 200
         let res = router
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -325,6 +325,27 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
 
     mkCacheSearchResponse json
 
+# rscApiCheckAuthorization: Checks if the provided authorization key is valid. 
+#
+# ```
+#  api | rscApiCheckAuthorization = Pass Unit
+# ```
+export def rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error =
+    require Some auth = api.getRemoteCacheApiAuthorization
+    else failWithError "rsc: Checking authorization requires authorization but none was provided"
+
+    require Pass response =
+        makeRoute api "auth/check"
+        | buildHttpRequest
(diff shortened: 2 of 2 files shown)


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -325,6 +334,27 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
...
     mkCacheSearchResponse json
 
+# rscApiCheckAuthorization: Checks if the provided authorization key is valid. 
+#
+# ```
+#  api | rscApiCheckAuthorization = Pass Unit
+# ```
+export def rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error =
+    require Some auth = api.getRemoteCacheApiAuthorization
+    else failWithError "rsc: Checking authorization requires authorization but none was provided"
```
ag-eitilt, line 344: I don't know what sort of logic chokepoints you have closer to the surface and with this being `export` it might indeed be the best way to handle it,[^1] but don't forget that if everything ultimately funnels through a small number of narrow necks, you can collect all the `rsc:` prefixes through `addErrorContext` and save a bit of typing/typo risk.

[^1]: Actually, why is it exported?  Doesn't seem like anything which would be helpful for general audiences, but I don't know what sort of package organization you're planning.

ag-eitilt (approved):
I tend to nerd out a bit over HTTP codes, but other than that this looks good so far as I can tell.

rust/rsc/src/rsc/main.rs
```diff
@@ -497,6 +504,36 @@ mod tests {
 
         assert_eq!(res.status(), StatusCode::BAD_REQUEST);
 
+        // Authorization check with invalid auth should 401
```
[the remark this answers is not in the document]
V-FEXrt: Yep, I agree that 401 means "unauthenticated" which imo is correct here since the point is to check if the provided auth key grants authentication :)

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: > why is it exported

I consider any HTTP web route as a public part of the `RemoteCacheApi` interface. Anyone with an instance of the apii should be able to call any function on the remote server.

Since `POST auth/check` is a function made available on the server, anyone should be able to call it.

 > If everything ultimately funnels through a small number of narrow necks, you can collect all the rsc: prefixes
 
I'll need you to give an example of what you mean, but maybe given the above comment this doesn't make sense anymore? The important thing to keep in mind is that this library is just a wrapper around what is available as the public interface on the server. What the consumer does with it (in this case remote_cache_runner) is entirely up to them

ag-eitilt: Yeah, part 1 makes sense and with that clarification, part 2 is indeed invalid.  I basically just meant that if you had a lot of potential errors which were not in directly-exported functions but instead were only accessible by going through a small number of entry points, you wouldn't have to add the `rsc:` prefix to each of the errors at the place you throw them, and could instead call `addErrorContext "rsc"` at those chokepoints, for what would likely be fewer times than you have errors.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -199,9 +199,18 @@ export def makeRemoteCacheApi (config: String): Result RemoteCacheApi Error =
     else failWithError "Remote cache config was set with non-integer port. Saw: {portStr}"
 
     def auth = if authStr ==* "" then None else Some authStr
+    def api = RemoteCacheApi domain port auth
 
-    RemoteCacheApi domain port auth
-    | Pass
+    # TODO: check for compatable wake version
```
colinschmidt, line 204: I thought this version check would be on the server side? What could we check for here?

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -325,6 +334,27 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
...
+# rscApiCheckAuthorization: Checks if the provided authorization key is valid. 
+#
+# ```
+#  api | rscApiCheckAuthorization = Pass Unit
+# ```
+export def rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error =
+    require Some auth = api.getRemoteCacheApiAuthorization
+    else failWithError "rsc: Checking authorization requires authorization but none was provided"
+
+    require Pass response =
```
colinschmidt, line 346: How frequently is this function called?
The db should cache the HTTPRequest job but I'm curious if the wake function should be a target?

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: see follow PR :)

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: > How frequently is this function called?

Two answers here
1. As much as the consuming client calls it
2. In the case of RemoteCacheRunner/default wake implementation, only once per wake invocation so a target doesn't change anything here

Aside: 
> The db should cache the HTTPRequest job

This isn't the case. Web requests are never safe to cache as the result can change at any time. Because of that web request jobs are never stored in the db

colinschmidt: SGTM.

PR #1550, opened 2024-04-26
rsc: Add client-server compatibility check

V-FEXrt:
Creates `GET /version/check` which a client can use to see if the server declares compatibility with the it. The server doesn't enforce anything based on the result but a client should expect issues if it continues to interact with the server when incompatible.

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -125,8 +125,8 @@ struct VersionCheck {
 }
 
 async fn check_version(check: axum::extract::Query<VersionCheck>) -> axum::http::StatusCode {
-    println!("{:?}", check.version);
-    if check.version != "sifive/wake/41.1.1" {
+    // During development, declare all version as compatible
+    if !check.version.starts_with("sifive/wake/") {
         return axum::http::StatusCode::FORBIDDEN;
     }
 
@@ -729,6 +729,34 @@ mod tests {
             .unwrap();
 
         assert_eq!(res.status(), StatusCode::OK);
+
+        // Allowed version should should 200
+        let res = router
+            .call(
+                Request::builder()
+                    .uri("/version/check?version=sifive/wake/1.2.3")
+                    .method(http::Method::GET)
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(res.status(), StatusCode::OK);
+
+        // Disallowed version should should 403
+        let res = router
+            .call(
+                Request::builder()
+                    .uri("/version/check?version=sifive/foo/1.2.3")
+                    .method(http::Method::GET)
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(res.status(), StatusCode::FORBIDDEN);
     }
 
     #[tokio::test]
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -119,6 +119,20 @@ async fn activate_stores(
     return active_stores;
 }
 
+#[derive(serde::Deserialize)]
+struct VersionCheck {
+    version: String,
+}
+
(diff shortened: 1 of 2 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -110,6 +110,20 @@ async fn activate_stores(
...
 }
 
+#[derive(serde::Deserialize)]
+struct VersionCheck {
+    version: String,
+}
+
+async fn check_version(check: axum::extract::Query<VersionCheck>) -> axum::http::StatusCode {
+    println!("{:?}", check.version);
+    if check.version != "sifive/wake/41.1.1" {
```
ag-eitilt, line 120: Unless I'm misunderstanding the workflow, I think it would be better to introduce and check against a RSC-specific API version rather than the Wake version; for the moment they'll be incremented more or less in sync given this is where the main development is, but once the work settles down having them decoupled would be one less place to worry about forgetting to update.

V-FEXrt: Ah this is a classic "web dev" problem that I suppose I shouldn't assume folks are familiar with.

We can't control what version of the software a given client is using, but we _can_ control what software is running on the server. Imagine a user that doesn't update to a newer version of wake for 6 months (maybe they are on a release branch or something) but in that time we update the server 10 times. That super old client should still be allowed to use the server as long as it is compatible. This means the server has to decide which clients are compatible.

Right now the code seems a bit trivial/silly but that's only because we have to get the api shape in before any official release since this is the one api route that may _never_ change in _any_ way.

A more real implementation of this would look something like 

```rust
fn check_version(..) {
  let application = "sifive/foo"; // magically parse out stuff before version number
  let version = 1.2.3; // magically parse the version out;
  
  // Test applications are always allowed regardless of version
  if application = "sifive/test" {
      return axum::http::StatusCode::OK;
  }

  // Too old, not supported
  if version < 45.0.0 {
      return axum::http::StatusCode::FORBIDDEN;
  }
  
  // random bug introduced by #PULL123
  if version == 46.1.2 {
      return axum::http::StatusCode::FORBIDDEN;
  }
  
  // Version >50 has some breaking changes we don't support yet
  if version >= 50.0.0 {
      return axum::http::StatusCode::FORBIDDEN;
  }

  return axum::http::StatusCode::OK;
}

```

ag-eitilt: Yeah, makes sense, and solves a fair number of questions.  It still seems like we could base the client's advertised version string on the API revision it (asserts that it) understands rather than whitelisting specific applications, but that's probably my decentralization bias rather than anything specifically applicable to the controlled ecosystem we have here.  (We're never going to get into a [user-agent](https://webaim.org/blog/user-agent-string-history/) situation, but it's still a horror story that's stuck with me.)

V-FEXrt: > probably my decentralization bias rather than anything specifically

FWIW, if the client disagrees, it can just ignore the servers recommendation. This route is merely the server informing the client of if it thinks they are compatible. It doesn't actually enforce anything and won't stop the client from sending other requests

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -201,7 +201,24 @@ export def makeRemoteCacheApi (config: String): Result RemoteCacheApi Error =
...
+
+        printlnLevel
+        logWarning
+        "RSC client is incompatable with server but continuing due to DEBUG enabled. Reason: {format err}"
+        | Pass
+
+    # If the client isn't compatiable with the server then give up on the cache
+    require Pass Unit =
+        api
+        | rscApiCheckClientVersion "sifive/wake/{version}"
```
colinschmidt, line 220: Where does the variable `version` come from here?

V-FEXrt: its a wake `prim` that returns the same number from `wake --version`

V-FEXrt: (in fact both cases source it from the same spot)

colinschmidt (approved):
LGTM. We should probably be adding a test for each new route.

colinschmidt (approved):
LGTM

PR #1552, opened 2024-04-29
rsc: Reenable minified version of trace logging

V-FEXrt:
Reenables tracing logs for the RSC. It does this by moving (and thus omitting) the spammy `sqx` logs and completely removing the very spammy parameter logs. 

This isn't as useful as the super spammy logs for debugging but its significantly better than the nothing we have right now.

In the long term, I'd like to see about writing the logs to a file, or more interestingly, writing the logs back into the postgres database so we can do structured analysis on them.

### DIFF
--- a/rust/rsc/src/rsc/add_job.rs
+++ b/rust/rsc/src/rsc/add_job.rs
@@ -17,7 +17,7 @@ pub async fn add_job(
 ) -> StatusCode {
     // First construct all the job details as an ActiveModel for insert
     let hash = payload.hash();
-    tracing::info!(hash, "Add Request Hash");
+    tracing::info!(hash);
 
     let vis = payload.visible_files;
     let output_files = payload.output_files;
--- a/rust/rsc/src/rsc/add_job.rs
+++ b/rust/rsc/src/rsc/add_job.rs
@@ -10,13 +10,15 @@ use tracing;
 #[path = "../common/database.rs"]
 mod database;
 
-#[tracing::instrument]
+#[tracing::instrument(skip_all)]
 pub async fn add_job(
     Json(payload): Json<AddJobPayload>,
     conn: Arc<DatabaseConnection>,
 ) -> StatusCode {
     // First construct all the job details as an ActiveModel for insert
     let hash = payload.hash();
+    tracing::info!(hash, "Add Request Hash");
+
     let vis = payload.visible_files;
     let output_files = payload.output_files;
     let output_symlinks = payload.output_symlinks;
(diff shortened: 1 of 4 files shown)


rust/rsc/src/rsc/add_job.rs
```diff
@@ -12,13 +12,15 @@ use tracing;
...
 
-#[tracing::instrument]
+#[tracing::instrument(skip_all)]
 pub async fn add_job(
     Json(payload): Json<AddJobPayload>,
     conn: Arc<DatabaseConnection>,
 ) -> StatusCode {
     // First construct all the job details as an ActiveModel for insert
     let hash = payload.hash();
+    tracing::info!(hash);
```
colinschmidt, line 22: Does this include the the function name or something?  It seems like it must otherwise just printing the hash seems hard to use.

rust/rsc/src/rsc/read_job.rs
```diff
@@ -59,7 +62,7 @@ pub async fn read_job(
         .transaction::<_, (Option<Uuid>, ReadJobResponse), DbErr>(|txn| {
             Box::pin(async move {
                 let Some(matching_job) = job::Entity::find()
-                    .filter(job::Column::Hash.eq(payload.hash()))
+                    .filter(job::Column::Hash.eq(hash))
                     .one(txn)
                     .await?
                 else {
```
colinschmidt, line 68: Can we add a `tracing::info!("miss")` here? or is that recorded somewhere else on the server side?

rust/rsc/src/rsc/read_job.rs
```diff
@@ -12,7 +12,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use tracing;
 
-#[tracing::instrument]
+#[tracing::instrument(skip_all)]
 async fn record_use(job_id: Uuid, conn: Arc<DatabaseConnection>) {
```
colinschmidt, line 16: This basically tracks "job hits" right? Can we just skip `conn` and not the `job_id`?
Is it interesting to trace `resolve_blob(uuid)` also, to see blobs that get reused across jobs?

colinschmidt (commented):
LGTM, some questions on what exactly we are logging.

rust/rsc/src/rsc/add_job.rs
V-FEXrt: Its just the hash which is mainly useful for seeing differences between repeated log lines, but also its a key into the database so we can look up the hash in the db if needed

colinschmidt (approved):
LGTM

PR #1553, opened 2024-04-30
rsc: Support dbonly small blobs on the client

V-FEXrt:
This change updates it so that `stdout` and `stderr` are optimized when small. It'll decrease the number of blobs stored externally, and the number of jobs needed to fetch those blobs.

Right now only `stdout`/`stderr` are supported but with careful consideration there is no reason generic blobs can't be supported.

The server still needs a change to dedup duplicate "small blobs" but that is a transparent change to the client that can be done later

### DIFF
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -334,7 +334,7 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
 # ```
 export def rscApiGetStringBlob ((CacheSearchBlob _ uri): CacheSearchBlob): Result String Error =
     require scheme, path, Nil = extract `([a-zA-Z][a-zA-Z0-9+.-]*)://(.*)` uri
-    else failWithError "rsc: uri has too many components: '{uri}'"
+    else failWithError "rsc: uri has unexpected format: '{uri}'"
 
     # maps path = "%46%6F%6F" to content = "foo"
     def dbScheme _scheme path =
@@ -342,11 +342,11 @@ export def rscApiGetStringBlob ((CacheSearchBlob _ uri): CacheSearchBlob): Resul
         require True = path.strlen > 0
         else Pass ""
 
-        # If not empty then it must as lease be %00
+        # If not empty then it must as least be %00
         require True = path.strlen > 3
         else failWithError "rsc: Invalid db path: '{path}'"
 
-        require _empty, byteStrs = tokenize `%` path
+        require "", byteStrs = tokenize `%` path
         else failWithError "rsc: Failed to extract bytes from db path: '{path}'"
 
         require Some bytes =
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -343,7 +343,7 @@ export def rscApiGetStringBlob ((CacheSearchBlob _ uri): CacheSearchBlob): Resul
         else Pass ""
 
         # If not empty then it must as least be %00
-        require True = path.strlen > 3
+        require True = path.strlen >= 3
         else failWithError "rsc: Invalid db path: '{path}'"
 
         require "", byteStrs = tokenize `%` path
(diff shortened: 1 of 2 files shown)


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -331,16 +333,53 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
 #  rscApiGetStringBlob (RemoteCacheBlob "asdf" "https://...") = Pass "foo\nbar\nbat" 
 # ```
 export def rscApiGetStringBlob ((CacheSearchBlob _ uri): CacheSearchBlob): Result String Error =
-    # TODO: use "small blob" for this and also parse the schema out
-    # require False = id ==* "00000000-0000-0000-0000-000000000000"
-    # else Pass  ""
+    require scheme, path, Nil = extract `([a-zA-Z][a-zA-Z0-9+.-]*)://(.*)` uri
```
ag-eitilt, line 336: Technically the `//` isn't required (without it the entire right side of the `:` is read as a simple payload+query+fragment;[^1] i.e. very likely something on the same machine), but since we'll likely not be dealing with any route lookups which do leave the authority off, and since the rest of the code doesn't support that case either, there's nothing strictly wrong with this as is.  Or is it an invariant you're checking when the `CacheSearchBlob` was originally created?

[^1]: You probably know this already, but for the peanut gallery, the breakdown of `https://github.com/sifive/wake/pull/1553/files#diff-UUID` winds up looking something like the below, and if we choose specific contexts to put the link in, all of them can be made optional:
    * Scheme: `https:`
    * Authority: `//github.com`
    * Path-like payload: `/sifive/wake/pull/1553/files`
    * Fragment: `#diff-UUID`

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -331,16 +333,53 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
...
+    else failWithError "rsc: uri has too many components: '{uri}'"
```
ag-eitilt, line 337: That does mean that if the URI doesn't specifically include a `://` substring we'll hit this error with a `Nil` returned, which is kind of the opposite of "too many components".

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -331,16 +333,53 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
...
+
+    # maps path = "%46%6F%6F" to content = "foo"
+    def dbScheme _scheme path =
+        # If path is empty we have the empty string
+        require True = path.strlen > 0
+        else Pass ""
+
+        # If not empty then it must as lease be %00
```
ag-eitilt, line 345:
```suggestion
        # If not empty then it must as least be %00
```

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -331,16 +333,53 @@ export def rscApiFindMatchingJob (req: CacheSearchRequest) (api: RemoteCacheApi)
...
+        require True = path.strlen > 3
+        else failWithError "rsc: Invalid db path: '{path}'"
+
+        require _empty, byteStrs = tokenize `%` path
+        else failWithError "rsc: Failed to extract bytes from db path: '{path}'"
```
ag-eitilt, line 350: I'd either `require "", byteStrs = ...` to enforce that we don't have unsupported text in the path or prepend the potentially-not-empty string to the start of the output; probably the first since the second opens up potential injection risks.

ag-eitilt (approved):
I am only slightly annoyed that in the case of a DB blob you're using a URI that includes something that *looks* like it should be an authority (`db://%00...`) but isn't (I'd have gone with `db:%00...`), but since individual schemes are given the power to define their own format it's not technically *wrong* to say that `db` URIs include a decorative `//` prefix.

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: yeah, that's better, I just didn't think about it

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: Yeah, I intentionally slightly abuse the spec here. There isn't any technical reason that it _has_ to be this way but since every other case we have uses `://` there is nice symmetry in keeping it. The parsing is more complicated otherwise

colinschmidt (approved):
LGTM

PR #1554, opened 2024-05-01
rsc: Don't cache the cache infra jobs

V-FEXrt:
Since the `localRunner` job don't report their inputs/outputs to the wake runtime they must rerun once any time the are encountered in a given invocation. Otherwise errors can occur where their outputs are missing and wake doesn't know how to recreate them.

Its not possible to teach wake about the inputs/outputs as that would cause the cached jobs to encounter "file output by multiple jobs" error.

However its totally acceptable for these to rerun since they are only requested from the runner when they aren't already in the local `wake.db`. If a job is in the db then wake won't try to reach out to the cache for it so the jobs don't unnecessarily rerun.


colinschmidt (approved):
LGTM.

PR #1557, opened 2024-05-07
rsc: Don't panic on invalid RSC

V-FEXrt:
Previously I just `panic`ed with the RSC config was invalid. At the time this was more reasonable because 1) it was only invalid when the config had a typo and 2) I (incorrectly) assumed it would only apply to the wake bootstrap build. 

With both of those no longer true, the `panic` needs to be replaced with a fallback to the non-cache runner


colinschmidt (approved):
LGTM

PR #1558, opened 2024-05-08
rsc: Page delete-all-jobs query for interruptable progress

V-FEXrt:
Running `rsc_tool danger-delete-all-jobs` on large databases takes a very long time and can peg the postgres processes at 100% for several hours. This seems to be a common postgres issue when issuing very large deletes.

Being a very rare thing we would do, its fine that the process is slow but we need to be able to interrupt the project if the machine is under heavy load however interrupting the process doesn't actually stop postgres from maxing out the cores.

To avoid this issue, the delete is issued in batches so when it is interrupted only the last (much smaller) batch needs to complete before postgres chills out.

As a bonus, this also gives some level of progress through the delete instead of hanging for a long time

### DIFF
--- a/rust/rsc/src/common/database.rs
+++ b/rust/rsc/src/common/database.rs
@@ -219,7 +219,7 @@ pub async fn delete_all_jobs<T: ConnectionTrait>(db: &T, chunk_size: u16) -> Res
         affected += exec.rows_affected();
         tracing::info!(%affected, %total, "Jobs deleted");
 
-        if exec.rows_affected() == 0 {
+        if exec.rows_affected() <= chunk_size.into() {
             break Ok(affected);
         }
     }
--- a/rust/rsc/src/common/database.rs
+++ b/rust/rsc/src/common/database.rs
@@ -1,7 +1,7 @@
 use chrono::NaiveDateTime;
 use data_encoding::BASE64;
 use entity::prelude::{
-    Blob, BlobStore, LocalBlobStore, OutputDir, OutputFile, OutputSymlink, VisibleFile,
+    Blob, BlobStore, Job, LocalBlobStore, OutputDir, OutputFile, OutputSymlink, VisibleFile,
 };
 use entity::{
     api_key, blob, blob_store, job, local_blob_store, output_dir, output_file, output_symlink,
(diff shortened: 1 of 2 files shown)


rust/rsc/src/common/database.rs
```diff
@@ -190,9 +191,38 @@ pub async fn read_dbonly_blob_store<T: ConnectionTrait>(
...
+                    LIMIT $1
+                )
+                "#,
+                [chunk_size.into()],
+            ))
+            .await?;
+        affected += exec.rows_affected();
+        tracing::info!(%affected, %total, "Jobs deleted");
+
+        if exec.rows_affected() == 0 {
```
colinschmidt, line 222: Shouldn't this be `<= chunk_size`? Otherwise we do an extra delete when there are already no jobs.

rust/rsc/src/common/database.rs
```diff
@@ -1,17 +1,18 @@
...
     api_key, blob, blob_store, job, local_blob_store, output_dir, output_file, output_symlink,
     visible_file,
 };
 use itertools::Itertools;
 use rand_core::{OsRng, RngCore};
+use sea_orm::ExecResult;
 use sea_orm::{
     prelude::Uuid, ActiveModelTrait, ActiveValue::*, ColumnTrait, ConnectionTrait, DbBackend,
-    DbErr, DeleteResult, EntityTrait, QueryFilter, Statement,
+    DbErr, DeleteResult, EntityTrait, PaginatorTrait, QueryFilter, QuerySelect, Statement,
```
colinschmidt, line 15: Does `Paginator` get used when you do the `LIMIT`? I would have expected to see some `page` function if we are using this trait. But I've been assuming there is some "unused import" check that would catch things like that.

V-FEXrt: imo Rust traits lean from kind of confusing to very confusing. This one is on the very side.

The function `.count(db)` from `let total = entity::job::Entity::find().count(db).await?;` is provided by the `PaginatorTrait` which is why it got pulled in

V-FEXrt: > But I've been assuming there is some "unused import" check that would catch things like that.

This is correct to assume :)

colinschmidt (approved):
LGTM

PR #1559, opened 2024-05-08
rsc: Optimize blob eviction query

V-FEXrt:
Optimizes the blob eviction query which was taking over 30 minutes to run every time on large databases.

Now runs in under half a second.


colinschmidt (approved):
LGTM

PR #1560, opened 2024-05-09
rsc: Only check auth/version once per wake invocation

V-FEXrt:
The `auth` and `version` checks were getting triggered far more than expected. This change makes them `target`s so they are only triggered once per wake invocation.


colinschmidt (approved):
LGTM. This is exactly what I expected to have to do.

PR #1561, opened 2024-05-09
rsc: Track job label in rsc db

V-FEXrt:
A job's label isn't relevant to a job's cachability so it was not previously tracked. However, since things like cmd/env are obscured in the db (due to storage format) it is very difficult to examine a specific job. Adding the label makes it easier to inspect the db so this change adds it as an option field when uploading a job.


colinschmidt (approved):
LGTM

PR #1564, opened 2024-05-13
rsc: Request max limit for open number of files

V-FEXrt:
The remote cache can fairly easily hit the 1024 default soft limit for number of open files. This change adds a step where it requests an increase to the max allowed by the system.

image


rust/rsc/.config.json
```diff
@@ -1,6 +1,6 @@
 {
   "database_url": "postgres://localhost:5433/test",
-  "server_addr": "127.0.0.1:3001",
+  "server_addr": "0.0.0.0:3002",
```
colinschmidt, line 3: Is this just an unrelated change sneaking in?

colinschmidt (approved):
rlimit change looks good to me. I thought about if we could calculate a limit but it is dependent on the number of clients so we can't really come up with an upper bound without limiting the number of clients which we don't want to do artifically.

rust/rsc/.config.json
V-FEXrt: Yeah, this is just the default config that I keep stashing/unstashing for my local testing. It made it in by accident and I figured I should just keep it since its the default testing values

PR #1565, opened 2024-05-14
rsc: Join chunked futures into single await

V-FEXrt:
Joining all the futures before awaiting them gives a small but noticeable performance improvement without much cost to code readability.
 
Going to move forward with this change and keep a close eye on it. This is something worth deeper profiling later but for now still provides a net win.


V-FEXrt:
More testing has shown that this increase the variance quite a bit while under heavy load and it was decided that its better to have all jobs finish a little slower than having some jobs finish faster and some jobs finish much slower.

Closing

PR #1566, opened 2024-05-14
rsc: Add profiling script

V-FEXrt:
I threw together a basic script for profiling the rsc. Nothing too fancy but useful for evaluating potential performance changes.

### DIFF
--- a/scripts/profile_rsc.py
+++ b/scripts/profile_rsc.py
@@ -1,0 +1,123 @@
+# Copyright 2024 SiFive, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You should have received a copy of LICENSE.Apache2 along with
+# this software. If not, you may obtain a copy at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


scripts/profile_rsc.py
```diff
@@ -0,0 +1,123 @@
...
+    output_files = [{
+        "path": "out." + blob_id +  "." + str(i),
+        "mode": 420,
+        "blob_id": blob_id
+    } for i in range(size) ]
+
+    return {
+        "cmd": [47,117,115,114],
+        "cwd":".",
+        "env":[80,65,84,72,61,47],
```
ag-eitilt, line 36: Are these seemingly-magic numbers documented anywhere?

ag-eitilt (commented):
Is this a PR you're hoping to get a review on, or did you just push it for future reference?  I can't quite tell from the description.

scripts/profile_rsc.py
V-FEXrt: These numbers themselves are made up. Its just the byte representation of the command/environment string of the job. The fact that you need to serialize those values into a byte array is documented in the api type

PR #1567, opened 2024-05-15
rsc: Allow blob stores to dedup blobs

V-FEXrt:
The dbonly blob store was creating hundreds of rows for the empty string blob. In solving this a more general implementation arose.

A blob store is now allowed to de-duplicate a blob by returning a previously return key. The database will automatically flatten that into one entry in the db (and update the TTL for blob eviction)

The uniqueness requirement is set for store id+key pair so two different stores are allowed to return the same key without issues

### DIFF
--- a/rust/rsc/src/common/database.rs
+++ b/rust/rsc/src/common/database.rs
@@ -8,6 +8,7 @@ use entity::{
     visible_file,
 };
 use itertools::Itertools;
+use migration::OnConflict;
 use rand_core::{OsRng, RngCore};
 use sea_orm::ExecResult;
 use sea_orm::{
@@ -382,6 +383,21 @@ pub async fn create_many_output_dirs<T: ConnectionTrait>(
 // --------------------------------------------------
 // ----------             Blob             ----------
 // --------------------------------------------------
+pub async fn upsert_blob<T: ConnectionTrait>(
+    db: &T,
+    blob: blob::ActiveModel,
+) -> Result<Uuid, DbErr> {
+    let result = blob::Entity::insert(blob)
+        .on_conflict(
+            OnConflict::columns(vec![blob::Column::Key, blob::Column::StoreId])
+                .update_column(blob::Column::CreatedAt)
+                .to_owned(),
+        )
+        .exec(db)
+        .await?;
+
+    Ok(result.last_insert_id)
+}
 
 // ----------            Create            ----------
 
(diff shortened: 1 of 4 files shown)


rust/rsc/src/common/database.rs
```diff
@@ -382,6 +383,21 @@ pub async fn create_many_output_dirs<T: ConnectionTrait>(
...
 // ----------             Blob             ----------
 // --------------------------------------------------
+pub async fn upsert_blob<T: ConnectionTrait>(
+    db: &T,
+    blob: blob::ActiveModel,
+) -> Result<Uuid, DbErr> {
+    let result = blob::Entity::insert(blob)
+        .on_conflict(
+            OnConflict::columns(vec![blob::Column::Key, blob::Column::StoreId])
+                .update_column(blob::Column::CreatedAt)
```
colinschmidt, line 393: We need to update the CreatedAt timestamp for eviction reasons?
It seems like it would be best to keep the the time of creation the same and just track the new use time

V-FEXrt: We do yeah. Every blob has an unreferenced TTL where we guarantee that the blob won't get evicted during.

If we didn't then the following would occur.

Job A -- reference --> Blob 1
Job B uploads a  blob but its a dupe so id = 1 is returned
Job A gets evicted and now blob 1 is unreferenced and beyond its TTL so its eligible for eviction so its deleted
Job B gets uploaded referencing id = 1 which doesn't exist so the upload fails


Reseting the created_at also resets the TTL window

colinschmidt: Is it worth the effort to create a second entry to separate the eviction data from the "metrics data"? Or renaming "CreatedAt" to note that it is just the most recent time this blob was reference by a new job?

V-FEXrt: Not opposed. Could be something like TTLWall or similar. Could be interesting to have both data points

colinschmidt: Ok I think its best to do that now so we bundle as many database updates together in a single release, but I'll leave it up to you whether you want to do that in this PR or a follow up.

colinschmidt (approved):
LGTM

rust/rsc/src/common/database.rs
V-FEXrt: I'll do it in a follow up after merging the two queued up PRs

PR #1568, opened 2024-05-16
rsc: Track blob size in database

V-FEXrt:
Blob size will be needed eventually for more complex eviction algorithms and is useful now for analyzing usage stats of given jobs.

This change adds sizes to blobs and requires the BlobStore to report the size used for an uploaded blob.

### DIFF
--- a/rust/migration/src/m20220101_000001_create_blob_tables.rs
+++ b/rust/migration/src/m20220101_000001_create_blob_tables.rs
@@ -35,6 +35,7 @@ impl MigrationTrait for Migration {
                     )
                     .col(ColumnDef::new(Blob::Key).string().not_null())
                     .col(ColumnDef::new(Blob::StoreId).uuid().not_null())
+                    .col(ColumnDef::new(Blob::Size).big_integer().not_null())
                     .foreign_key(
                         ForeignKeyCreateStatement::new()
                             .name("fk-store_id-blob_store")
@@ -121,6 +122,7 @@ enum Blob {
     Id,
     Key,
     StoreId,
+    Size,
 }
 
 #[derive(DeriveIden)]
--- a/rust/migration/src/m20220101_000001_create_blob_tables.rs
+++ b/rust/migration/src/m20220101_000001_create_blob_tables.rs
@@ -51,6 +51,18 @@ impl MigrationTrait for Migration {
             .await?;
 
         manager
+            .create_index(
+                Index::create()
+                    .name("idx-key-store_id-unq")
+                    .table(Blob::Table)
+                    .col(Blob::Key)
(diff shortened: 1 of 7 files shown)


rust/migration/src/m20220101_000001_create_blob_tables.rs
```diff
@@ -35,6 +35,7 @@ impl MigrationTrait for Migration {
                     )
                     .col(ColumnDef::new(Blob::Key).string().not_null())
                     .col(ColumnDef::new(Blob::StoreId).uuid().not_null())
+                    .col(ColumnDef::new(Blob::Size).big_integer().not_null())
```
colinschmidt, line 38: Does this really align with the `size: i64` in the model?
I would have assume big_integer was designed to be something potentially much larger.

rust/rsc/src/rsc/blob_store_impls.rs
```diff
@@ -32,17 +32,25 @@ impl BlobStore for LocalBlobStore {
...
 
         let key = create_temp_filename();
         let path = std::path::Path::new(&self.root).join(key.clone());
         let mut file = BufWriter::new(File::create(path).await?);
 
-        tokio::io::copy(&mut reader, &mut file).await?;
+        let written = tokio::io::copy(&mut reader, &mut file).await?;
 
-        Ok(key)
+        let size = match i64::try_from(written) {
```
colinschmidt, line 45: Is there a reason to not align the database type with the file writing type and make the database store `u64`?
The signed-ness here also doesn't seem to make logical sense to me.

rust/migration/src/m20220101_000001_create_blob_tables.rs
V-FEXrt: It does. The largest datatype supported by Postgres is `i64`. I actually wanted a `u64` but its not supported. 

Also good to remember that the `entity` structs are generated from the actual database schema so if it was a different type then it would have been generated differently

aside: i64::MAX is still something like 1888 Petabytes which we will hopefully never have a single file that large

rust/rsc/src/rsc/blob_store_impls.rs
V-FEXrt: Yeah, postgres doesn't natively support `u64`. I actually really wanted `u64`

rust/migration/src/m20220101_000001_create_blob_tables.rs
colinschmidt: Yeah not concerned about the actual size I agree its plenty large enough with or without the extra power of two :)

colinschmidt (approved):
LGTM

PR #1569, opened 2024-05-17
rsc: Include migration tool in wake release

V-FEXrt:


build.wake
```diff
@@ -107,12 +107,17 @@ def doInstall dest kind =
     # kick off in parallel, but rsc and rsc_tool cannot be built at the same time right now.
     # Force them to be serial relative to each other.
     def install_rsc_bins =
+        require Pass rsc_migration = buildRSCMigration Unit
+
+        require Pass rsc_migration =
+            installAs "{dest}/bin/wake-remote-cache-migration" rsc_migration
```
V-FEXrt, line 113: Tempted to name these `wake-rsc`, `wake-rsc-tool`, and  `wake-rsc-migration` instead if anyone has any opinions

colinschmidt: I think using `rsc` in the tool names is a good idea.

build.wake
```diff
@@ -107,12 +107,17 @@ def doInstall dest kind =
     # kick off in parallel, but rsc and rsc_tool cannot be built at the same time right now.
     # Force them to be serial relative to each other.
     def install_rsc_bins =
+        require Pass rsc_migration = buildRSCMigration Unit
```
colinschmidt, line 110: I think the comment above should be updated to mention that RSCMigration is also part of the serial chain, or if thats not a constraint move the require on RSCMigraiton later.

colinschmidt (approved):
LGTM

PR #1570, opened 2024-05-17
rsc: Add updated_at to blob

V-FEXrt:
Adds an `update_at` column to Blob to track when updated. Blob eviction TTL now uses updated_at instead of created_at

### DIFF
--- a/rust/rsc/src/rsc/main.rs
+++ b/rust/rsc/src/rsc/main.rs
@@ -492,6 +492,7 @@ mod tests {
         let active_key = entity::blob::ActiveModel {
             id: NotSet,
             created_at: Set((Utc::now() - Duration::days(5)).naive_utc()),
+            updated_at: Set((Utc::now() - Duration::days(5)).naive_utc()),
             key: Set("InsecureKey".into()),
             size: Set(11),
             store_id: Set(store_id),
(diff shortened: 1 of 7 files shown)


rust/rsc/src/rsc/main.rs
```diff
@@ -492,6 +492,7 @@ mod tests {
         let active_key = entity::blob::ActiveModel {
             id: NotSet,
             created_at: Set((Utc::now() - Duration::days(5)).naive_utc()),
+            updated_at: Set((Utc::now() - Duration::days(5)).naive_utc()),
```
colinschmidt, line 495: I think you also need to update the `ttl_eviction` test later in this file, since it is only setting `created_at` right now. I'm a little surprised that test still passed so maybe I'm missing something or we might not be running the tests I think we are.

V-FEXrt: > We might not be running the tests I think we are.

This one :) The TTL eviction test is for _job_ eviction. Blob eviction is a separate system with no eviction "policy". For blobs we just delete any unreferenced blob where updated_at older than X. The only way for a blob to become unreferenced is for a job pointing at it to be evicted

colinschmidt (approved):
LGTM

PR #1571, opened 2024-05-20
rsc: Correct structure to match cargo expectations

V-FEXrt:
Previously the `rsc` package had a "weird" folder structure from the perspective of cargo. This caused over 20 incorrect `not used` warnings which then went on to hide ~10 real ` not used` warnings.

In order to maintain project health its important that incorrect warnings don't hide the real ones. 

Therefore, the files have been moved around to match the expected layout of a multi-binary package as described https://doc.rust-lang.org/cargo/guide/project-layout.html which fixed all the incorrect `not used` warnings.


colinschmidt (approved):
LGTM

PR #1572, opened 2024-05-21
db: Generate sql for db inspection

V-FEXrt:
Previously database inspection was created by unioning & intersecting several smaller sqlite queries together. This was mostly a bandaid over the previous system.

It was good enough for awhile but couldn't support nice features like negation, precise exclusion (for internal jobs) or custom user filers.

With a (fair) bit of massaging all the queries can be sent as a single generated query with significantly more flexibility. 

This PR makes that change and also adds support for negation with `!` in front of any filter variable. 

Followup PRs will handing a better system for tagging and filtering out internal jobs

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -32,6 +32,7 @@
 #include <sstream>
 #include <unordered_map>
 #include <unordered_set>
+#include <vector>
 
 #include "status.h"
 #include "wcl/iterator.h"
@@ -70,14 +71,6 @@ struct Database::detail {
   sqlite3_stmt *find_prior;
   sqlite3_stmt *update_prior;
   sqlite3_stmt *delete_prior;
-  sqlite3_stmt *find_job;
-  sqlite3_stmt *find_owner;
-  sqlite3_stmt *find_last_exe;
-  sqlite3_stmt *find_last_use;
-  sqlite3_stmt *find_failed;
-  sqlite3_stmt *find_label;
-  sqlite3_stmt *find_tag;
-  sqlite3_stmt *find_canceled;
   sqlite3_stmt *fetch_hash;
   sqlite3_stmt *delete_jobs;
   sqlite3_stmt *delete_dups;
@@ -123,14 +116,6 @@ struct Database::detail {
         find_prior(0),
         update_prior(0),
         delete_prior(0),
-        find_job(0),
-        find_owner(0),
-        find_last_exe(0),
-        find_last_use(0),
-        find_failed(0),
-        find_label(0),
-        find_tag(0),
-        find_canceled(0),
         fetch_hash(0),
         delete_jobs(0),
         delete_dups(0),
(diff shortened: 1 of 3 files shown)


src/runtime/database.cpp
```diff
@@ -1407,49 +1319,142 @@ static std::vector<FileDependency> get_all_file_dependencies(const Database *db,
...
+  // The subtable is constructed by joining the jobs table with the minimal set of other dependent
+  // tables with the following extra processing.
+  // 1. files table is split into 2 columns, input_files & output_files
+  // 2. tags are flattened from two columns (uri, content) to one column (tags) with a = separator
+  // 3. input_files, output_files, and tags are group_concat'd into a single row per job. <d> is
+  //    used as a deliminator between each value. The deliminator is also places at the beginning
+  //    and end of each row so that queries don't need to special case the first/last entry.
+  //
+  // Any inspection flag/user code may add any WHERE expression conditions to the main query using
+  // the columns of the subtable for fine grain filters.
```
colinschmidt, line 1382: How did you select the columns to include? Is it just the minimum set needed to implement the previous high-level database inspection filters?

src/runtime/database.cpp
```diff
@@ -1407,49 +1319,142 @@ static std::vector<FileDependency> get_all_file_dependencies(const Database *db,
...
+              WHERE access = 2
+          ) ft_output
+          ON j.job_id=ft_output.job_id
+          GROUP BY
+              j.job_id
+)delim";
+
+  // This query wraps the subtable, applies the requested filters, and returns the matching jobs
+  std::string id_query = "SELECT job_id\nFROM (\n" + subtable + ")\nWHERE\n\t";
+  id_query += collapse_and(and_or_filters);
```
colinschmidt, line 1429: If `collapse_and` returns the empty string does `WHERE` still end the id_query correctly? Or do we need a guard on this `WHERE`s inclusion? Or maybe there is no way to end up here with zero filters?

tools/wake/main.cpp
```diff
@@ -176,75 +165,61 @@ void inspect_database(const CommandLineOptions &clo, Database &db, const std::st
...
-  }
+  // --label
+  make_and_group(clo.labels, "label", "", collect_ands);
 
-  if (!clo.output_files.empty()) {
-    intersected_job_ids = apply_inspection_query(
-        captured_jobs, std::move(intersected_job_ids), clo.output_files,
-        [&](auto a) { return db.output_files_matching(wcl::make_canonical(wake_cwd + a)); });
-  }
+  // --input_files
```
colinschmidt, line 176: This is a nit but I think the flags are `--input` and `--output`

tools/wake/main.cpp
```diff
@@ -176,75 +165,61 @@ void inspect_database(const CommandLineOptions &clo, Database &db, const std::st
...
   }
 
-  if (clo.last_exe) {
-    intersected_job_ids =
-        upkeep_intersects(captured_jobs, std::move(intersected_job_ids), db.last_exe());
+  // --last-use
+  if (clo.last_use) {
+    collect_ands.push_back({"use_id == (select max(run_id) from jobs)"});
+    collect_ands.push_back({"substr(cast(commandline as text),1,1) <> '<'"});
+    collect_ands.push_back({"label <> '<hash>'"});
```
colinschmidt, line 196: Maybe we should factor this pair into a `hide_system_jobs` function or something?

tools/wake/main.cpp
```diff
@@ -176,75 +165,61 @@ void inspect_database(const CommandLineOptions &clo, Database &db, const std::st
...
+    collect_ands.push_back({"endtime = 0"});
   }
 
-  std::vector<JobReflection> intersected_jobs = {};
-  for (auto &it : captured_jobs) {
-    if (intersected_job_ids(it.first)) {
-      intersected_jobs.push_back(std::move(it.second));
-    }
+  // Convert the collected parts into a meaningful query
+  if (collect_ands.empty()) {
```
colinschmidt, line 210: Ah this answer my previous question about the potentially dangling `WHERE`

tools/wake/main.cpp
```diff
@@ -176,75 +165,61 @@ void inspect_database(const CommandLineOptions &clo, Database &db, const std::st
...
-  Set<long> intersected_job_ids = suniversal<long>();
+  std::vector<std::vector<std::string>> collect_ands = {};
 
-  if (!clo.job_ids.empty()) {
-    intersected_job_ids =
-        apply_inspection_query(captured_jobs, std::move(intersected_job_ids), clo.job_ids,
-                               [&](auto a) { return db.job_ids_matching(a); });
-  }
+  // Process --job
+  make_and_group(clo.job_ids, "cast(job_id as TEXT)", "", collect_ands);
```
colinschmidt, line 171: It just turns out that all these command line options happen to be `vector<vector<string>>`? What is the second layer of `vector` for in something like `--job`?

colinschmidt (commented):
Overall looks good just a few nits/questions.

src/runtime/database.cpp
V-FEXrt: yep

tools/wake/main.cpp
V-FEXrt: > It just turns out  ...

Nope, was intentionally designed that way!

so all inspection flags that take params have the following format

`--flag 'foo,bar' --flag 'baz,bat'`

which gets parsed into `[[foo, bar], [baz, bat]]` The inner vectors represent `or` ops while the outer vector is one big `and` op

V-FEXrt: in `--job`s case, `--job '1?,2?' --job '?5'` is perfectly valid to say jobs 15 and 25

src/runtime/database.cpp
V-FEXrt: It's guarded as you saw in a comment below

src/runtime/database.cpp
```diff
@@ -1407,49 +1319,142 @@ static std::vector<FileDependency> get_all_file_dependencies(const Database *db,
...
-    std::cerr << "Tag query must have exactly one '=' per tag" << std::endl;
+std::vector<JobReflection> Database::matching(
+    const std::vector<std::vector<std::string>> &and_or_filters) {
+  // This query creates a subtable of the following shape:
+  //
+  // | job_id | label | run_id | use_id | endtime | commandline | status |       input_files   |
+  // output_files  |       tags       |
+  // --------------------------------------------------------------------------------------------------------------------------------
+  // |    1   |  foo  |   1    |    1   |  1234   | ls lah .    |   0    | <d>a.txt<d>b.txt<d> |
+  // <d>c.o<d>d.o<d> | <d>a=b<d>c=d<d>  | |    1   |  foo  |   1    |    1   |  1234   | ls lah . |
```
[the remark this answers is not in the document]
V-FEXrt: oh wow no. clang-format messed this up

tools/wake/main.cpp
colinschmidt: Is this and/or documented in the usage text?

PR #1574, opened 2024-05-24
Add a few list helper functions

ag-eitilt:
There's been a few times I and @maniltongya have lamented not having some less-obvious List helpers we'd reach for in Haskell; this adds the two we had talked about previously, and I quickly scanned through Data.List but didn't see anything else immediately relevant -- feel free to suggest others, though, if I did miss something.

Additionally, this implements the note that [we'd eventually add `single`](https://github.com/sifive/wake-extras/blob/bb9e17d6b1b492da968b89ff6efd73b876675164/query.wake#L3-L6) for similar limited benefit as `prepend` (i.e. not having to write a parenthetical underscore-lambda when we want to pass it to some other function).  I deliberately *haven't* brought over [the similarly-notated `lookup`](https://github.com/sifive/wake-extras/blob/bb9e17d6b1b492da968b89ff6efd73b876675164/json.wake#L14-L23) since it's not the type signature or semantics people would expect from that name, people should probably use a `Map` if they need to make a lot of `mlookup` calls, and there's enough variation in whether they'd want just the value or the key *and* the value (since the predicate might be written to match multiple keys) that I didn't feel a single helper was helpful when we already have `find`.

### DIFF
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -823,9 +823,9 @@ export def seq: Integer => List Integer =
 #
 # Example:
 # ```
-#   seq 5 "a" = "a", "a", "a", "a", "a", Nil
-#   seq (-1) "a" = Nil
-#   seq 0 "a" = Nil
+#   replicate 5 "a" = "a", "a", "a", "a", "a", Nil
+#   replicate (-1) "a" = Nil
+#   replicate 0 "a" = Nil
 # ```
 export def replicate (n: Integer) (element: a): List a =
     tab (\_ element) n
--- a/share/wake/lib/core/list.wake
+++ b/share/wake/lib/core/list.wake
@@ -77,6 +77,19 @@ export def empty: List a => Boolean = match _
     Nil -> True
     _ -> False
 
+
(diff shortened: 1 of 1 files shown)


share/wake/lib/core/list.wake
```diff
@@ -790,6 +819,17 @@ export def tab (f: Integer => a): Integer => List a =
 export def seq: Integer => List Integer =
     tab (_)
 
+# Create a list identical elements, with a given length.
+#
+# Example:
+# ```
+#   seq 5 "a" = "a", "a", "a", "a", "a", Nil
+#   seq (-1) "a" = Nil
+#   seq 0 "a" = Nil
```
V-FEXrt, line 828:
```suggestion
#   replicate 5 "a" = "a", "a", "a", "a", "a", Nil
#   replicate (-1) "a" = Nil
#   replicate 0 "a" = Nil
```

V-FEXrt (approved):

share/wake/lib/core/list.wake
ag-eitilt: Yeah, that would be the thing I had a feeling I was forgetting.

PR #1575, opened 2024-05-25
rsc: Track jobs history over time keyed by hash

V-FEXrt:
Adds a table to the database for tracking job history. The job hash is used so that even when a job is evicted its history stays around. If the job is ever recached it'll be picked back up


rust/rsc/src/database.rs
```diff
@@ -247,6 +251,38 @@ pub async fn delete_job<T: ConnectionTrait>(
...
+{
+    let hashes = JobHistoryHash::find_by_statement(Statement::from_sql_and_values(
+        DbBackend::Postgres,
+        r#"
+        DELETE FROM job 
+        WHERE created_at <= $1
+        RETURNING hash
+        "#,
+        [ttl.into()],
+    ))
```
ag-eitilt, line 271: For my own knowledge, what' s the benefit to direct SQL here rather than the `entity::job::Column::CreatedAt.lte(ttl)` you had previously?

ag-eitilt (approved):
All seems reasonable.  This is stat tracking to give you/people in general a bit more info for debugging?

rust/rsc/src/database.rs
V-FEXrt: The `RETURNING hash` part is required for the job history table and I couldn't get a combo of both "delete many" and "return hash" in the same query. This was the only way (that I could find) to get the correct data out and have it be in only one query

colinschmidt (approved):
LGTM

PR #1576, opened 2024-05-28
db: Use tags instead of magic to hide internal jobs

V-FEXrt:
Manually and directly tag jobs as hidden then filter those jobs out from inspection instead of relying on magic character in the jobs commandline/label

### DIFF
--- a/share/wake/lib/system/io.wake
+++ b/share/wake/lib/system/io.wake
@@ -179,6 +179,7 @@ target writeImp inputs mode path content =
     | setPlanOnce False
     | setPlanEnvironment Nil
     | runJobWith writeRunner
+    | setJobInspectVisibilityHidden
     | getJobOutput
 
 # Create all directories and the named file. The `content` string is written verbatim with no
@@ -277,6 +278,7 @@ def mkdirImp inputs mode path =
     | setPlanKeep False
     | setPlanEnvironment Nil
     | runJobWith mkdirRunner
+    | setJobInspectVisibilityHidden
     | getJobOutput
 
 # Create a directory in the parent
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -156,6 +156,11 @@ export def setJobTag (key: String) (value: String) (job: Job): Job =
 
     job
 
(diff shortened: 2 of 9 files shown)


ag-eitilt (changes requested):
Much better than the old hack!  I do think this as it is will probably start showing `<write>`, `<mkdir>`, `<source>`, and `<claim>` jobs (caught by the previous line 160, but not decorated with the new tag yet).  I'd argue that showing `<write>` and `<mkdir>` jobs is a good thing so I'm fine with those two (though open to discussion if either of you feel otherwise), but we should probably still hide `<source>` and `<claim>`.

V-FEXrt:
Went ahead and hid all those jobs along with the http jobs (I still consider it an implementation detail that http requests are jobs)

We can discuss `<write>`/`<mkdir>` here or separately. I don't have a strong opinion in the matter but the fact that they are mostly noise and somewhat trivially inspectable leads me to lean towards leaving them out.

aside: we could/should probably add an `--internal` flag that only matches internal jobs

share/wake/lib/system/job.wake
```diff
@@ -156,6 +156,11 @@ export def setJobTag (key: String) (value: String) (job: Job): Job =
 
     job
 
+# Set a special tag to tell db inspection that this job shouldn't be shown by default
+export def setJobInspectVisibilityHidden (job: Job): Job =
+    job
+    | setJobTag "inspect.visibility" "hidden"
```
ag-eitilt, line 162: Good call.

share/wake/lib/system/io.wake
ag-eitilt, line 1: I'd like to talk about these two in the meeting before merging the PR.

ag-eitilt (approved):

share/wake/lib/system/io.wake
ag-eitilt: Demonstrated in the meeting: The filter is only applied to `--last`/`--last-used` and `--last-executed`, so any other DB queries will still display the hidden jobs; this fine-grained control was my ideal solution (but I wasn't sure it was possible), so I'm quite happy with the current state.

colinschmidt (approved):
LGTM

PR #1577, opened 2024-05-29
rsc: Cleanup TODOs

V-FEXrt:
Cleans up TODOS, updates the strToBytes function to use `byteToInteger` which shouldn't break serialization on non-ASCII strings

### DIFF
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -451,10 +451,6 @@ export def rscApiGetStringBlob ((CacheSearchBlob _ uri): CacheSearchBlob): Resul
 #  rscApiGetFileBlob (RemoteCacheBlob "asdf" "https://...") "foo/bar" 0644 = Pass "foo/bar" 
 # ```
 export def rscApiGetFileBlob ((CacheSearchBlob _ uri): CacheSearchBlob) (path: String) (mode: Integer): Result String Error =
-    # TODO: use "small blob" for this and also parse the schema out
-    # require False = id ==* "00000000-0000-0000-0000-000000000000"
-    # else Pass  ""
-
     require Pass downloadPath =
         buildHttpRequest uri
         | setMethod HttpMethodGet
@@ -499,8 +495,8 @@ export def rscApiGetFileBlob ((CacheSearchBlob _ uri): CacheSearchBlob) (path: S
 def strToBytes (str: String): List Integer =
     str
     | explode
-    # FIXME: unicodeToInteger may return values larger than u8 which will break the serialization
-    | map (unicodeToInteger _)
+    # TODO: this only supports ASCII, not unicode. This needs to be enforced somewhere.
+    | map (byteToInteger _)
 
 # listFlattenWithNull ((1,2,3,), (4,5,6,), (7,8,9,),) = (1,2,3,0,4,5,6,0,7,8,9,0,)
 def listFlattenWithNull (parts: List (List Integer)): List Integer =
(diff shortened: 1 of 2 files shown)


ag-eitilt:
I'm a bit concerned about `byteToInteger` -- or maybe the problem's in the deserialization?
```
$ wake -x 'explode "[aɪ̯ pʰiː eɪ̯]" | map byteToInteger | map integerToByte'
"[", "a", "�", "�", " ", "p", "�", "i", "�", " ", "e", "�", "�", "]", Nil
$ wake -x 'explode "[aɪ̯ pʰiː eɪ̯]" | map unicodeToInteger | map integerToUnicode'
"[", "a", "ɪ", "̯", " ", "p", "ʰ", "i", "ː", " ", "e", "ɪ", "̯", "]", Nil
```

If that's a false negative, then it's all good, but I think we might need those single-quote raw strings that got simplified away long ago for `explode` to give you the expected bytes.

V-FEXrt:
You are right to be concerned, I was/am too. I tested this specific case using various emoji but I guess I didn't do a deep enough test.

There isn't really a good solution here. I have to send a stream of _bytes_ to the server and `unicodeToInteger` will very happily return values larger than a byte.

```
$ wake -x 'explode "[aɪ̯ pʰiː eɪ̯]" | map unicodeToInteger'
91, 97, 618, 815, 32, 112, 688, 105, 720, 32, 101, 618, 815, 93, Nil
            ^^^^^^^^ -- these don't fit in a byte
```

which will break the RSC. We might be able to use it if we could find the largest return value from `unicodeToInteger` and update the RSC to use that type

Maybe if we can figure out the largest possible value that `unicodeToInteger` will return but thats

V-FEXrt:
Okay, I found the max size and what I need to change everything too. I'd like to do that is a separate PR after this one lands. This PR is a little weird to land then immediately change to something else but the PR brings the world into a more consistent state (that is equally broken as the previous state)

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -499,8 +494,7 @@ export def rscApiGetFileBlob ((CacheSearchBlob _ uri): CacheSearchBlob) (path: S
 def strToBytes (str: String): List Integer =
     str
     | explode
-    # FIXME: unicodeToInteger may return values larger than u8 which will break the serialization
```
ag-eitilt, line 502: May still be worth leaving a `FIXME` with updated text related to high-code-point values.  I'm actually surprised emoji worked; I don't actually know how that would have happened unless there's a bug in `explode`.

ag-eitilt (commented):
Maybe we just need a specific `toUTF8Bytes` encoding function rather than overloading `explode`?  It likely requires introducing a new prim that we might not want to expose to users at large, but that's the actual task you're currently trying to solve rather than splitting the string into code points and having to send those over the wire.

Either way, I agree merging this now should be fine -- like you said, it's broken but it at least brings both ends into alignment with each other and makes it obvious from the names that we don't expect it to work for anything outside the lowest-code-point characters.

ag-eitilt (approved):

V-FEXrt:
Ah actually the larger integer type doesn't work since we are just streaming bytes to files. We have to make this work using just a byte stream.

From an offline chat with Sam, turns out the issue is actually that `explode` works on code points (which makes sense) so that it the core problem.

Wake is perfectly happy to reconstruct a UTF8 string from the `cat` of a bunch of invalid UTF8 strings.

```shell
$ wake -x '((integerToByte 0xc3), (integerToByte 0xa7), Nil) | cat'
"ç"
```

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: I've marked it in my task list. Its a bit easier to keep track of there

PR #1578, opened 2024-05-29
rsc: Use a singular env var to debug shared caches

V-FEXrt:
I unnecessarily introduced a new env var for debugging the remote cache. This complicated things without actually having an real value. This change makes the old env var the standard and warns when the new one is being used.

### DIFF
--- a/share/wake/lib/system/remote_cache_runner.wake
+++ b/share/wake/lib/system/remote_cache_runner.wake
@@ -417,6 +417,14 @@ def postJob (rscApi: RemoteCacheApi) (job: Job) (_wakeroot: String) (hidden: Str
     | rscApiPostJob req
 
 # Determines if the user has requested that debug info be emitted from the cache
-target shouldDebugRemoteCache Unit: Boolean = match (getenv "DEBUG_WAKE_REMOTE_CACHE")
-    Some _ -> True
-    None -> False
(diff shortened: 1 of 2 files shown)


share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -417,6 +417,14 @@ def postJob (rscApi: RemoteCacheApi) (job: Job) (_wakeroot: String) (hidden: Str
     | rscApiPostJob req
 
 # Determines if the user has requested that debug info be emitted from the cache
-target shouldDebugRemoteCache Unit: Boolean = match (getenv "DEBUG_WAKE_REMOTE_CACHE")
-    Some _ -> True
-    None -> False
+target shouldDebugRemoteCache Unit: Boolean =
+    match (Pair (getenv "DEBUG_WAKE_REMOTE_CACHE") (getenv "DEBUG_WAKE_SHARED_CACHE"))
+        Pair None None -> False
+        Pair (Some _) None ->
```
ag-eitilt, line 423: I feel like at some point we'll want to write an official `getBooleanEnv` to unify all of our handling of `DEBUG_WAKE_REMOTE_CACHE=0` explicit falses.

ag-eitilt (approved):

PR #1579, opened 2024-05-30
rsc: Support push/pull/pushpull RSC configs

V-FEXrt:
Adds support the the RSC client configuration to specify if only pushing to the cache, only pulling from the cache, or both should be allowed. It is entirely enforced on the client side but allows for certain actors (such as CI) to limit their interactions with the cache server.

This introduces a breaking change to the `WAKE_REMOTE_CACHE` env var. It now requires one of `push:`, `pull:`, `pushpull:` to be prepended to the previous env value


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -266,6 +322,9 @@ export def rscApiPostFileBlob (name: String) (file: String) (api: RemoteCacheApi
 #  (RemoteCacheApi "foo" 1 None) | rscApiPostJob (CachePostRequest ...) = Fail "authorization required"
 # ```
 export def rscApiPostJob (req: CachePostRequest) (api: RemoteCacheApi): Result Unit Error =
+    require True = api.getRemoteCacheApiCanPush
+    else failWithError "rsc: RSC Client not configured for pushing to the server"
+
     require Some auth = api.getRemoteCacheApiAuthorization
     else failWithError "rsc: Posting a job requires authorization but none was provided"
```
ag-eitilt, line 329: That particular invalid state isn't representable any more (at least not with the old semantics).  I think you should be able to do the same inside `uploadBlobRequest` and avoid having to add the other `require True` checks through the PR.
```suggestion
    require Some auth = api.getRemoteCacheApiAuthorization
    else failWithError "rsc: RSC Client not configured for pushing to the server"
```

ag-eitilt (approved):
Other than simplifying the error tests, this all looks good to me.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -18,14 +18,21 @@ package wake
 
 from http import _
 
+# RSC Policy on engaging with the cache. Can be confired to only push to the cache, only pull from
```
colinschmidt, line 21:
```suggestion
# RSC Policy on engaging with the cache. Can be configured to only push to the cache, only pull from
```

colinschmidt (approved):
LGTM

PR #1583, opened 2024-06-05
db: Add --json and --simple-metadata

V-FEXrt:
Adds `--simple-metadata` and `--json` renders.

`--json` outputs a structured json object that describes the matching jobs while `--simple-metadata` is the same as `--metadata` but without stdout/stderr


ag-eitilt (approved):
Code's correct, but your PR description is saying the wrong things are dropped from `--simple-metadata`.  If you have a sample `--json` output handy, what's it look like?, I can't visualize it from the C++.  (If you don't have it handy, no need to go out and get some, I'll be able to look at it myself soon.)

V-FEXrt:
I don't have one handy but its roughly
```
{
  "jobs": [
    {
      "id": 1234,
      "label": "foo bar",
      "commandline": ["echo", "foo", "bar"]
    }
  ]
}
```
and so on

PR #1585, opened 2024-06-07
rsc: Move client code to dedicated package

V-FEXrt:
Moves all of the RSC client code into the `remote_cache` package for better isolation and deals with the associated fallout

### DIFF
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -13,10 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# TODO: change this to package remote_cache and deal with all the churn
-package wake
+package remote_cache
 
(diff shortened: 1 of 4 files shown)


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -561,10 +570,10 @@ def asBytesDelimedByNull (parts: List String): List Integer =
     | listFlattenWithNull
 
 # getPathAsJson (Path "foo" "asdf") = (JObject ...)
```
ag-eitilt, line 572: Just noticed and absolutely not important, but hashes can't contain `s`. 😉

ag-eitilt (approved):
I'm completely in favour of the ongoing de-monolithization of `package wake`.

colinschmidt (approved):
LGTM

PR #1586, opened 2024-06-07
stdlib: Support (de)serialize for unicode strings

V-FEXrt:
To fully support unicode strings in the command/environment parts of a job in the RSC we need to fully serialize them. Before this change it wasn't possible to extract the underlying bytes of a unicode string (only the first byte was available). This updates the stdlib to have two functions. One from string to backing bytes and one from bytes to backing string.

No processing should be done on the bytes or it will likely result in an invalid UTF8 string


share/wake/lib/core/string.wake
```diff
@@ -207,6 +207,28 @@ export def byteToInteger (firstByteToConvert: String): Integer =
...
+# bytesToUnicode: Convert a list of exact backing bytes to the representative unicode string. If
+# you are using this function you are probably using it wrong. Consider using integerToUnicode.
+# The only safe way to use this function is when pairing it with unicodeToBytes without *any*
+# modification of the underlying bytes.
+#
+# bytesToUnicode (97, 201, 170, 204, 175, 112, 202, 176, 105, Nil) = "aɪ̯pʰi"
+export def bytesToUnicode (bytes: List Integer): String =
+    bytes
+    | map integerToByte
+    | cat
```
ag-eitilt, line 230: I'm surprised this works, but I guess it goes to show how unsafe breaking the String contract would be.

src/runtime/string.cpp
```diff
@@ -646,6 +646,30 @@ static PRIMFN(prim_str2bin) {
...
+  list[0].unify(Data::typeInteger);
+  return args.size() == 1 && args[0]->unify(Data::typeString) && out->unify(list);
+}
+
+static PRIMFN(prim_str2bytes) {
+  EXPECT(1);
+  STRING(arg0, 0);
+
+  // worst-case estimate
+  size_t need = reserve_list(arg0->size()) + arg0->size() * Integer::reserve(1);
```
ag-eitilt, line 661: Mind explaining this a bit more?  (Maybe even in the comment?)  Since we're on the C++ side and from looking at what's below I assume we're already safely dealing with the UTF8 representation rather than Unicode code points, but I don't know exactly what all the `reserve`s correspond to.  Calculate the size of the overhead plus the maximum size of an Integer, both for every list item?

ag-eitilt (approved):
It obviously works correctly, I'm just not fully able to review the C++.

ag-eitilt:
Actually, do we have anything in the tests which would cover this?

src/runtime/string.cpp
V-FEXrt: I'm mostly going off other code reference to derive this value (actually one of the reasons I put @terpstra on the review) but my vague understanding is the following

`reserve_list(arg0->size())` -> Reserve enough memory for a list of `size` elements
`arg0->size() * Integer::reserve(1)` -> Reserve enough memory for `size` number of `Integer`s. Might actually be sufficient to instead do `Integer::reserve(arg0->size())` but I'm not confident enough to assert that.

PR #1587, opened 2024-06-14
Use simpler string syntax table in Emacs

seldridge:
This minimally fixes a bug at the cost of slightly worse syntax highlighting for strings.  The bug is that, as written, any of `"`, `'`, `{`, or `}` are treated as the `|` (generic string delimiter) syntax class.  This means that these act like "string toggles".  This has the benefit of allowing things like string interpolation to work:

    "foo{bar}"
    ^   ^   ^^
    |   |   ||
    |   |   |End string
    |   |   Start string
    |   End string
    Start string

However, this creates problems if the strings are ever unbalanced.  E.g., the following is mis-highlighted without this patch:

    "foo'bar"
    ^   ^   ^
    |   |   |
    |   |   |
    |   |   Start string
    |   End string
    Start string

Avoid all this by just forcing `'` and `"` to make everything inside them strings.  This loses the syntax highlighting for string interpolation. However, this seems like a win overall as some internal Wake files run into this bug and I'm seeing the entire file show up as a string.

### DIFF
--- a/share/doc/wake/syntax/emacs/wake-mode.el
+++ b/share/doc/wake/syntax/emacs/wake-mode.el
@@ -167,10 +167,8 @@ repeated key presses."
     (modify-syntax-entry ?\) ")(" table)
     (modify-syntax-entry ?_ "w" table)
     (modify-syntax-entry ?\\ "\\" table)
-    (modify-syntax-entry ?\" "|" table)
-    (modify-syntax-entry ?\' "|" table)
-    (modify-syntax-entry ?\{ "|" table)
-    (modify-syntax-entry ?\} "|" table)
+    (modify-syntax-entry ?\" "\"" table)
+    (modify-syntax-entry ?\' "\"" table)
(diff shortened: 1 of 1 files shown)


seldridge:
I'm tagging @mikeurbach on this as I know he is an Emacs user and I think has higher elisp proficiency than I do.

seldridge:
Hold off on review/merging this as this has some issues with quote escaping.

mikeurbach (commented):
Ah, thanks for the tag. I'm honestly not an elisp expert, but happy to take a look. I've been annoyed by this, and I do agree with the premise of this PR. Sacrificing highlighting of interpolated strings in favor of not highlighting the rest of the file as a string if a `'` is used in a string seems like a win.

ag-eitilt (commented):
I know absolutely nothing about Emacs highlighting, but that does seem to be a better strategy, if you can get the quote escaping to work.

PR #1590, opened 2024-07-02
prebuilt: Remove CentOS

V-FEXrt:
CentOS is officially and fully EOL which is causing CI to break. Remove support for the prebuilt image


colinschmidt (approved):
LGTM

PR #1596, opened 2024-07-11
rsc: Decrease job TTL

V-FEXrt:
Out of an abundance of caution, this PR dials the job TTL down to 24 hours. As a fairly quick followup this will be changed into a config variable but for now I'd like to lean on the conservative side for the reason we are making later today


ag-eitilt (approved):
Rubber stamp

colinschmidt (approved):
LGTM

PR #1598, opened 2024-07-16
rsc: Add json route that returns stats

V-FEXrt:
This PR adds another route to the rsc `GET /dashboard` which returns a json object summarizing basic stats about the cache. 

A follow up PR will be added that renders it (somewhat) nicely as an HTML page

Example response:

```json
{
  "job_count": 1234,
  "blob_count": 9999,
  "size": 1000000000,
  "savings": 5432,
  "oldest_jobs": [
    {
      "label": "echo foo",
      "created_at": "2024-07-16T18:02:01.083651",
      "savings": 5
    }
  ],
  "most_reused_jobs": [
    {
      "label": "echo foo",
      "reuses": 2,
      "savings": 5
    }
  ],
  "most_space_efficient_jobs": [
    {
      "label": "foo bar",
      "runtime": 1,
      "disk_usage": 50,
      "ms_saved_per_byte": 50
    }
  ],
  "most_space_use_jobs": [
    {
      "label": "foo bar",
      "runtime": 1,
      "disk_usage": 50,
      "ms_saved_per_byte": 50
    }
  ],
  "lost_opportunity_jobs": [
    {
      "label": "foo bar",
      "reuses": 1,
      "misses": 50,
      "real_savings": 1,
      "potential_savings": 51
    }
  ]
}
```

### DIFF
--- a/rust/rsc/src/bin/rsc/types.rs
+++ b/rust/rsc/src/bin/rsc/types.rs
@@ -178,6 +178,7 @@ pub struct GetUploadUrlResponse {
 pub struct DashboardStatsOldestJob {
     pub label: String,
     pub created_at: DateTime,
+    pub reuses: i32,
     pub savings: i64,
 }
 
--- a/rust/rsc/src/bin/rsc/types.rs
+++ b/rust/rsc/src/bin/rsc/types.rs
@@ -1,4 +1,5 @@
 use blake3;
+use sea_orm::prelude::DateTime;
 use sea_orm::prelude::Uuid;
 use serde::{Deserialize, Serialize};
 
@@ -171,4 +172,28 @@ pub enum ReadJobResponse {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GetUploadUrlResponse {
     pub url: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsOldestJob {
+    pub label: String,
+    pub created_at: DateTime,
+    pub savings: i64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsMostReusedJob {
+    pub label: String,
+    pub reuses: i32,
+    pub savings: i64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsResponse {
+    pub job_count: u64,
+    pub blob_count: u64,
+    pub size: i64,
+    pub savings: i64,
+    pub oldest_jobs: Vec<DashboardStatsOldestJob>,
+    pub most_reused_jobs: Vec<DashboardStatsMostReusedJob>,
 }
--- a/rust/rsc/src/bin/rsc/types.rs
+++ b/rust/rsc/src/bin/rsc/types.rs
@@ -190,6 +190,23 @@ pub struct DashboardStatsMostReusedJob {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsLostOpportunityJob {
+    pub label: String,
+    pub reuses: i32,
+    pub misses: i32,
+    pub real_savings: i64,
+    pub potential_savings: i64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsSizeRuntimeValueJob {
+    pub label: String,
+    pub runtime: i64,
+    pub disk_usage: i64,
+    pub ms_saved_per_byte: i64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 pub struct DashboardStatsResponse {
     pub job_count: u64,
     pub blob_count: u64,
@@ -197,4 +214,7 @@ pub struct DashboardStatsResponse {
     pub savings: i64,
     pub oldest_jobs: Vec<DashboardStatsOldestJob>,
     pub most_reused_jobs: Vec<DashboardStatsMostReusedJob>,
+    pub lost_opportunity_jobs: Vec<DashboardStatsLostOpportunityJob>,
+    pub most_space_efficient_jobs: Vec<DashboardStatsSizeRuntimeValueJob>,
+    pub most_space_use_jobs: Vec<DashboardStatsSizeRuntimeValueJob>,
 }
--- a/rust/rsc/src/database.rs
+++ b/rust/rsc/src/database.rs
@@ -219,7 +219,7 @@ pub async fn time_saved<T: ConnectionTrait>(db: &T) -> Result<Option<TimeSaved>,
     TimeSaved::find_by_statement(Statement::from_string(
         DbBackend::Postgres,
         r#"
-        SELECT CAST(round(sum(savings)) as INT8) as savings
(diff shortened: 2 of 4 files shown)


rust/rsc/src/bin/rsc/types.rs
```diff
@@ -172,3 +173,27 @@ pub enum ReadJobResponse {
 pub struct GetUploadUrlResponse {
     pub url: String,
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsOldestJob {
+    pub label: String,
```
colinschmidt, line 179: `label` is an `Option<String>` in the job payload are we storing the empty string in the DB for none?

rust/rsc/src/bin/rsc/types.rs
```diff
@@ -172,3 +173,27 @@ pub enum ReadJobResponse {
...
+    pub reuses: i32,
+    pub savings: i64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DashboardStatsResponse {
+    pub job_count: u64,
+    pub blob_count: u64,
+    pub size: i64,
+    pub savings: i64,
```
colinschmidt, line 196: Can savings and size be negative? Why wouldn't they also be `u64`?
EDIT: oh I think this is just the SQL doesn't have a u64 type again.

rust/rsc/src/database.rs
```diff
@@ -193,19 +192,92 @@ pub async fn read_dbonly_blob_store<T: ConnectionTrait>(
...
+#[derive(Debug, FromQueryResult)]
+pub struct TimeSaved {
+    pub savings: i64,
+}
+
+pub async fn time_saved<T: ConnectionTrait>(db: &T) -> Result<Option<TimeSaved>, DbErr> {
+    TimeSaved::find_by_statement(Statement::from_string(
+        DbBackend::Postgres,
+        r#"
+        SELECT CAST(round(sum(savings)) as INT8) as savings
```
colinschmidt, line 222: Huh `INT8` is an 8-byte integer not an 8-bit integer TIL.

rust/rsc/src/database.rs
```diff
@@ -193,19 +192,92 @@ pub async fn read_dbonly_blob_store<T: ConnectionTrait>(
...
+        FROM (
+            SELECT h.hits * j.runtime as savings
```
colinschmidt, line 224: `hits` starts at 0 when the job is posted right? Don't want to have an off-by-one error here.

rust/rsc/src/database.rs
```diff
@@ -193,19 +192,92 @@ pub async fn read_dbonly_blob_store<T: ConnectionTrait>(
...
+    pub label: String,
+    pub created_at: DateTime,
+    pub savings: i64,
+}
+
+pub async fn oldest_jobs<T: ConnectionTrait>(db: &T) -> Result<Vec<OldestJobs>, DbErr> {
+    OldestJobs::find_by_statement(Statement::from_string(
+        DbBackend::Postgres,
+        r#"
+        SELECT j.label, j.created_at, CAST(round(CAST(h.hits * j.runtime as numeric)) as INT8) as savings
```
colinschmidt, line 246: Why didn't we need the `numeric` cast in the previous query?

rust/rsc/src/database.rs
```diff
@@ -193,19 +192,92 @@ pub async fn read_dbonly_blob_store<T: ConnectionTrait>(
...
+    ))
+    .one(db)
+    .await
+}
+
+#[derive(Debug, FromQueryResult)]
+pub struct OldestJobs {
+    pub label: String,
+    pub created_at: DateTime,
+    pub savings: i64,
```
colinschmidt, line 239: Is it basically free to include `reuses` here? Seems like we might as well throw it in if so.

colinschmidt (approved):
LGTM

rust/rsc/src/bin/rsc/types.rs
V-FEXrt: yep! I'm not actually sure why job_count/blob_count are allowed to be u64 here, I think its because summing over an i64 can overflow the container

rust/rsc/src/database.rs
V-FEXrt: Yep, could also use `BigInt` here since its aliases `INT8` if you want

rust/rsc/src/database.rs
V-FEXrt: yep! hits starts at 0 and is only incremented when a "cache hit" is returned from the cache lookup

rust/rsc/src/database.rs
V-FEXrt: I think there was an implicit cast in the subtable? Not entirely sure. I can try this one again without the inner cast but I remember having issues where everything was getting rounded down to 0

rust/rsc/src/database.rs
V-FEXrt: Yeah it is

rust/rsc/src/bin/rsc/types.rs
V-FEXrt: Yeah, `label` is non-optional in the database model so it picks up the default of empty string.

https://github.com/sifive/wake/blob/master/rust/entity/src/job.rs#L32

rust/rsc/src/database.rs
V-FEXrt: ah looks like the cast isn't actually needed here

rust/rsc/src/database.rs
V-FEXrt: Yeah it is!

rust/rsc/src/database.rs
```diff
@@ -249,7 +249,7 @@ pub async fn oldest_jobs<T: ConnectionTrait>(db: &T) -> Result<Vec<OldestJobs>,
         INNER JOIN job j
         ON j.hash = h.hash
         ORDER BY j.created_at
-        LIMIT 50
+        LIMIT 30
```
colinschmidt, line 252: This change makes me think it would be useful to be able to get more or less of these as part of the request parameters, but that seems like it would add a lot of complications, and is relatively low priority

rust/rsc/src/database.rs
```diff
@@ -272,7 +272,111 @@ pub async fn most_reused_jobs<T: ConnectionTrait>(db: &T) -> Result<Vec<MostReus
...
+           SELECT o.job_id, sum(b.size) as blob_size
+           FROM output_file o
+           INNER JOIN blob b
+           ON o.blob_id = b.id
+           GROUP BY o.job_id
+        ) b
+        INNER JOIN job j
+        ON j.id = b.job_id
+        INNER JOIN blob stdout
+        ON j.stdout_blob_id= stdout.id
```
colinschmidt, line 343: missing a space before the `=`

rust/rsc/src/database.rs
V-FEXrt: Yeah, I'd like to parameterize it eventually but I think for a first pass I'm happy with this being constant

PR #1599, opened 2024-07-18
rsc: Write logs to rolling files

V-FEXrt:
Previously logs were being dumped to stdout which meant there was quite limited lookback. This change updates the server to write logs to a rolling file with a new file being created for each day


rust/rsc/src/bin/rsc/main.rs
```diff
@@ -385,8 +385,9 @@ fn request_max_fileno_limit() {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // setup a subscriber for logging
-    let subscriber = tracing_subscriber::FmtSubscriber::new();
-    tracing::subscriber::set_global_default(subscriber)?;
+    let file_appender = tracing_appender::rolling::daily("./rsc_logs", "rsc.log");
```
V-FEXrt, line 388: Making a mental note that the log directory here should be another thing that is a config variable

colinschmidt (approved):
LGTM

PR #1601, opened 2024-07-18
rsc: Hotfix small rsc issues

V-FEXrt:
Fixes two small RSC issues

1) RSC dashboard failed with `div by zero` if a job doesn't output any files/stdout/stderr
2) running wake in `-v` mode would cause curl to way over spam the terminal


colinschmidt (approved):
LGTM

PR #1604, opened 2024-07-22
rsc: Fix filename escaping bug

V-FEXrt:
Filenames with uncommon characters just as `,` would previously fail to be cached. This change escapes the filename so that they can be cached.


colinschmidt (approved):
LGTM

ag-eitilt (commented):
Late review, but just to double-check, there's not going to be any further problems if any of the three strings contain `' ` (single-quote, space), right?  I know that the way Wake passes the argument through it'll be treated as a single token by the first script, but since that apparently passes it to something deeper in a way which winds up being a bit more sensitive, it seems like there might be opportunities for more breakage.  (Mainly asking because I have that exact bug as a known issue in a personal script at the moment.)

EDIT: Or for that matter, problems might arise even with just an odd number of single quotes between them, or *any* single quotes if any of them have to match the source string exactly (e.g. for file lookup).

V-FEXrt:
`'` will probably still break it yeah that's a harder thing to manage

PR #1606, opened 2024-07-25
rsc: Split configs for rsc/rsc_tool

V-FEXrt:
This is a simplification and split of the RSC config files in preparation for a larger push for config changes.

Unneeded configs and arg flags were removed and the `rsc` and `rsc_tool` were given dedicated config types


rust/rsc/src/bin/rsc/main.rs
```diff
@@ -502,6 +430,36 @@ mod tests {
         Ok(inserted_key.id)
     }
 
+    async fn create_insecure_api_key(
```
colinschmidt, line 422: Assuming these are just moves from above.

colinschmidt (approved):
LGTM

PR #1609, opened 2024-07-29
rsc: Convert various constants to config vars

V-FEXrt:
Many "I'll make that a config later" promises were made. This PR makes good on those promises and sets some reasonable defaults. LRU configs were added but are not yet used


colinschmidt (approved):
LGTM

PR #1613, opened 2024-07-30
lsc: Delete flaky and unused test

V-FEXrt:
The LSC isn't really used or supported anymore and the TTL test relies on time which has made it notorious flaky. Probably 65% of the CI failures we have come from this test flaking. 

I'm just going to delete the test since it doesn't give us anything and it actively harms productivity


ag-eitilt (approved):

colinschmidt (approved):
LGTM

PR #1614, opened 2024-07-31
rsc: Precalcuate job size as part of cron job

V-FEXrt:
Recalculating the job size every time it is needed is very expensive. This was less of an issue when only being used for dashboarding but the size is also needed for eventual LRU eviction. This PR sets up a cron task to calculate and store the size of the job after the job has been inserted so that the calculated value can be used directly


ag-eitilt (approved):
Rubber stamp.

colinschmidt (approved):
LGTM. Glad to see the same should_sleep loop used again.

PR #1615, opened 2024-07-31
rsc: Optimize blob eviction deletion

V-FEXrt:
- Adds index to blob(updated_at) which speeds up the TTL check for old blobs
- Completes the blob eviction lookup and delete in one query to not waste time repeating work.
- Makes delete query much more static instead of passing in 16k variables

### DIFF
--- a/rust/rsc/src/database.rs
+++ b/rust/rsc/src/database.rs
@@ -583,44 +583,6 @@ pub async fn upsert_blob<T: ConnectionTrait>(
 
 // ----------             Read             ----------
 
-// Reads blobs from the database that are unreferenced and have surpassed the allocated grace
-// period to be referenced.
-//
-// For new blobs this allows the client to create several blobs and then reference them all at
-// once. Existing blobs whose job was just evicted will likely be well past the grace period and
-// thus quickly evicted themselves.
-pub async fn read_unreferenced_blobs<T: ConnectionTrait>(
-    db: &T,
-    ttl: NaiveDateTime,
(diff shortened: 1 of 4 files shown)


ag-eitilt (approved):
Definitely seems like a reasonable optimization.  I imagine we used to do something else with the blobs between the fetch and the deletion?

rust/rsc/src/database.rs
```diff
@@ -645,37 +607,45 @@ pub async fn total_blob_size<T: ConnectionTrait>(db: &T) -> Result<Option<TotalB
...
             )
-            .exec(db)
-            .await?;
-
-        affected += result.rows_affected;
-    }
-
-    Ok(affected)
+            DELETE from blob b
+            WHERE b.id IN (SELECT id FROM eligible_blob_ids)
```
colinschmidt, line 642: Is it not possible to just do `IN (eligible_blob_ids.id)` or something like that? I'm surprised to see a select is needed again.

colinschmidt (approved):
LGTM

rust/rsc/src/database.rs
V-FEXrt: Nope :( I tried that. As fair as I could tell this didn't actually effect the query plan since `WITH` is kind of just writing an alias for a subquery

V-FEXrt:
>  I imagine we used to do something else with the blobs between the fetch and the deletion?

Not really. It was previously just split into two queries for simplicity reasons

PR #1616, opened 2024-07-31
rsc: Fix type for hidden_info

V-FEXrt:
Hidden info was suppose to be a `string` not an `ezblob`. It wasn't really hurting anything besides using more space but since db breaking changes are going in anyways it should be fixed


colinschmidt (approved):
LGTM

PR #1617, opened 2024-07-31
rsc: Tier blobs into nested dirs to avoid file limits

V-FEXrt:
Instead of dumping all blobs into the root of the local blob store, nest them into sub folders to avoid hitting files limits

### DIFF
--- a/rust/rsc/src/bin/rsc/blob_store_impls.rs
+++ b/rust/rsc/src/bin/rsc/blob_store_impls.rs
@@ -11,14 +11,25 @@ use tokio_util::bytes::Bytes;
 use tokio_util::io::StreamReader;
 
 fn create_random_blob_path() -> std::path::PathBuf {
-    let mut parts = [0u8; 32];
+    // 2 deep @ 8 bytes wide
+    let mut parts = [0u8; 10];
     OsRng.fill_bytes(&mut parts);
+
     let mut buf = std::path::PathBuf::from("");
-    for part in parts {
+
+    // First 2 bytes represent the containing directories
+    for i in 0..2 {
         let mut s = String::new();
-        write!(&mut s, "{:02X}", part).unwrap();
+        write!(&mut s, "{:02X}", parts[i]).unwrap();
         buf.push(s);
     }
+
+    // Next 8 bytes represent the file name
+    let mut s = String::new();
+    for i in 2..10 {
+        write!(&mut s, "{:02X}", parts[i]).unwrap();
+    }
+    buf.push(s);
 
     return buf;
 }
--- a/rust/rsc/src/bin/rsc/blob_store_impls.rs
+++ b/rust/rsc/src/bin/rsc/blob_store_impls.rs
@@ -1,20 +1,26 @@
 use crate::blob::*;
 use async_trait::async_trait;
-use data_encoding::BASE64URL;
 use futures::stream::BoxStream;
 use rand_core::{OsRng, RngCore};
 use sea_orm::prelude::Uuid;
+use std::fmt::Write;
 use tokio::fs::File;
 use tokio::io::AsyncReadExt;
 use tokio::io::BufWriter;
 use tokio_util::bytes::Bytes;
 use tokio_util::io::StreamReader;
 
-fn create_temp_filename() -> String {
-    let mut key = [0u8; 16];
-    OsRng.fill_bytes(&mut key);
-    // URL must be used as files can't contain /
-    BASE64URL.encode(&key)
+fn create_random_blob_path() -> std::path::PathBuf {
+    let mut parts = [0u8; 32];
+    OsRng.fill_bytes(&mut parts);
+    let mut buf = std::path::PathBuf::from("");
+    for part in parts {
+        let mut s = String::new();
+        write!(&mut s, "{:02X}", part).unwrap();
+        buf.push(s);
+    }
+
+    return buf;
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


rust/rsc/src/bin/rsc/blob_store_impls.rs
```diff
@@ -1,20 +1,26 @@
...
 use tokio_util::bytes::Bytes;
 use tokio_util::io::StreamReader;
 
-fn create_temp_filename() -> String {
-    let mut key = [0u8; 16];
-    OsRng.fill_bytes(&mut key);
-    // URL must be used as files can't contain /
-    BASE64URL.encode(&key)
+fn create_random_blob_path() -> std::path::PathBuf {
+    let mut parts = [0u8; 32];
```
V-FEXrt, line 14: Getting the depth right here is pretty important. I kind of randomly chose 32 but we should probably think about it a little bit.

We can't have the number be too small because we risk hitting the birthday paradox but we also don't want it too large because the storage overhead comes into play. (ex: the key per blob stored in the db is larger _and_ the linux file system overhead is much larger). 

One kind of annoying thing is that evicting a blob still leaves the probably completely empty path of directories to leading to it. We might be able to delete the directories when we delete the file but figuring out which ones are safe to delete is a non-trival tree walk. Otherwise we may be able to get away with a cleanup cron task

colinschmidt: Is it worth considering combining these strategies? A smaller number of directories plus a longer filename to avoid any collisions at the leaves? I feel like you could get away with just 4 or 8 directory levels then which would gives you 2^32  or 2^64 leaf directories so you won't hit the maximum number of files in a directory until we are running into other limits on the system.
It will be hard to delete the directories because they will be much less sparse but also a lower overhead.
Doing some math on the overhead assuming 4096 bytes/directory a fully populated 4 layer structure is more than 17TB. A fully populated 3 layer structure takes only 68GB and still lets us have over 83 trillion files, which again I think puts us safely in the realm of "having other problems".

V-FEXrt: Yeah, that's a pretty good tradeoff imo. Do you have a preference between 4 and 8 layers? Given that we made it this far with just 1 layer @ 16 byte filenames I think we could probably get away with 4 layers @ 4 byte filenames. I don't think making the filename any longer than 4 bytes helps since 256^4 = 4294967296 which is already almost 900 times the max number of files. 

If we want to be cautious, 8 @ 4  is probably something we'd never be able to fully fill


Aside: I'm realizing the super smart algorithm here would be something huffman encoding adjacent. Since directory names and file names are different there isn't any reason that all the files have to be at the leaf directory. basically every directory along the chain can hold ~6 milllion - 256 (one lot for each child directory) number of files. If we did it that way we could certainly get away with 4 @ 4 but we'd need to be much more careful about picking and up keeping file names

V-FEXrt: I wrote a quick little wake script to see how quickly we hit the birthday paradox with this. We might want to play with the numbers a bit. 4 byte file names and _specific_ directory is more likely than not to see a duplicate after 77164 files. That isn't the full picture since we don't consider the likelihood of a given file getting picked for a specific directory. In the worst case (assuming every file gets placed at depth = 1) that's 1/256 for a file to land in the directory so we'd (quite conservatively) expect a collision after 9 million blobs which doesn't seem far fetched for us.

Going to 5 byte filenames makes it 158 million. 6 is taking a while to actually calculate will update with that later

```
$ cat do.wake 
package do
    
target birthday (n: Double) (y: Double) =
    if n -. 1.0 <=. 0.0001 then
        0.0
    else
        def np = n -. 1.0
        def nm1 = birthday np y
        nm1 +. ((np *. (1.0 -. nm1)) /. y)

def searchWhenLikely n y =
    def chance = birthday n y

    if chance >=. 0.5 then
        n
    else 
        searchWhenLikely (n +. 1.0) y
    

export def do _ =
    def slots = 256 ^ 4

    searchWhenLikely 1.0 (dint slots) 
$ wake --in do do
77164e0
```

aside: writing out recursive formulas in wake is always very pleasant and fun to do

V-FEXrt: Final thoughts: 4 byte file names is *probably* enough and 5 byte filenames is certainly enough

colinschmidt: The filename length is one of the cheaper things to play with here so I am fine with your scheme but would suggest just making file names 8 bytes.
I'm also not convinced the random depth allocation is worth the complexity because there are 256 times more directories at each level the vast majority of the file storage is in the final layer anyways.

V-FEXrt: Decided offline to go with 2 @ 8 when we can trivially upgrade to 3 @ 8 if we fill up the 2 slots

rust/rsc/src/bin/rsc/blob_store_impls.rs
```diff
@@ -36,20 +53,29 @@ impl BlobStore for LocalBlobStore {
...
+            Err(err) => {
+                tracing::error!(%err, %written, "Size overflows i64, setting to i64::MAX instead");
                 i64::MAX
             }
             Ok(size) => size,
         };
 
+        let key = match rel_path.into_os_string().into_string() {
+            Err(path) => {
+                tracing::error!("Cannot convert path to string, returning loosy path instead");
```
colinschmidt, line 73:
```suggestion
                tracing::error!("Cannot convert path to string, returning lossy path instead");
```

colinschmidt (approved):
I'm surprised we didn't have to update the tests more. I guess we don't have a test for the specific path we use which is probably fine.

PR #1618, opened 2024-08-01
rsc: Refine curl escape

V-FEXrt:
The previous escape changes regressed small blob handling. This PR refines the escape so that it works for both small/blob and files with `,` in them


colinschmidt (approved):
LGTM

PR #1621, opened 2024-08-02
rsc: Enable RSC to build with rust 1.8

V-FEXrt:
One of the RSC's dependencies doesn't build against rust 1.8 which GHActions is now using by default. This PR updates the deps so that CI can pass again

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -96,13 +96,13 @@ jobs:
       strategy:
         matrix:
           include:
-            - target: alpine
-              dockerfile: alpine
-              extra_docker_build_args: ''
-              extra_docker_run_args: ''
-              build_cmd: make -C wake-* static
-              test_cmd: tar xvJf wake-*/wake-static_* && cd tests && ../wake-*/bin/wake -x Unit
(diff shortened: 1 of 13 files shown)


ag-eitilt (approved):
Looks like this doesn't update `rust/log_viewer/Cargo.lock`, is that deliberate?  It's falling a bit behind on version pinning (e.g. it asks for `mio` v0.8.8, where this bumps that to both v0.8.11 or v1.0.1, depending on where the dependency occurs) but I don't know if it would actually cause the same issues we're running into with CI.

V-FEXrt:
> is that deliberate

nope, just forgot about it

Though looks like rocky8 requires an older version of rust? this might be a pain

ag-eitilt:
Yeah, that will definitely make things interesting.  And the only reason I noticed the missing file was because I went through the list of dependabot warnings to see what this would clear up (answer: a fair number of the Rust ones, but since the majority of issues are with JavaScript we'll still definitely have a good few left).

V-FEXrt:
> but since the majority of issues are with JavaScript

I swear one day I'm going to remake the VSCode extension now that I know the proper way to do it

PR #1622, opened 2024-08-02
rsc: Fix dashboard for jobs with no file outputs

V-FEXrt:
Some jobs don't have any stdout/stderr/file outputs which causes divide by zero bugs in the dashboard. This PR fixes some of those cases


colinschmidt (approved):
LGTM

PR #1623, opened 2024-08-05
rsc: Log failures to upload blob

V-FEXrt:
We've observed blob uploads that silently failed which caused a delay is response time. This updates the `POST /blob` route to log an error when blob upload fails for some reason

### DIFF
--- a/rust/rsc/src/bin/rsc/blob.rs
+++ b/rust/rsc/src/bin/rsc/blob.rs
@@ -66,16 +66,18 @@ pub async fn create_blob(
             })))
             .await;
 
-        if let Err(msg) = result {
-            return (
-                StatusCode::INTERNAL_SERVER_ERROR,
-                Json(PostBlobResponse::Error {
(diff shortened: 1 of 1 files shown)


colinschmidt (approved):
It seems like there is still one more failure case we aren't logging anything for in the multipart section. Otherwise looks good.

colinschmidt:
> Ah that's intentional. That only happens when the client make a bad request, its nothing related to the server.
> 
> My thought there was that we shouldn't allow a misbehaving client to spam GBs of data into the server logs

So that error would show up in the client log? If so thats ok.

V-FEXrt:
> So that error would show up in the client log? If so thats ok.

Yep. It's pretty much a "client developer only" error because its the kind of thing that once you get it right the code never changes and you never hit that failure case again. Its the same kind of error as checking the shape of the `POST`'d JSON. You might have the client shape wrong a few times during development but once you get the shape right its always correct until the server changes it

PR #1624, opened 2024-08-05
rsc: Add blob_id indexes for fks

V-FEXrt:
Blob eviction queries were running unacceptably slow. This is because the FK checker had to linear search over the very large `output_file` table to ensure a deleted job wasn't reference. Adding an index converts the linear search into a much faster index lookup


colinschmidt (approved):
LGTM

PR #1626, opened 2024-08-07
rsc: Consider job before blindly caching

V-FEXrt:
There are a few reasons do have a precheck before pushing a job to the cache
- A job might become cached while being ran, uploading it is then a waste of resources (original purpose of this PR)
- A job might not be all that useful to cache so its eaiser to just skip/rebuild later
- The server may be overloaded and would prefer less requests, this provides an opportunity for it to shed load

This PR implement those reasons into a route that the client should check before pushing a new job to the cache


rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -203,3 +206,58 @@ pub async fn read_job(
...
+    let current_load = match system_load.read() {
+        Ok(lock) => *lock,
+        Err(err) => {
+            tracing::error!(%err, "Unable to lock system load for reading. Returning target load");
+            target_load
+        }
+    };
+
+    // Determine the chance of shedding the job and clamp to 0.0-1.0
+    let mut shed_chance = current_load / target_load - 1.0;
```
colinschmidt, line 250: Do we have a guard against `target_load == 0` somewhere?

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -203,3 +206,58 @@ pub async fn read_job(
...
+    if payload.runtime < 0.1 {
+        return StatusCode::NOT_ACCEPTABLE;
+    }
+
+    // Statistically reject jobs when the system should shed load
+    let current_load = match system_load.read() {
+        Ok(lock) => *lock,
+        Err(err) => {
+            tracing::error!(%err, "Unable to lock system load for reading. Returning target load");
+            target_load
```
colinschmidt, line 245: I wonder if there is some edge case here where the machine becomes to loaded to read this and then gets stuck accepting all requests. Perhaps in the failure to read we should assume current_load requires some shedding?

rust/rsc/src/bin/rsc/types.rs
```diff
@@ -122,6 +122,51 @@ impl ReadJobPayload {
...
+    pub status: i32,
+    pub runtime: f64,
+    pub cputime: f64,
+    pub memory: u64,
+    pub obytes: u64,
+    pub label: String,
+}
+
+impl AllowJobPayload {
+    // TODO: Figure out a way to de-dup this with AddJobPayload somehow
```
colinschmidt, line 145: 😬 Looks to be a clean copy paste right now

colinschmidt (commented):
Most of the thinking is about when we want to see the different exits from allow_job in the server log.
We won't be able to record how many jobs were too short to be included or how many were duplicates without some sort of breadcrumb. However for the too short exit condition we would end up spamming the log since all of those jobs will no longer be cached. The duplicate case seems potentially useful to log.
Finally in the high load case we are probably doing the right thing by not logging exactly when we shed a request just that we are very likely too at the moment. 
So I think the tracing summary is I would want to see an info/warn trace on the `CONFLICT` status return.

rust/rsc/src/bin/rsc/types.rs
V-FEXrt: Its a bit trickier than it seems on the surface

The 3 types (Post, Read, Allow) are all subtly different but they do all have the necessariy members to hash _just_ the job key. 

We can't easily have a regular function that does it since they are all different types. Using something like a `trait` would work but IIRC traits only allow you to specify methods so we'd be moving the boilerplate to a bunch of hand copied `fn get_foo_bar()` getter functions.

One thing that would probably work but I didn't consider until now is a generic function but I haven't tried that out yet. It does get rid of the super nice `object.hash()` api we currently have though

rust/rsc/src/bin/rsc/read_job.rs
V-FEXrt: how do you feel about making a 50% load shed chance in that case?

V-FEXrt:
> We won't be able to record how many jobs were too short to be included or how many were duplicates without some sort of breadcrumb. However for the too short exit condition we would end up spamming the log since all of those jobs will no longer be cached

We could maybe a few more columns to the job history table. Then we'd have `hits`, `misses`, `denied`, `conflicted`, `load_shed` or something along those lines. We already have the `hash` here so it doesn't cost much (besides slightly more load and extra disk) to track

> Finally in the high load case we are probably doing the right thing by not logging exactly when we shed a request just that we are very likely too at the moment.

Yeah I want to avoid too much extra work when under load. I'm actually somewhat tempted to temper the warning even further since if we are under load that print is likely be triggered a very large amount. Tempted to do something like 

```
if shed_chance > 0.5 && gen_bool(0.1) { LOG }
```

which would have a 10% chance of showing the the "heavy load" log. Another improvement would be to add a dashboard graph for "chance to shed load"

> So I think the tracing summary is I would want to see an info/warn trace on the CONFLICT status return.

On it

rust/rsc/src/bin/rsc/read_job.rs
colinschmidt: That seems reasonable to me.

colinschmidt:
DB tracking for these exit conditions sounds great and shouldn't be too expensive 🤞 
Yeah only showing the load shedding message sometimes sounds good to me.

rust/rsc/src/bin/rsc/types.rs
colinschmidt: The current boilerplate is at least much easier to check.

colinschmidt (approved):
LGTM

PR #1627, opened 2024-08-08
ci: Fix broken release

V-FEXrt:
When disabling alpine I missed turning off the download which broke the release pipeline. This should fix the pipeline while we work on fixing the build


colinschmidt (approved):
LGTM, based on what we discussed earlier.

PR #1628, opened 2024-08-09
rsc: Add config var for min runtime to cache

V-FEXrt:
Allows the config to specify what the minimum runtime should be for a cache job. Any job that completes faster than that runtime isn't worth caching due to various overheads


colinschmidt (approved):
LGTM, seems like you need to update the unit test

PR #1629, opened 2024-08-12
rsc: Recover from missing blobs

V-FEXrt:
The client side of the RSC is required to be very tolerant to all sorts of faults. One that was always planned but somehow missed was being tolerant of a failure to download the blob listed as part of the cache hit. This PR fixes/implements that missing feature.

Main things of note:
  - We `touch` the destination after `rm`ing in in the http library so the curl job always "outputs" a file to be hashed
  - When any part of the job hydration process fails we fall back to manually running the job without attempting to cache it


share/wake/lib/system/http.wake
```diff
@@ -303,15 +303,17 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
     #   2. change the permissions of that new path to write-protected
     # which is then refected in the permissions of the path generated by this function preventing
     # it from being overwritten. To work around this we need for force remove the file before
-    # downloading over the path.
+    # downloading over the path. After removing the file, it is touched so that the job always
+    # outputs a file even if curl fails to download the target. Without this wake-hash fails to
+    # hash a non-existant file.
```
colinschmidt, line 308:
```suggestion
    # hash a non-existant file, in the curl failure case.
```

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -59,6 +59,167 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: Result RunnerIn
 
         Pass output
 
+    def rehydrateJob response label input job =
```
colinschmidt, line 62: This looks like a clean refactor into a new function, but I did read through it again just in case anything else stood out.

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -106,168 +267,11 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: Result RunnerIn
...
-                    outputSymlinksDownload
-                    | findFail
-                    | addErrorContext "rsc: Failed to create a symlink"
-
-                Pass (RunnerOutput inputs outputs predict)
+                match (rehydrateJob response label input job)
+                    Pass x -> Pass x
+                    # If the job hydration fails for any reason just run the job as normal.
+                    # There is no point in attempting to push since the server just said its cached
+                    Fail _ -> baseDoIt job (Pass input)
```
colinschmidt, line 274: Should we not call `runJobAndUpload` here? Or do we think that will be problematic if the job is currently in a state where its matching cache response can't be rehydrated?

colinschmidt (approved):
LGTM except for one question in the `else`

share/wake/lib/system/remote_cache_runner.wake
V-FEXrt: It would be strictly wasted work. Since we are inside the "match" case then the server already thinks it has the job. If we then go and repost it we'll just get rejected by the `allowed` route

Not a whole lot we can do in this case but maybe would be nice for the client to somehow tell the server that the specific job is acting up

PR #1630, opened 2024-08-13
ci: Disable VSCode Extension

V-FEXrt:
_extremely temporary_

Disables VSCode Extension build to unblock critical RSC fixes. Immediately after the release  work will start to fix the extension build


.github/workflows/build.yaml
```diff
@@ -345,7 +345,6 @@ jobs:
             repo_token: '${{ secrets.GITHUB_TOKEN }}'
             files: |
               tarball/wake_*.tar.xz
-              vscode/wake-*.vsix
```
colinschmidt, line 348: Don't we need to disable the "download vscode" portion as well?

colinschmidt (approved):
LGTM

PR #1634, opened 2024-08-21
rsc: Allow file blobs in DbOnly Store

V-FEXrt:
Allows generic blobs to also be created as DbOnly blobs, including the deduping feature.

This change does require that the blob be `stat`d but that should essentially be free because `stat` is a `target` and we have to stat every file before upload anyways

### DIFF
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -316,11 +316,7 @@ export def makeRemoteCacheApi (config: String): Result RemoteCacheApi Error =
 #  (RemoteCacheApi "foo" 1 None) | rscApiPostStringBlob "foo" "my foo contents" = Fail "authorization required"
 # ```
 export def rscApiPostStringBlob (name: String) (value: String) (api: RemoteCacheApi): Result String Error =
-    def contentType =
-        if value.strlen < 95 then
-            Some "blob/small"
-        else
-            None
+    def contentType = contentTypeFromSize value.strlen
 
     require Pass temp = writeTempFile name value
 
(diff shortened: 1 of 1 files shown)


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -631,6 +632,10 @@ export def rscApiGetFileBlob ((CacheSearchBlob _ uri): CacheSearchBlob) (path: S
     Pass path
 
 ## --- Helper functions ---
+
+def contentTypeFromSize size =
+    if size < 95 then Some "blob/small" else None
```
colinschmidt, line 637: Did we determine ~100 was the right limit still?

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -336,7 +332,12 @@ export def rscApiPostStringBlob (name: String) (value: String) (api: RemoteCache
 export def rscApiPostFileBlob (name: String) (file: String) (api: RemoteCacheApi): Result String Error =
     # We must use unsafe here since we cannot elevate *file* to a Path without either copying it
     # or triggering the 'job output by multiple files' error.
-    uploadBlobRequest api (unsafe_addFormData name file None)
+
+    def contentType = match (unsafe_stat file)
+        Pass (Stat _ _ size) -> contentTypeFromSize size
+        Fail _ -> None
```
colinschmidt, line 338: Wondering if there is some benefit to failing earlier here rather than waiting for the `uploadBlobRequest` to fail if it can't stat the file?

colinschmidt (approved):
LGTM. A few small questions but nothing blocking.

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: nope, we talked about 256 but I'm kicking that down the road

share/wake/lib/system/remote_cache_api.wake
V-FEXrt: Maybe, I'm not sure which one happens first but I'll at it to the list of things to give a look

PR #1635, opened 2024-08-21
rsc: Optimize blob eviction query

V-FEXrt:
`UNION` requires that duplicate ids be removed which requires a sort + unique linear scan. We were previously doing this on the largest table in the database. Instead its more performant if we move the id deduplication to the last step where the id list is as small as possible. This results in a query plan where the sort + unique linear scan happens on a much smaller set of rows


colinschmidt (approved):
LGTM

PR #1638, opened 2024-08-23
rsc: Support db only uri scheme for output files

colinschmidt:
Hoist the RSC URI scheme dispatch into a separate function and then invoke that in the get file code path as well.


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -533,78 +533,99 @@ export target rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error
         Some x -> failWithError "Invalid auth key. Status code: {str x}"
         None -> failWithError "Invalid auth key. Unable to determine status code"
 
+# Converts db only blob uris into actual data
+# maps path = "%46%6F%6F" to content = "foo"
+def rscDbBlobToContent path =
```
V-FEXrt, line 538: nit: Rename to something like `fetchFromDbScheme`

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -533,78 +533,99 @@ export target rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error
...
+        | findNone
+    else failWithError "rsc: Failed to parse bytes strings into bytes from db path: '{path}'"
+
+    bytes
+    | map integerToByte
+    | cat
+    | Pass
+
+# Wrap scheme dispatch behind uri interface
+def uriSchemeFn dbBlobFn dbFileFn: String => Result a Error =
```
V-FEXrt, line 562: I'm not quite happy with the structure of this function. I think the ordering is a bit confusing.

Is there a reason we can't reorder it? something like

```ruby
def unsupportedScheme = ...

require scheme, path, Nil = uri | extract `...`
else failWithError "..."

match scheme
  "db" -> Fn
  ...
  _ -> unsupprtedScheme
```


The other question I have, both places calling this function want a return type of `Result String Error` is there any reason we can't just call `dbBlobfn`/`dbFileFn` on their behalf?

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -533,78 +533,99 @@ export target rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error
...
 # ```
 export def rscApiGetFileBlob ((CacheSearchBlob _ uri): CacheSearchBlob) (path: String) (mode: Integer): Result String Error =
-    require Pass downloadPath =
-        buildHttpRequest uri
+    def dbBlobScheme _scheme uripath =
+        require Pass content = rscDbBlobToContent uripath
+
+        writeTempFile "rsc.output_file.blob" content
+
+    def dbFileScheme scheme uripath =
```
V-FEXrt, line 612: `content` here is the raw bytes from the return? Do we know for sure the bytes are valid for being in a linux path name? Would prefer if we hash it just to be safe

V-FEXrt: nvm I'm just tired, hashing is done for you

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -558,65 +558,56 @@ def rscDbBlobToContent path =
     | cat
     | Pass
 
-# Wrap scheme dispatch behind uri interface
-def uriSchemeFn dbBlobFn dbFileFn: String => Result a Error =
-    def unsupportedScheme scheme uri =
+# Dispatches the content resolution for uri based on its scheme
+def resolveUriResponse (dbSchemeFn: (String => String => Result a Error)) (fileSchemeFn: (String => String => Result a Error))  (uri: String) =
```
colinschmidt, line 562: much better name :)

colinschmidt (commented):
New operator uses and names look nice.
Changes LGTM, thanks for the cleanup.

share/wake/lib/system/remote_cache_api.wake
colinschmidt: I got caught up thinking we shouldn't call it because one wanted a "Path" as a string and the other wanted content as the string. But then I did all the work to make it polymorphic so there wasn't a reason not to just return the `Result a Error`. Thanks for the suggestion and fix.

V-FEXrt (approved):

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -533,78 +533,88 @@ export target rscApiCheckAuthorization (api: RemoteCacheApi): Result Unit Error
...
+
+    bytes
+    | map integerToByte
+    | cat
+    | Pass
 
-        Pass response.getHttpResponseBody
+# Dispatches the content resolution for uri based on its scheme
+def resolveUriResponse (dbSchemeFn: (String => String => Result a Error)) (fileSchemeFn: (String => String => Result a Error)) (uri: String) =
+    require scheme, path, Nil = (uri | extract `([a-zA-Z][a-zA-Z0-9+.-]*)://(.*)`)
```
ag-eitilt, line 563: There's a couple places where this formation is used, but it feels like using the function as normal might be easier to read than the parenthesized piping.
```suggestion
    require scheme, path, Nil = extract `([a-zA-Z][a-zA-Z0-9+.-]*)://(.*)` uri
```

ag-eitilt (approved):
I don't have enough context to review this in terms of everything else, but what I can tell of it all looks good.

PR #1640, opened 2024-08-30
stdlib: Significantly refactor the Runner api

V-FEXrt:
I'll edit this later with a longer description but wanted to kick of ci.

tl;dr:
 - breaking change
 - score is gone
 - makeRunner -> depreMakeRunner
 - new makeRunner takes a single run function instead of two split functions with convoluted apis


share/wake/lib/system/io.wake
```diff
@@ -130,20 +130,30 @@ export def read (path: Path): Result String Error =
         Fail f -> Fail (makeError f)
 
 target writeImp inputs mode path content =
+    # TODO: move this runner out of writeImp and don't directly depent on the parameters passed in to writeImp
+    #       See mkdirRunner for example of proper implementation
```
ag-eitilt, line 134: Is this a TODO you're wanting to get in as part of this PR, or as a followup?
```suggestion
    # TODO: move this runner out of writeImp and don't directly depend on the parameters passed in to writeImp
    #       See mkdirRunner for example of proper implementation
```

share/wake/lib/system/job.wake
```diff
@@ -15,41 +15,111 @@
...
+    export dir: String
+    export stdin: String
+    export env: String
+    export cmd: String
+    export signature: Integer
+    export visible: String
+    export isatty: Integer
+
+# Create/reserve a job handle, parameters aren't necessarily finalized
+export def primJobCreate (label: String) ((JobKey dir stdin env cmd signature visible isatty): JobKey) (keep: Integer) (echo: String) (stdout: String) (stderr: String): Job =
```
ag-eitilt, line 28: Likely personal preference (especially since these are internal-only functions), but it might be slightly cleaner to not deconstruct the JobKey on the top line:
```suggestion
export def primJobCreate (label: String) (jobKey: JobKey) (keep: Integer) (echo: String) (stdout: String) (stderr: String): Job =
    JobKey dir stdin env cmd signature visible isatty = jobKey

```

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -216,79 +213,72 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: Result RunnerIn
...
+            | rscApiFindMatchingJob (mkSearchRequest input hashKey)
 
-            # Search the cache for a match
-            def requestTask =
-                rscApi
-                | rscApiFindMatchingJob (mkSearchRequest input hashKey)
+        require Pass response = requestTask
+        else
+            require Fail err = requestTask
+            else unreachable "Result must be either Pass or Fail"
```
ag-eitilt, line 241: I know this is just deindenting the existing code rather than being something new, but since you're touching this already, and especially given how complex both branches are, mind refactoring it into a match with helper functions?
```suggestion
        match requestTask
            Pass response -> runMatchedJob response
            Fail err -> runUncachedJob err
```

share/wake/lib/system/runner.wake
```diff
@@ -55,97 +56,70 @@ export tuple RunnerOutput =
...
-
-export def makeRunner name score pre post (Runner _ _ run) =
-    def doit job preInput = match (pre preInput)
-        Pair runInput state ->
-            def runOutput = run job runInput
-            def final _ = post (Pair runOutput state)
+# makeRunner: Hides some of the boiler plate required to create a runner
+#
+# This function requires very advanced wake expierence and should be used with the greatest amount of caution.
+# Callers must ensure at the very least than run calls primJobLaunch and one of the many job "wait" functions.
```
ag-eitilt, line 64:
```suggestion
# This function requires very advanced wake experience and should be used with the greatest amount of caution.
# Callers must ensure at the very least that `run` calls primJobLaunch and one of the many job "wait" functions.
```

share/wake/lib/system/runner.wake
```diff
@@ -55,97 +56,70 @@ export tuple RunnerOutput =
...
+# Runners should not wrap other runners instead they should precisely do what they need to complete their task.
```
ag-eitilt, line 65: I'm not sure I agree with this line; it's certainly a summation of how this architecture differs from the old one, but at this point we shouldn't have to talk about runners wrapping others going forward, and someone coming at this without historical context might think we're forbidding sharing code between runners or something (that's not just strawmanning the reading, it's the only way my mind makes sense of what it's saying once I "forget" the history).

share/wake/lib/system/runner.wake
```diff
@@ -55,97 +56,70 @@ export tuple RunnerOutput =
...
 
-            Fail e
-        Pass (RunnerInput _ _ vis _ _ _ _ _ predict _) ->
-            def Usage status runtime cputime mem in out = predict
+    makeRunner "virtual" run
 
-            def _ =
-                virtual job "" "" status runtime cputime mem in out # sets predict+reality
+# depreMakeRunner: Deprecated. Do not use
+export def depreMakeRunner name pre post (Runner _ run) =
```
ag-eitilt, line 107: There's definitely arguments either way for making the deprecation obvious in the name, but if we're saying we want to deprecate this rather than deleting it immediately, I'd still prefer having a bit more descriptive a name of what it's actually doing.  The comment and conversation should be sufficient to let people know not to use it.
```suggestion
export def wrapRunner name pre post (Runner _ run) =
```

ag-eitilt (commented):
Assuming CI and general integration are happy, I greatly prefer this architecture.  Everything looks good at a first pass, but there's enough moved blocks of code that I'm going to have to use other methods of reading the diff to fully see what it's changed (the by-commit reading you mentioned, undoing indentation changes for better diffs, ...) so I don't want to click approve immediately -- let me know if you want to get this in sooner than I get around to looking at it at that level of detail, I don't expect to find anything unpleasantly surprising.

share/wake/lib/system/runner.wake
V-FEXrt: Ah sure. FWIW it probably only keeping this for one release. I've already written the internal changes to migrate off of this. Its more as a warning for anyone external who depends on it

share/wake/lib/system/runner.wake
V-FEXrt: Yeah fair. I guess its supposed to be more of a "word of caution"/"don't make the same mistake we did" type thing

share/wake/lib/system/io.wake
V-FEXrt: Either TBH. I'll probably kick to a different PR since this one is getting large

share/wake/lib/system/runner.wake
ag-eitilt: It definitely is a great word of caution to have, and after the weekend away I do see good use to keeping something like it in, I just think that we might want to rephrase it a bit to emphasize that it's a historical/advisory note rather than something which describes what is now the new natural inclination.

share/wake/lib/system/runner.wake
```diff
@@ -55,97 +56,70 @@ export tuple RunnerOutput =
...
-    def badlaunch job error = prim "job_fail_launch"
+    def run (job: Job) ((RunnerInput _ _ vis _ _ _ _ _ predict _): RunnerInput): Result RunnerOutput Error =
+        def _ = primJobVirtual job "" "" predict
 
-    def doit job = match _
-        Fail e ->
-            def _ = badlaunch job e
+        job
+        | getJobReality
+        |< (\reality RunnerOutput (vis | map getPathName) Nil reality)
```
ag-eitilt, line 102: Very minor, but given my functional instincts I'd be inclined to eta(?) reduce this.
```suggestion
        job.getJobReality
        |< RunnerOutput (vis | map getPathName) Nil
```

ag-eitilt (approved):
Yeah, that all looks good even after a deeper readthrough.

share/wake/lib/system/remote_cache_runner.wake
V-FEXrt: The two arms here aren't Hit vs Miss, its "Got a valid server response" vs "Got an invalid server response" and the two extracted functions in the match would be kind of silly imo. TBH this is something I **_really_** want `when` for

```python
    # If a failure occurs making the request to the server log it and exit
    when Fail err = requestTask
        # Always leave a breadcrumb since this should be a rare error.
        def _ = breadcrumb "{label}: Failed to search for job in the cache"

        # Leave detailed info if debugging is enabled
        def _ =
            require True = shouldDebugRemoteCache Unit

            def _ =
                writeTempFile "remote.cache.lookup.fail" "label: {input.getRunnerInputLabel}\nerror: {err | format}"

            True

        # This job isn't getting cached. That's probably preferable since the server
        # request failed but good to keep in mind.
        baseDoIt job (Pass input)

    require Pass response = requestTask

    # If a match was found use it
    require NoMatch = response
    else
        ... rehydrate the job from the cache ...

    ... run and cache the job ...
```

Another also kind of weird way to do this would be a call to `rmapFail`. I don't really see this branch as "do one of two things" instead its "do one thing, or on failure recover and log"

share/wake/lib/system/runner.wake
```diff
@@ -55,97 +56,70 @@ export tuple RunnerOutput =
 # A Runner describes a way to invoke a Plan to get a Job
 export tuple Runner =
     export Name: String
-    export Score: Plan => Result Double String
     Fn: Job => Result RunnerInput Error => Result RunnerOutput Error
```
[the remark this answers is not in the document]
V-FEXrt: Yeah, I mostly didn't change this because it increases the surface area of the refactor. Certainly something to consider for later

share/wake/lib/system/job_cache_runner.wake
```diff
@@ -55,199 +55,194 @@ def getPath input =
...
+
+        def _ =
+            require True = isDebugOn
+
+            def _ = write ".cache-misses/read.{prefix}.json" "//{label}\n{jobCacheJsonIn}"
+
+            True
+
+        # Now we need to run the job
+        require Pass (RunnerOutput inputs outputs useage) = baseDoIt job (Pass input)
```
colinschmidt, line 179:
```suggestion
        require Pass (RunnerOutput inputs outputs usage) = baseDoIt job (Pass input)
```

share/wake/lib/system/remote_cache_runner.wake
ag-eitilt: Yeah, `when` might be nice for this.  My view is just that (A) the `else` case here is large enough that it's beginning to get difficult to tell at a glance where it ends, and (B) that the series of two immediate `require`s and an `unreachable` looks a bit silly when the ultimate goal (connotations aside) is the same as an all-constructors `match`.  I think it probably comes down to how we individually think of `match`/`require` -- to my eye, `match` doesn't imply that its cases are all equally prominent, so having the Fail case just be recovery/logging is entirely natural; meanwhile I tend to look at `require` as something which should "return control to the enclosing function" (despite that being *entirely* the incorrect summary of what's behind the scenes) as quickly as possible, with neither the RHS of the `=` nor the `else` clause ideally containing any significant code.

Not a critical issue either way, though.

share/wake/lib/system/io.wake
```diff
@@ -129,22 +129,43 @@ export def read (path: Path): Result String Error =
...
 
-        def post = match _
-            Pair (Fail f) _ -> Fail f
-            Pair (Pass output) Unit ->
-                if mode < 0 || mode > 0x1ff then
-                    Fail (makeError "write {path}: Invalid mode ({strOctal mode})")
-                else match (imp mode path content)
-                    Fail f -> Fail (makeError f)
-                    Pass path -> Pass (editRunnerOutputOutputs (path, _) output)
+        # Wait for the virtual job to complete
```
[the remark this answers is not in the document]
V-FEXrt: No clue. I also noticed that but I wanted to keep it the same as the old implementation since it was a large refactor. Happy to test it out though

share/wake/lib/system/io.wake
```diff
@@ -129,22 +129,43 @@ export def read (path: Path): Result String Error =
...
+        require Pass reality = job.getJobReality
 
-        makeRunner "write" (\_ Pass 0.0) pre post virtualRunner
+        # Actually trigger the effects required by the job
+        def mode =
+            int smode
+            | getOrElse 0x200
```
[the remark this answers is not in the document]
V-FEXrt: This was just copied over to match what we had before. If I had to guess, this code predates `require` and was just never updated. Should be totally fine to make that change

share/wake/lib/system/io.wake
```diff
@@ -248,29 +269,39 @@ export def installIn (toRoot: String) (fromRoot: String) (sourcePath: Path): Res
...
-
-            if mode < 0 || mode > 0x1ff then
-                Fail (makeError "mkdir {dir}: Invalid mode ({smode})")
-            else match (imp mode dir)
-                Fail f -> Fail (makeError f)
-                Pass path -> Pass (editRunnerOutputOutputs (path, _) output)
-
-    makeRunner "mkdir" (\_ Pass 0.0) pre post virtualRunner
+    def primMkdir (mode: Integer) (path: String): Result String String =
+        (\_ \_ prim "mkdir") mode path
```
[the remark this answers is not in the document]
V-FEXrt: Same, I can never mentally parse the other style. IIRC they both are the exact same once parsed but the mapping never makes sense to me

share/wake/lib/system/io.wake
```diff
@@ -248,29 +269,39 @@ export def installIn (toRoot: String) (fromRoot: String) (sourcePath: Path): Res
...
+
+    def run (job: Job) ((RunnerInput _ cmd vis _ _ _ _ _ predict _): RunnerInput): Result RunnerOutput Error =
+        # Command must be ("<mkdir>", "-m", "{string mode}", "{string path}", Nil)
+        require "<mkdir>", "-m", smode, path, Nil = cmd
+        else panic "mkdirImp violated command-line contract"
+
+        # Insert the <mkdir> job into the database
+        def _ = primJobVirtual job "" "" predict
```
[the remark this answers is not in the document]
V-FEXrt: > is there some special case that virtual jobs don't count towards wake's consumption of system resources

Yep, virtual jobs immediately complete without anything ever being scheduled. In this case `predict` is what gets returned as `reality` when inspecting the job later

share/wake/lib/system/job.wake
```diff
@@ -15,41 +15,122 @@
...
+
+# Complete a job successfully by providing to the runtime the inputs/outputs/usage of the job
+export def primJobFinish (job: Job) (inputs: String) (outputs: String) (all_outputs: String) (usage: Usage): Unit =
+    def Usage status runtime cputime membytes ibytes obytes = usage
+
+    (\_ \_ \_ \_ \_ \_ \_ \_ \_ \_ prim "job_finish")
+    job
+    inputs
+    outputs
+    all_outputs
```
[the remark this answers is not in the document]
V-FEXrt: This is actually something I was trying to look into/figure out for the next PR because once we apply the filter at the runner level `outputs` and `all_outputs` will always be the same

PR #1645, opened 2024-09-06
rsc: Failing jobs shouldn't be cached

V-FEXrt:
Not sure how I forget this, but we can't cache jobs that fail. It wastes space but more importantly can "lock in" a flow failure for other users/flow runs


colinschmidt (approved):
LGTM

ag-eitilt (approved):
Yep, that'll be important.

PR #1646, opened 2024-09-09
stdlib: Move file filters into runners

V-FEXrt:
Moves the fnInput/fnOutput functions into the runners so they can directly control them. Allows things like the cache runners to cache only actual output files

### DIFF
--- a/share/wake/lib/system/job_cache_runner.wake
+++ b/share/wake/lib/system/job_cache_runner.wake
@@ -60,7 +60,7 @@ export def mkJobCacheRunner (hashFn: RunnerInput => Result String Error) (wakero
     def job_cache_add str = prim "job_cache_add"
 
     def run (job: Job) (input: RunnerInput): Result RunnerOutput Error =
-        def (RunnerInput label cmd vis env dir stdin _ prefix _ _) = input
+        def (RunnerInput label cmd vis env dir stdin _ prefix _ _ _ _) = input
 
         def mkVisJson (Path path hash) =
             JObject (
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -47,6 +47,12 @@ export tuple RunnerInput =
     export Record: Usage
     # Determines if job should run in psuedoterminal
     export IsAtty: Boolean
+    # Modify the Runner's reported inputs (files read).
+    # May only be called once for a given job. Wrapping runners are required to maintain this invariant
+    export FnInputs: (List String => List String)
+    # Modify the Runner's reported outputs (files created).
+    # May only be called once for a given job. Wrapping runners are required to maintain this invariant
+    export FnOutputs: (List String => List String)
 
 export tuple RunnerOutput =
     export Inputs: List String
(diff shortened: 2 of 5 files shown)


share/wake/lib/system/runner.wake
```diff
@@ -47,6 +47,12 @@ export tuple RunnerInput =
     export Record: Usage
     # Determines if job should run in psuedoterminal
     export IsAtty: Boolean
+    # Modify the Runner's reported inputs (files read).
+    # May only be called once for a given job. Wrapping runners are required to maintain this invariant
```
colinschmidt, line 51: I think this "May" implies it could be run 0 or 1 times, but it really should be called exactly once.

share/wake/lib/system/job_cache_runner.wake
```diff
@@ -60,7 +60,7 @@ export def mkJobCacheRunner (hashFn: RunnerInput => Result String Error) (wakero
     def job_cache_add str = prim "job_cache_add"
 
     def run (job: Job) (input: RunnerInput): Result RunnerOutput Error =
-        def (RunnerInput label cmd vis env dir stdin _ prefix _ _) = input
+        def (RunnerInput label cmd vis env dir stdin _ prefix _ _ _ _) = input
```
colinschmidt, line 63: Is it valid not to use `fnInputs` or `fnOutputs`? Is that not part of the runner's contract with the `Plan` and `Job`?
Or are we not using them here because this  function calls another `run` function that does call those input/output functions?

share/wake/lib/system/runner.wake
ag-eitilt: That's a curious difference between "may" and "may *only*" -- with the right emphasis, "*may* only" could even be saying that it's expected to be multiple times, but in certain cases it could get as low as one.  Probably why there's an entire RFC dedicated to defining the word for use in specifications.

colinschmidt: https://datatracker.ietf.org/doc/html/rfc2119

I suggest "Must" but you could use "Shall" if you want some spice in your vocabulary
And maybe include "exactly once"

ag-eitilt: Also, we should probably discuss *why* the invariant's required somewhere, and that the proper way to modify the list when wrapping a runner is to modify the function being passed down.

ag-eitilt (approved):
Seems like a reasonable thing for it to own.  Slight documentation expansion would be nice, but the code seems good.

share/wake/lib/system/job_cache_runner.wake
V-FEXrt: > Is it valid not to use fnInputs or fnOutputs?

Nope, it must always be called at least once. In this case, the cache runners don't call it because the runner they are wrapping is expected to call it

colinschmidt (approved):
LGTM

colinschmidt (approved):
LGTM

ag-eitilt (commented):
How does the cleanable list interact with the FnOutputs, for existing code?  I definitely prefer how much more explicit this setup is, but I want to be sure that existing code which obviously doesn't interact with CleanableOutputs doesn't break, seeing as the function only seems to be applied as `def fileOutputs = fnOutputs cleanable` in the runners:

https://github.com/sifive/wake/blob/05d35baefada9d0a35abbc7448243d2fb2b1a58f/share/wake/lib/system/path.wake#L278-L293

And, actually, do we want to introduce a new Plan.FnCleanable to provide end users with more control over that list as well?

colinschmidt:
> How does the cleanable list interact with the FnOutputs, for existing code? I definitely prefer how much more explicit this setup is, but I want to be sure that existing code which obviously doesn't interact with CleanableOutputs doesn't break, seeing as the function only seems to be applied as `def fileOutputs = fnOutputs cleanable` in the runners:

Existing code calls `CleanableOutputs` `All_Outputs` or something and is always populated with the unfiltered outputs in `runAlways`. User code that implements its own runner will now have to fill in this field in the RunnerOutput to update to this version of wake which should not be a high burden given they are maintaining their own Runner already.

> And, actually, do we want to introduce a new Plan.FnCleanable to provide end users with more control over that list as well?

I think this is a good idea but needs to be carefully described as something even less likely the user wants to play with than Plan.FnOutputs.

V-FEXrt:
> How does the cleanable list interact with the FnOutputs, for existing code? I definitely prefer how much more explicit this setup is, but I want to be sure that existing code which obviously doesn't interact with CleanableOutputs doesn't break

Yep its a tiny bit subtle but the tl;dr is that previously `Runners` were required to return "All outputs" that they knew about then `runAlways` did two things. 1. Pass the unfiltered list of all outputs into the now renamed but identical "AllOutputs" 2. run fnOuputs over the all outputs list and pass that into "outputs". Now the runner directly controls both lists but trivially does the exact same thing.

One subtle and interesting thing is that "AllOutputs" was a very strong misnomer since that list is _always_ empty for the localRunner while the "filtered" outputs of the local runner has the actual discovered outputs. So in the runtime we actually get something like 

```
cleanable_outputs = "all outputs as reported by runner"
hashed_real_outputs = FnOutputs cleanable_outputs (which make increase the size of the list)
unhashed_real_cleanable = cleanable_outputs - hashed_real_outputs
```
So the only things actually processed separately are items that are in cleanable_outputs but not in "filtered" outputs


> And, actually, do we want to introduce a new Plan.FnCleanable to provide end users with more control over that list as well?

Lets punt on that for a bit. I can see it being useful but this change is already large, not one is currently asking for it, and its rather hard to explain _what_ should go into it (esp since the cache won't store files in that list).

PR #1648, opened 2024-09-19
stdlib: Add list functions

V-FEXrt:
I'm working on fixes to breakages in `v44.0.0` and it felt better to split things up instead of trying to make one larger disparate PR.

This PR adds list `intersect` and `subset` both of which are needed for the upcoming fixes. Care was taken to ensure they didn't run `N^2`. They roughly run in `2*nlogn + n` where `n = max(lhs.len, rhs.len)` though I am assuming `sortBy` is `nlogn` which I've not verified


share/wake/lib/core/list.wake
```diff
@@ -889,3 +889,62 @@ export def groupBy (cmpFn: a => a => Order) (list: List a): List (List a) =
...
+# Examples:
+# ```
+#   intersect icmp (seq 3) (seq 4) = (0, 1, 2, Nil)
+#   intersect icmp (seq 3) (seq 3) = (0, 1, 2, Nil)
+#   intersect icmp (seq 3) (4, 2, 0, Nil) = (0, 2, Nil)
+#   intersect scmp ("", "bar", "bat", Nil) ("", "foo", "bat", Nil) = ("", "bat", Nil)
+# ```
+export def intersect (cmpFn: a => a => Order) (lhs: List a) (rhs: List a): List a =
+    def lhsSort = sortBy cmpFn lhs
+    def rhsSort = sortBy cmpFn rhs
```
ag-eitilt, line 910: You probably weighed this already, but we could also maintain the ordering according to one of the lists by turning the other into a Tree and testing membership; at that point we've got `O((n+m) log n)` or thereabouts, which is probably not too bad of a complexity, either.  Though if we still want to limit repeated members (and I do think we do) with that implementation, we'd probably need to go through Map with the value being a counter, which would be a bit trickier than the very straightforward head-comparison you're able to use here.

ag-eitilt (approved):

share/wake/lib/core/list.wake
V-FEXrt: I think its worth keeping the simpler implementation here. Its not typical for intersect to maintain order and I don't know of any code I've seen that needs both order and intersect.

The other question is _which_ order is correct. Both lists could have completely different orders so you have to arbitrarily pick on to maintain order

ag-eitilt: > I don't know of any code I've seen that needs both order and intersect.

I can come up with some examples of where that might be helpful, but I don't know if I've actually *seen* it in practice, either.  So, yeah, probably not necessary.

> The other question is *which* order is correct. Both lists could have completely different orders so you have to arbitrarily pick on to maintain order.

This isn't a strong enough argument to block anything (particularly given Wake's propensity for pipelines -- it would be the order of the second list), but since there's not a strong argument for maintaining order, I do agree it's moot.

share/wake/lib/core/list.wake
```diff
@@ -900,26 +900,25 @@ export def groupBy (cmpFn: a => a => Order) (list: List a): List (List a) =
...
+#   intersect icmp (seq 3) (seq 3) = 0, 1, 2, Nil
+#   intersect icmp (seq 3) (4, 2, 0, Nil) = 0, 2, Nil
+#   intersect scmp ("", "bar", "bat", Nil) ("", "foo", "bat", Nil) = "", "bat", Nil
 # ```
 export def intersect (cmpFn: a => a => Order) (lhs: List a) (rhs: List a): List a =
-    def lhsSort = sortBy cmpFn lhs
-    def rhsSort = sortBy cmpFn rhs
-
-    match lhsSort rhsSort
+    def intersectImp cmpFn lhsSort rhsSort = match lhsSort rhsSort
```
ag-eitilt, line 909: Ah, whoops, yes, that's very important.

share/wake/lib/core/list.wake
```diff
@@ -938,13 +937,44 @@ export def intersect (cmpFn: a => a => Order) (lhs: List a) (rhs: List a): List
...
+#  - All items in output list will be present in lhs
+#  - No items in output list will be present in rhs
+#  - Performance is considered and exceeds the naive N^2 algorithm
+#
+# Examples:
+# ```
+#   difference icmp (seq 4) (seq 3) = 3, Nil
+#   difference icmp (seq 3) (seq 3) = Nil
+#   difference icmp (seq 3) (2, Nil) = 0, 1, Nil
+#   difference scmp ("", "bar", "bat", Nil) ("", "foo", "bat", Nil) = "bar", Nil
```
ag-eitilt, line 964: Intersection is symmetric, and "subset" could easily be read as "has as a subset", so I wasn't overly concerned about order.  Here, though, I wonder if we might want to subtract the *left* side from the right -- yes, it's backwards when written traditionally on a single line, but the current order comes across as backwards in pipelines, and it's harder to come up with a reading of the name which would switch the arguments.
```suggestion
#   seq 4 | difference icmp (seq 3) = 3, Nil
#   seq 3 | difference icmp (seq 3) = Nil
#   seq 3 | difference icmp (2, Nil) = 0, 1, Nil
#   ("", "bar", "bat", Nil) | difference scmp ("", "foo", "bat", Nil) = "bar", Nil
```

ag-eitilt: I guess `tsubtract`/`msubtract` have already pinned what order we'd want to use, and it does look like it's `rhs - lhs`.  Let's do follow that model, and maybe bring the names into alignment -- though since List doesn't have a letter prefix and `subtract` doesn't invoke the right mental image, we might want to change to `difference`/`tdifference`/`mdifference` instead, at least if the next release will be a breaking change.

colinschmidt (approved):
LGTM

share/wake/lib/core/list.wake
V-FEXrt: I'll just go with `subtract` for now and then we can rename everything to `difference` later. Don't want to mess with even more breaking changes for this release

V-FEXrt: > and it does look like it's rhs - lhs.

Can you double check that? My reading from both is that it is `lhs - rhs`

"Remove all keys from the left which occur in the right"

ag-eitilt:
```
sammay@pe01:/scratch/sammay/wake$ wake -x 'tsubtract (listToTree scmp ("a", "b", Nil)) (listToTree scmp ("b", "f", Nil)
) | treeToList'
"a", Nil
sammay@pe01:/scratch/sammay/wake$ wake -x 'msubtract (listToMap scmp ("a" :-> 1, "b" :-> 2, Nil)) (listToMap scmp ("b" :-> 11, "f" :-> 15, Nil)) | mapToList'
Pair "a" 1, Nil
```

My reading of that is "take the left map and remove some keys", and the results support that, but you are definitely right that the wording could be clearer.

PR #1649, opened 2024-09-19
rsc: Disallow hidden file dependencies

V-FEXrt:
When a job creates both a file and a symlink to that file but only outputs the symlink it creates a hidden dependency on that untracked file. 

This is generally not a huge problem but now is since the RSC is moving to only track true dependencies. The symlink would be restored but not the "hidden" file.

This PR panics when that situation occurs for a cached job. Broken jobs can avoid this by opting out of caching or by also outputting the hidden file

### DIFF
--- a/share/wake/lib/system/remote_cache_runner.wake
+++ b/share/wake/lib/system/remote_cache_runner.wake
@@ -216,6 +216,26 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: RunnerInput => 
 
         def _ = primJobVirtual job stdout stderr predict
 
+        ## ----------------------------- Filtered Output Commentary ----------------------------- ##
+        # outputs below is currently filtered via FnOutputs. This is done to drastically decrease
+        # the number of output files uploaded to the remote server. It causes several side effects
+        # worth highlighing. Specifically there are certain "hidden" outputs that a job may
+        # generate. The most obvious once is the job `mkdir foo && touch foo/bar.txt` with
+        # FnOutputs = (\_ "foo/bar.txt", Nil). As this job does *not* list the OutputDirectory foo
+        # we can't expect it to exist via normal rehydration so special handling is required. A
+        # less obvious hidden outputs is a symlink that points to a file, both created by the same
+        # job where only the symlink is output. The symlink would be retored but would be invalid
+        # since the target file doesn't exists.
+        #
+        # The current implementation does the following:
+        #   - When uploading a job, check for a "hidden" directory output then add it as hidden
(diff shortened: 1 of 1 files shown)


share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -216,6 +216,26 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: RunnerInput =>
...
 
+        ## ----------------------------- Filtered Output Commentary ----------------------------- ##
+        # outputs below is currently filtered via FnOutputs. This is done to drastically decrease
+        # the number of output files uploaded to the remote server. It causes several side effects
+        # worth highlighing. Specifically there are certain "hidden" outputs that a job may
+        # generate. The most obvious once is the job `mkdir foo && touch foo/bar.txt` with
+        # FnOutputs = (\_ "foo/bar.txt", Nil). As this job does *not* list the OutputDirectory foo
+        # we can't expect it to exist via normal rehydration so special handling is required. A
+        # less obvious hidden outputs is a symlink that points to a file, both created by the same
+        # job where only the symlink is output. The symlink would be retored but would be invalid
```
colinschmidt, line 227:
```suggestion
        # job where only the symlink is output. The symlink would be restored but would be invalid
```

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -408,6 +428,32 @@ def postJob (rscApi: RemoteCacheApi) (job: Job) (_wakeroot: String) (hidden: Str
...
+    require True =
+        require False = symlinksUpload.len == 0
+        else True
+
+        # Get the list of potential targets the job may have uploaded
+        require Pass symlinkTargets =
+            symlinksUpload
+            | findFail
+            |< map getCachePostRequestOutputSymlinkPath
+        else True
```
colinschmidt, line 443: If we are unable to get the target for an output symlink, we are assuming that means this job didn't output it?
Or are we just choosing to ignore that error here and let some other part of the code handle that failure.

colinschmidt (commented):
Looks correct to me, just one question about error handling.

share/wake/lib/system/remote_cache_runner.wake
V-FEXrt: > we just choosing to ignore that error here and let some other part of the code handle that failure

This one. The only failure here is a failure to `readlink` and if that fails then another part of the code (just a few lines below) will fully abort the job push

colinschmidt (approved):
LGTM

PR #1651, opened 2024-09-23
rsc: Track hidden output dirs

V-FEXrt:
Updates the RSC and clients to track "hidden" directories which are directories created by a job but not output by the job.

This is necessary as some jobs will output a file and depends on the directory being created without also outputting said directory

### DIFF
--- a/share/wake/lib/system/remote_cache_runner.wake
+++ b/share/wake/lib/system/remote_cache_runner.wake
@@ -88,7 +88,7 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: RunnerInput => 
         def stdoutDownload = rscApiGetStringBlob stdoutBlob
         def stderrDownload = rscApiGetStringBlob stderrBlob
 
-        def doMakeDirectory (CacheSearchOutputDirectory path mode) =
+        def doMakeDirectory (CacheSearchOutputDirectory path mode _hidden) =
             # wake-format off
             def cmd =
                 "mkdir",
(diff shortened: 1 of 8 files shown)


share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -350,30 +351,25 @@ def thirdBy (acceptFn: a => ThirdByGroup): (list: List a) => Triple (one: List a
...
-        Pass x -> Pass (Pair path x)
-
-    require Pass outputsStat =
-        allOutputs
-        | map rmapStat
-        | findFail
-        | addErrorContext "rsc: Failed to stat files to upload"
+    def filteredOutputs = output.getRunnerOutputOutputs
+    def cleanableOutputs = output.getRunnerOutputCleanableOutputs
+    def hiddenOutputs = subtract scmp cleanableOutputs filteredOutputs
```
colinschmidt, line 356: Should we do this subtract after the grouping so one of the lists is smaller?

colinschmidt (approved):
Looks correct to me.

share/wake/lib/system/remote_cache_runner.wake
V-FEXrt: I don't think that helps, we have to `stat` the full union of `filteredOutputs` and `cleanableOutputs` in order to determine which are symlinks and I think that will be the majority of the time spent in the function (fwiw we were always `stat`ing the full list so it shouldn't be a time regression)

PR #1652, opened 2024-09-24
rsc: Delete multiple blob files per task by config

V-FEXrt:
Blob file deletion previously launched a task per file. This has significant overhead when deleting 100s of thousands of files. 

This change allows the config to specify how many files should be deleted per task

### DIFF
--- a/rust/rsc/src/bin/rsc/config.rs
+++ b/rust/rsc/src/bin/rsc/config.rs
@@ -25,7 +25,7 @@ pub struct RSCBlobTTLConfig {
     pub tick_rate: u64,
     // How long an object is allowed to live
     pub ttl: u64,
-    // Maximum number of objects to delete from the db at a time. Must be 1 >= x <= 16000
+    // Maximum number of blobs to delete from the db at a time. Must be 1 >= x <= 64000
     pub chunk_size: u32,
     // Maximum number of files to delete from the disk per task
     pub file_chunk_size: usize,
--- a/rust/rsc/src/bin/rsc/config.rs
+++ b/rust/rsc/src/bin/rsc/config.rs
@@ -15,8 +15,20 @@ pub struct RSCTTLConfig {
     pub tick_rate: u64,
     // How long an object is allowed to live
     pub ttl: u64,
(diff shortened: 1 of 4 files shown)


rust/rsc/src/bin/rsc/config.rs
```diff
@@ -15,8 +15,20 @@ pub struct RSCTTLConfig {
...
+    pub chunk_size: u32,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct RSCBlobTTLConfig {
+    // How often to run the eviction check in seconds
+    pub tick_rate: u64,
+    // How long an object is allowed to live
+    pub ttl: u64,
+    // Maximum number of objects to delete from the db at a time. Must be 1 >= x <= 16000
```
colinschmidt, line 28:
```suggestion
    // Maximum number of blobs to delete from the db at a time. Must be 1 >= x <= 64000
```

colinschmidt: Not sure what the actual limit is but we've tested at least 64k so might as well note that.

rust/rsc/src/bin/rsc/config.rs
```diff
@@ -15,8 +15,20 @@ pub struct RSCTTLConfig {
...
     pub chunk_size: u32,
+    // Maximum number of files to delete from the disk per task
+    pub file_chunk_size: usize,
```
colinschmidt, line 31: Is `usize` always going to be less than `u32`?

colinschmidt (approved):
LGTM

rust/rsc/src/bin/rsc/config.rs
V-FEXrt: no `usize` is the number of bits a pointer on the host has (so pretty much `u64` everywhere). The API that the value goes into expects a `usize` so it was easier to let the config system do validation instead  of pushing it all the way into the eviction task

colinschmidt: Sure thats fine.

PR #1653, opened 2024-09-25
rsc: Add config var for max number of pool connections

V-FEXrt:
Adds a new config var `connection_pool_max_connect` which is used to determine the max number of connections the server will make to the database server

### DIFF
--- a/rust/rsc/src/bin/rsc/main.rs
+++ b/rust/rsc/src/bin/rsc/main.rs
@@ -204,7 +204,8 @@ async fn connect_to_database(
     let timeout = config.connection_pool_timeout;
     let mut opt = ConnectOptions::new(&config.database_url);
     opt.sqlx_logging_level(tracing::log::LevelFilter::Debug)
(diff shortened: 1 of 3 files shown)


rust/rsc/src/bin/rsc/main.rs
```diff
@@ -204,7 +204,8 @@ async fn connect_to_database(
     let timeout = config.connection_pool_timeout;
     let mut opt = ConnectOptions::new(&config.database_url);
     opt.sqlx_logging_level(tracing::log::LevelFilter::Debug)
-        .acquire_timeout(std::time::Duration::from_secs(timeout));
+        .acquire_timeout(std::time::Duration::from_secs(timeout))
+        .max_connections(config.connection_pool_max_connect);
```
colinschmidt, line 208: It seems prudent to also add a `tracing:info` for this new config setting.

colinschmidt (approved):
LGTM, expect for the extra trace line which would be nice to get in.

PR #1655, opened 2024-09-27
stdlib: Fix call timing for FnOutputs

V-FEXrt:
Runners need to wait for jobs to complete before calling `FnOutputs` so that all files are written to disk before any kind of magic file detection occurs


colinschmidt (approved):
LGTM

PR #1658, opened 2024-09-30
rsc: Hidden directories should be returned as cleanable outputs

V-FEXrt:
When rehydrating from the cache, any "hidden" directories should be listed as cleanable outputs since they aren't listed as actual outputs by the runner. If they aren't listed then they won't get cleaned up as expected by `wake --clean`

### DIFF
--- a/share/wake/lib/system/remote_cache_runner.wake
+++ b/share/wake/lib/system/remote_cache_runner.wake
@@ -186,9 +186,16 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: RunnerInput => 
             outputSymlinks
             | map getCacheSearchOutputSymlinkLink
 
-        def resolvedDirectories =
+        def Pair hiddenDirectories publishedDirectories =
             outputDirs
-            | filter (!_.getCacheSearchOutputDirectoryHidden)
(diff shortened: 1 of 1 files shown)


share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -237,7 +244,7 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: RunnerInput =>
         #     by the same job. If so, upload the target as a "hidden" output file. On rehydration
         #     retore the file as normal but don't list it in the outputs.
         ## -------------------------------------------------------------------------------------- ##
-        Pass (RunnerOutput inputs outputs Nil predict)
+        Pass (RunnerOutput inputs outputs cleanableOutputs predict)
```
colinschmidt, line 247: Ah this Nil should have triggered something in our brains during code review I guess.

colinschmidt (approved):
LGTM

share/wake/lib/system/remote_cache_runner.wake
V-FEXrt: Yeah, I had the same thought. It was subtle but it should have been a flag

PR #1664, opened 2024-10-15
rsc: Track job events in job audit table

V-FEXrt:
Events pertaining to a specific job are added to an append only table for long term tracking and auditing. This data is not currently published outside of the database but will eventually replace the queries provided by the `job_history` table. Once that is done that table may be deleted but until then data integrity is maintained in both locations.


colinschmidt (approved):
LGTM, once we start using this in the dashboard we'll want some indexes probably but just waiting until we know exactly what we need is fine.

PR #1665, opened 2024-10-16
rsc: Tag dbblob files with their blob id

V-FEXrt:
When writing a db backed blob file to disk, include the blob id in the file name for better inspection.


ag-eitilt (approved):
I know I'm not requested for review, but seems reasonable to me.

PR #1666, opened 2024-10-18
ci: Switch to available formatter

V-FEXrt:
clang-format-12  is no longer available in CI now that `ubuntu-latest` points to 24.04. Switch to a newer one


ag-eitilt (approved):
Can't say I think the formatting is an improvement, but it's nothing I'll have to interact with too often, and updating to the current version is definitely good.

colinschmidt (approved):
LGTM

PR #1667, opened 2024-11-05
cli: Add new history command

AbrarQuazi:
Added new `wake --history` command

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -80,6 +80,7 @@ struct Database::detail {
   sqlite3_stmt *tag_job;
   sqlite3_stmt *get_tags;
   sqlite3_stmt *get_all_tags;
+  sqlite3_stmt *get_runs_table;
   sqlite3_stmt *get_edges;
   sqlite3_stmt *get_file_dependency;
   sqlite3_stmt *get_output_files;
@@ -125,6 +126,7 @@ struct Database::detail {
         tag_job(0),
         get_tags(0),
         get_all_tags(0),
+        get_runs_table(0),
         get_edges(0),
         get_file_dependency(0),
         get_interleaved_output(0) {}
@@ -396,6 +398,7 @@ std::string Database::open(bool wait, bool memory, bool tty) {
   const char *sql_tag_job = "insert into tags(job_id, uri, content) values(?, ?, ?)";
   const char *sql_get_tags = "select job_id, uri, content from tags where job_id=?";
   const char *sql_get_all_tags = "select job_id, uri, content from tags";
+  const char *sql_get_runs_table = "select run_id, time, cmdline from runs order by time ASC";
   const char *sql_get_edges =
       "select distinct user.job_id as user, used.job_id as used"
       "  from filetree user, filetree used"
@@ -470,6 +473,7 @@ std::string Database::open(bool wait, bool memory, bool tty) {
   PREPARE(sql_tag_job, tag_job);
   PREPARE(sql_get_tags, get_tags);
   PREPARE(sql_get_all_tags, get_all_tags);
+  PREPARE(sql_get_runs_table, get_runs_table);
   PREPARE(sql_get_edges, get_edges);
   PREPARE(sql_get_file_dependency, get_file_dependency);
   PREPARE(sql_get_output_files, get_output_files);
@@ -528,6 +532,7 @@ void Database::close() {
   FINALIZE(tag_job);
   FINALIZE(get_tags);
   FINALIZE(get_all_tags);
+  FINALIZE(get_runs_table);
   FINALIZE(get_edges);
   FINALIZE(get_file_dependency);
   FINALIZE(get_output_files);
@@ -1582,6 +1587,25 @@ std::vector<JobTag> Database::get_tags() {
   }
   finish_stmt("Could not retrieve tags", imp->get_all_tags, imp->debugdb);
   return out;
+}
+
+static std::vector<RunReflection> get_all_runs(const Database *db, sqlite3_stmt *query) {
+  std::vector<RunReflection> out;
+  db->begin_txn();
+  while (sqlite3_step(query) == SQLITE_ROW) {
+    RunReflection run;
+    run.id = sqlite3_column_int(query, 0);
+    run.time = Time(sqlite3_column_int64(query, 1));
+    run.cmdline = rip_column(query, 2);
+    out.emplace_back(run);
+  }
+  finish_stmt("Could not retrieve runs", query, db->imp->debugdb);
+  db->end_txn();
+  return out;
+}
+
+std::vector<RunReflection> Database::get_runs() const {
+  return get_all_runs(this, imp->get_runs_table);
 }
 
 std::vector<std::pair<std::string, int>> Database::get_interleaved_output(long job_id) const {
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -80,7 +80,7 @@ struct Database::detail {
   sqlite3_stmt *tag_job;
   sqlite3_stmt *get_tags;
   sqlite3_stmt *get_all_tags;
-  sqlite3_stmt *get_runs_table;
+  sqlite3_stmt *get_all_runs;
   sqlite3_stmt *get_edges;
   sqlite3_stmt *get_file_dependency;
   sqlite3_stmt *get_output_files;
@@ -126,7 +126,7 @@ struct Database::detail {
         tag_job(0),
         get_tags(0),
         get_all_tags(0),
-        get_runs_table(0),
+        get_all_runs(0),
         get_edges(0),
         get_file_dependency(0),
         get_interleaved_output(0) {}
@@ -398,7 +398,7 @@ std::string Database::open(bool wait, bool memory, bool tty) {
   const char *sql_tag_job = "insert into tags(job_id, uri, content) values(?, ?, ?)";
   const char *sql_get_tags = "select job_id, uri, content from tags where job_id=?";
   const char *sql_get_all_tags = "select job_id, uri, content from tags";
-  const char *sql_get_runs_table = "select run_id, time, cmdline from runs order by time ASC";
+  const char *sql_get_all_runs = "select run_id, time, cmdline from runs order by time ASC";
   const char *sql_get_edges =
       "select distinct user.job_id as user, used.job_id as used"
       "  from filetree user, filetree used"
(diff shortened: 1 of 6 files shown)


AbrarQuazi:
How the history command looks like. I used the default date time formatting from the `Time` library that we have for consistency, but if needed, I can change the formatting to something else like [Mon Nov 27 12:17:37 PST 2023] (but to do this new logic needs to be added to get the abbreviated days and month for processing as sqlite library doesnt do this out of the box)
Screenshot 2024-11-05 at 9 34 13 AM

src/runtime/database.cpp
```diff
@@ -396,6 +398,7 @@ std::string Database::open(bool wait, bool memory, bool tty) {
   const char *sql_tag_job = "insert into tags(job_id, uri, content) values(?, ?, ?)";
   const char *sql_get_tags = "select job_id, uri, content from tags where job_id=?";
   const char *sql_get_all_tags = "select job_id, uri, content from tags";
+  const char *sql_get_runs_table = "select run_id, time, cmdline from runs order by time ASC";
```
AbrarQuazi, line 401: I could have gotten sqlite to process the time to appropriate datetime format like SELECT strftime('%Y-%m-%d %H:%M:%S', time / 1000000000, 'unixepoch'), but decided to just default to the `Time` Library https://github.com/sifive/wake/blob/b4689381eb3bde982e8bad7502b1aca272d9c730/src/runtime/database.h#L55

tools/wake/main.cpp
```diff
@@ -168,15 +168,16 @@ void hide_internal_jobs(std::vector<std::vector<std::string>> &out) {
...
-  if (clo.tagdag) {
-    JAST json = create_tagdag(db, clo.tagdag);
-    std::cout << json << std::endl;
-    return;
+void query_runs(Database &db)
+{
+  const auto runs = db.get_runs();
+  for (const auto run : runs)
+  {
+    std::cout << run.time.as_string() << " " << run.cmdline << std::endl;
```
AbrarQuazi, line 176: Decided to just output the string via `std::cout`, but I could have used the `Describe` library instead like we do with query_jobs : https://github.com/sifive/wake/blob/b4689381eb3bde982e8bad7502b1aca272d9c730/tools/wake/describe.cpp#L371 . But to do this, it would require additional refactoring the whole file to accept not only a `JobReflection` but now a `RunReflection` as well (could make a `TableReflection` parent class that describe takes in for polymorphism). Would that type of change be of benefit and needed now?

colinschmidt:
>  I used the default date time formatting from the `Time` library that we have for consistency

Where else do we print timestamps? I'm not sure we need the fancier formatting you mentioned but was thinking we could just drop the subsecond precision. Most wake invocations take longer than a second so I don't think we would lose any meaningful precision.

AbrarQuazi:
> Where else do we print timestamps
The only other time we use `Time` 's string function is for `wake --script`. Don't think subsecond is required for that right? 
Screenshot 2024-11-05 at 1 34 36 PM

src/runtime/database.cpp
ag-eitilt, line 401: Bikshedding nit: the `_table` suffix feels a bit out of place.  Yeah, you're effectively selecting the entire table, but the line just above this is doing something similar and is marked as such with an `_all_` infix instead.
```suggestion
  const char *sql_get_all_runs = "select run_id, time, cmdline from runs order by time ASC";
```

src/runtime/database.cpp
```diff
@@ -1584,6 +1589,25 @@ std::vector<JobTag> Database::get_tags() {
   return out;
 }
 
+static std::vector<RunReflection> get_all_runs(const Database *db, sqlite3_stmt *query) {
```
ag-eitilt, line 1592: ... but here you're using that pattern already ...

(Side note, this seems to be the only implementation here which uses a separate `static` function rather than having it all in the `Database::` method.)

src/runtime/database.h
```diff
@@ -169,6 +178,8 @@ struct Database {
   std::vector<JobEdge> get_edges();
   std::vector<JobTag> get_tags();
 
+  std::vector<RunReflection> get_runs() const;
```
ag-eitilt, line 181: ... and here you've dropped the markers entirely.  Looking at the surroundings, I think this one is correct, but the other two could do with standardization in their own ways.

tools/wake/main.cpp
```diff
@@ -168,15 +168,14 @@ void hide_internal_jobs(std::vector<std::vector<std::string>> &out) {
   out.push_back({"tags NOT LIKE '%<d>inspect.visibility=hidden<d>%'", "tags IS NULL"});
 }
 
-void inspect_database(const CommandLineOptions &clo, Database &db, const std::string &wake_cwd) {
-  // tagdag is technically a db inspection, but its very different from the
-  // rest, just handle it and exit.
```
ag-eitilt, line 173: Could you copy this comment over as well?  Maybe update it slightly to mention `history` as well, since it is indeed very different from both.

tools/wake/main.cpp
```diff
@@ -228,6 +227,18 @@ void inspect_database(const CommandLineOptions &clo, Database &db, const std::st
...
+  if (clo.tagdag) {
+    JAST json = create_tagdag(db, clo.tagdag);
+    std::cout << json << std::endl;
+    return;
+  } else if (clo.history) {
+    query_runs(db);
+  } else {
+    query_jobs(clo, db);
+  }
+}
```
ag-eitilt, line 240: The tagdag printing out here but the two queries embedding their printing internally feels a bit off from a functional point of view -- my instincts are saying that `inspect_database` should be either a thin dispatcher (and the tagdag printing should be done inside `create_tagdag`) or that the `query_*` should be pure DB (and `inspect_database` would handle the printing for all three).  This isn't functional, though, so it's a bit of an overreaction.

Whether you do anything with that or not, the `return` isn't necessary any more with the `else` structure.

ag-eitilt (commented):
+0.75 for dropping the subseconds.  Even for jobs which do take less than a second, that's still not affecting that statistic directly (which *is* helpful to have), it's just saying when the job was launched, and I don't see much reason for that to matter.  The big caveat I have is that subsecond resolution should still be visible when the `--timeline` is zoomed in far enough (I think that's how it can be interacted with, right?  Shows how little I use that feature.) -- if dropping the resolution at the struct level means the timeline gets quantized to seconds, or to a lesser degree if the JSON is affected in the same way, then we'd want to be a lot more hesitant about it.

src/runtime/database.cpp
AbrarQuazi: Sounds good, yeah removed the static function ( I was just following the get_file_dependencies pattern: https://github.com/sifive/wake/blob/4d62b1d4a50d7e6ff99c32abbc6d8a3999669dce/src/runtime/database.cpp#L1626 , but realized that is unnecessary for get_runs)

src/runtime/database.h
```diff
@@ -60,6 +60,15 @@ struct Time {
   std::string as_string() const;
 };
 
+struct RunReflection {
+  int id;
```
colinschmidt, line 64: This is the same data type as in the DB right?

src/runtime/database.h
```diff
@@ -60,6 +60,15 @@ struct Time {
...
+  Time time;
+  std::string cmdline;
+  RunReflection() = default;
```
colinschmidt, line 67: Do we need this constructor?

tools/wake/main.cpp
```diff
@@ -228,6 +227,19 @@ void inspect_database(const CommandLineOptions &clo, Database &db, const std::st
   describe(matching_jobs, get_describe_policy(clo), db);
 }
 
+void inspect_database(const CommandLineOptions &clo, Database &db) {
+  // tagdag and history are db inspection queries, but are very different from the
+  // rest of the queries which operate on the jobs table.
+  if (clo.tagdag) {
+    JAST json = create_tagdag(db, clo.tagdag);
+    std::cout << json << std::endl;
```
colinschmidt, line 235: I feel like we should just move the printing into `create_tagdag` since all the other functions do it that way

tools/wake/main.cpp
```diff
@@ -639,7 +652,7 @@ int main(int argc, char **argv) {
   }
 
   if (is_db_inspection) {
-    inspect_database(clo, db, wake_cwd);
```
colinschmidt, line 642: This argument was just dead code already?

tools/wake/main.cpp
AbrarQuazi: Got rid of the return statement, wasn't too sure about printing inside `create_tagdag` as the name suggests to return a tagdag, so would have to change the name as well (also dont know if this function will be useful later or not). But if you have strong opinions, I can definitely change the name and make it print inside the function.

AbrarQuazi:
> e `--timeline` is zoomed in far enough (I think that's how it can be interacted with, right

I dont see --timeline really interacting with the --script command or calling `Time`'s `as_string()` function

tools/wake/main.cpp
AbrarQuazi: Looks like @colinschmidt  agrees with printing inside the function, so will go ahead and do that

src/runtime/database.h
AbrarQuazi: Yep!
`CREATE TABLE runs(  run_id  integer primary key autoincrement,  time    integer not null,  cmdline text    not null);`

tools/wake/main.cpp
AbrarQuazi: Yeah it was already dead code

src/runtime/database.h
AbrarQuazi: We dont necessarily need it, but I used the constructor so its clearly understood what exactly I am querying: 
https://github.com/sifive/wake/pull/1667/files#diff-ce97135c77b7c227bcd205cb4d3449133611307c21539f37b8c422bde9780252R1591-R1593

colinschmidt: Ok thats fair. It we don't have any style convention here thats fine.

tools/wake/main.cpp
```diff
@@ -268,6 +279,7 @@ void print_help(const char *argv0) {
     << "    --last     -l      See --last-used"                                            << std::endl
     << "    --last-used        Capture all jobs used by last build. Regardless of cache"   << std::endl
     << "    --last-executed    Capture all jobs executed by the last build. Skips cache"   << std::endl
+    << "    --history          Report the cmndline history of all wake commands recorded"  << std::endl
```
colinschmidt, line 282:
```suggestion
    << "    --history          Report the command-line history of all wake commands"  << std::endl
```
To match the spelling for `--in`
We don't need to specify "recorded" in the same way we don't specify that for things like "--list-outputs"

colinschmidt (approved):
Some small spelling nits but otherwise LGTM.

PR #1670, opened 2024-12-05
Fix potential "infinite recursion" panics in runners

ag-eitilt:
We've seen a couple problems where the proper reporting of an issue got masked by an "Infinite recursion detected" panic, and we verified that it was due to `primJobFinish`/`primJobFailFinish` being called before the respective `primJobLaunch`/`primJobFailLaunch` -- which for all their other benefits is a new risk with v44's flattened runners.  Ultimately it would be best to implement an architecture solution for solving this, but in the mean time it's sufficient to patch the existing runners and to keep a careful eye on future code which touches them.

### DIFF
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -234,7 +234,7 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
 
     def run (job: Job) ((RunnerInput label command visible environment directory stdin res prefix record isatty fnInputs fnOutputs): RunnerInput): Result RunnerOutput Error =
         require True = executeOk
-        else failWithError "Runner {script} is not executable"
+        else markJobSetupFailure job "Runner {script} is not executable".makeError
 
         def Usage status runtime cputime membytes inbytes outbytes = record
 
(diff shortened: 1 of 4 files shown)


share/wake/lib/system/runner.wake
```diff
@@ -263,18 +263,14 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
                 Nil
             )
 
-        require Pass build =
-            mkdir ".build"
```
colinschmidt, line 267: Where does this mkdir live now?

colinschmidt (commented):
One question about the mkdir removal, otherwise LGTM

share/wake/lib/system/runner.wake
ag-eitilt: Inside the `write`:

```
export def write (path: String) (content: String): Result Path Error =
    def spath = simplify path

    require Pass parent =
        simplify "{spath}/.."
        | mkdir # Note that this is recursive, as well.

    writeImp (parent, Nil) 0644 spath content
```

My guess is that the `mkdir` was originally split out purely for the error message, but the shell's `mkdir` already advertises the path it tried unsuccessfully to create, so keeping it separate didn't seem to have any benefit.

Also, CI apparently succeeded?  I'll look into the supposedly end-of-lifed image in a bit.

ag-eitilt:
Ah, apparently the deprecation is happening [on jan 30](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/), so we've got another month and a half.  Merging!

PR #1671, opened 2024-12-17
rsc: update read_job transaction hit path

AbrarQuazi:


rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -63,152 +63,130 @@ pub async fn read_job(
...
-                        memory: matching_job.memory as u64,
-                        ibytes: matching_job.i_bytes as u64,
-                        obytes: matching_job.o_bytes as u64,
-                    },
-                ))
-            })
-        })
-        .await;
+    // Step 6: TODO before recording a hit and returning, need verification that objects from transaction did not change in
+    // database, or convert a hit into a miss
```
AbrarQuazi, line 179: I have questions on how best to accomplish this verification from chaning a hit to a miss. Do I just query (output_files, output_symlinks, output_dirs) again and verify they are the same with the database? Seems kind of wasteful (what happens when another update happens while I am doing the verification equivalence check?). 

Can we instead potentially rely on job::Entity's created_at field (or create a new one called updated_at) where we just quickly check timestamps are same before returning?

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -34,26 +35,51 @@ async fn record_miss(hash: String, conn: Arc<DatabaseConnection>) {
...
-            blob.store_id, id
-        ));
-    };
+    // Build a map of blob_id -> blob model for quick lookup
+    let blob_map: HashMap<Uuid, entity::blob::Model> = blobs.into_iter().map(|b| (b.id, b)).collect();
+
+    // Ensure we have all requested blobs
+    for &id in ids {
+        if !blob_map.contains_key(&id) {
+            return Err(format!("Unable to find blob {} by id", id));
```
colinschmidt, line 56: How does this error condition differ from the failed to query blobs check above?

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -34,26 +35,51 @@ async fn record_miss(hash: String, conn: Arc<DatabaseConnection>) {
...
+        }
     });
+
+    let results = join_all(futures).await;
+
+    let mut resolved_map = HashMap::new();
+    for res in results {
+        let (id, resolved_blob) = res?;
+        resolved_map.insert(id, resolved_blob);
+    }
```
colinschmidt, line 80: Is there no simpler constructor for this?
We only to do this iteratively because we wanted to resolve the urls in parallel, rather than with a map on the original `blob_map`?

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -63,152 +89,138 @@ pub async fn read_job(
...
-                        return Ok((None, ReadJobResponse::NoMatch))
-                    },
-                    Ok(files) => files,
-                };
+                Ok(Some((matching_job, output_files, output_symlinks, output_dirs)))
+            })
+        })
+        .await;
+    
+    let hash_copy = hash_for_spawns.clone();
```
colinschmidt, line 117: Why isn't this just `hash.clone()`?

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -34,26 +35,51 @@ async fn record_miss(hash: String, conn: Arc<DatabaseConnection>) {
 }
 
 #[tracing::instrument(skip(db, stores))]
-async fn resolve_blob(
-    id: Uuid,
-    db: &DatabaseTransaction,
+async fn resolve_blobs<T: ConnectionTrait>(
+    ids: &[Uuid],
```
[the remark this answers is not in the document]
AbrarQuazi: Yep, this should be a vec

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -34,26 +35,51 @@ async fn record_miss(hash: String, conn: Arc<DatabaseConnection>) {
...
+    db: &T,
     stores: &HashMap<Uuid, Arc<dyn blob::DebugBlobStore + Sync + Send>>,
-) -> Result<ResolvedBlob, String> {
-    let Ok(Some(blob)) = entity::prelude::Blob::find_by_id(id).one(db).await else {
-        return Err(format!("Unable to find blob {} by id", id));
-    };
+) -> Result<HashMap<Uuid, ResolvedBlob>, String> {
+    // Fetch all blobs in a single query
+    let blobs = Blob::find()
+        .filter(entity::blob::Column::Id.is_in(ids.to_vec()))
```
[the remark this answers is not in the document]
AbrarQuazi: We have logging turned on for slow queries that go over 1 second. Looking at logs from 2025-01-09 when I tested 10 PRs at once, I search for slow queries related to this:`WARN read_job:resolve_blobs`, here are some of the queries with longest number of IDs:
`rows_affected=17443 rows_returned=17443 elapsed=1.009696364s elapsed_secs=1.009696364 slow_threshold=1s`

`rows_affected=3102 rows_returned=3102 elapsed=1.477475543s elapsed_secs=1.477475543 slow_threshold=1s`
 
`rows_affected=922 rows_returned=922 elapsed=1.188434243s elapsed_secs=1.1884342430000001 slow_threshold=1s`

 `rows_affected=1533 rows_returned=1533 elapsed=1.470887866s elapsed_secs=1.470887866 slow_threshold=1s`

 `rows_affected=5423 rows_returned=5423 elapsed=6.464703056s elapsed_secs=6.464703056 slow_threshold=1s (when load was higher)`

 `rows_affected=3984 rows_returned=3984 elapsed=6.72935904s elapsed_secs=6.72935904 slow_threshold=1s`

From the logs and statements above, I saw that scaling the number of IDs had minimal performance impact (with a slow query having 17,000 IDs often having the same performance as a slow query with 5 IDs). Rather the impact on performance was based on system load and potentially contention with other transactions.

Looking at the postgres documentation: https://www.postgresql.org/docs/current/limits.html, looks like the query parameter limits are 65,535, and currently the largest parameter limit we are seeing from above is 17443. We can definitely chunk the query closer to this limit, if you feel that is necessary

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -34,26 +35,51 @@ async fn record_miss(hash: String, conn: Arc<DatabaseConnection>) {
...
+        .all(db)
+        .await
+        .map_err(|e| format!("Failed to query blobs: {}", e))?;
```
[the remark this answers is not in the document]
AbrarQuazi: Yep, Ill change that

rust/rsc/src/bin/rsc/read_job.rs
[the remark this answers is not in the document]
colinschmidt: Yes chunking based on max parameters is needed so thanks for doing that.

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -63,152 +93,138 @@ pub async fn read_job(
...
-                        obytes: matching_job.o_bytes as u64,
-                    },
-                ))
+    // Construct ResolvedBlobFile for each output file
+    let output_files = output_files
+        .into_iter()
+        .map(|m| {
+            let blob_id = m.blob_id;
+            let resolved_blob = resolved_blob_map.get(&blob_id).cloned().ok_or_else(|| {
+                format!("Missing resolved blob for {}", blob_id)
```
colinschmidt, line 149: Do we want to track the job that had a blob fetch failure? I can see that being useful debug info when something is failing.

rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -63,152 +93,138 @@ pub async fn read_job(
...
+        status: matching_job.status,
+        runtime: matching_job.runtime,
+        cputime: matching_job.cputime,
+        memory: matching_job.memory as u64,
+        ibytes: matching_job.i_bytes as u64,
+        obytes: matching_job.o_bytes as u64,
+    };
+
+    let job_id = matching_job.id;
+    let hash_copy = hash_for_spawns.clone();
```
colinschmidt, line 218: Again it feels like we didn't need to create this separate `hash_for_spawns` variable and could have instead continued to clone `hash`

colinschmidt (commented):
Logic seems correct, a few error message and style nits left, but LGTM

colinschmidt (approved):
LGTM

PR #1674, opened 2025-01-16
Migrated github actions to v4

AbrarQuazi:


colinschmidt (approved):
We weren't using any non-support or backwards incompatible options for upload-artifact?
Otherwise LGTM, we can update back to ubuntu24 at some point in the future.

PR #1676, opened 2025-02-19
Add better profiling of heap for --profile-heap

AbrarQuazi:
`--profile-heap` currently lacks proper metrics to debug OOM issues and varies from the information provided from valgrind. This PR adds more relevant information about the garbage collector for each GC cycle, while also adding extra information in profiling csv file.

Example generated CSV file: 
[heap_log.csv](https://github.com/user-attachments/files/18872526/heap_log.csv)

### DIFF
--- a/src/runtime/gc.cpp
+++ b/src/runtime/gc.cpp
@@ -26,19 +26,19 @@
 #include <string.h>
 
 #include <algorithm>
+#include <chrono>
+#include <fstream>
 #include <iomanip>
 #include <map>
 #include <sstream>
 #include <vector>
-#include <chrono>
-#include <fstream>
 
 #include "status.h"
 
 #define INITIAL_HEAP_SIZE 1024
 
 bool HeapAgeTracker::heapAgeTracker = false;
-std::unordered_map<const HeapObject*, uint32_t> HeapAgeTracker::age_map;
+std::unordered_map<const HeapObject *, uint32_t> HeapAgeTracker::age_map;
 static std::ofstream g_csv;
 
 HeapObject::~HeapObject() {}
@@ -161,7 +161,9 @@ Heap::~Heap() {
 
 size_t Heap::used() const { return (free - imp->spaces[imp->space].array) * sizeof(PadObject); }
 
-size_t Heap::alloc() const { return (imp->spaces[0].alloc + imp->spaces[1].alloc) * sizeof(PadObject); }
+size_t Heap::alloc() const {
+  return (imp->spaces[0].alloc + imp->spaces[1].alloc) * sizeof(PadObject);
+}
 
 size_t Heap::avail() const { return (end - free) * sizeof(PadObject); }
 
@@ -208,16 +210,14 @@ struct StatOrder {
 void Heap::GC(size_t requested_pads) {
   auto gc_start = std::chrono::system_clock::now();
   std::time_t current_t = std::chrono::system_clock::to_time_t(gc_start);
-  std::tm* local_tm = std::localtime(&current_t);
+  std::tm *local_tm = std::localtime(&current_t);
   std::ostringstream curent_time_string;
-  
-  auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
-                  gc_start.time_since_epoch()) % 1000;
-  
-  curent_time_string << std::put_time(local_tm, "%H:%M:%S") <<
-                  '.' << std::setfill('0') << std::setw(3) << ms.count();
- 
 
+  auto ms =
+      std::chrono::duration_cast<std::chrono::milliseconds>(gc_start.time_since_epoch()) % 1000;
+
+  curent_time_string << std::put_time(local_tm, "%H:%M:%S") << '.' << std::setfill('0')
+                     << std::setw(3) << ms.count();
   imp->gc_count += 1;
 
   Space &from = imp->spaces[imp->space];
@@ -255,8 +255,8 @@ void Heap::GC(size_t requested_pads) {
       ++s.objects;
       total_objs++;
       s.pads += (static_cast<PadObject *>(next.obj) - static_cast<PadObject *>(progress.obj));
-      
-      HeapObject* ho = static_cast<HeapObject*>(progress.obj);
+
+      HeapObject *ho = static_cast<HeapObject *>(progress.obj);
       auto obj_age = HeapAgeTracker::getAge(ho);
       if (obj_age < 2) {
         ++young_objects;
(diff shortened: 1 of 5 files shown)


src/runtime/gc.h
```diff
@@ -55,6 +57,42 @@ struct HeapStep {
   HeapObject **found;
 };
 
+struct HeapAgeTracker {
```
AbrarQuazi, line 60: Instead of having this entire class, and using an unordered_map to store heap age, I could have simply had a field inside HeapObject that keeps track of age for all HeapObjects: https://github.com/sifive/wake/blob/2fe5d90f53e626e4cd22640f014e6ed22576a9b0/src/runtime/gc.h#L98

However, when I did this I had issues with alignment, as we have a strict requirement of having 8 bytes alignment for all GCObjects: https://github.com/sifive/wake/blob/2fe5d90f53e626e4cd22640f014e6ed22576a9b0/src/runtime/gc.h#L362 
and I felt hesitant to change this requirement.

src/runtime/gc.cpp
```diff
@@ -209,50 +244,100 @@ void Heap::GC(size_t requested_pads) {
...
-  free = progress.free;
-  imp->last_pads = free - to.array;
+  end = to.array + elems; // elems doesn't include the extra 50% from resize
+  free = progress.free; // The place to append new things on the heap
+  imp->last_pads = free - to.array; // how many bytes were copied to the to space
   // Contain heap growth due to no_gc_overrun pessimism
   size_t desired_sized = imp->heap_factor * imp->last_pads + requested_pads;
   if (desired_sized < elems) {
-    end = to.array + desired_sized;
+    end = to.array + desired_sized; // Update the end to be smaller if we don't need that make space
```
colinschmidt, line 300:
```suggestion
    end = to.array + desired_sized; // Update the end to be smaller if we don't need that much space
```

src/runtime/gc.h
colinschmidt: I think having another data structure to capture the age of heap objects is a better separation of concerns. Since we only want to track this during profiling.

src/runtime/gc.h
```diff
@@ -354,21 +392,31 @@ size_t GCObject<T, B>::reserve() {
 template <typename T, typename B>
 template <typename... ARGS>
 T *GCObject<T, B>::claim(Heap &h, ARGS &&...args) {
-  return new (h.claim(sizeof(T) / sizeof(PadObject))) T(std::forward<ARGS>(args)...);
+  T* obj = new (h.claim(sizeof(T) / sizeof(PadObject))) T(std::forward<ARGS>(args)...);
+  HeapAgeTracker::setAge(obj,0);
```
colinschmidt, line 396: I wonder if we would be able to only increase an objects age when we are moving the objects during GC.
I think this would further put this profiling data out of the non-profiling path.

colinschmidt (commented):
Hopefully we can remove the allocation changes by putting the age tracking into the GC profiling movement loop

src/runtime/gc.cpp
```diff
@@ -265,6 +350,27 @@ void Heap::GC(size_t requested_pads) {
           << std::endl;
       }
       s << "------------------------------------------" << std::endl;
+      g_csv << std::fixed << std::setprecision(2)
+            << imp->gc_count << ", " 
+            << curent_time_string.str() << ", "
+            << gc_duration_ms << ", "
```
ag-eitilt, line 356: Two questions: are `fixed` and `setprecision` for the current statement being executed, or for the handle overall?, and would it help or hinder readability to interleave these with the print-to-terminal?
```
      s << "GC Number: " << imp->gc_count << std::endl;
      g_csv << imp->gc_count << ", ";

      s << "Current Time Stamp: " <<  curent_time_string.str() << std::endl;
      g_csv << curent_time_string.str() << ", ";

      s << "Current GC Duration: " << gc_duration_ms << " ms" << std::endl;
      g_csv << gc_duration_ms << ", ";
```

(Also, double-check your whitespace -- there's apparently a loose space at the end of the `gc_count` line, and where there's one there might be more.)

src/runtime/gc.h
AbrarQuazi: We cant really increase the object age during the GC loop itself, as we only have access to the root objects, but only when we recurse through the newly moved root objects via the `descend` BFS call where we actually move all of the roots children: https://github.com/sifive/wake/blob/master/src/runtime/gc.cpp#L213
```  
while (progress.obj != progress.free) {
    auto next = progress.obj->descend(progress.free);
    if (profile) {
      ObjectStats &s = stats[progress.obj->type()];
      ++s.objects;
      s.pads += (static_cast<PadObject *>(next.obj) - static_cast<PadObject *>(progress.obj));
    }
    progress = next;
  }
  ```
  
  Thus the only appropriate place to increment age is the `moveto` call which I have done (`descend` calls `moveto`)
  
We need to also initialize the Age of a new GC Object, and that can only be done on object creation (which are done through the `claim` and `alloc` calls) and unfortunately can't be done during the actual garbage collection, thus why I had to change it for every GCObject type

File: src/runtime/gc.cpp [hunk 5]
```diff
@@ -162,6 +178,8 @@ void Heap::report() const {
     std::stringstream s;
     s << "------------------------------------------" << std::endl;
     s << "Peak live heap " << (imp->most_pads * 8) << " bytes" << std::endl;
+    s << "Peak System Alloc: " << imp->peak_alloc << std::endl;
+    s << "Current Alloc: " << alloc() << std::endl;
```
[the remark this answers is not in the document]
AbrarQuazi: Yep this is the end of the report run, I'll go ahead and delete the call

src/runtime/gc.cpp
```diff
@@ -146,14 +161,15 @@ Heap::~Heap() {
 
 size_t Heap::used() const { return (free - imp->spaces[imp->space].array) * sizeof(PadObject); }
 
-size_t Heap::alloc() const { return (end - imp->spaces[imp->space].array) * sizeof(PadObject); }
```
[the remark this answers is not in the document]
AbrarQuazi: Nope, this function wasn't used before, but I am using it now to keep track of whats currently allocated

src/runtime/gc.cpp
AbrarQuazi: Yeah the `fixed` and `setprecision` are being executed, and they remain in effect to the stream until the next time they are called.

In terms of interleaving, I could do that if you wanted to, but I dont think it would have too much difference on readability to be honest - as I am only writing to stdout a subset of the information that I want thats in the csv. I also tried to make the variable names as clear as possible, so the reader knows whats being streamed out - but if its still not clear, let me know and I can definitely change that

ag-eitilt: Nah, if they're deliberately distinct sets of data then there's not much reason to shove them together.  The biggest thing which would have suggested joining the two would be if we were trying to keep them in sync, since it would be a lot easier to notice if one of the output formats got updated but not the other.

ag-eitilt (approved):
Semi-rubber-stamp -- I'm happy with this and the other comments were addressed.

PR #1678, opened 2025-02-25
Added endpoint to see caching metrics from grafana/prometheus

AbrarQuazi:
This PR exposes an endpoint to see cache metrics such as cache hits, misses and latency information which prometheus can scrape for grafana dashboarding


rust/rsc/src/bin/rsc/read_job.rs
```diff
@@ -17,6 +17,49 @@ use sea_orm::{
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 use tracing;
+use std::time::Instant;
+use lazy_static::lazy_static;
+use prometheus::{register_counter, register_histogram, Counter, Histogram};
+
+lazy_static! {
```
AbrarQuazi, line 24: prometheus should be able to hit the `/metrics` endpoint and scrape data for the following pomql queries to show in grafana dashboards:

# query for average hits per second over specified times
rate(cache_hits[5m]) #or alternatively total hits over specified time: increase(cache_hits[5m])

# query for average misses per second over specified times
rate(cache_misses[5m])

# query for hit ratio over specified time
sum(increase(cache_hits[5m])) / sum(increase(cache_hits[5m]) + increase(cache_misses[5m]))

#query for rolling average latency
rate(hit_latency_ms_sum[5m]) / rate(hit_latency_ms_count[5m])
rate(miss_latency_ms_sum[5m]) / rate(miss_latency_ms_count[5m])

PR #1680, opened 2025-03-07
[result] Fix documentation for getWhenFail and getWhenPass methods.

richardxia:
The examples in the documentation showed passing the same arguments and getting different results. The examples are incorrect because they forgot to change the arguments to demonstrate the other case. This fixes the examples and also replaces the type of the other case to more clearly indicate that it is totally unused.


AbrarQuazi (approved):

AbrarQuazi:
Hmmm seems like the update to github actions to v4 for Docs was missed, thats why the build is failing. Will create a PR to fix this issue

richardxia:
Rebasing to pick up the other patch (I tried rerunning within GitHub Actions, but it looks like it was still using the version of the workflows from this branch and not `master`).

PR #1681, opened 2025-03-07
updated github actions for docs

AbrarQuazi:
Updating github actions for docs was missed. According to https://github.com/actions/upload-pages-artifact/releases , we should update `actions/upload-pages-artifact` to the latest `v3` and update `actions/deploy-pages` to `v4`


richardxia (approved):
LGTM! Yeah, I've been doing some similar things with some personal projects as well, since GitHub dropped support for a bunch of their older Actions.

PR #1683, opened 2025-04-01
rsc: rsc client hardening with timeouts

AbrarQuazi:
In order to avoid situations where the RSC is too congested with high system load, and client requests are taking too long thus causing long build times, it is important to disable the RSC for that specific wake invocation.

To be able to achieve this, the following has been done:
1. Introduce the `--max-time` and `--connect-timeout` flags into the curl request and `http` library code: https://everything.curl.dev/usingcurl/timeouts.html
2. Append to the RSC API config string the optional timeout parameters, so that they can be configurable (is backwards compatible)
3. Setting those timeout flags for each of the requests to the RSC endpoints in `remote_cache_api.wake`
4. Handling of the timeout by creating a sentinel file in the client side, that is written into a uuid corresponding to the wake invocation that caused the timeout, and is used as a check to disable the cache for further requests from the cache from that wake invocation


share/wake/lib/system/http.wake
```diff
@@ -243,13 +255,24 @@ def makeCurlCmd ((HttpRequest url method headers body formData): HttpRequest) (e
 #
 # WARNING: This function will likely break the sandbox and make your build unreliable
 export def makeRequest (request: HttpRequest): Result HttpResponse Error =
+    def maxTimeFlag =
```
AbrarQuazi, line 258: move this down one layer to `makeCurlCmd`

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -366,14 +374,27 @@ export def rscApiPostJob (req: CachePostRequest) (api: RemoteCacheApi): Result U
...
-    require Pass response =
+    def request =
         makeRoute api "job"
         | buildHttpRequest
         | setMethod HttpMethodPost
         | addContentTypeJsonHeader
         | addAuthorizationHeader auth
         | setBody jsonStr
-        | makeRequest
+        | setTimeout 300 10
```
AbrarQuazi, line 384: Potentially think about having a `runRequest` function that is called by `makeRequest` and @ here, to be able to match on job status, if it makes sense

otherwise add comment in `http.wake` about the fragile error cause equality

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -366,14 +374,27 @@ export def rscApiPostJob (req: CachePostRequest) (api: RemoteCacheApi): Result U
...
+    def attempt = makeRequest request
+
+    require Pass response = match attempt
+        Pass r -> Pass r
+        Fail err ->
+            def cause = err.getErrorCause
+            if cause ==* "curl timed out (exit code 28)"
+            then
+                def _ = disableRemoteCache
+                Fail  (makeError "rsc: Remote cache disabled due to timeout")
```
AbrarQuazi, line 395: `def _ = println "message here"` for debugging/testing purposes 

or printlnLevel logDebug "message here" if I want to include it here

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -645,6 +727,18 @@ export def rscApiGetFileBlob ((CacheSearchBlob blobId uri): CacheSearchBlob) (pa
 
     Pass path
 
+export def disableCacheFilePath: String = ".build/tmp/disable_remote_cache"
+
+export def isRemoteCacheDisabled: Boolean =
```
AbrarQuazi, line 732: dont use unsafe_stat since its a target and is cached. 

options:
* copy unsafe_stat and make it a def
* use makejob with setplanpersistance as rerun, and use local runner for "ls file"

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -36,6 +36,23 @@ export target rscRunner (rscApi: RemoteCacheApi): Runner =
...
+            | setPlanPersistence Once
+            | setPlanFnOutputs (\_ Nil)
+            | runJobWith localRunner
+            | setJobInspectVisibilityHidden
+            | isJobOk
+        else failWithError "Could not remove {disableCacheFilePath}"
+
+        Pass Unit
+
+    def _ = cleanupDisableRemoteCacheFile
```
AbrarQuazi, line 54: change this to require for "parallelism sync point" , so we know for sure that this is run.

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -36,6 +36,23 @@ export target rscRunner (rscApi: RemoteCacheApi): Runner =
...
+
```
AbrarQuazi, line 55: The problem with removing the file is that it is difficult to guarantee the ordering of what happens first (file creation or deletion), so its difficult to gurantee whether you are in the current state or the previous state run. 

The alternative solution is to always have the disableCacheFile exist, and instead write to the file a unique id  (set plan persistance once so that it can be saved in memory for that wake invocation)for that specific wake invocation- this will allow us to know if its actually the current state of the world.

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: In the case where Im generating a new uid for each wake invocation I need to compare the following in `isRemoteCacheDisabled`
* generate uid func -> with setplanpersistance once
* read from disableRemoteCache file

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: Instead of a runRequest function, I created a helper within `remote_cache_api`, because I can't have rsc logic code in `http.wake` - which would have happened if I had this specific `runRequest` function

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -645,6 +694,49 @@ export def rscApiGetFileBlob ((CacheSearchBlob blobId uri): CacheSearchBlob) (pa
...
+    require Pass currentWakeUuid = randomUUID Unit
+    else False
+
+    match (readTmpFile disableCacheFilePath)
+        Pass writtenUuid ->
+            if writtenUuid ==* currentWakeUuid
+            then
+                True
+            else
+                False
```
ag-eitilt:
```suggestion
        Pass writtenUuid -> writtenUuid ==* currentWakeUuid
```

That said, I wonder if it would be better to have this function be `guardRemoteCacheDisabled` or something, with this `match` block forming an inner def, and then move the `require False = ... \ failWithError cascadeTimeoutErrorMessage` into this function as well, so that calling it you just need `require Pass _ = guardRemoteCacheDisabled Unit`.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -645,6 +694,49 @@ export def rscApiGetFileBlob ((CacheSearchBlob blobId uri): CacheSearchBlob) (pa
...
+    uuidJob
+    | getJobStdout
+
+def disableRemoteCache Unit: Result Unit Error =
+    require Pass uuid = randomUUID Unit
+    require Pass _ = write disableCacheFilePath uuid
+    Pass Unit
+
+def readTmpFile (pathStr: String): Result String Error =
+    def imp p = prim "read"
```
ag-eitilt, line 728: Definitely add a doc comment here saying why we can't use the standard `read` (`target` *and* `Path`).

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -673,12 +767,30 @@ def getPathAsJson path =
...
     )
 
+# Helper function for further processing the result of a http request in relation to the rsc
+def matchRequestResult (label: String) (attempt: Result HttpResponse Error): Result HttpResponse Error =
+    match attempt
+        Pass r -> Pass r
+        Fail err ->
+            def cause = err.getErrorCause
+
+            if cause ==* "curl timed out (exit code 28)" then
```
ag-eitilt, line 777: I've said before that I don't like matching against the Stringy cause, but given the fact that we'd need to open a hole in `package http` in order to get anything else, I think this is good enough here.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -969,3 +1083,114 @@ def mkCacheSearchResponse (json: JValue): Result CacheSearchResponse Error =
     obytes
     | Match
     | Pass
+
+## --- Test functions ---
+export def testDisablingCacheWithEndpoints (api: RemoteCacheApi): Result Unit Error =
```
ag-eitilt, line 1088: I don't know what the standard pattern for test functions is, but unless you're using package-private functions here, it might be better to pull this into a separate file with a separate `*_test` package, to keep the API that little bit cleaner?

tests/remote-cache/disable-cache-cascade/test.wake
ag-eitilt, line 1: This and the `pass.sh` are missing end-of-file newlines.

ag-eitilt (approved):
I think the most blocking comment would be getting rid of the `localRunner`, but overall this looks good!

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -705,11 +693,13 @@ def timeoutErrorMessage: String =
 def cascadeTimeoutErrorMessage: String =
     "rsc: Remote cache previously disabled due to timeout"
 
-def randomUUID Unit: Result String Error =
+# Generates a UUID (using date in nano seconds) for the current wake invocation.
```
ag-eitilt, line 696: Hurrah for portability!  Slight nitpick that this makes it not a UUID, but just a reasonably-unique string.  (It's kind of impressive how much design work has gone into avoiding accidental collisions for true UUIDs.)

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -645,6 +694,49 @@ export def rscApiGetFileBlob ((CacheSearchBlob blobId uri): CacheSearchBlob) (pa
...
+
+def timeoutErrorMessage: String = "rsc: Remote cache disabled due to timeout"
+def cascadeTimeoutErrorMessage: String = "rsc: Remote cache previously disabled due to timeout"
+
+def randomUUID Unit: Result String Error =
+    def uuidJob =
+        makeExecPlan ("uuidgen", "-r", Nil) Nil
+        | setPlanLabel "generate uuid"
+        | setPlanPersistence Once # Exacally one UUID per wake invocation
+        | runJobWith localRunner
```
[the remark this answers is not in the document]
ag-eitilt: Per external discussion, since `defaultRunner` winds up delegating to the RSC in certain cases, this actually *should* be `localRunner` to avoid any chance of accidental loop.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -719,26 +709,32 @@ def randomUUID Unit: Result String Error =
...
 
 def disableRemoteCache Unit: Result Unit Error =
-    require Pass uuid = randomUUID Unit
+    require Pass uuid = generateUUID Unit
     require Pass _ = write disableCacheFilePath uuid
 
     Pass Unit
 
+# This is only used to read temporary files that are purposefully not tracked by wake (i.e the sentinel disable cache file).
+# This is used instead of the read function from io.wake as that takes in a Path.
```
ag-eitilt, line 718:
```suggestion
# This is used instead of the read function from io.wake as that takes in a Path, and caches the file contents for the rest of the wake invocation.
```

ag-eitilt (approved):

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -253,27 +257,41 @@ export def getRemoteCacheApiCanPull (api: RemoteCacheApi): Boolean =
...
         | int
     else failWithError "Remote cache config was set with non-integer port. Saw: {portStr}"
 
+    # Defualt maxTime to 300 and connTime to 10 if empty
+    def maxT =
+        if maxStr ==* ""
+        then Some 300
+        else match (int maxStr)
+            Some i -> Some i
+            None -> Some 300
```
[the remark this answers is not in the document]
AbrarQuazi: I actually decided to make this include error handling of bad input, otherwise this would affect the parsing of the API key, and any invalid key can become valid (as maxT and connT would take any input)

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: I dont believe `read` caches as its not a `target`:https://github.com/sifive/wake/blob/2d320d71dbff8c1ef149e1d7511716ffa11b3028/share/wake/lib/system/io.wake#L125 , from our previous discussions from before that was actually `unsafe_stat` from the previous implementation

tests/remote-cache/disable-cache-cascade/test.wake
AbrarQuazi: For some reason, I am getting a weird issue, where git is not recognizing the end-of-file new lines, this is how my file looks like:
Screenshot 2025-04-11 at 9 48 59 AM

share/wake/lib/system/remote_cache_api.wake
ag-eitilt: Huh, I knew stat was, but I thought I had looked up read as well and saw the target there.  Probably a good thing, though, that could easily get to be a lot om memory.

tests/remote-cache/disable-cache-cascade/test.wake
ag-eitilt: Weird.  In the VS code settings, what are `editor.renderFinalNewline`, `files.insertFinalNewline`, and `files.trimFinalNewlines`?  Mine are "off", enabled, and disabled, respectively.

PR #1689, opened 2025-04-24
Update broken README.md URLs after migration to sifiveinc

davidbiancolin:
There are probably more of these i did not make any attempt to be thorough.


AbrarQuazi (approved):
LGTM, thanks David!

PR #1691, opened 2025-05-07
bsp-wake: only build the requested project

terpstra:
The build server spams requests for compile of no projects. Don't build in that case.

Also: pass the desired project to wake vscode.
That requires a matching change.


AbrarQuazi (approved):
LGTM, Thanks!

Issue #1692, opened 2025-05-09
PID namespace function returns last child process status instead of user cmd process


ngraybeal:
The [child wait loop](https://github.com/sifiveinc/wake/blob/cd26789d4dc2582593d0f9648baa6849454913d9/src/wakefs/namespace.cpp#L521) appears to assume the user launched cmd (pid 0) will be the last child process to finish and therefore uses the last child status as the main process exit status. This can lead to the incorrect exit status being reported if the main cmd is not the last process to finish. I believe this is large in part because any orphaned process will be reparented to the pid namespace. 

This issue was discovered when running an `npm build` command where `npm` exited with status 1, but `node` processes where orphaned and the final child exit has status 0, giving us an incorrect exit status result.

PR #1693, opened 2025-05-13
#1692 Specifically wait on user_cmd_process and return its result

ngraybeal:
Intends to address [issue](https://github.com/sifiveinc/wake/issues/1692). 

This change ensures that the user_cmd_child process result is the final result returned from `pidns_init`. Before it was returning the last child process, which could result in false positives.


AbrarQuazi (approved):
If you can show that this change resolves the npm build issue, that would be great. Other  than that, LGTM

PR #1701, opened 2025-08-04
Bugfix: Fix PID namespace init hanging due to improper cleanup of zombie processes

AbrarQuazi:
`pidns_init` was incorrectlying waiting for the main user command first, then cleaning up other children or orphaned processes. This caused hangs when child processes exited before the main process, as they weren't being reaped immediately.

To fix this, this PR improves `pidns_init` by:

- using seperate pid variables so that values are not overwritten
- fix waitpid loop to reap all orphaned processes, including the user command

### DIFF
--- a/src/wakefs/namespace.cpp
+++ b/src/wakefs/namespace.cpp
@@ -516,7 +516,8 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
     exit(1);
   }
 
-  int user_cmd_status = 0;
+  int user_cmd_status = -1;
+  bool user_cmd_finished = false;
 
   // Reap all children as they exit
   while (true) {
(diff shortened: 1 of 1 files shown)


src/wakefs/namespace.cpp
```diff
@@ -507,30 +507,40 @@ bool setup_user_namespaces(int id_user, int id_group, bool isolate_network,
...
     // Execute the user-specified command.
     const auto *args = reinterpret_cast<const pidns_args *>(arg);
     err = execve_wrapper(args->command, args->environment);
     std::cerr << "execve " << args->command[0] << ": " << strerror(err) << std::endl;
     exit(1);
   }
 
-  int user_cmd_status;
-  waitpid(pid, &user_cmd_status, 0);
+  int user_cmd_status = 0;
```
ngraybeal, line 519: Should the default be healthy? What if `user_pid` status is never returned for some reason.

AbrarQuazi: Yeah, we should check for that case as well

PR #1703, opened 2025-08-06
Wake db open version error bug fix

AbrarQuazi:
When doing schema updates to the wake.db, we would get a confusing error message like `Failed to open wake.db: no such column: runner_status`. The reason is because wake is doing version checking after applying the schema. This check needs to be done before to get a clear error message - however since we update our version within the schema table this is not possible with the current implementation.

To properly address this, we should start deprecating the schema version table, and instead have the version in our DB header (via `PRAGMA user_version`) for quick access and allows us to check for the version without executing the schema (and without additional locks).

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -284,7 +284,7 @@ std::string Database::open(bool wait, bool memory, bool tty) {
 
     // since older wake.db files may not have the header set, we need to check the legacy table,
     // to differentiate between a brand-new DB and an old one.
-    // Todo: remove this check once all wake.db instances have migrated to newer versions.
+    // TODO: remove this check once most wake.db instances have the newer version.
     int legacy_ver = 0;
     if (header_ver == 0) {
       sqlite3_stmt *st = nullptr;
@@ -320,7 +320,7 @@ std::string Database::open(bool wait, bool memory, bool tty) {
       close_db(imp.get());
       return db_ver > atoi(SCHEMA_VERSION)
                  ? "wake.db was created by a newer version of Wake; please upgrade Wake."
-                 : "wake.db was created by an older version of Wake; remove it.";
+                 : "wake.db was created by an older version of Wake; please remove it.";
     }
 
     char *fail = nullptr;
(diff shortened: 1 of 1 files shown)


AbrarQuazi:
Proof of it working. First is a call to show the error message people were seeing, and then testing with the updated changes
Screenshot 2025-08-06 at 5 02 20 PM

src/runtime/database.cpp
```diff
@@ -272,16 +272,77 @@ std::string Database::open(bool wait, bool memory, bool tty) {
...
+      sqlite3_stmt *st = nullptr;
+      if (sqlite3_prepare_v2(imp->db, "PRAGMA user_version;", -1, &st, nullptr) == SQLITE_OK &&
+          sqlite3_step(st) == SQLITE_ROW)
+        header_ver = sqlite3_column_int(st, 0);
+      sqlite3_finalize(st);
+    }
+
+    // since older wake.db files may not have the header set, we need to check the legacy table,
+    // to differentiate between a brand-new DB and an old one.
+    // Todo: remove this check once all wake.db instances have migrated to newer versions.
```
ngraybeal, line 287: I can't say I agree with this TODO since I don't believe we can guarantee at point in time when this will be true. 

One Nit: 
```suggestion
    // TODO: remove this check once all wake.db instances have migrated to newer versions.
```

src/runtime/database.cpp
```diff
@@ -272,16 +272,77 @@ std::string Database::open(bool wait, bool memory, bool tty) {
...
+      }
+    }
+
+    int db_ver = header_ver ? header_ver : legacy_ver;
+
+    if (db_ver && db_ver != atoi(SCHEMA_VERSION)) {
+      close_db(imp.get());
+      return db_ver > atoi(SCHEMA_VERSION)
+                 ? "wake.db was created by a newer version of Wake; please upgrade Wake."
+                 : "wake.db was created by an older version of Wake; remove it.";
```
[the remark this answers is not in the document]
AbrarQuazi: 😂 just going off of how we did it previously below: "produced by an incompatible version of wake; remove it.", but yes I'll be more nice

src/runtime/database.cpp
AbrarQuazi: Removing the version table in the schema will help clean out the code, so we dont have to maintain it and check it in two places. I dont think we need all instances of the wake.db to have been migrated, just a decent amount of time to have been passed so that most users haver the user_version in their wake.db header. The only fallout of removing this check when there are still wake.db instances of the older version, is that they will still get a wake.db error related to the actual schema error- but i'll go ahead and add an error message that is less confusing for that error path as well.

PR #1704, opened 2025-08-11
Add wake.db migration tool

AbrarQuazi:
Added the `wake-migration` tool so that users can now easily update and migrate their wake.db whenever wake releases have database version upgrades/schema changes without having to delete their current wake db instance.


ngraybeal (approved):
LGTM

PR #1707, opened 2025-08-26
"Modifiable" input files for incremental builds

ag-eitilt:
Much as we'd like to have Wake be the only build system managing compilation, that's not a realistic assumption to make when working with languages that have strong ecosystems.  In those cases, a solution for passing multiple files at once to a build tool and having that tool determine which need to be recompiled is often optimal, but Wake's sandboxing prevents that determination from occurring.  This allows Runners to selectively poke holes in the sandbox to pick up artifacts from previous compilation, *and* to write to those same files if the tool determines the artifact needs to be *re*compiled.  Crucially, that's *only* exposed at the level of the Runner, and user-facing Plans don't have any way of adding modifiable files -- the intent is to allow defining an incremental runner for each tool (a la `incrementalCRunner` for C+Make in the test[^make]), but not to allow this knob as a free-for-all footgun.

[^make]: C+Make isn't actually something that should be implemented using this method; *that* setup, at the least, is well-served by just using Wake's caching -- better-served, actually, since it enables hash-based recompilation rather than timestamp-based.  It is, however, a very simple model to build a test around.

Additionally, the PR properly marks files in the previous `visible` list as not being writable within the sandbox, regardless of their permissions outside it.  That is the existing *behaviour* as enforced by the FUSE implementation, and why simply having the Runners adding files to `RunnerInputs.Visible` rather than the new `RunnerInputs.Modifiable` wasn't sufficient.

Side note: I wouldn't have had any hope of finding the second tier of sandbox file in 289bdc1abe08fb1bbd0149c6b73425c4822a8574 if it weren't for the AI assistant.

I'd still want to go back through before merging this to add documentation, double-check the generated C test code meets my coding style, etc., but I wanted to get this posted first to get discussion going; if we determine that this is too unsafe, I wouldn't want to have to throw out all the time to polish it up.


share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -337,7 +337,7 @@ def mkPostJobRequest ((RunnerInput label cmd vis env dir stdin _res _prefix _ is
     obytes
 
 # Creates a CacheAllowedRequest from the various inputs and outputs of a runner
-def mkCacheAllowedRequest ((RunnerInput label cmd vis env dir stdin _res _prefix _ isAtty _ _): RunnerInput) (output: RunnerOutput) (hidden: String) =
+def mkCacheAllowedRequest ((RunnerInput label cmd vis _ env dir stdin _res _prefix _ isAtty _ _): RunnerInput) (output: RunnerOutput) (hidden: String) =
     def Usage status runtime cputime mem _ibytes obytes = output.getRunnerOutputUsage
 
     CacheAllowedRequest label cmd dir env hidden isAtty stdin vis status runtime cputime mem obytes
```
ag-eitilt, line 343: Most of the existing runners seem fine enough ignoring the `modifiable` inputs.  This along with `mkJobCacheRunner` are the pair I'm most concerned about.  So long as the invariant that the same outputs are produced from the same inputs holds up, it *should* be fine to cache, but it feels a bit too easy to have a buggy/malicious artifact slip through and poison the cache.

AbrarQuazi (commented):
Don't have any issue with this first pass for implementation, other than some questions. I don't think that this is an unsafe change, and am ok with you going further along with this, as I see this as a value add to what the sandbox should be allowed to do (ignoring or writing to incremental compile collaterals) as long as we properly gate what gets to be defined as a Modifiable input file

PR #1709, opened 2025-09-10
Add content_hash field to verify cached blobs when downloaded

AbrarQuazi:
In order to protect against cache poisoning of the blob store, this PR introduces the `content_hash` field that sends a blake2b hash of the contents of the output file blobs of a cached job to the RSC server for it to store in the DB, as a way for clients to check against the hash when downloading the blob from the store during a cache hit.

### DIFF
--- a/rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
+++ b/rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
@@ -1,0 +1,40 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Blob::Table)
+                    .add_column(
+                        ColumnDef::new(Blob::ContentHash)
+                            .string()
+                            .not_null()
+                            .default("")
+                    )
+                    .to_owned(),
+            )
+            .await
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Blob::Table)
+                    .drop_column(Blob::ContentHash)
+                    .to_owned(),
+            )
+            .await
+    }
+}
+
+#[derive(Iden)]
+enum Blob {
+    Table,
+    ContentHash,
+}
--- a/rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
+++ b/rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
@@ -14,7 +14,6 @@ impl MigrationTrait for Migration {
                         ColumnDef::new(Blob::ContentHash)
                             .string()
                             .not_null()
-                            .default("")
                     )
                     .to_owned(),
             )
--- a/rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
+++ b/rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
@@ -33,7 +33,7 @@ impl MigrationTrait for Migration {
     }
 }
 
-#[derive(Iden)]
+#[derive(DeriveIden)]
 enum Blob {
     Table,
     ContentHash,
--- a/share/wake/lib/system/http.wake
+++ b/share/wake/lib/system/http.wake
@@ -120,8 +120,9 @@ export def setBody (body: String): HttpRequest => HttpRequest =
 export def addFormData (name: String) (value: Path) (contentType: Option String): HttpRequest => HttpRequest =
     unsafe_addFormData name (HttpFormDataFile value.getPathName) contentType
 
-# Adds a string to the request's form data
-export def addStringFormData (name: String) (value: String) (contentType: Option String): HttpRequest => HttpRequest =
+# Adds a string to the request's form data. This is marked unsafe as it is the callers responsibility to ensure the string
+# is shell safe.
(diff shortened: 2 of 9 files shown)


AbrarQuazi:
Proof of this working, I modified blobs in the backend blob store, and got the builds passing with the following logs in wake.log. Previously, poisoning the cache would cause builds to fail, now with these changes the client builds the job locally.


`{"hostname":"pe01.internal.sifive.com","pid":"2165006","source":"wake","time":"2025-09-15T18:00:41.598570953","level":"info","message":"rsc: Hash mismatch for blob .build/wake/stdlib/http/binary.141a84a14150b5d722d7c94b88f66a0abf3656bcd7defbb206408821a7841a6e. Expected: 759b001593352b43daa49fc0661c4dcf68b715beda4ae06b5b40e12cb228d9db, Got: 00d6fef8ea9079f661f64ecb9463cf62fabba1f669c18e85ee6e23ac4cc70104"}`

Screenshot 2025-09-12 at 2 24 40 PM

share/wake/lib/system/http.wake
```diff
@@ -21,18 +21,25 @@ tuple HttpHeader =
     Name: String
     Value: String
 
-# A form data entry. It consists of a user defined name and a file on disk to upload.
+# A form data entry. It consists of a user defined name and either a file on disk or string to upload.
 #
-# Note: while raw strings are generally accepted as http form data the actual encoding of the
-# string results in shell escape headaches. Strings should be written to a file before upload.
```
AbrarQuazi, line 27: I am extending the library to include strings as I dont want to overload the client with thousands of writes of hashes for each output file just for curl to send to the server. Since we are dealing with hashes which are hex, we dont need to worry about shell escapes for this current use case. In general, I think we should allow string as form data, and let it be the responsibility of the user to encode/ handle shell escapes properly

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -1084,3 +1142,31 @@ def mkCacheSearchResponse (json: JValue): Result CacheSearchResponse Error =
     obytes
     | Match
     | Pass
+
+def hashUsage =
+    Usage 0 0.0 0.1 0 0 0
+
+# RSC version of computing hashes for files, similar to Hashcode target in path.wake
+def rsciHashFile (file: String): Result String Error =
```
ag-eitilt, line 1150: Is the `i` meaningful or a typo?  And you say "similar to", but how do they differ?  As on the meeting notes, any way to merge the two?

rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
```diff
@@ -0,0 +1,40 @@
...
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Blob::Table)
+                    .add_column(
+                        ColumnDef::new(Blob::ContentHash)
+                            .string()
+                            .not_null()
+                            .default("")
```
ag-eitilt, line 17: I always greatly prefer `None`/`NULL` as a fallback for strings rather than `""`.  Though the client requirement does make `NULL` a bit more of a hassle here.

ag-eitilt (commented):
Definitely not opposed to this overall, but there might be some things we want to clean up.  Also, I might have missed it, but do we have (do we even want?) any server-side verification of blob hashes?  Whether that's just after upload to be sure the communication with the client wasn't interrupted/corrupted, as part of eviction passes in case anything gets corrupted at rest (unlikely, but catches collisions) and/or to avoid latency from hashing as part of the client handovers, or before download to be sure it's sending the right thing?  It's more compute on the server, particularly for large files, and probably overkill unless we see the sort of corruption that prompted this more often despite this PR, but it did seem a reasonable question.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -65,6 +65,8 @@ tuple CacheSearchBlob =
     Id: String
     # The uri where the blob may be downloaded from. Not guaranteed to be http/https
     Uri: String
+    # The content hash of the blob used for verification
+    ContentHash: String
```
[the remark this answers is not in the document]
AbrarQuazi: we want it to be a hard requirement for blobs to have hashes for this verification to work, so Ill be cleaning the postgres prod db and the blob store for prod before turning RSC on again.

share/wake/lib/system/http.wake
```diff
@@ -207,9 +219,14 @@ def methodToString = match _
...
-        Some ct -> "--form", "{name}=@\"{file}\";type={ct}", Nil
-        None -> "--form", "{name}=@\"{file}\"", Nil
+    def formDataToCurlFlag (HttpFormData name content contentType) =
+        def typeFlag = match contentType
+            Some ct -> ";type={ct}"
+            None -> ""
+
+        match content
+            HttpFormDataFile file -> "--form", "{name}=@\"{file}\"{typeFlag}", Nil
+            HttpFormDataString str -> "--form", "{name}={str}{typeFlag}", Nil
```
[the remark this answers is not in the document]
AbrarQuazi: for our specific use case, `typeFlag` is used to set if the blob is `blob/small` or `blob/large` for the RSC to determine if the blob should be saved in the database or the blob store. I dont think this field really matters for strings (or our case hashes), and people should be using the file format anyways if they need more specification with the typeFlag, so I think we can go ahead and use `--form-string` instead.

rust/migration/src/m20250900_000000_add_content_hash_to_blob.rs
ag-eitilt: > we want it to be a hard requirement for blobs to have hashes for this verification to work

Let's not provide a default value here, then?  Prevent the inconsistent state from being specified to begin with, and all.

PR #1712, opened 2025-09-17
Handle orphaned blobs better in blob eviction logic for blob storage

AbrarQuazi:
Currently there is an issue with the RSC where blobs are not evicted properly and remain orphaned in the blob store. This can lead to disk space issues over time. To properly handle this, we should modify the blob eviction logic to delete from the blob store first before deleting from the database, so that we can handle retries if deletion fails


ag-eitilt (approved):
Got a couple questions, but mostly about *how* things work rather than *why* they're this way.  Definitely a good catch for why the blobs were getting orphaned; let's get this in and see if there's any other leaks!

PR #1716, opened 2025-10-06
Allow for multiple concurrent readers to the wake.db

AbrarQuazi:
This PR introduces introduces WAL (Write-Ahead Log) checkpointing, and a wake lock file. This enables multiple read-only inspection commands to run safely while builds are active, while preventing conflicting concurrent builds. This PR also resolves the issue of querying the DB in multiuser environments (ex: CI machines), allowing for easier debugging outlined in the issue: https://github.com/sifiveinc/wake/issues/501. 

By setting `locking_mode=normal` and having Write-ahead logging enabled, we will be able to have further features that builds upon this like multiple wake invocations for building targets in the same workspace.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -137,12 +137,40 @@ struct Database::detail {
         get_runner_status(0) {}
 };
 
+class WaitingIndicator {
+ private:
+  bool waiting = false;
+  bool tty;
+
+ public:
+  WaitingIndicator(bool tty_) : tty(tty_) {}
+
+  void show_waiting(const std::string &message) {
+    if (!tty) return;
+
+    if (!waiting) {
+      waiting = true;
+      std::cerr << message << " .";
+    } else {
+      std::cerr << ".";
+    }
+  }
+
+  void finish() {
+    if (waiting && tty) {
+      std::cerr << std::endl;
+      waiting = false;
+    }
+  }
+
+  ~WaitingIndicator() { finish(); }
+};
+
 static void close_db(Database *db) {
   if (!db || !db->imp || !db->imp->db) {
     return;
   }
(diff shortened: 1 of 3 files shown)


src/runtime/schema.h
```diff
@@ -1,15 +1,16 @@
...
+#define SCHEMA_VERSION "9"
 
 // Increment the SCHEMA_VERSION every time the below string changes.
 // Also add migrations to the wake-migration tool if needed.
 inline const char* getWakeSchemaSQL() {
   return "pragma auto_vacuum=incremental;"
          "pragma journal_mode=wal;"
          "pragma synchronous=0;"
-         "pragma locking_mode=exclusive;"
+         "pragma locking_mode=normal;"
```
AbrarQuazi, line 12: having `journal_mode=wal` and `locking_mode=normal;` allows concurrent access to the database. Since we are changing from `exclusive`, we now need to prevent the case of multiple writers - hence having file based locks for non-introspection commands.

src/runtime/database.cpp
```diff
@@ -965,6 +1108,11 @@ void Database::finish_job(long job, const std::string &inputs, const std::string
   end_txn();
 
   if (fail) exit(1);
+
+  static int job_count = 0;
+  if (++job_count % 50 == 0) {  // Keep WAL manageable by checkpoinging every 50 jobs
```
ngraybeal, line 1113: Can you please make magic number 50 a named variable please.

ngraybeal (approved):
Perhaps we should figure out the minimum version of sqlite3 is required to support this.

Otherwise, this seems like a solid foundation towards "multi-wake" and enabling people to do "no-harm" reads of the db.

PR #1721, opened 2025-11-25
Use `map ... | findFail` in preference to `findFailFn`

ag-eitilt:
In debugging some odd behaviour around Jobs, we've noticed some slowdown caused by `sources` pulling in *large* lists of files for hashing; specifically, the use of `findFailFn` serializes the hash calls, whereas allowing them to be run in parallel would saturate the Job queue but still allow a faster overall completion.  This applies the already-existing guidance to only use `findFailFn` for trivial getters to a narrower definition of "trivial".

For the moment, this might make operations in certain environments slightly worse.  Specifically, as `sources` can now saturate the Job queue, wake execution may wind up with larger CPU/memory spikes, which might be noticeable on constrained systems where Wake is allowed full access to the resources (though the 90%-by-default limit should prevent systems from *fully* grinding to a halt).  We're hoping to transition to a better solution to hashing files in general (i.e. a `prim` rather than a Job), though there are still questions open on how to best handle certain elements of that.  In the mean time, `--jobs` and `--memory` may be used to limit execution if necessary.

### DIFF
--- a/share/wake/lib/core/json.wake
+++ b/share/wake/lib/core/json.wake
@@ -480,7 +480,7 @@ def deduplicateJObjectKeys (simplifyValueList: List JValue => Result JValue Erro
         _ -> simplifyValueList values
 
     groupBy cmpKeysOnly dict
-    | findFailFn simplifyKeyGroup
+    | findFailSerial simplifyKeyGroup
 
 # Attempt to determine equality across simple values, concatenate `JArray`s, and
 # recurse into `JObject`s.
--- a/share/wake/lib/core/result.wake
+++ b/share/wake/lib/core/result.wake
@@ -157,20 +157,20 @@ export def rfoldr (combiningFn: element => accumulator => Result accumulator err
 #   findFail (Pass 456, Pass 123, Nil) = Pass (456, 123, Nil)
 #   findFail (Pass 456, Fail "xyz", Fail "abc", Nil) = Fail "xyz"
 export def findFail: List (Result a b) => Result (List a) b =
-    findFailFn (_)
+    findFailSerial (_)
 
-# findFailFn: if fn returns Pass for all List elements, return the outputs else Fail
+# findFailSerial: if fn returns Pass for all List elements, return the outputs else Fail
 # Once a Fail is found, fn is not evaluated on further elements.
 # This means that fn is applied to the List mostly sequentially.
 # If more parallelism is desired, use 'map fn | findFail' instead.
 # In particular, this function is most appropriate for cases where the Result is simply a member
-# of a tuple (e.g. `findFailFn getPairSecond`) or for other *trivial* transformations rather than
-# anything that does meaningful work.
+# of a tuple (e.g. `findFailSerial getPairSecond`) or for other *trivial* transformations rather
+# than anything that does meaningful work.
 #
 #   def toInt x = int x | getOrFail "not an Integer ({x})"
-#   findFailFn toInt ("456", "123", Nil) = Pass (456, 123, Nil)
-#   findFailFn toInt ("_56", "123", Nil) = Fail "not an Integer (_56)"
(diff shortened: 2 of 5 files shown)


ag-eitilt:
In addition to the reverted JSON prints, I'm also seeing some flaky behaviour locally with the cancelation test.  In one run, the `timeout` call used seems to have killed things before it had a chance to print anything out:
```shellsession
sammay@host:wake$ make test
WAKE_PATH=/usr/lib64/ccache ./bin/wake --in test_wake runTests
[warning] : inspection/canceled failed! Stdout: 
[warning] : 
[warning] : inspection/canceled failed! Stderr: 
[warning] : Early termination requested
[warning] : No jobs matched query
[warning] : 
[warning] : tmp/bin/wake unit testing results:
[warning] :   - command-line: 2/2 tests pass
[warning] :   - config: 6/6 tests pass
[warning] :   - dst: 3/3 tests pass
[warning] :   - inspection: 0/1 tests pass
[warning] :     - canceled: Test failed. Expected: Ok, Actual: Exited 1. See above for details
[warning] :   - job-cache: 6/6 tests pass
[warning] :   - parser: 1/1 tests pass
[warning] :   - remote-cache: 1/1 tests pass
[warning] :   - runtime: 1/1 tests pass
[warning] :   - standard-library: 11/11 tests pass
[warning] :   - type-system: 11/11 tests pass
[warning] :   - wake-format: 3/3 tests pass
[warning] :   - wakebox: 8/8 tests pass
[warning] : ---
Fail (Error "Test failed. Expected: Ok, Actual: Exited 1. See above for details" Nil)
make: *** [Makefile:77: test] Error 1
sammay@host:wake$ cat tests/inspection/canceled/pass.sh 
#! /bin/sh

set -e
WAKE="${1:+$1/wake}"

# Start from any empty db every time for stable job ids
rm -f wake.db

# Use || true to ignore the expected non-0 return from timeout
timeout 1 ${WAKE} test || true

${WAKE} --canceled --simple
```
And in another the job ID didn't remain stable as the expected `1`:
```shellsession
sammay@host:wake/tests/inspection/canceled$ ./pass.sh ../../../bin 
[ #    ] sleep 1000
Early termination requested
#  (2)
$ sleep 1000
```
(I didn't capture the database after that run, but for comparison this is what comprises a successful test; I'm not sure how anything could have slipped ahead of the `sleep` Job:)
```shellsession
sammay@host:wake/tests/inspection/canceled$ ../../../bin/wake --job 1
#  (1)
$ sleep 1000
sammay@host:wake/tests/inspection/canceled$ ../../../bin/wake --job 2
# mkdir: .build 0755 (2) [inspect.visibility=hidden,]
$ <mkdir> -m 0755 .build
sammay@host:wake/tests/inspection/canceled$ ../../../bin/wake --job 3
# hash: #2 (3) [inspect.visibility=hidden,]
$ /scratch/sammay/wake/tmp/bin/../lib/wake/wake-hash .build
0000000000000000000000000000000000000000000000000000000000000000
sammay@host:wake/tests/inspection/canceled$ ../../../bin/wake --job 4
# write: .build/spec-1.json 0644 (4) [inspect.visibility=hidden,]
$ <write> -m 0644 .build/spec-1.json
sammay@host:wake/tests/inspection/canceled$ ../../../bin/wake --job 5
# hash: #4 (5) [inspect.visibility=hidden,]
$ /scratch/sammay/wake/tmp/bin/../lib/wake/wake-hash .build/spec-1.json
f5c72cc0965fae2784a7626c8939dedfbf76e9bbaea0e4d3b18486510c9080bf
sammay@host:wake/tests/inspection/canceled$ ../../../bin/wake --job 6
No jobs matched query
```

It's suspicious that I got *both* of those errors while working with this PR, but I don't expect it to have directly impacted that test.  I'd like to go ahead and merge this before leaving on vacation, but we should keep an eye on CI and be aware that we might want to stabilize it before making a release if it keeps being an issue.

PR #1737, opened 2026-02-03
Workspace virtualization fuse changes

AbrarQuazi:
This PR:
- expects the fuse client to have a hash along with the path as an input for visible files (from wake's `Path` object)
- Fuse daemon adds files,directories, symlinks, hardlinks to metadata mapping and staging directory (files only)
- Implements changes to fuse I/O operations to properly read from CAS store and write to staging. Currently there is an ENV variable `WAKE_CAS` to be able to switch between CAS mode and writing directly to workspace (for backwards compatibility purposes).
- Fuse daemons output JSON now has `staging_files` key, which has relevant metadata and distinction between files, directories, symlinks, and hardlinks like so:
```
{
  "staging_files": {
    "build/foo.o": {
      "type": "file",                                   //can be hardlink, symlink, dir
      "staging_path": ".cas/staging/1",
      "mode": 493,
      "mtime_sec": 1234567890,
      "mtime_nsec": 123456789
    },
...
  }
}
```
- Fuse client is able to parse the `staging_files` output, hash the staged output files and insert them into the CAS
- Fuse client stores the staged file metadata along with its output hash in result.json files, so that the adding to CAS and materialization of outputs to workspace can be handled downstream (by the wake runtime):
```
  "hashed_files": {
    "build/foo.o": {
      "type": "file",
      "mode": 493,
      "mtime_sec": 1234567890,
      "mtime_nsec": 123456789,
      "hash": "a1b2c3d4e5f6..."          //newly added hash
    },
  },
  ```
  
  Future PR's will need to remove the ability to directly read from the workspace (once we implement `Source` input files directly being added to the CAS)

### DIFF
--- a/tests/wakebox/cas-file-write/input.json
+++ b/tests/wakebox/cas-file-write/input.json
@@ -1,0 +1,24 @@
+{
+  "command": [
+    "/bin/sh",
+    "-c",
+    "echo 'hello world' > test_output.txt\ncat test_output.txt\nls -la test_output.txt"
+  ],
+  "environment": [
+    "USER=root",
+    "HOME=/root",
+    "PATH=/usr/bin:/bin:/usr/sbin:/sbin"
+  ],
+  "directory": ".",
+  "stdin": "",
+  "user-id": 0,
+  "group-id": 0,
+  "mount-ops": [
+    {
+      "type": "workspace",
+      "destination": "."
+    }
+  ],
+  "visible": []
+}
+
--- a/tools/fuse-waked/main.cpp
+++ b/tools/fuse-waked/main.cpp
@@ -41,6 +41,7 @@
 #include <set>
 #include <sstream>
 #include <string>
+#include <variant>
 
 #include "compat/nofollow.h"
 #include "compat/utimens.h"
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

File: tools/fuse-waked/main.cpp [hunk 1]
```diff
@@ -92,14 +288,46 @@ void Job::parse() {
...
+  for (auto &x : jast.get("visible").children) {
+    std::string path;
+    std::string hash;
+
+    if (x.second.kind == JSON_OBJECT) {
+      path = x.second.get("path").value;
+      hash = x.second.get("hash").value;
+    } else {
+      // Legacy format: just a string path
+      path = x.second.value;
```
dtzSiFive: When will "legacy" format be used/reach this code?

Is this reachable? Is this for manual use pointing at old JSON or thereabouts?

Not breaking things is nice, but also a liability.  Is it expected this will be removed at some point?

tools/fuse-waked/main.cpp
```diff
@@ -132,7 +360,64 @@ void Job::dump() {
...
+  //     "mtime_sec": ..., "mtime_nsec": ... }, "path/to/link": { "type": "hardlink",
+  //     "staging_path": ".cas/staging/123_1", "mode": ..., "mtime_sec": ..., "mtime_nsec": ... },
+  //     "path/to/sym":  { "type": "symlink", "target": "..." },
+  //     "path/to/dir":  { "type": "directory", "mode": .. }
+  //   }
+  if (g_use_cas) {
+    s << "],\"staging_files\":{";
+    first = true;
+    if (auto *job_staged = g_staged_files.get_job(job_id)) {
+      for (auto &entry : *job_staged) {
```
dtzSiFive: Nit: Consider, e.g.: `for (const auto &[_, sf] :` or thereabouts 👍 .

(And maybe `si` for "StagedItem" instead of `sf` (?)?)

tools/fuse-waked/main.cpp
```diff
@@ -132,7 +360,64 @@ void Job::dump() {
...
+        const StagedItem &sf = entry.second;
+
+        size_t start = 0;
+        size_t lastslash = sf.dest_path.rfind("/");
+        if (lastslash != std::string::npos) start = lastslash + 1;
+        if (sf.dest_path.compare(start, prefix.length(), prefix) == 0) continue;
```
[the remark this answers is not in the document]
dtzSiFive: Anyway consider pulling this into a helper, and/or have a comment -- I had to go looking to grok what "prefix" might possibly be here s.t. we skip them.

Probably any files Job creates with this name are not supported / badness ?

Should this be gated on being a file specifically, or can this be anything?
If directory, do we ensure has trailing slash? So that there's no `/path/to/.fuse-hidden` (if it's a directory on disk)

tests/wakebox/cas-file-write/input.json
```diff
@@ -0,0 +1,24 @@
+{
```
dtzSiFive: These tests I think can land separately and directly to master?

Is anything about them CAS-specific?   I see they set USE_CAS, but they are testing wakebox in a CAS-agnostic way, right?

Wonder if could reasonably run tests both with and without USE_CAS or something, instead of having specific tests for each?

That said, once we're done we will only have the one path (is that right?), so it'll just be wakebox tests.

For your consideration 👍 .

I got nervous about all the hardcoded things like PATH and USER/HOME but haven't checked if there's precedent.  Probably that's standard.

dtzSiFive (commented):
First round of comments!

Have you looked at executing some filesystem tests (e.g., https://github.com/kdave/xfstests or possibly https://github.com/tuxera/pjd-fstest)?

Mostly: There are a /lot/ of details to get right.

Just a thought re:above.
At least kicking the tires by some jobs that explicitly exercise filesystem behavior seems useful, even if full compliance is not our primary objective or realizable.

That said, those tests (without studying them) may also be insufficient, as they may not be written to expose the sorts of issues our implementation has due to its nature that other filesystems may not.

May help identify unexpected/unintended behavior, however, and likely there are others/better test suites.

AbrarQuazi:
Accidentally merged these changes. I reverted this merge in the feature branch, and continuing work on: https://github.com/sifiveinc/wake/pull/1787

PR #1738, opened 2026-02-03
Workspace virtualization wake runtime changes

AbrarQuazi:
Wake Runtime Changes for workspace virtualization: 
- Adds typed path metadata to Wake’s runtime and database flow so paths carry type information alongside hashes, which is needed to represent regular files, directories, and symlinks cleanly in CAS-aware execution.
- Extends hashing and path-resolution plumbing to work with typed outputs and mode bits, including the CAS-side helpers needed to ingest already-hashed workspace content without recomputing hashes.
-  Registers the new CAS primitives in the Wake runtime so Wake code can query CAS state and ingest/materialize CAS-backed items.
- Introduces a staged-output model in Wake with OutputSet, OutputPaths, and OutputStaged, so runners can distinguish between ordinary workspace outputs and FUSE/CAS-style staged outputs.
- Adds staging_outputs.wake to parse staged runner metadata, resolve staged files/symlinks/directories into final Path values, and normalize them for job finalization.
- Updates job finalization so all staged items can be ingested/materialized while only the filtered published subset is stored as the job’s official outputs.
- Keeps non-published staged outputs tracked as cleanable outputs so they remain visible to Wake’s cleanup flow without being promoted to published hashed outputs.
- Preserves the legacy string-based visible format in the review branch so the changes stay independent of the pending wakefs/FUSE transport changes.


share/wake/lib/system/cas.wake
```diff
@@ -0,0 +1,34 @@
+# Copyright 2024 SiFive, Inc.
```
dtzSiFive: Please make pass updating Copyright year 👍 .

I'm not sure our policy here, but seems a little unexpected to say Copyright a year before this was created.

src/runtime/cas_prim.h
```diff
@@ -0,0 +1,53 @@
...
+
+// Context for CAS operations, manages the Cas lifecycle
+class CASContext {
+ public:
+  CASContext() = default;
+  ~CASContext() = default;
+
+  // Get or create the CAS store for a workspace
+  // Returns nullptr if CAS initialization fails
+  cas::Cas* get_store(const std::string& workspace);
```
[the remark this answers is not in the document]
AbrarQuazi: Yeah definitely we can do this 👍

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -1144,8 +1144,13 @@ def hashUsage =
 
 # RSC version of computing hashes for files, similar to Hashcode target in path.wake
 def rscHashFile (file: String): Result String Error =
+    def addCasHashEnv = match (getenv "WAKE_CAS")
+        Some value -> setPlanEnvVar "WAKE_CAS" value
+        None -> identity
```
ag-eitilt: You've already got this defined elsewhere.

ag-eitilt (changes requested):
Only reviewed (most of) the Wake files so far.  I'll check the C++ eventually, but this should give you plenty to look at/discuss in the mean time, while I turn to my own deadlines for this afternoon.

tools/wake-migrate/main.cpp
```diff
@@ -190,6 +207,51 @@ static std::vector<Migration> get_migrations() {
...
+                                nullptr) != SQLITE_OK) {
+           sqlite3_finalize(select_stmt);
+           return false;
+         }
+
+         bool ok = true;
+         while (sqlite3_step(select_stmt) == SQLITE_ROW) {
+           const unsigned char* path_txt = sqlite3_column_text(select_stmt, 0);
+           const unsigned char* hash_txt = sqlite3_column_text(select_stmt, 1);
+           std::string path = path_txt ? reinterpret_cast<const char*>(path_txt) : std::string();
```
dtzSiFive: When might these might not have values? (that seems like an error?)
Does "BadHash" currently ever reach our DB?

share/wake/lib/system/path.wake
```diff
@@ -110,142 +110,165 @@ export def basename (file: String): String =
...
+                    else
+                        Pass None
+        else
+            Pass (Some cachedPathInfo)
+
+export def getCachedPathMode (file: String): Result (Option Integer) Error =
+    match (getCachedPathInfo file)
+        Pass None -> Pass None
+        Pass (Some pathInfo) -> Pass (Some pathInfo.getPathInfoMode)
+        Fail err -> Fail err
```
ag-eitilt: I can see why you'd want a helper for this "monad" stack, but there's an easier way than unwrapping manually:
```suggestion
export def getCachedPathMode (file: String): Result (Option Integer) Error =
    file.getCachedPathInfo
    | rmap (omap getPathInfoMode)
```

share/wake/lib/system/runner.wake
```diff
@@ -142,12 +170,179 @@ export def localRunner: Runner =
...
+        def stageInputsLen = len normAllOutputPaths
+        def stageInputsBytes = sum (map strlen normAllOutputPaths)
+        def useStageStdinFile = stageInputsBytes >= 1900000 || stageInputsLen >= 12000
+        def stageStdinContent = catWith "\n" normAllOutputPaths
+        def stageStdinFilePath = ".build/tmp/workspace.stage.{hashString stageStdinContent}.stdin"
+        def stageCommandBase = "{workspaceStageTool}", "--staging-base", workspaceStageBase, Nil
+
+        require Pass stagePlanPair = match useStageStdinFile
+            True ->
+                require Pass stdinFile = write stageStdinFilePath stageStdinContent
```
ag-eitilt: Same comment on a small bit less bloat by putting `stageStdin(Content|FilePath)` here rather than outside the block.

share/wake/lib/system/runner.wake
```diff
@@ -330,24 +535,76 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
...
+        # Extract string from either a plain string or an object with "path" field
+        # This handles both old format (strings) and new format (objects with path/hash)
+        def extractPath (jval: JValue): Option String = match jval
+            JString s -> Some s
+            JObject fields ->
+                # find returns Option (Pair (Pair String JValue) Integer)
+                match (find (\(Pair k _) k ==* "path") fields)
+                    Some (Pair (Pair _ pathVal) _) -> getJString pathVal
+                    None -> None
+            _ -> None
```
ag-eitilt: I know we aren't going to be switching everything to the CAS for *storage* for a while, but isn't this still a layer we'll be able to unconditionally use the new format for?  Or is this here primarily for backwards compatibility with old `.build` files (reading or branch-switching)?

ag-eitilt (changes requested):
*Still* not had a chance to look at the C++, but here's my next (overlarge, sorry 😅) flurry of Wake-side comments.  Lot fewer critical concerns, lot more nits and cleanup as we move toward a good version.

share/wake/lib/system/cas.wake
```diff
@@ -0,0 +1,78 @@
...
+    Some value -> setPlanEnvVar "WAKE_CAS" value
+    None -> identity
+
+export def primCasHasBlob (hash: String): Boolean =
+    def raw hash = prim "cas_has_blob"
+
+    raw hash
+
+# Materialize an item from CAS to workspace.
+export def primCasMaterializeItem (destPath: String) (itemType: String) (hashOrTarget: String) (mode: Integer) (mtimeSec: Integer) (mtimeNsec: Integer): Result Unit Error =
```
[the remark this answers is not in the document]
AbrarQuazi: So I do think it makes sense to go to the database directly for these arguments, but that means that I will have to mix in my `CasContext` defined in `cas_prim.h`, which  currently only manages the CAS lifecycle - to now account for the database as well - which leads into thinking about separation of concern (it would be nice for this CAS runtime wrapper to not know about the database, and leave CAS prims to deal with CAS only, but I don't have a super hard opinion on this).

Would it be potentially okay to just leave the `prim` versions of these as is, and not export them and have them as local helpers? Export only `ingestStagedItemIntoCas`, `materializeStagedItemFromCas` and have them take the `PreparedStagingItem` as arguments instead, so we minimize the falling out of sync problem?

We can then revisit this once we have the multi-wake DB and CAS cleanup changes, to see if we really do want to have the CAS runtime wrapper to have details about the DB as well.

src/runtime/database.cpp
```diff
@@ -1266,6 +1293,35 @@ std::string Database::get_hash(const std::string &file, long modified) {
   return out;
 }
 
+std::pair<std::string, std::string> Database::get_hash_type(const std::string &file,
+                                                            long modified) {
+  std::pair<std::string, std::string> out;
```
dtzSiFive: Please wrap these in `begin_ro_txn()` / `end_ro_txn()` for consistency.

For single statement this should be equivalent, but these all should be aligned and I'd prefer we were explicit.  We especially don't want this to have two separate statements / transactions as they might observe different data (MVCC).

share/wake/lib/system/path.wake
```diff
@@ -110,142 +110,165 @@ export def basename (file: String): String =
...
-    def to_hash_file_bytes = sum (map strlen to_hash)
+    if hash ==* "" then
+        Pass None
+    else
+        require Pass pathType = pathTypeFromString pathTypeString
+        else failWithError "Failed to decode cached path type '{pathTypeString}' for {file}"
+
+        def cachedPathInfo = PathInfo file pathType hash mode
+
+        # TODO: Remove the non-CAS cache acceptance path once WAKE_CAS is the default.
```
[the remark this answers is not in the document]
AbrarQuazi: Yeah I think this is unecessary. Was being defensive against if DB has the path and CAS didn't (previous run didnt use WAKE_CAS, and the new run enables CAS) - but people shouldnt be switching between CAS and no CAS mode (makes things more complex), and should just blow away their DB's.

The other question is on failure modes, if somehow we add to the DB but fail to add to the CAS, if that would cause an issue - but I think on the next rerun it would just self heal and fix itself

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -1144,26 +1144,36 @@ def hashUsage =
...
-        stdout
-        | extract `(.{64}).*`
-
-    match fileHash
-        (hash, Nil) -> Pass hash
-        _ -> failWithError "rsc: No hash found in output for file {file}"
+    if cached !=* "" then
+        Pass cached
+    else
+        # Fall back to shim job for files not yet in DB (e.g. downloaded blobs).
```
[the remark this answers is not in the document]
AbrarQuazi: I actually think this is not needed anymore, as the post-job finalization callback takes care of the hash being in the DB for the upload cases, and for the download case we resolve the blobPath into the DB before we call rscHashFile

AbrarQuazi (commented):
responding to feedback

share/wake/lib/system/job.wake
```diff
@@ -201,14 +292,40 @@ def runAlways cmd env dir stdin res uusage finputs foutputs vis keep run echo st
...
     def cache = primJobCache jobKey
 
+    # TODO: Replace nested-Pair return from C++ claim_tree with a proper Wake tuple.
+    def parseCachedPathInfo (Pair name (Pair pathTypeString (Pair hash (Pair mode mtimeNs)))) =
+        require Pass pathType = pathTypeFromString pathTypeString
+        else failWithError "Failed to parse cached output type '{pathTypeString}' for {name}"
+
+        Pass (PathInfo name pathType hash mode mtimeNs)
+
+    require Pass cachedOutputs =
```
dtzSiFive: I'm not sure we should silently ignore primitive encoding errors (arguments or return results)?

That seems like a Panic and something that by construction we avoid?
Just like calling a primitive with incompatible arguments.

share/wake/lib/system/job.wake
```diff
@@ -201,14 +292,40 @@ def runAlways cmd env dir stdin res uusage finputs foutputs vis keep run echo st
...
+
     match cache
-        Pair (job, _) last -> confirm True last job
-        Pair Nil last -> confirm False last (build Unit)
+        Pair (job, _) _ ->
+            if useCasHashes then
+                require Pass _ = materializeCachedOutputs cachedOutputs
+                else build Unit
+
+                confirm True cachedOutputs job
```
dtzSiFive, line 338: This `confirm` check seems borderline useless, probably we should check /before/ materializing?

WDYT?

I'm not sure this checks against CAS corruption even, since we will set mtime to known value when materializing I believe (?).

--------

That said, I think `confirm` has issues for multi-wake:

What stops another run from changing the file (here, post-materialization, so small window)?

Or if we check before materializing, another run could .... discussion taken elsewhere.

share/wake/lib/system/http.wake
```diff
@@ -403,8 +403,8 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
         # Curl jobs don't use many resources, most time is spent waiting for a response
         | setPlanUsage (Usage 0 0.0 0.1 0 0 0)
         | setPlanEcho logDebug
-        # Use localRunner to help avoid potential high namespace overhead.
-        | runJobWith localRunner
+        # Use workspaceRunner to help avoid potential high namespace overhead.
+        | runJobWith workspaceRunner
```
ag-eitilt, line 407: We might want to think about moving this into a `stagingDirRunner` to centralize the directory handling, but that's definitely polish to think about *after* this PR lands.

share/wake/lib/system/io.wake
```diff
@@ -135,10 +135,16 @@ export def read (path: Path): Result String Error =
 # Ideally content would be part of RunnerInputCmd however this gets tracked exactly in the database
 # which means all writes would use 2x the total storage in the database.
 def writeRunner (content: String) =
+    def primStageMkdir (path: String): Result String Error =
+        def raw mode path = prim "mkdir"
+
+        raw 0755 path
+        | rmapError makeError
```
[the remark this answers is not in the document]
ag-eitilt: Definitely something to think about, and we could pass the mode through from the single callsite in `writeImp` easily enough.  Currently our only internal uses of `writeIn` (which is the only entrypoint which might set a different mode) are all still set to enable global read, so it's not a critical issue, but it could still be a security leak if someone does come up with something where the path to the file is the sensitive data.  Like you said, detail for later.

share/wake/lib/system/job.wake
```diff
@@ -170,14 +172,123 @@ def runAlways cmd env dir stdin res uusage finputs foutputs vis keep run echo st
...
-
-            def _ = unsafe_primJobSetRunnerStatusMessage job runnerError
-
-            primJobFinish job inputs.implode hashedOutputs cleanable.implode reality
+            def needsWorkspaceMaterialization =
+                outputs.getStagedOutputsNeedsWorkspaceMaterialization
+
+            # Maybe materialize a prepared published item to the workspace.
+            def maybeMaterialize preparedItem =
+                if needsWorkspaceMaterialization then
```
ag-eitilt: Not critical, but it might be nice to pull this block out into a couple separate functions -- at least the materialization logic, maybe some of the finalization, etc.  `runAlways` was already a complex function, and it would probably read better to avoid putting even more stuff inline for it.

share/wake/lib/system/path.wake
```diff
@@ -110,156 +110,256 @@ export def basename (file: String): String =
...
-            hashPlan ("{wakePath}/../lib/wake/wake-hash", to_hash) Nil
+            hashPlan ("{wakePath}/../lib/wake/wake-hash", "--typed", simpleFiles) Nil
             | Pass
-    else mergeSelect which_files_to_hash (map (add _ "BadHash") to_hash) not_to_hash
+    else failWithError "Failed to prepare batched hash job"
 
     def job =
         plan
+        | setPlanLabel "hash: {str toHashLen} paths"
         | setPlanEcho logDebug
```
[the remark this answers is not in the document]
ag-eitilt: This previous comment is a nit, but I feel it's an important nit.  To add on, I definitely agree that passing the `prefix` to the wrapper just for the single purpose of putting it in the label does come across as a bit silly, but it's also the only thing we have which connects the hash job to the one it's being run for, so it might still be nice to do.

share/wake/lib/system/path.wake
```diff
@@ -110,151 +114,169 @@ export def basename (file: String): String =
...
+            def stdinFilePath = ".build/tmp/hash.{str wakePid}.{hashString stdinContent}.stdin"
+
+            require Pass stdinFile = write stdinFilePath stdinContent
 
-            hashPlan ("{wakePath}/../lib/wake/wake-hash", "@", Nil) (stdin_file,)
-            | setPlanStdin stdin_file.getPathName
+            hashPlan ("{wakePath}/../lib/wake/wake-hash", "@", Nil) (stdinFile,)
+            | setPlanStdin stdinFile.getPathName
             | Pass
+            |< (\p Pair p (Some stdinFilePath))
```
ag-eitilt, line 244: Not for this PR, but `Plan.Stdin` is one of the fields I'd *really* like to update to contain an `Option Path` instead of a `String`; the other is the `Plan.Directory`, with just a plain `Path`.  We'd be able to automatically add them to the visible list rather than requiring the user add that manually, and you'd be able to `getPlanStdin` specifically rather than dealing with this `Pair` sidecar.  Plus, for the directory, that then ensures that we aren't trying to have any weird, likely-to-escape interactions with files/locations outside the workspace.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -756,9 +756,26 @@ def generateID Unit: Result String Error =
...
+
+    Pass Unit
+
 def disableRemoteCache Unit: Result Unit Error =
-    require Pass id = generateID Unit
-    require Pass _ = write disableCacheFilePath id
+    require Pass uuid = generateID Unit
+    require Pass _ = writeUntrackedTmpFile disableCacheFilePath uuid
 
     Pass Unit
```
ag-eitilt, line 780: This is something we're going to need to revisit for multi-wake.  We're writing to a single file, and expecting that the contents exactly match the ID of the current run, when there could actually now be multiple runs all trying to disable the remote cache logic.  *Maybe* not blocking the beta release, but if it doesn't go in now, it should be one of the first things we get in after it.

Side note, I previously deliberately renamed `uuid`->`id`.  If you want something more specific, go for `runId` or `rscCacheId` or something, but this is directly derived from a timestamp and is *not* a proper UUID.

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -347,158 +382,145 @@ def mkCacheAllowedRequest ((RunnerInput label cmd vis env dir stdin _res _prefix
...
+    def outputUploadsResult =
+        def stagedOutputPaths =
+            stagingItems
+            | map getStagingItemDestPath
+
+        require True = subset scmp filteredOutputPaths stagedOutputPaths
+        else failWithError "rsc: Missing staging metadata for published outputs"
+
+        def uploadStagingFile (StagingFileOutput destPath _stagingPath mode _ _) =
+            # Post-finalization: staging file has been consumed. Upload from workspace dest path.
```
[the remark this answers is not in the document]
ag-eitilt: Bumping this as at least a yellow flag.

share/wake/lib/system/runner.wake
```diff
@@ -142,12 +169,183 @@ export def localRunner: Runner =
...
+def workspaceStageBase =
+    ".build/staging"
+
+# Stages a declared workspace-output list with wake-stage and returns the full staged output
+# metadata plus any helper paths that should be cleaned later.
+#
+# allOutputPaths: every output path that should be represented in staging
+# filteredOutputPaths: the published subset of allOutputPaths
+# baseCleanableOutputs: non-staging cleanables to preserve alongside helper artifacts
+export def stageWorkspaceOutputs (allOutputPaths: List String) (filteredOutputPaths: List String) (baseCleanableOutputs: List String) (alreadyInWorkspace: Boolean): Result (Pair StagedOutputs (List String)) Error =
```
ag-eitilt: Nits: `alreadyInWorkspace` isn't described in the docstring, and we probably want to explicitly say what "published" means (i.e. exposed in `getJobOutputs`) for `filteredOutputsPaths`.  "cleanables to preserve" is also a bit unclear.

share/wake/lib/system/runner.wake
```diff
@@ -142,12 +169,183 @@ export def localRunner: Runner =
...
+        def filteredInputs = localOutput.getRunnerOutputInputs
+        def runnerUsage = localOutput.getRunnerOutputUsage
+        def baseRunnerError = localOutput.getRunnerOutputRunnerError
+
+        # Preserve the base command's status. Stage outputs regardless of exit code so that
+        # getJobFailedOutputs can access them.
+        match baseRunnerError
+            Some err -> RunnerOutput filteredInputs emptyStagedOutputs Nil runnerUsage (Some err)
+            None ->
+                match (stageWorkspaceOutputs publishedOutputPaths publishedOutputPaths publishedOutputPaths False)
```
ag-eitilt: Nit: Probably cleaner to extract this function call to a `def`, though it does make sense to keep that `def within the `None` block.

Also not important: Do we really want to keep `publishedOutputPaths` for all three lists?  I guess it wouldn't *hurt*, but it feels extraneous for at least the cleanable outputs (see my comment last set on trying to reduce the overlap between them).

share/wake/lib/system/runner.wake
```diff
@@ -242,7 +449,8 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
                 "label" :-> JString label,
                 "command" :-> command | map JString | JArray,
                 "environment" :-> environment | map JString | JArray,
-                "visible" :-> visible | map (_.getPathName.JString) | JArray,
+                #TODO: Add updated visible transport fileds {path,type,hash,mode}
+                "visible" :-> visible | map getPathName | map JString | JArray,
```
ag-eitilt, line 459: I'm guessing this is a TODO that's solved in https://github.com/sifiveinc/wake/pull/1787?

share/wake/lib/system/runner.wake
```diff
@@ -330,24 +538,83 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
...
-        def filteredOutputs = fnOutputs jsonOutputs
-        def cleanableOutputs = specFile, resultFile, jsonOutputs
-
-        match jobUsageResult
-            Pass jobUsage ->
+        def filteredOutputPaths = fnOutputs jsonOutputs
+        def defaultCleanableOutputs = specFile, resultFile, jsonOutputs
+
+        # `fnOutputs` is not always a pure filter over `jsonOutputs`. Runners that use
+        # `TrackingDisabled` (e.g. `bindmount`) tell wakebox not to record writes, so
```
ag-eitilt: Nit 1: We can still bindmount other things other places without setting `TrackingDisabled`.
Nit 2: `TrackingDisabled` is a feature of the internal-only sandbox, not the open source -- it's something that we've *meant* to decouple from expecting specific technologies and bring over here, but we've never gotten around to doing so.
```suggestion
        # `TrackingDisabled` (e.g. `bindmount` the workspace itself) tell wakebox not to record writes, so
```

share/wake/lib/system/runner.wake
```diff
@@ -330,24 +538,83 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
...
+            | distinctBy scmp
+
+        # TODO: Remove the non-CAS post-staging branch once WAKE_CAS is the default.
+        def outputSetResult = match useCasHashes
+            True -> match (parseStagingFiles content)
+                Pass stagingItems ->
+                    # If the runner reported no staged items but `fnOutputs` discovered
+                    # outputs on disk (the TrackingDisabled case), CAS cannot ingest from
+                    # the empty `staging_files` set. Fall back to the workspace-staging
+                    # path so the discovered outputs are picked up from the workspace.
```
ag-eitilt, line 594: Same note here.

Really, though, this is just a mismatch between models where the user (`fnOutputs`) is declaring outputs unknown to the runner (`stagingItems`, not `staging_files`).  The case where the `stagingItems` is empty (an entirely non-sandbox-aware runner) is probably the only time we want to allow `filteredOutputPaths` to contain things that aren't in `stagingItems` -- which obviously isn't the case since we're running within the JSON runner -- but that level of change is something we'll have to more carefully evaluate.  If nothing else, it *seems* like it would be nice to modify the output (and input?) Strings from that function to be relative to the staging directory rather than to the workspace root.  This is clearly working for now, let's not mess with it for beta, but let's mark it to revisit before MVP.

share/wake/lib/system/sources.wake
```diff
@@ -185,6 +185,6 @@ export def claimFileAsPathIn (outputDirectory: Path) (existingFile: String) (des
         | makePlan "claim: {existingFile} -> {desiredWorkspacePath}" visible
         | setPlanPersistence Once
         | setPlanFnOutputs (\_ desiredWorkspacePath, Nil)
-        # This needs to be localRunner in case the existingFile is already inside the workspace.
-        | runJobWith localRunner
+        # This needs to be workspaceRunner in case the existingFile is already inside the workspace.
+        | runJobWith workspaceRunner
```
ag-eitilt, line 189: I think this probably winds up getting a more complex runner -- eventually.  We're currently copying the original file somewhere within the workspace, then copying *that* to a staging directory to be hashed, and finally to the CAS.  We should be able to get away with copying directly to the staging directory instead of stopping over in the workspace first.

share/wake/lib/system/staging_outputs.wake
```diff
@@ -0,0 +1,322 @@
...
+# Reads the staged item type string from the JSON payload for one destination path.
+def stagingItemType (destPath: String) (itemInfo: JValue): Result String Error =
+    require Some itemType =
+        itemInfo // `type`
+        | getJString
+    else failWithError "Missing type for {destPath}"
+
+    Pass itemType
+
+# Parses the JSON metadata shared by staged file-like entries.
```
ag-eitilt: Nit: I'm sure someone's going to wonder what's "file-*like*", so maybe mention hardlinks and anything else getting rolled into this explicitly?

share/wake/lib/system/staging_outputs.wake
```diff
@@ -0,0 +1,322 @@
...
+# Note: Symlinks are excluded here because their hashes are computed directly from the
+# target string.
+def stagingHashInput (item: StagingItem): Option (Triple String String Integer) = match item
+    StagingFile (StagingFileOutput destPath stagingPath _ mtimeSec mtimeNsec) ->
+        Some (Triple destPath stagingPath (mtimeNsFromParts mtimeSec mtimeNsec))
+    _ -> None
+
+# Converts staged metadata into PreparedStagingItem values by hashing staged files from their
+# staging paths and deriving hashes for symlinks/directories. Checks the dest-path hash cache
+# first to avoid re-hashing files whose content is already known (e.g. source files).
```
ag-eitilt:
```suggestion
# first to avoid re-hashing files whose content is already known (e.g. unchanged source files).
```

share/wake/lib/system/staging_outputs.wake
```diff
@@ -0,0 +1,322 @@
...
+                None -> None
+        )
+
+    def uncachedInputs =
+        hashInputs
+        | filter (
+            \(Triple destPath _ mtimeNs) match (tryCachedFileHash destPath mtimeNs)
+                Some _ -> False
+                None -> True
+        )
```
ag-eitilt, line 220: Pretty sure this will trigger twice the database lookups we actually need.  Let's `map` this at the end of calculating `hashInputs`, and then add it as a fourth value on each entry; then in these `mapPartial`/`filter`, we can just check that stored value.  Or we make `getCachedPathInfo` a `target` and deal with the memory overhead.

share/wake/lib/system/staging_outputs.wake
```diff
@@ -0,0 +1,322 @@
...
+
+    # Only hash files that missed the cache.
+    def freshHashesResult: Result (List String) Error =
+        computeHashes (map getTripleSecond uncachedInputs)
+
+    match freshHashesResult
+        Fail err -> Fail err
+        Pass freshHashes ->
```
ag-eitilt: There's not actually any parallelism to be gained by delaying the pattern-match on the result (and even if you want to keep a separate `def` for future plans or something, you can just use a separate `require` rather than the `match`).

```suggestion
    require Pass freshHashes: List String =
        computeHashes (map getTripleSecond uncachedInputs)
```

share/wake/lib/system/staging_outputs.wake
```diff
@@ -0,0 +1,322 @@
...
+
+    def publishedDestPathSet =
+        filteredOutputPaths
+        | map (\destPath destPath :-> Unit)
+        | listToMap scmp
+
+    def stagedDestPathSet =
+        stagingItems
+        | map (\item getStagingItemDestPath item :-> Unit)
+        | listToMap scmp
```
ag-eitilt: The type you're looking for here is (admittedly confusingly) `Tree`.
```suggestion
    def publishedDestPathSet =
        filteredOutputPaths
        | listToTree scmp

    def stagedDestPathSet =
        stagingItems
        | map getStagingItemDestPath
        | listToTree scmp
```

share/wake/lib/system/staging_outputs.wake
```diff
@@ -0,0 +1,322 @@
...
+        filteredOutputPaths
+        | mapPartial (\destPath mlookup destPath stagingItemByDestPath)
+
+    def unpublishedStagingItems =
+        stagingItems
+        | filter (
+            \item match (mlookup (getStagingItemDestPath item) unpublishedDestPathSet)
+                Some _ -> True
+                None -> False
+        )
```
ag-eitilt:
```suggestion
        | filter (mcontains _.getStagingItemDestPath unpublishedDestPathSet)
```

ag-eitilt (approved):
Fair bit of nits, still, but I don't think any of these are critical.  I'd be happy with you applying as much as you feel a need for, merging, and dealing with the rest afterward.

share/wake/lib/system/http.wake
AbrarQuazi: Yeah definitely agree the need of a `stagingDirRunner` as a base runner for specific runners like these and potentially RSC

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: Went ahead with the name change, thanks!

We will definitely need to revisit RSC implementation for both CAS and multiwake post beta, so will keep this in mind

share/wake/lib/system/remote_cache_runner.wake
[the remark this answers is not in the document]
AbrarQuazi: Yeah I plan on including this apart of the work for RSC post beta, shouldnt affect beta though as CAS should not be on with RSC for beta users

share/wake/lib/system/runner.wake
AbrarQuazi: yep!

share/wake/lib/system/runner.wake
AbrarQuazi: put it to the def, 

we can revisit the `publishedOutputPaths` later

share/wake/lib/system/sources.wake
AbrarQuazi: yep, I agree!

share/wake/lib/system/runner.wake
```diff
@@ -197,30 +197,31 @@ def workspaceStageTool =
     "{wakePath}/../lib/wake/wake-stage"
 
 def workspaceStageBase =
-    ".build/staging"
+    ".build/staging/{str wakePid}"
```
ag-eitilt, line 200: Per discussion on #1760, we'll probably eventually want to have this be the run (row) ID rather than the PID, for consistency with the database lock files if nothing else, but I'm not sure that's particularly easy to check at the moment and PID is definitely sufficient for beta.

ag-eitilt (approved):
Fixes look good, merge ahead!

PR #1740, opened 2026-02-06
Blake3 migration

AbrarQuazi:
Updates our hashing algorithm from blake2b to blake3, which should lead to faster hashing performance: https://github.com/BLAKE3-team/BLAKE3. This will bring wake and RSC to utilize the same hashing algorithm.

Blake3 C and header files were taken from https://github.com/BLAKE3-team/BLAKE3/tree/master/c version: 1.8.2


tools/wake-migrate/main.cpp
```diff
@@ -190,6 +190,22 @@ static std::vector<Migration> get_migrations() {
        },
        "Convert runner_status from INTEGER to TEXT"},
 
+      // Version 9 -> 10: Hash algorithm changed from BLAKE2b to BLAKE3
+      // Migration is not supported - user must delete wake.db and reinitialize
+      {9, 10,
+       [](sqlite3* db) -> bool {
+         (void)db;  // unused
+         std::cerr << std::endl;
```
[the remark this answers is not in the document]
dtzSiFive: Certainly not if we want to drop blake2 support :).

We /could/ support this if we wanted -- promote to one of the schemes (maybe like SRI hashes) that specifies the hash kind with the hash value/result, but I'm on board with the break.  Just saying it's not necessary, but also without costs of course.

dtzSiFive (approved):
LGTM, nice!

I did not review the contents of the `blake3` directory, I assume it was almost entirely (other than blake3.wake) imported from a release tarball (or equivalent).

Do we want / need to sort out how to enable vector support for x86_64 (probably essentially all of our users), conceivably dynamically detecting feature support?

ngraybeal (approved):
LGTM

AbrarQuazi:
Some stats from initial testing on sourcing all files from repo:

Binary | Mean wall-clock | Std dev | User CPU | System
-- | -- | -- | -- | --
blake3 (blake3-migration) | 108.5 s | ±3.8 s | 173.6 s | 135.1 s
blake2 (feature/multi-wake) | 113.0 s | ±1.0 s | 269.8 s | 141.3 s

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(112, 122, 140); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(31, 36, 48); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Summary:</strong> blake3 is 1.04× faster — ~4.4 s less wall-clock per cold run over 61,694 files.</p>

I think the most benefit here is User CPU time due to SIMD support

tools/wake-migrate/main.cpp
```diff
@@ -521,9 +521,24 @@ static std::vector<Migration> get_migrations() {
        },
        "Add files.deleted column to retain metadata while allowing CAS space recovery"},
 
+      // Version 15 -> 16: Hash algorithm changed from BLAKE2b to BLAKE3.
+      // Every stored file hash, job signature, and CAS blob is BLAKE2b-based and
```
AbrarQuazi, line 525: Not supporting migrating, and telling users to remove their wake.db

vendor/blake3/blake3.wake
```diff
@@ -0,0 +1,81 @@
...
+    | getJobOutputs
+
+# Get the list of SIMD assembly files based on architecture
+def blake3AsmFiles: List String =
+    if machine ==~ "x86_64" then
+        "blake3_sse2_x86-64_unix.S",
+        "blake3_sse41_x86-64_unix.S",
+        "blake3_avx2_x86-64_unix.S",
+        "blake3_avx512_x86-64_unix.S",
+        Nil
```
ag-eitilt, line 38: The machine architecture isn't a sufficient test for the instruction flags -- for example, my personal system has an Alder Lake CPU which doesn't support AVX-512.  If the `*.S` files include the relevant feature tests, then all good; if they're meant to be added as the result of external testing, we're going to need to be a bit more specific.  (I'm not seeing feature tests in the files which suggests the latter, but I could be missing something.)

Makefile
```diff
@@ -34,18 +34,48 @@ COMMON_CPP  := $(foreach dir,$(COMMON_DIRS),$(wildcard $(dir)/*.cpp))
...
+# BLAKE3 C files (always compiled)
+BLAKE3_C_COMMON := vendor/blake3/blake3.c vendor/blake3/blake3_portable.c vendor/blake3/blake3_dispatch.c
+
+# BLAKE3 platform-specific files and flags
+ifeq ($(UNAME_M),x86_64)
+    # x86_64: Use SSE2/SSE4.1/AVX2/AVX-512 assembly implementations
+    BLAKE3_ASM := vendor/blake3/blake3_sse2_x86-64_unix.S \
+                  vendor/blake3/blake3_sse41_x86-64_unix.S \
+                  vendor/blake3/blake3_avx2_x86-64_unix.S \
+                  vendor/blake3/blake3_avx512_x86-64_unix.S
```
ag-eitilt, line 49: Same comment here.

vendor/blake3/blake3.wake
AbrarQuazi: So feature detection happens at runtime in `blake3_dispatch.c:114` via `get_cpu_features(),` which uses cpuid/xgetbv to detect what the actual CPU supports. The `.S` files are then gated based on which files are actually supported, so including them shouldn't be an issue

PR #1742, opened 2026-02-10
Add debug statments capturing error message from client mount connection

ngraybeal:
In order to help with debugging `Could not contact FUSE daemon` issues, this adds some debug logging in the client daemon connect loop based on the DEBUG_FUSE_WAKE env var.

When enabled, it will now print the attempt number and then `open()` call error code/str.
In the message, I include the `[fuse_client]` prefix to identify where its coming from. I updated the `Could not contact FUSE daemon` to include it as well.


dtzSiFive (approved):
LGTM

AbrarQuazi (approved):
Thanks!

PR #1744, opened 2026-02-10
fuse-daemon startup error logging

ngraybeal:
This creates a macro to print the stderr messages during Fuse-Daemon startup to both stderr and the log file if debug is enabled.

Default behavior is to just print to stderr.

### DIFF
--- a/tools/fuse-waked/main.cpp
+++ b/tools/fuse-waked/main.cpp
@@ -1262,6 +1262,18 @@ static struct fuse *fh;
 static struct fuse_chan *fc;
 static sigset_t saved;
 
+// Print to stderr, and optionally to a debug log.
+struct DebugLogger {
+  template <typename... Args>
+  void operator()(const char *fmt, Args... args) {
+    fprintf(stderr, fmt, args...);
+    if (debug && fd != -1) dprintf(fd, fmt, args...);
+  }
+
+  bool debug = false;
+  int fd = -1;
+};
+
 static struct fuse_operations wakefuse_ops;
 
 static void *wakefuse_init(struct fuse_conn_info *conn) {
@@ -1408,6 +1420,7 @@ int main(int argc, char *argv[]) {
   int log, null;
   bool madedir;
   struct rlimit rlim;
+  DebugLogger debug_log{debug, -1};
 
   if (argc != 3) {
     fprintf(stderr, "Syntax: fuse-waked <mount-point> <min-timeout-seconds>\n");
@@ -1437,6 +1450,8 @@ int main(int argc, char *argv[]) {
     close(log);
     log = STDOUT_FILENO;
   }
+
+  debug_log.fd = log;
 
   umask(0);
 
@@ -1492,13 +1507,6 @@ int main(int argc, char *argv[]) {
     goto term;
   }
 
-// Macro to write to stderr, and also to log if debug is enabled
-#define startup_fprintf(fmt, ...)                \
-  do {                                           \
-    fprintf(stderr, fmt, ##__VA_ARGS__);         \
-    if (debug) dprintf(log, fmt, ##__VA_ARGS__); \
-  } while (0)
-
   // Open the logfile and use as lock on it to ensure we retain ownership
   // This has to happen after fork (which would drop the lock)
   memset(&fl, 0, sizeof(fl));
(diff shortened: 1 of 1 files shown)


dtzSiFive (commented):
This is the first in-tree use of preprocessor macros for helpers?

While prevalent style is rather C-oriented, WDYT about some variant of:

```c++
// Print to stderr, and optionally to a debug log.
template <typename... Args>
void debug_log(bool debug, int fd, const char *fmt, Args... args) {
  fprintf(stderr, fmt, args...);
  if (debug && fd != -1) dprintf(fd, fmt, args...);
}
```

?

The following wraps it in a class, with state constructed as-we-go due to use of goto:

```patch
From 0ff4e2366f573281add5c0a6b4f5dfae06ae38f0 Mon Sep 17 00:00:00 2001
From: Will Dietz <will.dietz@sifive.com>
Date: Fri, 13 Feb 2026 10:21:07 -0600
Subject: [PATCH] DebugLogger helper.

---
 tools/fuse-waked/main.cpp | 36 +++++++++++++++++++++---------------
 1 file changed, 21 insertions(+), 15 deletions(-)

diff --git a/tools/fuse-waked/main.cpp b/tools/fuse-waked/main.cpp
index e758bb5d..e6c00b4f 100644
--- a/tools/fuse-waked/main.cpp
+++ b/tools/fuse-waked/main.cpp
@@ -1367,6 +1367,18 @@ static void handle_exit(int sig) {
   }
 }

+// Print to stderr, and optionally to a debug log.
+struct DebugLogger {
+  template <typename... Args>
+  void operator()(const char *fmt, Args... args) {
+    fprintf(stderr, fmt, args...);
+    if (debug && fd != -1) dprintf(fd, fmt, args...);
+  }
+
+  bool debug = false;
+  int fd = -1;
+};
+
 int main(int argc, char *argv[]) {
   bool debug = getenv("DEBUG_FUSE_WAKE");

@@ -1408,6 +1420,7 @@ int main(int argc, char *argv[]) {
   int log, null;
   bool madedir;
   struct rlimit rlim;
+  DebugLogger debug_log{debug, -1};

   if (argc != 3) {
     fprintf(stderr, "Syntax: fuse-waked <mount-point> <min-timeout-seconds>\n");
@@ -1437,6 +1450,7 @@ int main(int argc, char *argv[]) {
     close(log);
     log = STDOUT_FILENO;
   }
+  debug_log.fd = log;

   umask(0);

@@ -1492,13 +1506,6 @@ int main(int argc, char *argv[]) {
     goto term;
   }

-// Macro to write to stderr, and also to log if debug is enabled
-#define startup_fprintf(fmt, ...)                \
-  do {                                           \
-    fprintf(stderr, fmt, ##__VA_ARGS__);         \
-    if (debug) dprintf(log, fmt, ##__VA_ARGS__); \
-  } while (0)
-
   // Open the logfile and use as lock on it to ensure we retain ownership
   // This has to happen after fork (which would drop the lock)
   memset(&fl, 0, sizeof(fl));
@@ -1509,12 +1516,12 @@ int main(int argc, char *argv[]) {
   if (fcntl(log, F_SETLK, &fl) != 0) {
       if (debug) {
-        startup_fprintf("fcntl(%s.log): %s -- assuming another daemon exists\n", path.c_str(),
-                        strerror(errno));
+        debug_log("fcntl(%s.log): %s -- assuming another daemon exists\n", path.c_str(),
+                  strerror(errno));
       }
       status = 0;  // another daemon is already running
     } else {
-      startup_fprintf("fcntl(%s.log): %s\n", path.c_str(), strerror(errno));
+      debug_log("fcntl(%s.log): %s\n", path.c_str(), strerror(errno));
     }
     goto term;
   }
@@ -1555,31 +1562,30 @@ int main(int argc, char *argv[]) {
 #else
   if (fuse_opt_add_arg(&args, "wake") != 0) {
 #endif
-    startup_fprintf("fuse_opt_add_arg failed\n");
+    debug_log("fuse_opt_add_arg failed\n");
     goto rmroot;
   }

   if (debug && fuse_opt_add_arg(&args, "-odebug") != 0) {
-    startup_fprintf("fuse_opt_add_arg debug failed\n");
+    debug_log("fuse_opt_add_arg debug failed\n");
     goto rmroot;
   }

   fc = fuse_mount(path.c_str(), &args);
   if (!fc) {
-    startup_fprintf("fuse_mount failed\n");
+    debug_log("fuse_mount failed\n");
     goto freeargs;
   }

   fh = fuse_new(fc, &args, &wakefuse_ops, sizeof(wakefuse_ops), 0);
   if (!fh) {
-    startup_fprintf("fuse_new failed\n");
+    debug_log("fuse_new failed\n");
     goto unmount;
   }

   fflush(stdout);
   fflush(stderr);

-#undef startup_fprintf
   dup2(log, STDERR_FILENO);

   if (null != STDIN_FILENO) {
--
2.50.1
```

"debug_log" might be better named.

dtzSiFive (approved):
LGTM

PR #1749, opened 2026-02-13
Fuse_daemon: Always attempt to clear out stale mounts before mounting

ngraybeal:
Create a persistant daemon started by Wake runtime, instead of wakebox to reduce the up/downtime costs and possible flakiness of those actions.

### DIFF
--- a/tools/fuse-waked/main.cpp
+++ b/tools/fuse-waked/main.cpp
@@ -53,6 +53,16 @@
 // We ensure STDIN is /dev/null, so this is a safe sentinel value for open files
 #define BAD_FD STDIN_FILENO
 
+#ifndef __APPLE__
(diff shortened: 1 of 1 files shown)


tools/fuse-waked/main.cpp
```diff
@@ -1557,6 +1567,11 @@ int main(int argc, char *argv[]) {
     goto rmroot;
   }
 
+#ifndef __APPLE__
+  // Attempt to clean up any stale mount before trying to mount
+  try_unmount_stale(path.c_str(), debug);
+#endif
+
   fc = fuse_mount(path.c_str(), &args);
```
AbrarQuazi: Any race conditions that can happen when we unmount and then immediately try to mount? What if the mount is still cleaning up?

AbrarQuazi (approved):
LGTM

PR #1760, opened 2026-02-26
database: add lock files for dead run detection

dtzSiFive:
Use per-run lock files (.wake/locks/run_N.lock) to detect dead wake
processes.  Each wake process creates and holds an exclusive fcntl
lock on its lock file for the duration of the run.  On startup, probe
lock files of incomplete runs -- if we can acquire the lock, the
original process is dead and we mark it as reaped (end_time == -1).

Additionally:
If for whatever reason lock files are missing, to avoid blocking
GC watermark indefinitely assumes such runs are dead if they
started more than 24 hours ago.
This is mostly to cover a strange case that shouldn't happen,
not a part of anyone's normal flow.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -39,6 +39,7 @@
 
 #include "schema.h"
 #include "status.h"
+#include "util/mkdir_parents.h"
 #include "wcl/iterator.h"
 
 #define VISIBLE 0
@@ -96,9 +97,11 @@ struct Database::detail {
   sqlite3_stmt *insert_run_job;
   sqlite3_stmt *get_gc_watermark;
   sqlite3_stmt *set_run_end_time;
+  sqlite3_stmt *get_incomplete_runs;
 
   long run_id;
   long gc_watermark;
+  int run_lock_fd;
   detail(bool debugdb_)
       : debugdb(debugdb_),
         db(0),
@@ -144,8 +147,10 @@ struct Database::detail {
         insert_run_job(0),
         get_gc_watermark(0),
         set_run_end_time(0),
+        get_incomplete_runs(0),
         run_id(0),
-        gc_watermark(0) {}
+        gc_watermark(0),
+        run_lock_fd(-1) {}
 };
 
 class WaitingIndicator {
@@ -459,6 +464,7 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_insert_run_job = "insert or ignore into run_jobs(run_id, job_id) values(?, ?)";
   const char *sql_get_gc_watermark = "select min(run_id) - 1 from runs where end_time is null";
   const char *sql_set_run_end_time = "update runs set end_time = ? where run_id = ?";
+  const char *sql_get_incomplete_runs = "select run_id, time from runs where end_time is null";
 
 #define PREPARE(sql, member)                                                                     \
   ret = sqlite3_prepare_v2(imp->db, sql, -1, &imp->member, 0);                                   \
@@ -515,6 +521,7 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   PREPARE(sql_insert_run_job, insert_run_job);
   PREPARE(sql_get_gc_watermark, get_gc_watermark);
   PREPARE(sql_set_run_end_time, set_run_end_time);
+  PREPARE(sql_get_incomplete_runs, get_incomplete_runs);
 
   return "";
 }
@@ -580,6 +587,14 @@ void Database::close() {
   FINALIZE(insert_run_job);
   FINALIZE(get_gc_watermark);
   FINALIZE(set_run_end_time);
+  FINALIZE(get_incomplete_runs);
+
+  // Close run lock file if held
+  if (imp->run_lock_fd >= 0) {
+    ::close(imp->run_lock_fd);
+    imp->run_lock_fd = -1;
+  }
+
   close_db(this);
   release_build_lock();
 }
@@ -808,6 +823,31 @@ void Database::entropy(uint64_t *key, int words) {
   end_txn();
 }
 
+static std::string run_lock_path(long run_id) {
+  return ".wake/locks/run_" + std::to_string(run_id) + ".lock";
+}
+
+// Returns true if lock acquired, false otherwise (sets errno).
+static bool acquire_lock(int fd, bool wait) {
+  struct flock fl;
+  memset(&fl, 0, sizeof(fl));
+  fl.l_type = F_WRLCK;
+  fl.l_whence = SEEK_SET;
+  fl.l_start = 0;
+  fl.l_len = 0;  // 0 = entire file
+  return fcntl(fd, wait ? F_SETLKW : F_SETLK, &fl) == 0;
+}
+
+static void release_lock(int fd) {
+  struct flock fl;
+  memset(&fl, 0, sizeof(fl));
+  fl.l_type = F_UNLCK;
+  fl.l_whence = SEEK_SET;
+  fl.l_start = 0;
+  fl.l_len = 0;
+  fcntl(fd, F_SETLK, &fl);
+}
+
 void Database::prepare(const std::string &cmdline) {
   struct timespec now;
   clock_gettime(CLOCK_REALTIME, &now);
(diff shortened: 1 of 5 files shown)


src/runtime/database.cpp
```diff
@@ -609,6 +616,14 @@ void Database::close() {
   FINALIZE(insert_run_job);
   FINALIZE(get_gc_watermark);
   FINALIZE(set_run_end_time);
+  FINALIZE(get_incomplete_runs);
+
+  // Close run lock file if held
+  if (imp->run_lock_fd >= 0) {
+    ::close(imp->run_lock_fd);
```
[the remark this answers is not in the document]
ag-eitilt: Nit: Mind adding the "only removed after..." comment to the code to point back to the `unlink` in `finish_run`, just to make it clear we *do* remove it?

It also feels slightly odd to me that we're embedding run-level locking in the `Database` type handling, but doing so using sidecar files entirely independent from the database, but this is admittedly an object which will be initialized at the start of every run and last until the end, and already involves a lot of transactional patterns, so it's not entirely out of place, either.

src/runtime/database.cpp
```diff
@@ -830,6 +845,31 @@ void Database::entropy(uint64_t *key, int words) {
   end_txn();
 }
 
+static std::string run_lock_path(long run_id) {
+  return ".wake/locks/run_" + std::to_string(run_id) + ".lock";
+}
```
ag-eitilt: Nit: Is "lock" the best terminology for this?  It seems to be acting more like a pidfile than a proper lock (do we want to lean into that and write the PID to the file to allow debugging/integrating externally?), though I do love the flock mechanism for detecting liveliness rather than needing to look through the process table after getting the number, so it's not *exactly* a basic pidfile.

src/runtime/database.cpp
```diff
@@ -830,6 +845,31 @@ void Database::entropy(uint64_t *key, int words) {
...
+
+// Returns true if lock acquired, false otherwise (sets errno).
+static bool acquire_lock(int fd, bool wait) {
+  struct flock fl;
+  memset(&fl, 0, sizeof(fl));
+  fl.l_type = F_WRLCK;
+  fl.l_whence = SEEK_SET;
+  fl.l_start = 0;
+  fl.l_len = 0;  // 0 = entire file
+  return fcntl(fd, wait ? F_SETLKW : F_SETLK, &fl) == 0;
```
ag-eitilt: How cross-platform is this?  Not that that's a critical concern, but it seems like Mac requires `flock` to be installed manually, and I'm not sure if the libraries are any different.

src/runtime/database.cpp
```diff
@@ -842,6 +882,81 @@ void Database::prepare(const std::string &cmdline) {
   single_step(why, imp->add_run, imp->debugdb);
   imp->run_id = sqlite3_last_insert_rowid(imp->db);
 
+  mkdir_with_parents(".wake/locks", 0755);
+
+  // Acquire lock before committing, so we never appear incomplete without one.
```
ag-eitilt, line 839: I guess my questions of scope come down to how tightly we mean to relate the lock/pid files to the DB.  If we *do* just treat it as a pidfile (and include the PID in the DB) then it becomes the overlying determiner, and should be created outside the `Database` context; no run entry appears incomplete since the pidfile is written out before the DB is ever opened.  On the other hand, if we treat the DB as the ultimate authority and the lockfile just as a FS-backed method of exposing the entry, then as you have it here does make sense; the row needs to be created first to have the ID for the name, but we definitely can't have an entry without a file or the liveliness-check logic has a racy hole poked in it.

I do think the DB-primary method is tighter and more in line with the rest of Wake's architecture, it's just not the first thing that comes to my mind when seeing these files.

src/runtime/database.cpp
```diff
@@ -842,6 +882,81 @@ void Database::prepare(const std::string &cmdline) {
...
+  if (!acquire_lock(imp->run_lock_fd, true)) {
+    std::cerr << "error: failed to acquire run lock: " << strerror(errno) << std::endl;
+    unlink(our_lock_path.c_str());
+    exit(1);
+  }
+  end_txn();
+
+  // Query incomplete runs (separate txn; snapshot doesn't need to match above).
+  why = "Could not query incomplete runs";
+  std::vector<std::pair<long, int64_t>> incomplete_runs;  // (run_id, start_time)
```
ag-eitilt: Nit: Mind pulling this (all the way through line 957) into a separate function?  The row+lockfile initialization makes sense in `prepare`, as does the watermark calculation, but the dead runs are technically a separate enough concern (outside of their role in getting the watermark) that it feels a bit like function creep to have them inlined here -- not that calling out to a separate function to do the same work is any different in practice, but it just feels better for some reason.

ag-eitilt (approved):
Nicely defensive programming here!  Biggest questions are about where code lives and what it's called, and we can bikeshed on that whenever.

src/runtime/database.cpp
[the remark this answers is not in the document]
dtzSiFive: I strongly agree this is weird to have as part of "Database"!!

I'll note we previously did the build lock here, but I would like this to live elsewhere.

Thanks for the reminder.

src/runtime/database.cpp
dtzSiFive: fcntl is the more portable of the two! POSIX

And FWIW we use `fcntl` elsewhere in wake today (build lock, and fuse-waked's one-daemon-only lock).

src/runtime/database.cpp
dtzSiFive: I'm happy to write the PID into the file.

I would really like to not have PID's in the database!

dtzSiFive: Also probably not a problem for use, but this works the way it does to not make assumptions about pid reuse.
 
 Even with configuration to use huge pid space, if the machine is shutdown pid's of unrelated processes could appear.
 
 Using a live lock seems the simple answer to these problems 👍 .

src/runtime/database.cpp
ag-eitilt: Ah, wasn't aware they were different means of handling it; I saw `struct flock fl` and that we pass it into `fcntl`, and went from there.

src/runtime/database.cpp
ag-eitilt: Yep, PIDs in the database (despite the later thought experiment) really doesn't feel like a good idea.  It's also not immediately helpful in this file, but I'm just thinking about how other things *might* want to interact with Wake -- I can imagine ways CI might inspect these files for tracking/management, etc.

Live lock being the primary method is wonderful; live lock with a way to find the process only seems more powerful.

dtzSiFive: Love it.  FWIW `lsof` is your friend :P but only if it's still live haha.

src/runtime/database.cpp
dtzSiFive: Aye aye agreed thanks.

dtzSiFive: Done.

src/runtime/database.cpp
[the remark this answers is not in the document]
dtzSiFive: Refactored a bunch of this out to `run_lock.{h,cpp}`.  LMK what you think!

dtzSiFive:
(This will make sense once our feature/multi-wake is updated to match this: https://github.com/dtzSiFive/wake/tree/feature/multi-wake )

src/runtime/run_lock.cpp
```diff
@@ -0,0 +1,164 @@
...
+static auto make_errno(Args&... args) {
+  return make_error<R>(std::forward<Args>(args)..., ": ", strerror(errno));
+}
+
+// RunLock implementation
+
+RunLock::RunLock(wcl::unique_fd&& fd, std::string path)
+    : fd(std::move(fd)), path(std::move(path)) {}
+
+RunLock::LockAcquireRetTy RunLock::create_and_acquire(long run_id, bool wait) {
```
ag-eitilt, line 99: Curious: What's `RetTy` stand for?

ag-eitilt (approved):
Looks good to me!

src/runtime/run_lock.cpp
dtzSiFive: Return Type! 🙃

PR #1761, opened 2026-02-26
multi-wake: remove .wake-build-lock mechanism

dtzSiFive:
The file-based build lock prevented concurrent wake invocations.
This is no longer needed, remove it!

Remove entropy table exclusive lock in schema.

---------

Should land at least after #1760 .

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -210,13 +210,6 @@ Database::~Database() { close(); }
 std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   if (imp->db) return "";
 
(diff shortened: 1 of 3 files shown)


ag-eitilt (approved):
Pretty sure this should be able to work (obviously after #1760) even without any of the other multi-wake/CAS changes; sorry I didn't review earlier to have allowed it to be rolled into v48.

ngraybeal (approved):
LGTM

dtzSiFive:
I'm just waiting for #1760, which is what makes removing these locks safe 👍 .

PR #1762, opened 2026-02-26
database: add --clean safety check for active builds

dtzSiFive:
Refuse to clean while builds are in progress.

First reap dead runs (refactor), then with RW lock held:
* Check if incomplete jobs, exit
* Delete files

Add test.

Test relies on https://github.com/sifiveinc/wake/pull/1761 , changes depend on #1760.

Commits unique to this are:
https://github.com/sifiveinc/wake/pull/1762/commits/ce75768bfc465bd584aa5298dd1de16e2a15e25a
https://github.com/sifiveinc/wake/pull/1762/commits/ff6153f6ab5987a69d22689e7c12f4d4ebc1aba2
https://github.com/sifiveinc/wake/pull/1762/commits/f086b718181fe71afddaf500376bbcd51b95b063

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -1214,6 +1214,41 @@ std::vector<std::string> Database::clear_jobs() {
   return out;
 }
 
+bool Database::clear_jobs_if_safe(std::function<void(std::vector<std::string>)> delete_files) {
+  const char *why = "Could not clear jobs";
+  std::vector<std::string> out;
+
+  // Ensure no new runs while we're cleaning, and that we have latest + consistent view.
+  begin_rw_txn();
+
+  if (sqlite3_step(imp->get_incomplete_runs) == SQLITE_ROW) {
+    finish_stmt(why, imp->get_incomplete_runs, imp->debugdb);
+    end_txn();
+    return false;
+  }
+  finish_stmt(why, imp->get_incomplete_runs, imp->debugdb);
+  while (sqlite3_step(imp->get_output_files) == SQLITE_ROW) {
+    out.emplace_back(rip_column(imp->get_output_files, 0));
+  }
+  finish_stmt(why, imp->get_output_files, imp->debugdb);
+
+  while (sqlite3_step(imp->get_unhashed_file_paths) == SQLITE_ROW) {
+    out.emplace_back(rip_column(imp->get_unhashed_file_paths, 0));
+  }
+  finish_stmt(why, imp->get_unhashed_file_paths, imp->debugdb);
+
+  single_step(why, imp->remove_all_jobs, imp->debugdb);
+  single_step(why, imp->remove_output_files, imp->debugdb);
+
+  // Delete files while still holding write lock. This prevents new
+  // builds from starting until cleanup is complete.
+  delete_files(std::move(out));
+
+  end_txn();
+
+  return true;
+}
+
 void Database::tag_job(long job, const std::string &uri, const std::string &content) {
   const char *why = "Could not tag a job";
   begin_rw_txn();
--- a/tests/runtime/clean-safety/pass.sh
+++ b/tests/runtime/clean-safety/pass.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+# --clean should refuse while a build is active, succeed after reaping dead runs
+
+set -eu
+
+WAKE="${1:+$1/wake}"
+WAKE="${WAKE:-wake}"
+
+cleanup() {
+  if [ -n "${WAKE_PID:-}" ]; then
+    kill $WAKE_PID 2>/dev/null || true
+    wait $WAKE_PID 2>/dev/null || true
+  fi
+  rm -rf wake.db* wake.log .wake .started
+}
+trap cleanup EXIT
+cleanup
+
+# longRunningBuild: first job creates .started, second sleeps forever
+"${WAKE}" -x 'longRunningBuild Unit' &
+WAKE_PID=$!
+
+# Wait for first job to complete (wake has recorded it)
+while [ ! -f .started ]; do sleep 1; done
+
+# --clean should refuse (second job still running)
+if "${WAKE}" --clean 2>&1 | grep -q "in progress"; then
+  echo "PASS: --clean refused while build active"
+else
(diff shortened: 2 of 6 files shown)


dtzSiFive:
I let augment help fix the test, I was already a little unsure about how it was written and now it's gotten a bit.. unwieldy.

Open to suggestions.  I'll see what I come up with.

src/runtime/database.cpp
```diff
@@ -829,8 +833,43 @@ void Database::prepare(const std::string &cmdline) {
   }
   end_txn();
 
-  // Query incomplete runs (separate txn; snapshot doesn't need to match above).
-  why = "Could not query incomplete runs";
+  // Reap any dead runs (our run_id is automatically excluded).
+  reap_dead_runs();
```
ag-eitilt, line 771: Ah, you've already implemented https://github.com/sifiveinc/wake/pull/1760#discussion_r3115132651. 😁

src/runtime/database.cpp
```diff
@@ -1274,6 +1285,42 @@ std::vector<std::string> Database::clear_jobs() {
   return out;
 }
 
+bool Database::clear_jobs_if_safe(std::function<void(std::vector<std::string>)> delete_files) {
+  const char *why = "Could not clear jobs";
+  std::vector<std::string> out;
+
+  // BEGIN IMMEDIATE acquires write lock, blocking other writers.
+  // This ensures no new runs can be inserted between our check and delete.
```
ag-eitilt: Mind adding a bit more on *why* we're preventing that?, i.e. that without the write lock (also prevents reads, right?) we run the risk of pulling these jobs as cache hits and so deleting them would corrupt the run that revived them?  As it is, we have to read between the lines of the `delete_files` comment a bit to realize why clearing jobs from dead/outdated runs would need this level of protection ("they're dead, it's not like that run is going to be resumed").

tests/runtime/clean-safety/pass.sh
```diff
@@ -0,0 +1,45 @@
...
+# --clean should refuse (second job still running)
+if "${WAKE}" --clean 2>&1 | grep -q "in progress"; then
+  echo "PASS: --clean refused while build active"
+else
+  echo "FAIL: --clean should have refused"
+  exit 1
+fi
+
+# Kill wake (simulates crash)
+kill -9 $WAKE_PID 2>/dev/null || true
```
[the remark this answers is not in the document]
ag-eitilt: Good incidental test for that behaviour, at least...

tools/wake/main.cpp
```diff
@@ -582,60 +582,69 @@ int main(int argc, char **argv) {
...
+    // Track if file deletion had errors.
+    bool delete_error = false;
+
+    // Atomically check for active builds and clear if safe.
+    // File deletion happens inside the transaction to hold the write lock,
+    // preventing new builds from starting until cleanup completes.
+    bool cleaned = db.clear_jobs_if_safe([&](std::vector<std::string> paths) {
+      // Sort so child directories come before parent directories
+      std::sort(paths.begin(), paths.end(),
+                [](const std::string &a, const std::string &b) { return a.size() > b.size(); });
```
ag-eitilt, line 646: Don't know if it's just me being unfamiliar with C++, but the inline function and the re-wrapping of the `sort` make this a fair bit harder for me to read.  Don't feel like you have any need to act on that, I just wanted to give my impressions.

tools/wake/main.cpp
```diff
@@ -582,60 +582,69 @@ int main(int argc, char **argv) {
...
       }
-    }
 
-    // Since the log is append only, we should clean it up from time to time.
-    // TODO: this is just "unlink_no_fail". Those functions should be moved to
-    // a more generic library
-    if (unlink("wake.log") < 0 && errno != ENOENT) {
-      wcl::log::error("unlink(wake.log): %s", strerror(errno)).urgent()();
+      // Clean up lock directory (only succeeds if empty, which is safe)
+      (void)rmdir(".wake/locks");
```
ag-eitilt, line 693: Ah, `(void)`! 😉  Looks like CI is fine with this one.

ag-eitilt (approved):
Only thing I'd ask for on this one before merging is expanding the comment I pointed out, everything else looks good to me.

tools/wake/main.cpp
dtzSiFive: I'll take a look, thanks! Mostly was just moving the existing code into the lambda.

src/runtime/database.cpp
dtzSiFive: Write does not prevent reads -- so you could inspect the database /during/ a `wake --clean` as implemented here.  Note that for `wake-migrate` safety I went to greater lengths to ensure exclusivity.

Write lock does two things:

Ensures we have the latest "version" of the data /throughout/ the deletion process.  With just a read lock new data could be added (I think technically "already present" but mostly I think that way to stay defensive) during the transaction.  This ensures, for example, the returned path list (`get_output_files`) includes all paths we remove from the database later in `remove_output_files`).

Second (well maybe it's part of the same 😉) it ensures we don't do any of this while a run is concurrently running! Lots of DB badness could occur otherwise, and perhaps most problematically: we might delete files out from under the run!

Finally, my new "rule" for DB is that literally everything that /might/ write uses `begin_rw_txn`, this avoids errors and badness that occur if two wake processes need to promote their read locks to write locks, which is considered a deadlock and by ruling out we do not need to handle rollback or trickier things.

Finally-finally, it's important that /if/ we write we do so as part of the same transaction that checks for incomplete runs.  That's what the comment is mostly addressing.  I'm not sure the comment is carrying its weight!

You raise an interesting point I hadn't really explicitly considered: what about a `--clean` that works concurrently with runs? I am not sure that could be made sufficiently safe and am inclined to prefer to disallow it 👍 .

For future "wake gc"-ish operations I am planning to allow working that way (holding RW lock but not rejecting concurrent runs being in-flight!).

---------

Thank you for the careful considerations!!

dtzSiFive: Tweaked comment! Thanks!

dtzSiFive:
Disabling test until issue is resolved 😞 👍 .

PR #1763, opened 2026-02-26
--history: show run status, end time, and duration 

dtzSiFive:
Display whether each run is running, crashed, or duration.
Adds end_time to RunReflection.
Migrated runs show 0s since we lack their actual end time.


dtzSiFive:
Test comes later, as it requires more pieces.

AbrarQuazi (approved):

dtzSiFive:
Thanks! We can iterate on the output formatting, let's get this in for now.

PR #1765, opened 2026-03-04
Add tests showing three ways wake reuses jobs that would have different results

dtzSiFive:
Three categories of reuse behavior where wake uses a previous cached job when running the job as described produces a different result.

* symlink: Job that has as input a symlink `a` that points to file `target`, will be reused when instead of a symlink `a` has the content `target`.
* negative input: Job reuse (use A instead of running B) is used when a job A's inputs are a subset of the new job B's visible set.  If A observes files in that visible set are not present, reusing A for B produces result different than running B would.
* permissions: Job reuse occurs despite input files having different properties such as permissions.

These are considered unsupported features, hence the tests passing.


dtzSiFive:
Moved tests to using `stdout` reference file testing 👍 .  They're "pass.sh" now.

dtzSiFive:
Moved all tests to a new `pending` directory to indicate these tests are expected to pass in the future!

AbrarQuazi (approved):

ngraybeal (approved):
LGTM

PR #1768, opened 2026-03-11
Add builtin type Time

ngraybeal:
Creates a builtin type `Time` for the main purpose of handling timestamps. 

Time is stored as a unit of nanoseconds. 

This adds functional support for adding, subtracting, comparing, parsing, and formatting time. Default functions for getting the date appear in UTC, but there is support for requesting date in specific TimeZones.

This uses the `chrono` library to help safely handle time. Since we are using c++17, I'm still using `strftime` to format the string.

I did rename the database `Time` struct to `DBTime` to avoid conflicts.

### DIFF
--- a/share/wake/lib/system/time.wake
+++ b/share/wake/lib/system/time.wake
@@ -1,0 +1,200 @@
+# Copyright 2026 SiFive, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You should have received a copy of LICENSE.Apache2 along with
+# this software. If not, you may obtain a copy at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package wake
+
+# Time Library
+#
+# The Time type is a built-in type representing a point in time with nanosecond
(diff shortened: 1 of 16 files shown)


share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,200 @@
...
+# Arguments:
+#   format   - strftime format string
+#   timezone - IANA timezone name (e.g., "America/Los_Angeles", "UTC", "PST8PDT")
+#   time     - Time value to format
+#
+# Example:
+#   def now = getTime Unit
+#   formatTimeTZ "%Y-%m-%d %H:%M:%S" "UTC" now                    # "2026-03-10 08:15:07"
+#   formatTimeTZ "%Y-%m-%d %H:%M:%S" "America/Los_Angeles" now    # "2026-03-10 00:15:07"
+#   formatTimeTZ "%Y-%m-%d %H:%M:%S %Z" "America/New_York" now    # "2026-03-10 03:15:07 EST"
```
ag-eitilt: LA is not UTC.

share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,200 @@
...
+
+# getTime: Get the current system time.
+#
+# Returns a Time value representing the current time with nanosecond precision.
+# This function observes the system clock and is marked as ORDERED to ensure
+# proper sequencing in the build system.
+#
+# Example:
+#   def now = getTime Unit
+export def getTime (_: a): Time =
```
ag-eitilt:
```suggestion
export def unsafe_getTime (_: a): Time =
```
Or, following the example of the previous comment:
```suggestion
export def unsafe_getCurrentTime (_: a): Time =
```

ag-eitilt (commented):
Only skimmed the C++, I'll try to look at that in more detail if I make a second pass through.

PR #1782, opened 2026-04-10
makeBinaryRequest: mkdir http dir for curl request

ngraybeal:
https://github.com/sifiveinc/wake/pull/1777 introduced a bug that caused RSC rehydration to fail.

This PR dropped a `mkdir` that was very much necessary otherwise the curl command will fail as the target directory doesn't exist. This was silently failing.

I also believe (but didn't test) that the original [change](https://github.com/sifiveinc/wake/pull/1729) to move to fuse runner would have introduced the same bug as the directory would be created, but wouldn't be in the wakebox command.

I have updated the code to create the dir and ties it to the job, even though it's a localRunner.

### DIFF
--- a/share/wake/lib/system/http.wake
+++ b/share/wake/lib/system/http.wake
@@ -334,7 +334,10 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
     # always rerun and we'd want to replace the file incase it changed on the remote.
     # Per-process directory avoids conflicts when multiple wake processes share a workspace.
     def wakePid: Integer = prim "pid"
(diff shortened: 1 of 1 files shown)


ag-eitilt (approved):
Weird, not sure how this *ever* worked (not just in the context of my PR) given that it's apparently why things were failing.  Maybe the directory was getting implicitly created by some earlier Job, which is no longer getting called with one of the various changes in v48?  Either way, going to be really nice to get this `curl`ing into the upcoming sandbox model to avoid (A) needing to worry about overwriting something that might already exist and (B) the hardlink in the first place.  Though, the edge case described in the comment is still a potential failure case we'll have to be aware of for our materialization logic.

PR #1787, opened 2026-04-13
Workspace virtualization fuse changes updated

AbrarQuazi:
This PR extends the fuse-waked sandbox daemon to support CAS-first (content-addressed storage) job execution when the WAKE_CAS environment variable is set.

**Staging during job execution**
Files written by a job are now redirected to `.cas/staging/` instead of the workspace directly. The daemon tracks each output as a typed StagedItem (file, symlink, directory, or hardlink) with its metadata (mode, timestamps). After the job exits, the daemon serializes a staging_files manifest for wakebox to hash and ingest into CAS.

**CAS-backed visible file reads**
The visible file protocol is upgraded from bare path strings to structured {path, type, hash, mode} objects. In CAS mode, files are served from .cas/blobs/<hash> rather than the workspace, so inputs are always read from the content store. The legacy string format is preserved for backward compatibility.

**Runner plumbing**
makeJSONRunner now serializes visible files as JSON objects and passes cas-blobs-dir (.cas/blobs) in the job spec. WAKE_CAS is propagated through to the daemon. CAS and non-CAS daemons use distinct mount paths (.fuse/<uid>.<gid>.cas vs .fuse/<uid>.<gid>) so they can coexist.


src/wakefs/daemon_client.cpp
```diff
@@ -126,8 +135,19 @@ bool daemon_client::connect(std::vector<std::string> &visible, bool close_live_f
...
+
+  // Add visible files with path, type, hash, and mode.
   auto &vis = for_daemon.add("visible", JSON_ARRAY);
-  for (auto &s : visible) vis.add(std::move(s));
+  for (auto &v : visible) {
+    auto &obj = vis.add(JSON_OBJECT);
+    obj.add("path", v.path);
+    obj.add("type", v.type);
+    obj.add("hash", v.hash);
+    if (v.mode) obj.add("mode", static_cast<long long>(*v.mode));
```
dtzSiFive: Is "mode" optional on the way in? `mkVisibleJSON` suggests it's not.

If you think it should be optional re:format specification at this layer, go for it.

src/wakefs/daemon_client.cpp
```diff
@@ -82,7 +83,15 @@ bool daemon_client::connect(std::vector<std::string> &visible, bool close_live_f
       std::string delayStr = std::to_string(exit_delay);
       const char *env[3] = {"PATH=/usr/bin:/bin:/usr/sbin:/sbin", 0, 0};
       if (getenv("DEBUG_FUSE_WAKE")) env[1] = "DEBUG_FUSE_WAKE=1";
-      execle(executable.c_str(), "fuse-waked", mount_path.c_str(), delayStr.c_str(), nullptr, env);
+      // Pass --use-cas to enable CAS-first staging when WAKE_CAS env var is set
+      // TODO: Remove the non-CAS daemon launch path once WAKE_CAS is the default.
+      if (getenv("WAKE_CAS")) {
```
dtzSiFive, line 91: If wake is launched inconsistently with WAKE_CAS set as well as not set, might they find themselves using a `fuse-waked` launched with wrong configuration?

Is there a tidy way to detect/avoid this?

Probably we want to error in this situation (unless it's sane to have both running in different directories or so?), WDYT?

src/wakefs/fuse.cpp
```diff
@@ -59,7 +60,36 @@ bool json_as_struct(const std::string &json, json_args &result) {
...
+      vf.path = x.second.get("path").value;
+      vf.type = x.second.get("type").value;
+      vf.hash = x.second.get("hash").value;
+      try {
+        const std::string &mode = x.second.get("mode").value;
+        if (!mode.empty()) vf.mode = std::stoi(mode);
+      } catch (const std::exception &) {
+      }
+      if (vf.type.empty()) {
+        std::cerr << "Visible entry object is missing required 'type' field\n";
```
dtzSiFive: Is there a reason to be permissive in this way; it must have path and hash too (and possibly mode), right?

while there's the "be liberal in what you accept" mindset, I'm not sure it's sane to have a visible file that doesn't have a path, right?

src/wcl/unique_fd.h
```diff
@@ -59,13 +59,13 @@ class unique_fd {
   }
 
   void close() {
-    if (fd > 0) {
+    if (fd >= 0) {
```
dtzSiFive, line 62: It's good to get into habit to break out these sorts of changes to separate PR's.

For a few reasons? Anyway not a blocker, just a comment for future consideration.

tests/wakebox/cas-directory/pass.sh
```diff
@@ -0,0 +1,34 @@
...
+
+export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
+export WAKE_CAS=1
+
+trap 'rm -rf output_dir' EXIT
+
+OUTPUT=$(${1}/wakebox -p input.json 2>&1)
+EXIT_CODE=$?
+
+if [ "$EXIT_CODE" != "0" ]; then
```
dtzSiFive: As @ag-eitilt pointed out on a PR of mine previously, no need to reinvent the wheel here -- we already can check exit codes/etc (if not use `set -eu` in opening?).

Similarly, where applicable consider reference stdout/stderr files instead?
WDYT?

This could save quite a bit of boilerplate, but I understand is unwanted churn immediately so not a blocker but something for future consideration.

The use of `-l` might involve details we'd like not to include, and sanitizing the output can be a pain.

tests/wakebox/cas-directory/pass.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+# Test that directory creation with files works in CAS mode.
+# Verifies that:
+# 1. Directories can be created
+# 2. Files can be created inside directories
+# 3. Directory listing works correctly
```
dtzSiFive, line 6: The next test confirms the output has the expected metadata-- here, perhaps we should check the directory is in the staging files or so? WDYT?

tests/wakebox/cas-file-write/pass.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+# Test that file creation and writing works in CAS mode.
+# Verifies that:
+# 1. Files can be created and written to
+# 2. File content can be read back immediately
+# 3. staging_files output contains the file with proper metadata
```
dtzSiFive: I don't see this happening?

tests/wakebox/cas-overwrite/pass.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+# Test that file overwriting works correctly in CAS mode.
+# When a file is overwritten, the old staging file should be deleted
+# and a new one created.
+# Verifies that:
+# 1. File can be overwritten multiple times
+# 2. Each read returns the current content
+# 3. Final content is correct
```
dtzSiFive: I don't see this being checked, that seems useful 👍 .

tests/wakebox/cas-symlink/pass.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+# Test that symlink creation works in CAS mode.
+# Verifies that:
+# 1. Symlinks can be created pointing to staged files
+# 2. Reading through symlink returns correct content
+# 3. readlink returns the correct target
```
dtzSiFive, line 6: So many things would be good to test:
* symlink that points to invalid
* writing through symlinks
* symlink to input
* symlink loop

Tests looks good for what it's doing.
In the future I think it's warranted to put together a bit of test infra to greatly lower friction of what's needed for each scenario and see what it's doing (single string in a boilerplate json is a bit rough, probably not fun to have written either?).

tools/wakebox/wakebox.wake
```diff
@@ -18,4 +18,4 @@ package build_wake
 from wake import _
 
 target buildWakeBox variant =
-    tool @here Nil variant "bin/wakebox" (json, gopt, wakefs, util, Nil) Nil Nil
+    tool @here Nil variant "bin/wakebox" (json, gopt, wakefs, util, casLib, Nil) Nil Nil
```
dtzSiFive: **Does** wakebox presently need CAS lib/objects? Removing the line here and similar from Makefile changes builds cleanly for my branch and I don't see this PR introducing such changes into wakebox?

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+  // Type name for JSON output
+  const char *type_name() const {
+    return std::visit(
+        overloaded{
+            [](const StagedFileData &) { return "file"; },
+            [](const StagedSymlinkData &) { return "symlink"; },
+            [](const StagedDirectoryData &) { return "directory"; },
+            // Serialize hardlinks as ordinary file entries. The shared staging_path still preserves
+            // the deduplication behavior Wake cares about during hashing and CAS ingestion.
+            [](const StagedHardlinkData &) { return "file"; },
```
dtzSiFive, line 129: It would be great to have this under test, if not actively documented somewhere.
Not immediately blocking.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+        },
+        data);
+  }
+
+  // Get staging_path (only valid for file and hardlink)
+  const std::string &staging_path() const {
+    if (auto *f = std::get_if<StagedFileData>(&data)) return f->staging_path;
+    if (auto *h = std::get_if<StagedHardlinkData>(&data)) return h->staging_path;
+    static const std::string empty;
```
dtzSiFive: Avoid function-local statics!
This results in a guard variable to ensure it's initialized in a thread-safe manner 😦 :

```
$ nm lib/wake/fuse-waked|grep empty|c++filt|grep guard
0000000000267b90 V guard variable for StagedItem::staging_path[abi:cxx11]() const::empty[abi:cxx11]
```

This can be optimized away, but I think std::string in our c++ version can't be constexpr.

For later.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+    }
+  }
+};
+
+// Encapsulates the nested map of staged files with helper methods
+// Structure: job_id -> (path -> StagedItem)
+// Also tracks fd -> StagedItem* for open file handles
+class StagedFilesStore {
+ public:
+  using JobMap = std::map<std::string, StagedItem>;
```
dtzSiFive: Unless we require (and generally it's a great property to have, so don't let me convince you to drop this unduly) ordering, and perhaps even then, a std::map over std::string is really painful:

std::map is a red-black tree or similar, commonly.
Lookups take O(log N).

String comparisons are expensive, and scale poorly with longer paths and especially with strings that have many common characters in them (e.g., file trees).

I'd love to go through all our string handling, but for this specifically a simple move to `std::unordered_map<std::string, T>` would be a big improvement.

Empirically I was profiling fuse-waked this morning and a fair number of samples were actually in inserting/accessing into maps with `std::string` keys.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+  using StoreMap = std::map<std::string, JobMap>;
+
+  // Find a staged item, returns nullptr if not found
+  StagedItem *find(const std::string &job_id, const std::string &path) {
```
dtzSiFive: Future: Consider std::string_view now that we're on c++17.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+    auto job_it = files_.find(job_id);
+    if (job_it == files_.end()) return nullptr;
+    auto it = job_it->second.find(path);
+    if (it == job_it->second.end()) return nullptr;
+    return &it->second;
+  }
+
+  // Get or create a staged item (for insertion)
+  StagedItem &operator()(const std::string &job_id, const std::string &path) {
```
dtzSiFive: I said this before but separating out whether we expect insertion or not.. consider it as part of defensively ensuring we don't accidentally critical state.

Future work.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+    if (job_it == files_.end()) return nullptr;
+    return &job_it->second;
+  }
+
+  // Iterate over all jobs
+  StoreMap::iterator begin() { return files_.begin(); }
+  StoreMap::iterator end() { return files_.end(); }
+
+  // FD tracking for open staged files
+  void register_fd(int fd, StagedItem *item) { fd_map_[fd] = item; }
```
dtzSiFive: Throughout could use defensive programming.  I'll avoid mentioning this again.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+  struct timespec mtime;
+  struct timespec atime;
+};
+
+using StagedItemData =
+    std::variant<StagedFileData, StagedSymlinkData, StagedDirectoryData, StagedHardlinkData>;
+
+struct StagedItem {
+  std::string dest_path;
+  std::string job_id;
```
dtzSiFive, line 111: Is there reason to treat/expect this as a string?

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+
+struct StagedDirectoryData {
+  mode_t mode;
+  struct timespec mtime;
+};
+
+// Hardlink: references source file, shares staging_path for dedup
+struct StagedHardlinkData {
+  std::string staging_path;
+  std::string source_path;
```
dtzSiFive: I don't see this updated anywhere; likely situations that rename:
* the source file itself
* the directory containing the source file

Do not work? WDYT?

tools/fuse-waked/main.cpp
```diff
@@ -92,14 +311,99 @@ void Job::parse() {
...
+    std::optional<mode_t> mode;
+
+    if (x.second.kind == JSON_OBJECT) {
+      // New format: {"path": "...", "type": "...", "hash": "...", "mode": ...}
+      path = x.second.get("path").value;
+      type = x.second.get("type").value;
+      hash = x.second.get("hash").value;
+      try {
+        const std::string &mode_value = x.second.get("mode").value;
+        if (!mode_value.empty()) mode = static_cast<mode_t>(std::stoul(mode_value, nullptr, 10));
```
dtzSiFive: I'm not sure the value of being tolerant of invalid input here?

The concern being bigger-picture this masks failures.

Not blocking.

tools/fuse-waked/main.cpp
```diff
@@ -92,14 +311,99 @@ void Job::parse() {
...
+      } catch (const std::exception &) {
+      }
+      if (type.empty()) {
+        fprintf(stderr, "Visible entry object is missing required 'type' field\n");
```
dtzSiFive: I thought I left comment on this already but can't find it.  Apologies if so:

Why only checking type and not the other fields?

tools/fuse-waked/main.cpp
```diff
@@ -92,14 +311,99 @@ void Job::parse() {
...
+        continue;
+      }
+    } else {
+      // Legacy format: just a string path
+      path = x.second.value;
+    }
+
+    if (!path.empty() && path[0] != '/') {
```
dtzSiFive, line 402: Empty path should be a hard error, right?

tools/fuse-waked/main.cpp
```diff
@@ -92,14 +311,99 @@ void Job::parse() {
...
+          files_visible.insert(parent);
+          visible_entries[parent] = VisibleEntry{"directory", "", std::nullopt};
+        }
+      }
+    }
+  }
 }
 
-void Job::dump() {
+static std::string cas_blob_path(const cas::ContentHash &hash) {
```
dtzSiFive, line 425: (This has bad ergonomics in various error conditions, and is in many places blindly used to access the filesystem....  )

For something like a filesystem, consider adopting a defensive programming mindset.  Assertions can help catch during testing (if they're elided in production? overrated IMO) "should not happen" cases which can easily be masked or very hard to debug elsewhere.

Think about what "should" be true, and construct things with those invariants in mind.

No change requested here.

tools/fuse-waked/main.cpp
```diff
@@ -92,14 +311,99 @@ void Job::parse() {
...
+  auto result = cas::ContentHash::from_hex(hash);
+  if (!result) return false;
+  *content_hash = *result;
+  return true;
+}
+
+// Extract file permission bits (rwxrwxrwx) from a VisibleEntry.
+// If the entry has a mode set, use it; otherwise use the fallback mode.
+// Masks with 07777 to ensure only permission bits are returned (strips file type bits).
+static mode_t visible_mode_or(const Job::VisibleEntry &entry, mode_t fallback) {
```
dtzSiFive, line 433: Pedantically you're stripping these to "file mode bits"[1]:

       POSIX refers to the stat.st_mode bits corresponding to the mask
       S_IFMT (see below) as the file type, the 12 bits corresponding to
       the mask 07777 as the file mode bits and the least significant 9
       bits (0777) as the file permission bits.

The comment makes it sound like perhaps we /mean/ to strip down to permissions bits, but then mask needs to be `0777` 👍 .

tools/fuse-waked/main.cpp
```diff
@@ -92,14 +311,99 @@ void Job::parse() {
...
+  return static_cast<mode_t>(entry.mode.value_or(fallback) & 07777);
+}
+
+// Read the entire contents of a CAS blob into a string.
+// This is primarily used for reading symlink targets stored in CAS.
```
dtzSiFive:
```suggestion
// This is only used for reading symlink targets stored in CAS.
```

tools/fuse-waked/main.cpp
dtzSiFive: If we want to support special bits, let's get tests going for them.  Can happen later.

tools/fuse-waked/main.cpp
```diff
@@ -57,15 +62,229 @@
...
+                      data);
+  }
+
+  // Set mode (for chmod support)
+  void set_mode(mode_t m) {
+    std::visit(overloaded{
+                   [m](StagedFileData &f) { f.mode = m; },
+                   [](StagedSymlinkData &) {},  // Symlinks don't have mode
+                   [m](StagedDirectoryData &d) { d.mode = m; },
+                   [m](StagedHardlinkData &h) { h.mode = m; },
```
dtzSiFive, line 161: Hardlinks should share a mode, I think?

If you want to punt on robust hardlink support we can do so but let's document that to either visit later or as known limitation 👍 .

tools/fuse-waked/main.cpp
```diff
@@ -335,6 +783,66 @@ static int wakefuse_access(const char *path, int mask) {
...
+      const std::string &type = visible_it->second.type;
+      if (type == "directory") {
+        if (mask & W_OK) return -EACCES;
+        return 0;
+      }
+      if (type == "symlink") {
+        return 0;
+      }
+      cas::ContentHash hash;
+      if (parse_visible_hash(type, visible_it->second.hash, &hash)) {
```
dtzSiFive: Can we just parse these on startup? Isn't that the form we want?

Not a dealbreaker.

tools/fuse-waked/main.cpp
```diff
@@ -347,6 +855,7 @@ static int wakefuse_access_trace(const char *path, int mask) {
   return out;
 }
 
+// Reads symlink target. For staged symlinks, returns the stored target string.
 static int wakefuse_readlink(const char *path, char *buf, size_t size) {
   if (is_special(path)) return -EINVAL;
```
dtzSiFive, line 897: if (!size) return -EINVAL;

?

From readlink(2):

> EINVAL bufsiz is not positive.

The use of EINVAL elsewhere seems incorrect but smaller detail.  I bring this up because without we writethrough to the caller's buffer which presumably is zero bytes, buffer overflow.

tools/fuse-waked/main.cpp
```diff
@@ -360,6 +869,38 @@ static int wakefuse_readlink(const char *path, char *buf, size_t size) {
...
+  if (g_use_cas) {
+    auto visible_it = it->second.visible_entries.find(key.second);
+    if (visible_it != it->second.visible_entries.end()) {
+      cas::ContentHash hash;
+      const std::string &type = visible_it->second.type;
+      if (parse_visible_hash(type, visible_it->second.hash, &hash)) {
+        if (type != "symlink") return -EINVAL;
+
+        std::string target;
+        if (read_cas_blob_bytes(hash, &target)) {
```
dtzSiFive: Future: cache this?

tools/fuse-waked/main.cpp
dtzSiFive: Update: The lower_bound tricks rely on this for JobMap, right?

Could be done for StoreMap, but also that likely is much smaller of a map in practice.

Let's let this go, even though shows up on profiling we have many other bigger bottlenecks.

dtzSiFive (commented):
Looking good!!

There are many small details that I think are a bit uneven.

Generally this looks great, and I think we can move ahead with it, once we bounce through a few things.

Testing results are encouraging which helps.

* WAKE_CAS - see comment about connecting to a fuse-waked with wrong cas-enabled state from two wake instances?
* hardlinks: source_path is never updated, I think?
* tests: The testing PR may help here but there's a lot that might be specific to our FUSE
* permissions: lots to be careful about here.

There's a small issue in readlink that could cause a buffer overflow but it's an unlikely path in practice, however it is nearly trivial to guard against.

LMK if you have questions, feel free to resolve comments especially those tagged for future consideration.

I ran out of steam, and would love for this to be /very/ robust and compliant to serve as strong foundation that we don't have to come back to.

I think generally this is solid, great work!

tools/fuse-waked/main.cpp
ag-eitilt: Yeah, them sharing a mode is the cause of some twisty logic in either the Wake-side RSC or curl library, can't remember which.  This'll be fine for any hardlinks which were entirely created in the course of the job, but unless we've recently added logic to prevent hardlinking an input file elsewhere this could wind up messing with the mode in the CAS.

dtzSiFive: I think mode tracking is entirely virtual, and also we don't produce on-disk hardlinks, they're all backed by the same staging file.

So the detail in question is being sure tracked hardlinks agree on their mode, where here they appear to be tracked per hardlink.

See also keeping `hardlinks` set up-to-date.

tools/fuse-waked/main.cpp
```diff
@@ -534,13 +1143,42 @@ static int wakefuse_create(const char *path, mode_t mode, struct fuse_file_info
...
   if (it->second.is_visible(key.second)) return -EEXIST;
 
-  if (!it->second.is_writeable(key.second)) (void)deep_unlink(context.rootfd, key.second.c_str());
+  if (g_use_cas) {
+    // CAS mode: write to staging directory (wakebox will hash and store in CAS)
+    // Check if this path was already staged by this job - if so, delete the old staging file
+    if (StagedItem *existing = g_staged_files.find(key.first, key.second)) {
+      const std::string &existing_staging_path = existing->staging_path();
+      if (!existing_staging_path.empty()) {
+        unlink(existing_staging_path.c_str());
```
dtzSiFive: Only unlink staging_path if it's the last use re:hardlinks, or so?

Consider a "wakebox" test with shell script line like so:

"touch link_file_src.txt\nln link_file_src.txt link_file_dst.txt\nrm link_file_src.txt"

We need the staging file for `link_file_dst.txt` around?

This works as a shell script run directly outside of wakebox:

```
sh -c "touch link_file_src.txt; ln link_file_src.txt link_file_dst.txt; rm link_file_src.txt; echo \"pass\" > link_file_dst.txt; cat link_file_dst.txt"
```

(one-liner just because that's how we do our wakebox tests presently?)

Using v48.0.2's wakebox, this works.

tools/fuse-waked/main.cpp
AbrarQuazi: Fuse gives paths like `/{job_id}/file` that we currently extract the string from, so its kept as a string and not converted

tools/fuse-waked/main.cpp
AbrarQuazi: making this an optional

tools/fuse-waked/main.cpp
AbrarQuazi: I'll go ahead and document this as a limitation for now, to come and revisit later for robust hardlink support.

Currently the way hardlinks are implemented, we reflink/copy the source's staging file to a new staging path. This means that querying information about a hardlink/modifying its metadata (via chmod or utimens for timestamps) only reflect each hardlink's copied staging file, and its not shared metadata between them. In the current implementation, we also currently reject modifying a hardlink's data (including the source file).

tools/fuse-waked/main.cpp
AbrarQuazi: Yeah currently I preserve the ordering so calls to `has_children` (used by `rmdir`) and `moved_staged_children` (used by `rename` to rename a dir's children), can benefit from this with `O(log N + k)` instead of having to scan every entry with `O(N)`.

I am open to other potential structures if we see this as a bigger bottleneck (or we find its acceptable tradeoff for `rmdir` and `rename`)

tools/fuse-waked/main.cpp
AbrarQuazi: Gotcha, will make a future PR with more comprehensive tests. I believe we will want to support uid/gid in the future so will definitely keep that in mind

tools/fuse-waked/main.cpp
AbrarQuazi: yep will take a look at this again in the future

tools/fuse-waked/main.cpp
AbrarQuazi: With CAS mode , this test is actually not going to pass, since currently I block all writes to the hardlinks set (which adds both src and destination), and an unlink via `rm` currently does not update the set.

I think to enable this we will need some sort of mapping from a hardlink path to all its co-hardlinked path. Let's make sure to revisit this when we think about adding proper hardlink fuse support post beta, and add proper tests for that

src/wakefs/daemon_client.cpp
AbrarQuazi: I think the easiest way would be to have separate `mount_path`'s between CAS mode and Non-CAS mode, so they dont mismatch and interfere - I think this would be fine

so the mount paths:
Non-cas: `.fuse/<uid>.<gid>/`
CAS: `.fuse/<uid>.<gid>.cas/`

src/wcl/unique_fd.h
AbrarQuazi: Sorry this slipped from me

tests/wakebox/cas-symlink/pass.sh
AbrarQuazi: Yep, I think we for sure have to discuss proper test methadologies and test infra before we transition CAS being the default

tools/fuse-waked/main.cpp
AbrarQuazi: yep, not used - removed this

tools/fuse-waked/main.cpp
AbrarQuazi: added assertions here

src/wakefs/daemon_client.cpp
dtzSiFive: Awesome, yes agreed, love it thanks!

src/wcl/unique_fd.h
dtzSiFive: NP

tools/fuse-waked/main.cpp
dtzSiFive: Works for me! Thanks

dtzSiFive (approved):
I have not re-walked this entirely but based on discussions and changes I've seen go by and answers to my comments, LGTM.

PR #1788, opened 2026-04-13
Add builtin type Time

ngraybeal:
Creates a builtin type Time for the main purpose of handling timestamps.

Time is stored as a unit of nanoseconds since linux epoch.

This uses the chrono library to help safely handle time. Since we are using c++17, I'm still using strftime to format the string. This does augment strftime with a `%f` formatter for specifying subsecond time.

Functions include:
getTime
formatTime in UTC
Time conversion to nanoseconds ,milliseconds and seconds (integer, double, double)
Println to specify format and timezone.

I did rename the database Time struct to DBTime to avoid conflicts.

This is a more narrowly scoped feature set for Time than original [PR](https://github.com/sifiveinc/wake/pull/1768). The biggest difference is the avoidance of Timezone due to its complexities. The only way timezone can be specified is through the print statement, otherwise everything is handled in UTC.

### DIFF
--- a/share/wake/lib/core/types.wake
+++ b/share/wake/lib/core/types.wake
@@ -78,3 +78,11 @@ from builtin export type RegExp
 #
 # Jobs are created via the `runJob` API.
 from builtin export type Job
+
+# The Time type is a builtin of the wake language itself.
+#
+# A Time object represents a point in time with nanosecond
+# precision, stored as nanoseconds since the Unix epoch
+# (January 1, 1970 00:00:00 UTC).
+#
+from builtin export type Time
--- a/share/wake/lib/system/time.wake
+++ b/share/wake/lib/system/time.wake
@@ -1,0 +1,129 @@
+# Copyright 2026 SiFive, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You should have received a copy of LICENSE.Apache2 along with
+# this software. If not, you may obtain a copy at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package time
+
+# Time Library
+#
+# The Time type is a built-in type representing a point in time since the Unix epoch
+# with nanosecond precision.
+#
+# Format strings use strftime syntax with additional sub-second specifiers:
+#   %f     - nanoseconds since last whole second (9 digits, zero-padded)
+#   %.f    - decimal fraction with leading dot, trailing zeros trimmed
+#   %.Nf   - decimal fraction with leading dot, fixed N digits (N = 1-9)
+#   %Nf    - fractional digits without dot, fixed N digits (N = 1-9)
+#
+# Examples:
+#   "%H:%M:%S%.3f"  => "08:15:07.026"    (milliseconds)
+#   "%H:%M:%S%.6f"  => "08:15:07.026490" (microseconds)
+#   "%H:%M:%S%3f"   => "08:15:07026"     (millis, no dot)
+#   "%H:%M:%S%.f"   => "08:15:07.02649"  (trimmed)
+
+# getTime: Get the current system time.
+#
+# Returns a Time value representing the current time since the Unix epoch
+# with nanosecond precision.
+#
+# Example:
+#   def now = getTime Unit
+export def getTime (_: a): Time =
+    prim "get_time"
+
+# formatTime: Format a Time value as a string in UTC.
+#
+# Converts a Time value to a human-readable string using strftime format
+# specifiers with sub-second support as %f. The output is always in UTC.
+#
+# Arguments:
+#   format - strftime format string
+#   time   - Time value to format
+#
+# Example:
+#   def now = getTime Unit
+#   formatTime "%Y-%m-%d" now                    # "2026-03-10"
+#   formatTime "%Y-%m-%d %H:%M:%S" now           # "2026-03-10 08:15:07"
+#   formatTime "%Y-%m-%d %H:%M:%S.%.3f" now      # "2026-03-10 08:15:07.026"
+export def formatTime (format: String) (time: Time): String =
+    (\f \t prim "format_time") format time
+
+# getTimeISO: Get the current time as an ISO 8601 string in UTC.
+#
+# Example:
+#   getTimeISO Unit  # "2026-03-10T08:15:07Z"
+export def getTimeISO (_: a): String =
+    formatTime '%Y-%m-%dT%H:%M:%SZ' (getTime Unit)
+
+# timeToNanoseconds: Convert a Time to nanoseconds since the Unix epoch.
+#
+# Example:
+#   def nanos = timeToNanoseconds (getTime Unit)
+export def timeToNanoseconds (time: Time): Integer =
+    (\t prim "time_to_nanos") time
+
+# timeToMilliseconds: Convert a Time to milliseconds since the Unix epoch.
+#
+# Example:
+#   def millis = timeToMilliseconds (getTime Unit)
+export def timeToMilliseconds (time: Time): Double =
+    dint (timeToNanoseconds time) /. 1.0e6
+
+# timeToSeconds: Convert a Time to seconds since the Unix epoch.
+#
+# Example:
+#   def secs = timeToSeconds (getTime Unit)
+export def timeToSeconds (time: Time): Double =
+    dint (timeToNanoseconds time) /. 1.0e9
+
+# printTimeLnLevel: Print a message with a timestamp prefix at a given log level.
+#
+# The timestamp is formatted using strftime format specifiers in the specified
+# timezone. Use "UTC" for UTC output, or an IANA timezone name. Empty string
+# uses the local machine timezone.
+#
+# Arguments:
+#   format   - strftime format string for the timestamp
+#   timezone - IANA timezone name (e.g., "UTC", "America/Los_Angeles")
+#   time     - Time value to format
+#   logLevel - LogLevel to print at
+#   message  - message to print
+#
+# Example:
+#   printTimeLnLevel "%H:%M:%S" "UTC" (getTime Unit) logWarning "something happened"
+#   printTimeLnLevel "%H:%M:%S" "America/Los_Angeles" (getTime Unit) logReport "hello"
+export def printTimeLnLevel (format: String) (timezone: String) (time: Time) (logLevel: LogLevel) (message: String): Unit =
+    def LogLevel name = logLevel
+
+    (\f \tz \t \s \m prim "time_print") format timezone time name "{message}\n"
+
+# printTsLn: Print a local-timezone ISO 8601 timestamped message at report level.
+#
+# Output: [2026-04-07T01:15:07-0700] message
+#
+# Example:
+#   printTsLn (getTime Unit) "build started"
+export def printTsLn (time: Time) (message: String): Unit =
+    printTimeLnLevel '[%Y-%m-%dT%H:%M:%S%.3f%z]' '' time logReport message
+
+# printCurrentTsLn: Print a local-timezone ISO 8601 timestamped message at report level.
+# Automatically captures the current time.
+#
+# Output: [2026-04-07T01:15:07-0700] message
+#
+# Example:
+#   printCurrentTsLn "build started"
+export def printCurrentTsLn (message: String): Unit =
+    printTsLn (getTime Unit) message
+
--- a/share/wake/lib/system/time.wake
+++ b/share/wake/lib/system/time.wake
@@ -101,22 +101,20 @@ export def printTimeLnLevel (format: String) (time: Time) (logLevel: LogLevel) (
 
     (\f \t \s \m prim "time_print") format time name "{message}\n"
 
-# printTsLn: Print a local-timezone ISO 8601 timestamped message at report level.
+# printTimeLn: Print a ISO 8601 timestamped message at report level.
 #
 # Output: [2026-04-07T01:15:07-0700] message
 #
 # Example:
-#   printTsLn (getTime Unit) "build started"
-export def printTsLn (time: Time) (message: String): Unit =
+#   printTimeLn (getTime Unit) "build started"
+export def printTimeLn (time: Time) (message: String): Unit =
     printTimeLnLevel '[%Y-%m-%dT%H:%M:%S%.3f%z]' time logReport message
 
-# printCurrentTsLn: Print a local-timezone ISO 8601 timestamped message at report level.
-# Automatically captures the current time.
+# printCurrentTimeLn: Print an ISO 8601 timestamped message at report level using the current time.
 #
 # Output: [2026-04-07T01:15:07-0700] message
 #
 # Example:
-#   printCurrentTsLn "build started"
-export def printCurrentTsLn (message: String): Unit =
-    printTsLn (getTime Unit) message
-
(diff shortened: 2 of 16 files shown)


share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,129 @@
...
+#   %f     - nanoseconds since last whole second (9 digits, zero-padded)
+#   %.f    - decimal fraction with leading dot, trailing zeros trimmed
+#   %.Nf   - decimal fraction with leading dot, fixed N digits (N = 1-9)
+#   %Nf    - fractional digits without dot, fixed N digits (N = 1-9)
+#
+# Examples:
+#   "%H:%M:%S%.3f"  => "08:15:07.026"    (milliseconds)
+#   "%H:%M:%S%.6f"  => "08:15:07.026490" (microseconds)
+#   "%H:%M:%S%3f"   => "08:15:07026"     (millis, no dot)
+#   "%H:%M:%S%.f"   => "08:15:07.02649"  (trimmed)
```
dtzSiFive: I can see this going either way, but found it a little surprising this explanation was before a primitive that doesn't use format strings 😁 .

Having all explanations at top is nice, however, so just a comment.

share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,129 @@
...
+#   format   - strftime format string for the timestamp
+#   timezone - IANA timezone name (e.g., "UTC", "America/Los_Angeles")
+#   time     - Time value to format
+#   logLevel - LogLevel to print at
+#   message  - message to print
+#
+# Example:
+#   printTimeLnLevel "%H:%M:%S" "UTC" (getTime Unit) logWarning "something happened"
+#   printTimeLnLevel "%H:%M:%S" "America/Los_Angeles" (getTime Unit) logReport "hello"
+export def printTimeLnLevel (format: String) (timezone: String) (time: Time) (logLevel: LogLevel) (message: String): Unit =
```
richardxia: I'm wondering if it makes sense to expose the `timezone` as a parameter. Other than UTC, you wouldn't want any code to hardcode a time zone as a string, since it will be incorrect for any person not in that time zone.

Should there just be two versions of this function: one for always printing UTC, and the other for using the `TZ` environment variable? Or perhaps always use `TZ` and default to `UTC` if it's not set?

share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,129 @@
...
+
+    (\f \tz \t \s \m prim "time_print") format timezone time name "{message}\n"
+
+# printTsLn: Print a local-timezone ISO 8601 timestamped message at report level.
+#
+# Output: [2026-04-07T01:15:07-0700] message
+#
+# Example:
+#   printTsLn (getTime Unit) "build started"
+export def printTsLn (time: Time) (message: String): Unit =
```
richardxia: Nit: It's inconsistent for one of these functions to use the word `Time` in its name while the other two functions uses `Ts`. Can we be consistent and pick one naming convention?

src/runtime/time.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include <string>
+
+#include "prim.h"
+#include "status.h"
+#include "types/data.h"
+#include "value.h"
+
+// Pre-process a format string to expand %f sub-second specifiers before passing to strftime.
+// Supported: %f (9 digits), %.f (trimmed), %.Nf (dot + N digits), %Nf (N digits) where N is 1-9
+static std::string expand_subsecond_formats(const char *fmt, int64_t nanoseconds) {
```
AbrarQuazi, line 38: I know that theres no proper library in `c++17` for subsecond support, but what's preventing us from just importing over https://github.com/howardhinnant/date instead of trying to roll out own implementation?

src/runtime/time.cpp
```diff
@@ -0,0 +1,206 @@
...
+      result += '%';
+    }
+  }
+  return result;
+}
+
+// Format a Time's nanoseconds using strftime with sub-second expansion.
+// If timezone is nullptr, uses UTC (gmtime_r). Otherwise sets TZ and uses localtime_r.
+// Empty string timezone means system default (unset TZ).
+static std::string format_time_str(const char *fmt, int64_t nanoseconds, const char *timezone) {
```
AbrarQuazi, line 93: How should we be handling errors like invalid timezone etc?

share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,129 @@
...
+#   "%H:%M:%S%.f"   => "08:15:07.02649"  (trimmed)
+
+# getTime: Get the current system time.
+#
+# Returns a Time value representing the current time since the Unix epoch
+# with nanosecond precision.
+#
+# Example:
+#   def now = getTime Unit
+export def getTime (_: a): Time =
```
AbrarQuazi, line 30: Should either mark this function as "unsafe" or add documentation that this breaks caching 

similar comment for `time_print`

src/runtime/time.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+static PRIMTYPE(type_time_to_nanos) {
+  return args.size() == 1 && args[0]->unify(Data::typeTime) && out->unify(Data::typeInteger);
+}
+
+void prim_register_time(PrimMap &pmap) {
+  prim_register(pmap, "get_time", prim_get_time, type_get_time, PRIM_ORDERED);
+  prim_register(pmap, "format_time", prim_format_time, type_format_time, PRIM_PURE);
+  prim_register(pmap, "time_to_nanos", prim_time_to_nanos, type_time_to_nanos, PRIM_PURE);
+  prim_register(pmap, "time_print", prim_time_print, type_time_print, PRIM_ORDERED);
```
AbrarQuazi: Shouldn't this be marked as `PRIM_IMPURE` similar to `prim_print` in `string.cpp`

AbrarQuazi: Since its writing to stream?

AbrarQuazi (commented):
Some comments/questions

src/runtime/time.cpp
ngraybeal: It doesn't support subsecond formatting. I feel the implementation is pretty straightforward and narrow in scope, so we can avoid the extra baggage.

src/runtime/time.cpp
ngraybeal: I'll make a note. We are following glibc behavior and falling back on UTC for any invalid timezone.

share/wake/lib/system/time.wake
ngraybeal: Your comment made sense, though i went with option 3 (not listed) and just followed glibc default behavior. Use TZ if set, otherwise default to system configuration.

This doesn't impact format time function in cpp or wake.

share/wake/lib/system/time.wake
ngraybeal: Went ahead and bumped it down to the format string prim and reformatted the comment.

share/wake/lib/core/types.wake
```diff
@@ -78,3 +78,11 @@ from builtin export type RegExp
...
 # Jobs are created via the `runJob` API.
 from builtin export type Job
+
+# The Time type is a builtin of the wake language itself.
+#
+# A Time object represents a point in time with nanosecond
+# precision, stored as nanoseconds since the Unix epoch
+# (January 1, 1970 00:00:00 UTC).
+#
+from builtin export type Time
```
ag-eitilt, line 88: Can we export this in `package Time` to keep it all in one place?

share/wake/lib/system/time.wake
ag-eitilt: Probably wouldn't hurt to have this be either `unsafe_getTime` or `getCurrentTime` just to drive the point home, yeah, but the prefix is more a means of flagging the functions during reviews than it is any concrete policy.  The main goal is just to make sure that anyone working with its values is aware of the potential unreproducibility, and getting the system time is already intuitively unreproducible.

Printing doesn't need the flag though -- it might have a *user*-observable side effect, but since it returns `Unit` nothing *within* wake can observe that it's any different than any other call to it.

share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,120 @@
...
+    (\f \t \s \m prim "time_print") format time name "{message}\n"
+
+# printTimeLn: Print a ISO 8601 timestamped message at report level.
+#
+# Output: [2026-04-07T01:15:07-0700] message
+#
+# Example:
+#   printTimeLn (getTime Unit) "build started"
+export def printTimeLn (time: Time) (message: String): Unit =
+    printTimeLnLevel '[%Y-%m-%dT%H:%M:%S%.3f%z]' time logReport message
```
ag-eitilt, line 111: Nit:
```suggestion
    printTimeLnLevel '[%Y-%m-%dT%H:%M:%S%.3f%z] ' time logReport message
```

src/runtime/time.cpp
ag-eitilt: Yeah, I recently gained a new appreciation for the "vendor or reimplement your (simple) dependencies" mentality.  Especially since we'd only be exposing a minimal surface area anyway, may as well keep the dependency tree well-trimmed.

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+  std::string result;
+  for (const char *p = fmt; *p; ++p) {
+    if (*p != '%') {
+      result += *p;
+      continue;
+    }
+
+    // Look ahead from '%'
+    if (p[1] == '.' && p[2] >= '1' && p[2] <= '9' && p[3] == 'f') {
+      // %.Nf — dot + fixed N digits (N = 1-9)
```
ag-eitilt, line 54: Sneaky AI even sneaks its em-dashes into otherwise-ASCII code comments. 😂

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <ctime>
+#include <string>
+
+#include "prim.h"
+#include "status.h"
+#include "types/data.h"
+#include "value.h"
+
+// Pre-process a format string to expand %f sub-second specifiers before passing to strftime.
+// Supported: %f (9 digits), %.f (trimmed), %.Nf (dot + N digits), %Nf (N digits) where N is 1-9
```
ag-eitilt:
```suggestion
// Supported: %f (9 digits), %.f (trimmed), %.Nf (trimmed + N digits), %Nf (N digits) where N is 1-9
```
Also, do we think we need to clarify the definition of "trimmed" or is it clear enough?  I guess this is in an internal-dev comment anyway, so that would probably be overkill.

share/wake/lib/system/time.wake
```diff
@@ -0,0 +1,120 @@
...
+#   time   - Time value to format
+#
+# Examples:
+#   formatTime "%Y-%m-%d" (getTime Unit)              # "2026-03-10"
+#   formatTime "%Y-%m-%d %H:%M:%S" (getTime Unit)     # "2026-03-10 08:15:07"
+#   formatTime "%H:%M:%S%.3f" (getTime Unit)          # "08:15:07.026" (milliseconds)
+#   formatTime "%H:%M:%S%.6f" (getTime Unit)          # "08:15:07.026490" (microseconds)
+#   formatTime "%H:%M:%S%3f" (getTime Unit)           # "08:15:07026" (no dot)
+#   formatTime "%H:%M:%S%.f" (getTime Unit)           # "08:15:07.02649" (trimmed)
+export def formatTime (format: String) (time: Time): String =
```
ag-eitilt, line 55: I don't like how the `%3f` case has seemingly-special behaviour.  Is that really how `sprintf` handles non-time floats?  And `%f` should be in the table, particularly since it's another of the no-dot cases.

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+      // Not a sub-second specifier — pass through for strftime
+      result += '%';
+    }
+  }
+  return result;
+}
+
+// Format a Time's nanoseconds using strftime with sub-second expansion.
+// If timezone is nullptr, uses UTC (gmtime_r). Otherwise sets TZ and uses localtime_r.
+// Empty string timezone means system default (unset TZ).
```
ag-eitilt: Is this line contradicting itself?  System default, but it also unsets `TZ`?

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+    char *old_tz = getenv("TZ");
+    std::string saved_tz;
+    bool had_tz = (old_tz != nullptr);
+    if (had_tz) saved_tz = old_tz;
+
+    if (timezone[0] == '\0') {
+      unsetenv("TZ");
+    } else {
+      setenv("TZ", timezone, 1);
+    }
```
ag-eitilt, line 116: Like I mentioned in the quick scan I did live, I would really like it if this syscall hadn't been implemented to depend on the user environment.  Nothing we can do, though.

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+      result.append(nanos_str, n);
+      p += 2;
+    } else if (p[1] == 'f') {
+      // %f — all 9 digits, no dot
+      result.append(nanos_str, 9);
+      p += 1;
+    } else {
+      // Not a sub-second specifier — pass through for strftime
+      result += '%';
+    }
```
ag-eitilt, line 84: We need a special case to skip over both percents in `%%foo`.

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+      if (last != std::string::npos) {
+        result += '.';
+        result.append(frac, 0, last + 1);
+      }
+      p += 2;
+    } else if (p[1] >= '1' && p[1] <= '9' && p[2] == 'f') {
+      // %Nf — fixed N digits, no dot (N = 1-9)
+      int n = p[1] - '0';
+      result.append(nanos_str, n);
+      p += 2;
```
ag-eitilt, line 72: It would be nice to have rounding, not truncation, but if the standard `sprintf` truncates we definitely want to match that behaviour.

src/runtime/time.cpp
```diff
@@ -0,0 +1,207 @@
...
+      setenv("TZ", saved_tz.c_str(), 1);
+    } else {
+      unsetenv("TZ");
+    }
+    tzset();
+  }
+
+  std::string expanded = expand_subsecond_formats(fmt, nanoseconds);
+  char buffer[256];
+  strftime(buffer, sizeof(buffer), expanded.c_str(), &tm_info);
```
ag-eitilt, line 130: Is there any chance this could be abused to construct malicious formatting?  So long as you make the `%%` fix, I'm having trouble constructing a format string which the expansion would collapse but the `strftime` would think was a length for some surrounding interpolation, and I'm also having trouble thinking of a case where even if that *did* happen it would be able to do anything significant.  I'm not a black hat, though.

ag-eitilt (approved):
Not sure how many of these are better to fix here, vs fixing in a followup, but there's no big issues I had with any of it.

src/runtime/time.cpp
ngraybeal: Interesting point. though does have the added complication of carrying over into seconds if it all rounds up.

share/wake/lib/system/time.wake
ngraybeal: I prefer not having the space to give more control to the caller.

src/runtime/time.cpp
ngraybeal: Good catch

src/runtime/time.cpp
AbrarQuazi: This is probably not multithread safe right? I know that since wakes runtime is single threaded - so its probably not a big deal right now - but in the future this might be an issue if want to make the runtime multihreaded some how. Would this be multiprocess/multiwake safe though?

AbrarQuazi (approved):
LGTM

src/runtime/time.cpp
ngraybeal: I went ahead and added a mutex lock.

c++20 does have a `localtime_rz` to avoid all this

PR #1789, opened 2026-04-13
wake --clean: recover freed space on disk

dtzSiFive:
We likely just deleted a large portion of the database.

To avoid pathological behavior asking sqlite to later do incremental vacuum on a mostly-empty database
(which can be very slow with WAL enabled),
take this opportunity to reclaim disk space
for the database and its log file.

Vacuum does compaction + defragmentation as well.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -814,16 +814,26 @@ void Database::clean() {
   // Add checkpoint after cleanup operations
   checkpoint(false);  // Non-blocking checkpoint for sync point
 
-  // This cannot be a prepared statement, because pragmas may run on prepare
-  char *fail;
-  int ret = sqlite3_exec(imp->db, "pragma incremental_vacuum;", 0, 0, &fail);
-  if (ret != SQLITE_OK) std::cerr << "Could not recover space: " << fail << std::endl;
+  vacuum();
+}
+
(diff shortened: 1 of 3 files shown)


src/runtime/database.cpp
```diff
@@ -814,16 +814,26 @@ void Database::clean() {
...
+void Database::vacuum(bool incremental) {
+  // Non-incremental requires extended exclusive access to entire DB.
+  auto *command = incremental ? "PRAGMA incremental_vacuum;" : "vacuum;";
+
+  char *fail = nullptr;
+  int ret = sqlite3_exec(imp->db, command, 0, 0, &fail);
+  auto *err = fail ? fail : "unknown error";
+  auto *kind = incremental ? "incremental" : "full";
+  if (ret != SQLITE_OK)
+    std::cerr << "Could not recover space (" << kind << "):" << err << std::endl;
```
ag-eitilt: Very minor:
```suggestion
    std::cerr << "Could not recover space (" << kind << "): " << err << std::endl;
```

ag-eitilt (approved):
Not familiar with this level of SQLite management, but this all seems reasonable.

src/runtime/database.cpp
dtzSiFive: You're absolutely right, thanks!

PR #1790, opened 2026-04-15
build: Limit version description to v* tags.

dtzSiFive:
This avoids other tags breaking wake expecting a version-like string to be produced here.


dtzSiFive:
Need to similarly touchup vendor/git.wake will fix that later.

ngraybeal (approved):
LGTM to align on a versioning schema

dtzSiFive:
Thanks!

PR #1792, opened 2026-04-15
database: retry setting writer pragmas

dtzSiFive:
These may require exclusive access in a way that ignores busy_timeout and so we need to explicitly handle.

I believe this should only occur when multiple wake runs concurrently attempt to set these (fresh DB case), and that steady-state there's no risk of locking.

Fixes issue encountered stress-testing multi-wake.


ngraybeal (approved):

dtzSiFive:
Just saw this, thanks for the review! 🙏

PR #1795, opened 2026-04-20
Stop making detectGccVersion to run by default

AbrarQuazi:
This PR fixes `detectGccVersion` from running eagerly by default whenever a wake command runs (wake -x Unit)

### DIFF
--- a/share/wake/lib/gcc_wake/gcc.wake
+++ b/share/wake/lib/gcc_wake/gcc.wake
@@ -37,10 +37,9 @@ target detectGccVersion Unit: Option Integer =
 # Determine if -lstdc++fs is needed based on system GCC version.
 # libstdc++ from GCC < 9 requires this flag for std::filesystem.
 # On Linux, both GCC and Clang use libstdc++ by default.
-def filesystemLFlag Unit: List String =
-    match (detectGccVersion Unit)
-        Some gccVersion if gccVersion < 9 -> "-lstdc++fs", Nil
-        _ -> Nil
+def filesystemLFlag Unit: List String = match (detectGccVersion Unit)
+    Some gccVersion if gccVersion < 9 -> "-lstdc++fs", Nil
+    _ -> Nil
 
 export def cpp11Flags =
     ("-std=c++11", Nil)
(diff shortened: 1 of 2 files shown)


share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -37,9 +37,10 @@ target detectGccVersion Unit: Option Integer =
 # Determine if -lstdc++fs is needed based on system GCC version.
 # libstdc++ from GCC < 9 requires this flag for std::filesystem.
 # On Linux, both GCC and Clang use libstdc++ by default.
-def filesystemLFlag Unit: List String = match (detectGccVersion Unit)
-    Some gccVersion if gccVersion < 9 -> "-lstdc++fs", Nil
-    _ -> Nil
+def filesystemLFlag Unit: List String =
```
dtzSiFive: Is this just a formatting change?
(I'm staring at it and only seeing the match broken off to new line as difference, yes?)

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -56,19 +57,19 @@ export def c11Flags =
...
     ("-lpthread", "-g", "-pg", "-Wl,--build-id", Nil) ++ filesystemLFlag Unit
 
 export def releaseCFlags =
     ("-Wall", "-Wno-format-security", "-g", "-O2", Nil)
 
-export def releaseLFlags =
+export def releaseLFlags Unit =
     ("-lpthread", "-Wl,--build-id", Nil) ++ filesystemLFlag Unit
 
 export def staticCFlags =
```
dtzSiFive, line 68: Little unexpected for some of these to be invoked with `Unit` and others not (lflags vs cflags?); also since these are exported is this a breaking change?

dtzSiFive (commented):
Thanks for addressing this.  Are these wake-internal-only?

Main doubt is that this change makes the various flag kinds have a different access mechanism/pattern.  Maybe they all should be uniform in that way?

share/wake/lib/gcc_wake/gcc.wake
AbrarQuazi: The cFlags are just lists and dont need the call a function (in this case a function that sets `-lstdc++fs` which is a linker time flag)


No other uses outside of this repo currently

share/wake/lib/gcc_wake/gcc.wake
AbrarQuazi: yep only difference

AbrarQuazi: which reminds me I have to run the wake-formatter ;)

dtzSiFive:
Okay, sure.  I would prefer the uniformity but if you'd rather not since it's not needed and that seems better, ship it!

dtzSiFive (approved):
Thanks!

PR #1798, opened 2026-04-22
Feature/workspace virtualization

AbrarQuazi:
This feature branch combines the entire changes needed for workspace virtualization beta release. It has the following changes:
* https://github.com/sifiveinc/wake/pull/1743
* https://github.com/sifiveinc/wake/pull/1738
* https://github.com/sifiveinc/wake/pull/1787
* https://github.com/sifiveinc/wake/pull/1795

It includes the following API updates:
* Update to `getPathParent` (takes in a string instead of a `Path`)
* Updates to `RunnerOutput` (includes a new `StagedOutputs` type)
* A new `workspaceRunner`
* Cas libraries to allow for materialization
* A new `staging_outputs` library to interact with staged files
* New fields in `Path`


share/wake/lib/system/cas.wake
```diff
@@ -0,0 +1,101 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package wake
+
+export def useCasHashes = match (getenv "WAKE_CAS")
+    Some _ -> True
+    None -> False
```
ngraybeal, line 20: Do we not want to check at least for `0` or `1`?

AbrarQuazi: Does it matter really? I dont really want people actively setting it and toggling between them like that, and hopefully its not user facing (going to be set by the module). We are going to remove it soon afterwards regardless

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -15,6 +15,32 @@
...
+        | replace `\s` ""
+
+    # -dumpversion may output versions like "8", "8.5", or "8.5.0"
+    match (extract `^([0-9]+)` output)
+        (majorStr, Nil) -> int majorStr
+        _ -> None
+
+# Determine if -lstdc++fs is needed based on system GCC version.
+# libstdc++ from GCC < 9 requires this flag for std::filesystem.
+# On Linux, both GCC and Clang use libstdc++ by default.
```
ag-eitilt, line 39: If we ever get serious about packaging Wake for broader use, it would be really nice to avoid hardcoding these things about the toolchain -- even before switching to FreeBSD, I was using Clang and libc++ for all the Gentoo building I could.  I don't think tying it specifically to GCC is a bad thing, particularly not for a beta branch, but it would get in the way of eventually restoring Alpine and non-Linux to CI.

share/wake/lib/gcc_wake/gcc.wake
```diff
@@ -95,8 +121,8 @@ def doLinkO variant linker flags objects targ extraFiles =
 export def makeCompileC variant gcc flags =
     Pair variant (\extraFlags doCompileC variant gcc (flags ++ extraFlags)), Nil
 
-export def makeLinkO variant linker flags =
-    Pair variant (\extraFlags doLinkO variant linker (flags ++ extraFlags)), Nil
+export def makeLinkO variant linker flagsFn =
+    Pair variant (\extraFlags doLinkO variant linker (flagsFn Unit ++ extraFlags)), Nil
```
ag-eitilt, line 125: It feels like it might be better at this point to pass the `extraFlags` as the `flagsFn` argument, since we're no longer treating it as a static list.  A bit more semantic than just a `Unit`, and offers a method for non-compiler-tied flag expansion in the future.
```suggestion
export def makeLinkO variant linker flagsFn =
    Pair variant (\extraFlags doLinkO variant linker (flagsFn extraFlags)), Nil
```

share/wake/lib/system/io.wake
```diff
@@ -150,14 +171,24 @@ def writeRunner (content: String) =
...
+            | RunnerOutput Nil emptyStagedOutputs Nil emptyUsage
 
         require True = mode >= 0 && mode <= 0x1ff
         else
             Some "write {path}: Invalid mode ({strOctal mode})".makeError
-            | RunnerOutput Nil Nil Nil emptyUsage
+            | RunnerOutput Nil emptyStagedOutputs Nil emptyUsage
+
+        def stagedFilePath =
+            "{workspaceStageBase}/direct/write.{prefix}.{hashString "{path}\n{content}"}"
```
ag-eitilt, line 182: I *think* `prefix` is unique per job, and since we'll never be `ReRun`ing write jobs, the `hashString` is probably unnecessary.  Equally, though, we'd ideally not be keeping so much data in memory that the hashing imposes a significant time penalty, so it's probably not too bad a thing to leave it in just for that extra safety.

share/wake/lib/system/io.wake
```diff
@@ -150,14 +171,24 @@ def writeRunner (content: String) =
...
 
-        def writeTask = primWrite mode path content
+        def writeTask =
+            require Pass _ = primStageMkdir ".build"
+            require Pass _ = primStageMkdir ".build/staging"
+            require Pass _ = primStageMkdir workspaceStageBase
```
ag-eitilt, line 187: This relies on `workspaceStageBase` being a direct child of `.build/staging`.  *Should* be safe, but might wind up fragile if we ever change that.

share/wake/lib/system/io.wake
```diff
@@ -168,14 +199,27 @@ def writeRunner (content: String) =
...
-        # Wait for the virtual job to complete
-        match writeTask
-            Fail msg -> RunnerOutput fileInputs Nil Nil usage (Some msg.makeError)
-            Pass path ->
-                def cleanable = (path, Nil)
-                def fileOutputs = fnOutputs cleanable
+        def stagedOutputSetResult = match writeTask
+            Fail err -> Fail err
+            Pass _ ->
+                def publishedOutputPaths = fnOutputs (path, Nil)
```
ag-eitilt, line 205: Nit, definitely not required for merging:
```suggestion
        def stagedOutputSetResult =
            require Pass _ = writeTask

            def publishedOutputPaths = fnOutputs (path, Nil)
```

share/wake/lib/system/job.wake
```diff
@@ -143,6 +145,115 @@ def implode strings =
...
+# Finalize a completed job: collect any runner/output errors, invoke the post-finalize callback,
+# clean up the staging root, and call primJobFinish to publish the job's outputs.
+def finalizeJob job inputs outputs cleanable reality runnerError =
+    def workspaceOutputsReadyResult = materializeJobOutputs outputs
+
+    def finalRunnerError: Option Error =
+        (runnerError, getFail workspaceOutputsReadyResult, Nil)
+        | find isSome
+        | omap getPairFirst
+        | getOrElse None
```
ag-eitilt, line 218: I think `findSome` would work?  Not sure if that actually exists, though, and it might not even have the type I'm thinking it does.
```suggestion
        | findSome
```

share/wake/lib/system/path.wake
```diff
@@ -110,77 +114,107 @@ export def basename (file: String): String =
...
     # things cause us to go over. These numbers appear to be safe in practice (found exact via binary search).
     # To account for any other variance I backed off these numbers quite a bit.
-    def use_file = to_hash_file_bytes >= 1900000 || to_hash_len >= 12000
-    def add f h = prim "add_hash"
+    def useStdinFile = toHashFileBytes >= 1900000 || toHashLen >= 12000
 
     def hashPlan cmd vis =
         Plan
-        "hash: #{prefix}"
+        "hash: {str toHashLen} paths"
```
ag-eitilt: I'm still on my horse about passing the `prefix` into `computeHashes` just in order to have *something* tying this back to the base Job being a good thing.  Even if you're using this in contexts other than immediately after Jobs, we can pass some other identifier instead of the Job ID for those cases.

share/wake/lib/system/path.wake
```diff
@@ -195,66 +229,67 @@ def computeHashes (prefix: String) (files: List String): List String =
...
 
-    def stdin_file_path = ".build/tmp/to_hash.{prefix}.stdin"
-
-    # We construct a different plan depending on if we could use command line arguments or not
-    require Pass plan = match use_file
+    require Pass (Pair plan maybeStdinFilePath) = match useStdinFile
         True ->
-            require Pass stdin_file = write stdin_file_path (catWith "\n" to_hash)
+            def stdinContent = catWith "\n" simpleFiles
+            def stdinFilePath = ".build/tmp/hash.{str wakePid}.{hashString stdinContent}.stdin"
```
ag-eitilt, line 239: This one, unlike the `write` name, *does* need the uniqueness granted by `hashString`. 👍🏻

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -1140,31 +1157,12 @@ def mkCacheSearchResponse (json: JValue): Result CacheSearchResponse Error =
...
-        stdout
-        | extract `(.{64}).*`
-
-    match fileHash
-        (hash, Nil) -> Pass hash
-        _ -> failWithError "rsc: No hash found in output for file {file}"
+    if cached !=* "" then
+        Pass cached
+    else
+        failWithError "rsc: hash not found in DB for {file}"
```
ag-eitilt, line 1168: Out of curiosity, what was the change which makes RSC's separate hashing unnecessary?  It's definitely a good thing, but there had to be a reason Ashley didn't just use `get_hash` to begin with, right?

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -353,160 +388,153 @@ def mkCacheAllowedRequest ((RunnerInput label cmd vis env dir stdin _res _prefix
...
+    def StagedOutputs filteredOutputPaths _ stagingItems _ _ = stagedOutputs
+
+    def publishedOutputSet =
+        filteredOutputPaths
+        | map (\p p :-> Unit)
+        | listToMap scmp
+
+    def isPublishedOutput (path: String): Boolean = match (mlookup path publishedOutputSet)
+        Some _ -> True
+        None -> False
```
ag-eitilt, line 403:
```suggestion
    def publishedOutputSet =
        filteredOutputPaths
        | listToTree scmp

    def isPublishedOutput (path: String): Boolean = tcontains path publishedOutputSet
```

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -353,160 +388,153 @@ def mkCacheAllowedRequest ((RunnerInput label cmd vis env dir stdin _res _prefix
...
+            | Pass
+
+        def publishedFileUpload (item: StagingItem): Option (Result CachePostRequestOutputFile Error) =
+            match item
+                StagingFile file ->
+                    if isPublishedOutput file.getStagingFileOutputDestPath then
+                        Some (uploadStagingFile file)
+                    else
+                        None
+                _ -> None
```
ag-eitilt, line 431: Not sure if this is something you'd find easier to read or not, but you *can* get rid of the duplicate `None`s.  Parentheses not required, but even I have trouble reading the line without them, and I'm familiar with the syntax already.
```suggestion
            match item
                (StagingFile file) if isPublishedOutput file.getStagingFileOutputDestPath ->
                    Some (uploadStagingFile file)
                _ -> None
```

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -353,160 +388,153 @@ def mkCacheAllowedRequest ((RunnerInput label cmd vis env dir stdin _res _prefix
...
+            def stdoutUpload =
+                rscApi
+                | rscApiPostStringBlob "stdout" stdout
+
+            def stderrUpload =
+                rscApi
+                | rscApiPostStringBlob "stderr" stderr
+
+            # Due to a side effect of filtering outputs its possible for a job to create both a file
+            # and a symlink to that file, but the only output said file. This will break the build
```
ag-eitilt, line 495: Nit since you're touching the old code slightly. 😉 
```suggestion
            # and a symlink to that file, but then only output said file. This will break the build
```

share/wake/lib/system/sources.wake
```diff
@@ -185,6 +185,6 @@ export def claimFileAsPathIn (outputDirectory: Path) (existingFile: String) (des
         | makePlan "claim: {existingFile} -> {desiredWorkspacePath}" visible
         | setPlanPersistence Once
         | setPlanFnOutputs (\_ desiredWorkspacePath, Nil)
-        # This needs to be localRunner in case the existingFile is already inside the workspace.
-        | runJobWith localRunner
+        # This needs to be workspaceRunner in case the existingFile is already inside the workspace.
+        | runJobWith workspaceRunner
```
ag-eitilt, line 189: `stagingDirRunner` is coming later, right?  That would also be sufficient, and safer.

ag-eitilt (approved):
I've *still* not reviewed the C++, and probably have a few more comments here than you want to see in the last dash before beta; sorry!  Really, though, the thing I'm most strongly pushing back on is just the `hash` Plan labeling, so this isn't really a blocking review despite the number.

share/wake/lib/system/cas.wake
ag-eitilt: I'm pretty sure there's other env var flags with this sort of behaviour already, but I don't remember what they look at or even what side of the OSS/internal divide they're on.  I definitely remember pointing at these and laughing in the past, but had been writing them off as the equivalent of `gcc -DFLAG` -- though apparently that sets `FLAG=1` so it's not exactly as solid a prior art as I thought it was.

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: Its because of the `PostFinalize` callback as well as because of the writeRunner changes that ensure things are hashed and are in the DB before being used

AbrarQuazi:
> I've _still_ not reviewed the C++, and probably have a few more comments here than you want to see in the last dash before beta; sorry! Really, though, the thing I'm most strongly pushing back on is just the `hash` Plan labeling, so this isn't really a blocking review despite the number.

I'll add the prefix to the label! Sorry about that! Did'nt really understand what you meant at first in in the initial reviews!

share/wake/lib/system/job.wake
```diff
@@ -201,14 +307,38 @@ def runAlways cmd env dir stdin res uusage finputs foutputs vis keep run echo st
...
+        cache.getPairSecond
+        | map parseCachedPathInfo
+
     match cache
-        Pair (job, _) last -> confirm True last job
-        Pair Nil last -> confirm False last (build Unit)
+        Pair (job, _) _ ->
+            if useCasHashes then
+                require Pass _ = materializeCachedOutputs cachedOutputs
+                else build Unit
```
dtzSiFive: Can we error instead (and show user information about /why/ materialization failed?).

As-is, we'll end up instead producing -- almost surely and in practice AFAICT 100% of the time -- "File output by multiple Jobs" -- this is because we've already marked the cached job as used by this run, and so running it again will problematically look like two jobs producing the same outputs.

I'm not sure silencing an underlying error is the right move here.  If we think we'd like to try to sweep this under the rug, then we should alter how this works to pro-actively validate CAS blobs before allowing jobs to be reused.

I think the better option is to error here as this should be unreachable and something rather wrong is happening if it is reached.

And in any case I don't think running the job "again" here will be useful only problematic?

ag-eitilt: Good point.  I do like the *idea* of materialization failures being a warning rather than an error (goes along well with the workspace not being guaranteed to be consistent with multiple concurrent runs, and with potentially detecting user modifications and deliberately not overwriting them), but that is indeed going to need a bit of associated touchup to the logic.

It looks to me like the `build Unit` is an incorrect carryover of the `else` case from the `keep` test, and I'd agree that we wouldn't want to rerun -- the job was successfully retrieved from cache, we *want* to use that version, and the only problem is that the workspace might not match that view.  Things might be left in an odd state if we've got some `localRunner` jobs in the mix (either as what was cached or what the outputs are being fed into), but those always have the potential of going odd.

dtzSiFive: Fixed with #1822 .

PR #1803, opened 2026-04-22
Support multiple file's with same path.

dtzSiFive:
* schema: 12: -stale, no longer unique on path
* Send in all path info to job prims
* Find files using richer information.
* Consider multiple candidates in reuse_job,
  since find_prior doesn't check specifics of visible.
* Update add_hash for multi-wake
  Don't remove records, but do support new mtime.
* Update detect_overlap, delete_overlap.
  Keep these for time being.
  
  Test locally often runs into the issue fixed by #1792 .


AbrarQuazi (approved):
LGTM, thanks!

PR #1804, opened 2026-04-22
Move modified into filetree (and add to Path!).

dtzSiFive:
Move modified into filetree (and add to Path!).
This means a Path doesn't truly exist until
a job records its access of it.

No longer attempt to fetch mtime raw from workspace,
it must be passed in.

Modified observed (recorded) originally for a job
persists and seeing otherwise same file does not alter this.
This is used in materialization.

Explicitly do not consider modified as part of job
reuse candidate selection.  Similar behavior to having
jobs point to a record with a mutable mtime field,
supports important use cases like git operations.

Must now check modified as well as file_id when matching
on filetree.

get_hash and get_cached_path only return information
for files that are recorded in filetree:
add_hash is not sufficient.

Bump to schema 13.


dtzSiFive:
Closing in favor of #1806 .

PR #1806, opened 2026-04-22
Add modified to Path, use on visible files in FUSE runner

dtzSiFive:
No longer attempt to fetch mtime raw from workspace, it must be passed in.

Do not update modified "last-seen", preserve
modified for materialization and records.

Explicitly do not consider modified as part of job reuse candidate selection.  Similar behavior to having jobs point to a record with a mutable mtime field, supports important use cases like git operations.

Bump to schema 13 for change to index.


dtzSiFive:
Alternative / replaces #1804

dtzSiFive:
I'd like to add a test exhibiting what observable behavior differences this makes possible, stay tuned.

AbrarQuazi:
Would you be to add this into https://github.com/sifiveinc/wake/pull/1798 instead since it requires a federation release (updating runners)?

src/runtime/database.cpp
```diff
@@ -408,9 +406,9 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
       "  and j2.job_id<>?2"
       "  and (select coalesce(max(run_id), 0) from run_jobs where job_id=j2.job_id) <= ?1"
       ")";
-  const char *sql_fetch_hash = "select hash from files where path=? and modified=?";
+  const char *sql_fetch_hash = "select hash from files where path=? and modified=? limit 1";
   const char *sql_fetch_cached_path =
-      "select hash, type, mode from files where path=? and modified=?";
+      "select hash, type, mode from files where path=? and modified=? limit 1";
```
ag-eitilt, line 411: If we weren't already looking at moving `modified` into `filetree`, I'd say `files` would be better with a uniqueness constraint on `(path, modified)` rather than this `limit` -- no matter how unlikely it is to have different hashes at the same modified time.

src/runtime/database.cpp
```diff
@@ -1047,10 +1056,10 @@ Usage Database::reuse_job(const std::string &directory, const std::string &envir
       auto hash = rip_column(imp->get_tree, 1);
       auto type = rip_column(imp->get_tree, 2);
       long mode = sqlite3_column_int64(imp->get_tree, 3);
-      PathInfo vis{path, hash, type, mode};
+      PathInfo vis{path, hash, type, mode, /*modified=*/0};
 
       auto it = vis_hashes->find(path);
-      if (it == vis_hashes->end() || it->second != vis) {
+      if (it == vis_hashes->end() || !it->second.same_identity(vis)) {
```
ag-eitilt, line 1062: Maybe an explicit comment saying that job reuse is *deliberately* calculated without regard to the modtime?  That's reasonably clear looking at the code, but I'm not sure it would be quite as obvious to future programmers (including ourselves) who don't have the discussions fresh in their minds.

src/runtime/schema.h
```diff
@@ -43,8 +43,8 @@ inline const char *getWakeSchemaSQLTxn() {
          "  type      text    not null,"
          "  mode      integer not null,"
          "  modified  integer not null);"
-         "create unique index if not exists file_path_hash_type_mode on files(path, hash, type, "
-         "mode);"
+         "create unique index if not exists file_path_hash_type_mode_modified on"
+         "  files(path, hash, type, mode, modified);"
```
ag-eitilt, line 47: Going back to the previous comment, is the `unique` constraint really helping us over this number of fields?
```suggestion
         "create index if not exists file_path_hash_type_mode_modified on"
         "  files(path, hash, type, mode, modified);"
```
...with a separate `unique` index over a better-targetted selection? `(path, hash, modified)` or some such?

tools/fuse-waked/main.cpp
```diff
@@ -372,6 +375,10 @@ void Job::parse() {
         fprintf(stderr, "Visible entry '%s' missing 'mode'\n", path.c_str());
         continue;
       }
+      if (mtime_value.empty()) {
+        fprintf(stderr, "Visible entry '%s' missing 'mtime'\n", path.c_str());
+        continue;
+      }
```
ag-eitilt, line 381: Is using the `<<` form in `fuse.cpp` but the `fprintf` here meaningful?  They seem to have parallel but subtly different check/conversion code as well (`stol` vs `stoul`) and it *seems* like the kind of thing we should have a shared library/helpers for to keep in sync -- unless the two uses are just that very slightly incompatible in ways we can't reconcile.

Not for this PR, but thought it would be worth suggesting for future cleanup.

ag-eitilt (approved):
Couple suggestions, and like Abrar said, might be nice to roll into the first RC, but overall looks good.

src/runtime/database.cpp
dtzSiFive: I need to distill my (our) thinking about whether we want this in modified (and Path becomes file_id + ..., some queries are more complicated, ...) or if we want it in the files table.

No more flip flopping :).

BUT I think that actually the right answer -- for our mtime optimization -- is that if we DO have multiple hashes for a (path, mtime) pair we just insist the file be rehashed.

That is, if we are limiting ourselves to mtime.  If we extend our observations for "do we need to rehash" to other details, perhaps such as those used by git index for the same purpose: https://github.com/git/git/blob/867b1c1bf68363bcfd17667d6d4b9031fa6a1300/Documentation/technical/index-format.txt .

tools/fuse-waked/main.cpp
dtzSiFive: I followed the print format of the surrounding code.  The use of `stoul` is incorrect and was addressed when splitting this off for workspace virtualization directly: https://github.com/sifiveinc/wake/pull/1811 .  Good call, thanks!

I agree these code sections very much look like they should be factored out / DRY !

src/runtime/schema.h
dtzSiFive: I assumed we would want to support the same file/etc with different mode's, as unlikely as that may be.

I think at this point the unique index is really just to clarify intent -- no duplicate records.

src/runtime/database.cpp
dtzSiFive: Good call, will add it thanks!

dtzSiFive:
Thank you very much for the reviews, esp @ag-eitilt !!

Closing in favor of filetree version: #1821 .

PR #1807, opened 2026-04-23
Revert parseJSONFile API change and claim result jsons

AbrarQuazi:
Make parseJSONFIle accept Path objects again to keep its safety

### DIFF
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -557,7 +557,14 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
             Some "Non-zero exit status ({str runnerStatusCode}) for JSON runner on {specFile}".makeError
             | RunnerOutput Nil emptyStagedOutputs (specFile, Nil) runnerUsage
 
-        require Pass content = parseJSONFile resultFile
+        def claimResult =
+            claim resultFile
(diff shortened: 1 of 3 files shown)


ngraybeal (approved):

share/wake/lib/system/runner.wake
```diff
@@ -557,7 +557,14 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
             Some "Non-zero exit status ({str runnerStatusCode}) for JSON runner on {specFile}".makeError
             | RunnerOutput Nil emptyStagedOutputs (specFile, Nil) runnerUsage
 
-        require Pass content = parseJSONFile resultFile
+        def claimResult =
+            claim resultFile
+            | addErrorContext "Failed to claim {resultFile} for JSON runner: "
```
ag-eitilt, line 562: I feel like between this and the talk of cleaning the `spec-*`/`result-*` files, we might be leaning toward some concept of "runner files" as supplemental to "job files".  Definitely needs a lot more thought, though, since the entire idea behind those runner files are to handle data (e.g. IPC) *within* the runner so delaying hashing these until alongside rest of the paths is a non-starter.  Let's get the CAS settled, along with some of the other pending Runner changes, and see if it's any more attractive at that point.

ag-eitilt (approved):

PR #1810, opened 2026-04-23
change wake-migrate to support version 10 migration

AbrarQuazi:
Updating wake-migrate to support version 10 schema changes

### DIFF
--- a/tools/wake-migrate/main.cpp
+++ b/tools/wake-migrate/main.cpp
@@ -234,16 +234,122 @@ static std::vector<Migration> get_migrations() {
        },
        "Convert runner_status from INTEGER to TEXT"},
 
-      // Version 9 -> 10: Not supported. The schema changes to the files table require a fresh
-      // database. Users must delete wake.db and let wake recreate it.
+      // Version 9 -> 10: Add 'type' and 'mode' columns to the files table.
+      // Backfills type/mode by lstat()-ing every stored path; evicts jobs whose
+      // input or output files no longer exist on the workspace.
       {9, 10,
-       [](sqlite3*) -> bool {
-         std::cerr << "Migration from version 9 to 10 is not supported.\n"
-                   << "Please delete wake.db and run wake again to create a fresh database."
-                   << std::endl;
-         return false;
+       [](sqlite3* db) -> bool {
+         // Recreate files table with the new type and mode columns.
+         static const char* ddl[] = {
+             "create table files_new("
+             " file_id integer primary key, path text not null,"
+             " hash text not null, type text not null default '',"
+             " mode integer not null default 0, modified integer not null)",
+             "insert into files_new select file_id,path,hash,'',0,modified from files",
+             "drop table files",
+             "alter table files_new rename to files",
+             "create unique index filenames on files(path)",
+         };
+         for (const char* sql : ddl)
+           if (!exec_sql(db, sql)) return false;
+
+         // Cursor-based stat loop: process 1000 rows at a time so memory usage
+         // stays bounded even on very large databases.
+         const char* sel_sql =
+             "select file_id,path,hash from files where file_id>? order by file_id limit 1000";
+         const char* upd_sql = "update files set type=?,mode=? where file_id=?";
+         const char* del_jobs_sql =
+             "delete from jobs where job_id in"
+             " (select job_id from filetree where file_id=? and (access=1 or access=2))";
+
+         sqlite3_stmt *sel = nullptr, *upd = nullptr, *del_jobs = nullptr;
+         if (sqlite3_prepare_v2(db, sel_sql, -1, &sel, nullptr) != SQLITE_OK ||
+             sqlite3_prepare_v2(db, upd_sql, -1, &upd, nullptr) != SQLITE_OK ||
+             sqlite3_prepare_v2(db, del_jobs_sql, -1, &del_jobs, nullptr) != SQLITE_OK) {
+           std::cerr << "Migration 9->10: prepare failed: " << sqlite3_errmsg(db) << "\n";
+           sqlite3_finalize(sel);
+           sqlite3_finalize(upd);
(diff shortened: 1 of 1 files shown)


ngraybeal (approved):

tools/wake-migrate/main.cpp
```diff
@@ -234,16 +234,122 @@ static std::vector<Migration> get_migrations() {
...
-         std::cerr << "Migration from version 9 to 10 is not supported.\n"
-                   << "Please delete wake.db and run wake again to create a fresh database."
-                   << std::endl;
-         return false;
+       [](sqlite3* db) -> bool {
+         // Recreate files table with the new type and mode columns.
+         static const char* ddl[] = {
+             "create table files_new("
+             " file_id integer primary key, path text not null,"
+             " hash text not null, type text not null default '',"
```
ag-eitilt, line 246: Do we want to have `default ''`, or are the insertion commands sufficiently updated to be able to enforce at the schema level that they always provide a `type`?

Also, probably too late to change things, but I just realized we're using a `text` enum in the new table rather than an `integer` one; the latter would probably be *slightly* more optimized, but I doubt there's enough difference to matter.  Might be something to look at for the future, though.

tools/wake-migrate/main.cpp
```diff
@@ -234,16 +234,122 @@ static std::vector<Migration> get_migrations() {
...
+             " mode integer not null default 0, modified integer not null)",
+             "insert into files_new select file_id,path,hash,'',0,modified from files",
+             "drop table files",
+             "alter table files_new rename to files",
+             "create unique index filenames on files(path)",
```
ag-eitilt, line 251: Just double-checking: this is an index from the previous DB version, and not a new overly-restrictive `unique` constraint, right?

tools/wake-migrate/main.cpp
```diff
@@ -234,16 +234,122 @@ static std::vector<Migration> get_migrations() {
...
+             sqlite3_prepare_v2(db, del_jobs_sql, -1, &del_jobs, nullptr) != SQLITE_OK) {
+           std::cerr << "Migration 9->10: prepare failed: " << sqlite3_errmsg(db) << "\n";
+           sqlite3_finalize(sel);
+           sqlite3_finalize(upd);
+           sqlite3_finalize(del_jobs);
+           return false;
+         }
+
+         // Wake stores directories with a fixed all-zero hash (dirHash in path.wake).
+         static const std::string dir_hash(64, '0');
```
ag-eitilt, line 277: Similarly, we might want to look into storing the hash as an `integer`... though since the biggest integers in [SQLite are eight bytes](https://sqlite.org/datatype3.html#storage_classes_and_datatypes) we'd either need to truncate the hash or store it across four columns to get the 32 bytes required -- and at that point we'd be dealing with signedness conversions as well.  Still, even the four-columns representation would drop the storage size of the hash to a bit over half (assuming I'm [doing the math](https://sqlite.org/fileformat2.html#record_format) correctly, from 65 B to 36 B), which depending on the length of the path names would probably come to something vaguely like 15%-20% savings for that table.  Probably wasn't worth it when there was only a single entry per path name, but might be more attractive now that there's effectively an entry per `(path, hash)` pair, though we'll have to see how much history we wind up storing in practice and how much `files` sharing does wind up being possible between runs.

tools/wake-migrate/main.cpp
```diff
@@ -234,16 +234,122 @@ static std::vector<Migration> get_migrations() {
...
+           if (!exec_sql(db, sql)) return false;
+
+         // Cursor-based stat loop: process 1000 rows at a time so memory usage
+         // stays bounded even on very large databases.
+         const char* sel_sql =
+             "select file_id,path,hash from files where file_id>? order by file_id limit 1000";
+         const char* upd_sql = "update files set type=?,mode=? where file_id=?";
+         const char* del_jobs_sql =
+             "delete from jobs where job_id in"
+             " (select job_id from filetree where file_id=? and (access=1 or access=2))";
```
ag-eitilt, line 263: What's the benefit to having the `access` value restriction?  Wouldn't we want to remove *any* reference to missing files, no matter which side of the job they're on?  After all, if the job is missing output files, it would be rerun anyway so removing it preemptively doesn't seem like it would have any more invasive an impact.

tools/wake-migrate/main.cpp
```diff
@@ -234,16 +234,122 @@ static std::vector<Migration> get_migrations() {
...
+               sqlite3_reset(del_jobs);
+             }
+             if (!ok) break;
+           }
+         }
+
+         // Foreign Key cascade is off during migration, so deleting jobs above left
+         // orphaned rows in every table that references jobs(job_id).
+         // Sweep them now; otherwise filetree rows with access=2 cause the
+         // runtime to report "File output by multiple Jobs" on the next run.
```
ag-eitilt, line 329: Good catch, I hadn't thought about this.

ag-eitilt (approved):
Looks good!  All of my comments are basically just thoughts for potential future development.

Out of curiosity, how much time *were* you seeing this take in your tests?  I know we were seeing anywhere from 30 seconds to 3 minutes with just the naive `find` statting, and I'm wondering how much overhead the SQL/C++ did wind up adding (vs how much the known-`files` filter offset that overhead).

tools/wake-migrate/main.cpp
AbrarQuazi: Yeah we always ensure insertions go through the `PathType` enum check

tools/wake-migrate/main.cpp
AbrarQuazi: Yep its the same filenames index and exactly from schema.h. Have to do it again here because we DROP TABLE

tools/wake-migrate/main.cpp
AbrarQuazi: Doesnt seem like there is any real value here, you are right

PR #1811, opened 2026-04-24
Add modified time (mtime) to Path, send to FUSE.

dtzSiFive:
Plumb mtime into fuse-waked and use it to report
mtime for visible files.  Add test.

Test fails without this change with WAKE_CAS=1.

This test was written for multi-wake but
checks this behavior for normal wake too.


share/wake/lib/system/runner.wake
```diff
@@ -465,6 +465,7 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
                     "type" :-> JString (pathTypeToString p.getPathType),
                     "hash" :-> JString p.getPathHash,
                     "mode" :-> JInteger p.getPathMode,
+                    "mtime" :-> JInteger p.getPathMtimeNs,
                     Nil
```
AbrarQuazi, line 469: We will need to update the runners in our internal repo as well

AbrarQuazi (approved):
Thanks! LGTM

tests/runtime/mtime-preservation/pass.sh
```diff
@@ -0,0 +1,34 @@
...
+set -eu
+
+WAKE="${1:+$1/wake}"
+WAKE="${WAKE:-wake}"
+
+export WAKE_CAS=1
+
+rm -rf .build .fuse wake.db* wake.log output.txt result-a.txt result-b.txt
+
+echo "Fresh concurrent (if supported) runs:"
```
dtzSiFive, line 15: This might trip on #1792 until the workspace-virtualization branch includes that (probably by rebasing onto master again at some point).

PR #1812, opened 2026-04-24
test: set mtime and observe (broken!)

dtzSiFive:
Staging files are lacking story for maintaining mtime/atime (as appropriate) throughout.  Particularly mtime.

A basic test is included to show this behavior.

Marking as draft as this only contains test demonstrating the bug not the appropriate set of fixes.

It looks like we try to let the underlying staging file track mtime/atime updating, which seems reasonable?


dtzSiFive:
Fixed via linked PR ( https://github.com/sifiveinc/wake/pull/1814 ), test included there.

PR #1818, opened 2026-04-28
wake_spawn: set PR_SET_PDEATHSIG on spawned processes

dtzSiFive:
This is for multi-wake, but should go to master as well to fix stale wakebox processes after `wake` has been killed.

### DIFF
--- a/src/compat/spawn.c
+++ b/src/compat/spawn.c
@@ -23,10 +23,17 @@
 
 #include <sys/types.h>
 #include <unistd.h>
+#ifdef __linux__
+#include <signal.h>
+#include <sys/prctl.h>
+#endif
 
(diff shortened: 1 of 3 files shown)


ngraybeal (approved):
It probably makes more sense to do this by default than not. It's also safer. We can revisit if we need to expose this to Jobs and allow for persistent processes (ie a job is uploading data, so we should let it finish).

dtzSiFive:
> It probably makes more sense to do this by default than not. It's also safer. We can revisit if we need to expose this to Jobs and allow for persistent processes (ie a job is uploading data, so we should let it finish).

Awesome, great.  My thinking as well, thanks 👍 .

PR #1819, opened 2026-04-28
FUSE virtual directory stat update

ngraybeal:
FUSE:
Updates `st_size` for a directory type to be a standard 4KB size.
Updates `st_nlink` for a directory type is set to 1 to indicate wake FUSE does not follow normal conventions.

### DIFF
--- a/tools/fuse-waked/main.cpp
+++ b/tools/fuse-waked/main.cpp
@@ -724,6 +724,7 @@ static int wakefuse_getattr(const char *path, struct stat *stbuf) {
                           [stbuf](const StagedDirectoryData &d) {
                             // Return synthetic directory stat
                             memset(stbuf, 0, sizeof(*stbuf));
+                            stbuf->st_size = 4096;
                             stbuf->st_mode = S_IFDIR | (d.mode & 07777);
                             stbuf->st_nlink = 2;
                             stbuf->st_uid = getuid();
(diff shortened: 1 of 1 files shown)


dtzSiFive (commented):
Interesting that we don't set st_ino at all for these?

I went looking because we use nlink=1 and ino=0 for the job subdir root and overall fuse mount root as well.
Maybe we should set st_ino=0 for these (those changed in this PR) also?

That also means there's a bit of precedent for our use of `nlink=1`.

EDIT: Oh we memset it to zero so they are set that way.  Okay.

dtzSiFive (approved):
LGTM.  We set nlink=1 for some other bits (above), so there's precedent.

Bits from coreutils' du: https://github.com/coreutils/coreutils/blob/0f85d79ed60de25b7530af7bbf1a97523fbbfca2/src/du.c#L598

findutils' (didn't quickly find permalink):
> File systems which don't have these semantics should simply return a
> value less than 2 in the @code{st_nlinks} member of @code{struct stat}
> in response to a successful call to @code{stat}.
( https://cgit.git.savannah.gnu.org/cgit/findutils.git/tree/doc/find.texi )

Didn't look at other filesystems, so between those two I think this makes sense especially for st_nlink.  Fixing it might allow for optimization in the future.

Size is wrong (/misleading!) as reported here, but the idea is it's "less wrong" than reporting zero so works for me.

PR #1820, opened 2026-04-28
Add modified time (mtime) to Path, send to FUSE.

dtzSiFive:
Plumb mtime into fuse-waked and use it to report
mtime for visible files.  Add test.

Test fails without this change with WAKE_CAS=1.

This test was written for multi-wake but
checks this behavior for normal wake too.

#1811 but for multi-wake, since not in base workspace-virtualization yet.

Handling of multiple modified's will be follow-up.


dtzSiFive:
I plan to send this through unreviewed as this portion has been looked at a few times now and is expected to be part of workspace-virtualization "shortly".

PR #1821, opened 2026-04-28
Move modified into filetree (and add to Path!).

dtzSiFive:
No longer attempt to fetch mtime raw from workspace, it must be passed in (!).

Modified observed (recorded) originally for a job
persists and seeing otherwise same file does not alter this. This is used in materialization.

Explicitly do not consider modified as part of job reuse candidate selection.  Similar behavior to having jobs point to a record with a mutable mtime field, supports important use cases like git operations.

Must now check modified as well as file_id when matching on filetree.

get_hash and get_cached_path only return information for files that are recorded in filetree:
add_hash is not sufficient.

Require modified to match for critpath updates,
but not for file dep tracking (visualization).

Bump to schema 13.

----

Replaces #1806, incorporates feedback given there (🙏 🙏 ).

New version of #1804, with upstreamed changes for fuse-waked split out and minor tweaks:
* Migrate was overkill
* Don't require modified matches for dep tracking

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -251,6 +251,7 @@ def finalizeJob job inputs outputs cleanable reality runnerError =
             hash,
             pathTypeToString (getStagingItemPathType item),
             str (getStagingItemMode item),
+            str (getStagingItemMtimeNs item),
         )
 
     def preparedOutputStrings = map pstr preparedOutputStagingItems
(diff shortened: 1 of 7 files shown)


dtzSiFive:
Okay so storePath constructs a Path from the staged item /before/ `modified` reaches filetree .

This means we have `Path`s floating around with mtime not backed by DB but that's only true as long as they're not used by jobs and I don't think is a real problem or threat to conceptual integrity.  Worth knowing about, however.

ngraybeal (approved):

PR #1822, opened 2026-04-29
Produce error on materialization failure, fix job label on reused jobs

dtzSiFive:
See each commit.  Test presently fails with "multiple jobs produced output"-style error.
See the reference `stdout` for the new error produced instead.

The runtime change goes upstream (and separate PR!); it does mean we capture the label string when we wouldn't today, FWIW.

Leaving here for now since I'd rather other PR's got review attention.
Marking draft for that reason.


ngraybeal (approved):
If this causes a different error, it makes sense to just alert users to the root cause.

dtzSiFive:
Landing, welp.

PR #1824, opened 2026-04-29
Wake 49.0.0 Release

ngraybeal:


ag-eitilt (approved):
Tests work locally, so I'm guessing the failure is flaky.

The `lib/wake/wake-stage{,.*}` binaries aren't actually in `.gitignore` and so clutter `git status`, but that will only affect the few people building Wake itself and so is not an issue we'll need to handle for the release.  Somewhat more concerningly, I ran into `.wake-build-lock` not getting removed somehow and subsequent runs never getting past the "database locked" warning until I removed it manually, but I've not been able to replicate that (it does indeed get cleaned up on standard Ctrl-C), so I don't think that's a *pressing* issue -- I'll keep trying to trigger that again and if I can figure it out we can roll that into a bugfix release.

dtzSiFive:
> The `lib/wake/wake-stage{,.*}` binaries aren't actually in `.gitignore` and so clutter `git status`, but that will only affect the few people building Wake itself and so is not an issue we'll need to handle for the release. Somewhat more concerningly, I ran into `.wake-build-lock` not getting removed somehow and subsequent runs never getting past the "database locked" warning until I removed it manually, but I've not been able to replicate that (it does indeed get cleaned up on standard Ctrl-C), so I don't think that's a _pressing_ issue -- I'll keep trying to trigger that again and if I can figure it out we can roll that into a bugfix release.

Between lack of #1818 on master and not setting O_CLOEXEC on the build lock (for better or for worse), I wonder if .....

Oh did you say /database locked/? 😢 .

Yikes well let's keep an eye on that.

ag-eitilt:
> Oh did you say /database locked/? 😢 .
>
> Yikes well let's keep an eye on that.

Pretty sure this was after I had removed the `wake.db` entirely, so at a wild guess it's something seeing the lockfile but *not* seeing any corresponding run entry in the table.  Definitely an edge case, it's just figuring out *which* edge.

...and then printing "database locked" due to the error message being a bit more abstract/out of date than the actual logic that's triggering it being printed.

PR #1835, opened 2026-05-04
Add --ps command to show currently running jobs; add --active capture filter.

dtzSiFive:
Sharing early, still needs more self-review.

For transparency but not responsibility, written with help from Claude Sonnet 4.6.

Idea is as discussed a few times:
* Set starttime when starting a job so have that information available.  Today this is committed as part of finish_job.
* Query DB For jobs that have started that haven't finished and are part of runs that aren't known to have finished.
* Present this information to the user.

See commit message for `--active`; need to sort out `--canceled` and then have a story for these together.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -103,6 +103,8 @@ struct Database::detail {
   sqlite3_stmt *set_run_end_time;
   sqlite3_stmt *get_incomplete_runs;
   sqlite3_stmt *clear_run_files;
+  sqlite3_stmt *set_starttime;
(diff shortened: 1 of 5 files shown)


dtzSiFive:
The `--active` filter here is trash, will build this up on top of #1837 : this lays groundwork for determining which runs are active for use in RO queries / scenarios.

This can be used to create focused queries that fit into our `matching` structure without excessive complication and make this work quite a bit faster in practice.

dtzSiFive:
Will open new PR(s).

Issue #1593, opened 2024-07-10
Overwriting a hardlinked file greatly broadens permissions


ag-eitilt:
Split from #1589: If a Wake job overwrites a file which has been hardlinked to elsewhere, the known output is delinked and overwritten as expected.  However, the remaining copy of the inode suddenly gains `777` permissions.

```console
$ wake -x 'makePlan "test" Nil "echo -n test > test.txt" | runJob' ; ln test.txt test2.txt ; chmod -w test2.txt ; wake -x 'makePlan "test" Nil "echo -n test with different contents > test.txt" | runJob'
Job 9666
Job 9679
$ ls -l test.txt test2.txt
-rwxrwxrwx 1 sammay sammay  4 Jul  8 12:48 test2.txt
-rw-r--r-- 1 sammay sammay 28 Jul  8 12:48 test.txt
```

Our guess during the meeting is that this is a problem with our FUSE sandboxing, likely somewhere in `deep_unlink` and *specifically* if the file isn't writable, trying to deinitialize the inode without realizing that it's still accessible.  So far as we can tell this doesn't provide a vector for permission escalation attacks as the Linux permission system would block any attempt to change permissions on files which the user running Wake wouldn't already have access to, but it is technically possible to construct a script which calls `ln [...]/target.file .` a victim *into* a Wake workspace and deliberately overwrite that to expose the original -- it's simply also possible to `chmod 777 [...]/target.file` directly.

Our determination is that this is definitely a problem which needs to be fixed, but isn't a *critical* problem to drop everything for.  It also points to the fact that we likely shouldn't be rolling our own FUSE implementation, and that switching that whole subsystem out for overlayfs would eliminate a whole category of potential bugs lurking in edge cases such as this.

PR #1836, opened 2026-05-05
fuse-waked: for CAS, hardlinks backed by hardlinks, share virtual mode.

dtzSiFive:
Modify hardlink test to run with CAS enabled, which before this change would cause it to fail (cannot write through hardlinks).

Extend test to check mode changes on one are observed on the other.

### DIFF
--- a/tests/wakebox/hardlink/pass.sh
+++ b/tests/wakebox/hardlink/pass.sh
@@ -5,6 +5,6 @@ export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
 
 trap 'rm link_file_src.txt link_file_dst.txt' EXIT
 
-STDOUT=$(${1}/wakebox -p input.json)
+export WAKE_CAS=1
 
-[ "$STDOUT" = "pass" ] || exit 1
+${1}/wakebox -p input.json
--- a/tools/fuse-waked/main.cpp
+++ b/tools/fuse-waked/main.cpp
@@ -79,7 +79,9 @@ overloaded(Ts...) -> overloaded<Ts...>;
 // Staged item types - written to .cas/staging/ during job execution, hashed by wakebox after
 struct StagedFileData {
   std::string staging_path;
-  mode_t mode;
+  std::shared_ptr<mode_t> mode;
+
+  bool is_hardlink() const { return mode.use_count() > 1; }
 };
 
 struct StagedSymlinkData {
@@ -92,15 +94,7 @@ struct StagedDirectoryData {
   struct timespec mtime;
 };
 
-// Hardlinks are not true filesystem hardlinks. On link(), we reflink/copy the source's staging
-// file so each entry owns an independent staging_path.
-struct StagedHardlinkData {
-  std::string staging_path;
-  mode_t mode;
-};
-
-using StagedItemData =
-    std::variant<StagedFileData, StagedSymlinkData, StagedDirectoryData, StagedHardlinkData>;
+using StagedItemData = std::variant<StagedFileData, StagedSymlinkData, StagedDirectoryData>;
 
 struct StagedItem {
   std::string dest_path;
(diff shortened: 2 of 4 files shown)


tools/fuse-waked/main.cpp
```diff
@@ -1630,6 +1596,8 @@ static int wakefuse_link(const char *from, const char *to) {
 
   if (it->second.is_visible(keyt.second)) return -EEXIST;
 
+  if (it->second.is_visible(keyf.second)) return -EACCES;
```
dtzSiFive, line 1599: Drive-by: ban hardlinks originating from visible files for lack of structure presently to enforce RO access this way.

Fixes existing bug where you can modify visible files through a hardlink.

Wake snippet demonstrating:
```
export def datFile Unit =
    write "datFile" "hello world"

export def modLink Unit =
    require Pass dat =
        datFile Unit
    makeExecPlan ("sh", "-c", "ln datFile modLink; echo asdf >> modLink; cat datFile", Nil) (dat, Nil)
    | setPlanLabel "test (hardlink): modLink"
    | setPlanShare False
    | runJobWith defaultRunner
    | getJobOutput

export def test _ =
  require Pass out = modLink Unit
  Pass Unit
```

Note that since we -- before this PR -- block writes through hardlinks in CAS mode, this issue today only exists in non-CAS wake.

cc #1593 .

dtzSiFive: This change might deny benign jobs from performing reasonable behavior and should be considered carefully and perhaps separately 👍 .

tools/fuse-waked/main.cpp
```diff
@@ -1641,31 +1609,25 @@ static int wakefuse_link(const char *from, const char *to) {
...
       g_staged_files.insert(keyt.first, keyt.second, std::move(sf));
 
       it->second.staged_paths.insert(keyt.second);
       it->second.files_wrote.insert(keyt.second);
-      // Both hardlink paths need direct_io to prevent kernel caching issues
-      hardlinks.insert(std::string(from));
-      hardlinks.insert(std::string(to));
       return 0;
     }
+    return -EEXIST;
```
dtzSiFive, line 1630: In CAS mode: never fallback to workspace!

This may be unreachable, it should be, but just be explicit here to avoid surprise behaviors while supporting both.

dtzSiFive:
Dare we set st_nlink based on this? 😉

tests/wakebox/hardlink/pass.sh
```diff
@@ -5,6 +5,6 @@ export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
 
 trap 'rm link_file_src.txt link_file_dst.txt' EXIT
 
-STDOUT=$(${1}/wakebox -p input.json)
+export WAKE_CAS=1
```
ag-eitilt, line 8: Probably don't want to hardcode this in the test?  I know we were saying that we'd rather push toward getting everything unified rather than supporting multiple code paths, but `make test ; WAKE_CAS=1 make test` should still leave us testing both paths over the full suite.

tools/fuse-waked/main.cpp
AbrarQuazi: Would ENONET or EPERM be a better return for if a file is not staged?

tools/fuse-waked/main.cpp
```diff
@@ -79,7 +79,9 @@ overloaded(Ts...) -> overloaded<Ts...>;
 // Staged item types - written to .cas/staging/ during job execution, hashed by wakebox after
 struct StagedFileData {
   std::string staging_path;
-  mode_t mode;
+  std::shared_ptr<mode_t> mode;
+
+  bool is_hardlink() const { return mode.use_count() > 1; }
 };
```
ag-eitilt, line 85: I don't like this C++ design choice... `struct` should be *data*, and if you want methods you reach for `class`/`interface`.  But that's me getting grumpy at the language, not at this implementation.

tools/fuse-waked/main.cpp
AbrarQuazi: Yeah I think this change is right, but don't know if there is any fallout due to this, so agreed that maybe we test and evaluate this in a separate PR

tools/fuse-waked/main.cpp
```diff
@@ -1641,31 +1609,25 @@ static int wakefuse_link(const char *from, const char *to) {
...
       sf.job_id = keyf.first;
-      sf.data = StagedHardlinkData{new_staging_path, src_mode};
+      sf.data = StagedFileData{new_staging_path, src_sf->mode};
       g_staged_files.insert(keyt.first, keyt.second, std::move(sf));
 
       it->second.staged_paths.insert(keyt.second);
       it->second.files_wrote.insert(keyt.second);
-      // Both hardlink paths need direct_io to prevent kernel caching issues
-      hardlinks.insert(std::string(from));
-      hardlinks.insert(std::string(to));
```
AbrarQuazi: Can the `hardlinks` set be completely removed? Or its still needed because the non-CAS mode still depends on it? If thats the case, maybe add a comment to remove this structure

ag-eitilt (approved):
This all looks good to me!

AbrarQuazi (approved):
LGTM, thanks!

We should merge this to master and then maybe cherry pick to the feature/multi-wake branch so we can get to a beta release faster

tests/wakebox/hardlink/pass.sh
ag-eitilt: Ah, I see what you're getting at with this guard -- this *is* a breaking change between CAS and non-CAS.  Can't say I like the different behaviour, but it's unlikely enough to not cause any issue.  Instead of setting `WAKE_CAS`, though, can we instead *check* it, print a stderr warning about skipping the test, and then `exit 0`?  (Also applies to https://github.com/sifiveinc/wake/pull/1838)

tools/fuse-waked/main.cpp
dtzSiFive: Actually, I think today attempts to hardlink visible "succeed" but fall through and don't do this in the staging directory but write through to workspace? The new `-EEXIST` check prevents this, as does the tentative `is_visible` explicit guard discussed here.

tests/wakebox/hardlink/pass.sh
dtzSiFive: Mainline WAKE doesn't propagate WAKE_CAS=1 into the tests, whether this is a bug or not I am not sure.  If I was certain I'd have addressed it.

But okay that works for me.  I removed it in the multi-wake variant.

Good call!

tools/fuse-waked/main.cpp
dtzSiFive: It's for non-CAS.  I'll add a comment.

tools/fuse-waked/main.cpp
dtzSiFive: I'll hoist it out to `is_hardlink` on StagedItem, that fits the current organization/style better anyway.  Meant to revisit this, thanks!

dtzSiFive:
Mixed signals on where we want to send these changes hahaha.

Applied y'all's feedback to multi-wake version.  I'll sync those changes back here later.

PR #1838, opened 2026-05-05
fuse-waked: for CAS, hardlinks backed by hardlinks, share virtual mode.

dtzSiFive:
Modify hardlink test to run with CAS enabled, which before this change would cause it to fail (cannot write through hardlinks).

Extend test to check mode changes on one are observed on the other.

Rebased (hardlink test was already updated to accommodate CAS behavior) version of #1836 for multi-wake branch directly.

### DIFF
--- a/tests/wakebox/hardlink/input.json
+++ b/tests/wakebox/hardlink/input.json
@@ -2,7 +2,7 @@
   "command": [
     "/bin/sh",
     "-c",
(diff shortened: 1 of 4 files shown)


ag-eitilt (approved):
Guess the approval I gave on the other PR should have been given here instead. 😁

dtzSiFive:
Thanks!

PR #1839, opened 2026-05-06
wake: fix --clean: don't attempt vacuum from within transaction.

dtzSiFive:
Touchup clean-safety test to check stdout and especially stderr.

Require no error output, which would fail before this change.

### DIFF
--- a/tests/runtime/clean-safety/stdout
+++ b/tests/runtime/clean-safety/stdout
@@ -1,0 +1,2 @@
+PASS: --clean refused while build active
+PASS: --clean succeeded after reaping dead run
--- a/tools/wake/main.cpp
+++ b/tools/wake/main.cpp
@@ -712,12 +712,6 @@ int main(int argc, char **argv) {
         }
       }
 
-      // Full vacuum, we likely just deleted most of the database.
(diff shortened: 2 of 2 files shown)


ngraybeal (approved):

ag-eitilt (approved):
Can't add this as a comment for some reason, but:

```c++
    // Blocking checkpoint, truncate the log file.
    db.checkpoint(/*blocking=*/true);
```
The error message is just the vaccuum.  Does this have to happen *after* the vaccuum (i.e. is it checkpointing the empty DB pages?), and if not can it still occur within the transaction?

Either way, that doesn't seem like a critical issue to me, so 🚀 .

dtzSiFive:
> The error message is just the vaccuum. Does this have to happen _after_ the vaccuum (i.e. is it checkpointing the empty DB pages?), and if not can it still occur within the transaction?
> 
> Either way, that doesn't seem like a critical issue to me, so 🚀 .

It needs to happen after the vacuum -- vacuum will create a giant WAL file and we'd like to truncate it while visiting.  Yeah it's what actually reclaims the free pages!

dtzSiFive:
Thanks, folks!

PR #1840, opened 2026-05-06
wake: record starttime early, add capture filters; fix --canceled

dtzSiFive:
Record starttime eagerly via start_job() at fork time rather than writing it in finish_job, so queued jobs (starttime==0) are distinguishable from running ones.

Also allows showing how long jobs have actually been running.

Add --active, --queued, --in-flight capture filters, and fix --canceled (now only jobs not part of live runs).

Tweak canceled (and all of these) to determine "unfinished" as "stat_id is null", which means finish_job was not called.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -1987,10 +1987,10 @@ std::vector<JobReflection> Database::matching(
   // This query creates a subtable of the following shape:
   //
   // clang-format off
-  // | job_id | label | run_id | endtime | commandline | runner_status | status | runtime |       tags       |
-  // -----------------------------------------------------------------------------------------------------------
-  // |    1   |  foo  |   1    |  1234   | ls lah .    | NULL          |   0    |   2.8   | <d>a=b<d>c=d<d>  |
-  // |    2   |  bar  |   1    |  0000   | cat f.txt   | "Job failed"  |   0    |   0.0   |      null        |
+  // | job_id | label | run_id | starttime | endtime | stat_id | commandline | runner_status | status | runtime |       tags       |
+  // -------------------------------------------------------------------------------------------------------------------------------
+  // |    1   |  foo  |   1    |   12340   |  12350  |   42    | ls lah .    | NULL          |   0    |   2.8   | <d>a=b<d>c=d<d>  |
+  // |    2   |  bar  |   1    |   0       |  0      |  NULL   | cat f.txt   | "Job failed"  |   0    |   0.0   |      null        |
   // clang-format on
   //
   // The subtable is constructed by joining the jobs table with the minimal set of other dependent
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -103,6 +103,7 @@ struct Database::detail {
   sqlite3_stmt *set_run_end_time;
   sqlite3_stmt *get_incomplete_runs;
   sqlite3_stmt *clear_run_files;
+  sqlite3_stmt *set_starttime;
 
   long run_id;
   long gc_watermark;
@@ -159,6 +160,7 @@ struct Database::detail {
         set_run_end_time(0),
         get_incomplete_runs(0),
         clear_run_files(0),
+        set_starttime(0),
         run_id(0),
         gc_watermark(0) {}
 };
@@ -387,8 +389,7 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_add_stats =
       "insert into stats(hashcode, status, runtime, cputime, membytes, ibytes, obytes)"
       " values(?, ?, ?, ?, ?, ?, ?)";
-  const char *sql_link_stats =
-      "update jobs set stat_id=?, starttime=?, endtime=?, keep=? where job_id=?";
+  const char *sql_link_stats = "update jobs set stat_id=?, endtime=?, keep=? where job_id=?";
   const char *sql_detect_overlap =
       "select f1.path, t2.job_id from filetree t1, filetree t2, files f1, files f2, run_jobs rj"
       " where t1.job_id=?1 and t1.access=2"
@@ -488,6 +489,7 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_set_run_end_time = "update runs set end_time = ? where run_id = ?";
   const char *sql_get_incomplete_runs = "select run_id, time from runs where end_time is null";
   const char *sql_clear_run_files = "delete from run_files where run_id = ?";
+  const char *sql_set_starttime = "update jobs set starttime=? where job_id=?";
 
 #define PREPARE(sql, member)                                                                     \
   ret = sqlite3_prepare_v2(imp->db, sql, -1, &imp->member, 0);                                   \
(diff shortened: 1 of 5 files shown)


src/runtime/database.cpp
```diff
@@ -387,8 +391,7 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_add_stats =
       "insert into stats(hashcode, status, runtime, cputime, membytes, ibytes, obytes)"
       " values(?, ?, ?, ?, ?, ?, ?)";
-  const char *sql_link_stats =
-      "update jobs set stat_id=?, starttime=?, endtime=?, keep=? where job_id=?";
+  const char *sql_link_stats = "update jobs set stat_id=?, endtime=?, keep=? where job_id=?";
```
ag-eitilt, line 392: This is an interesting question about `keep`.  It's something we *could* insert at the start, but by waiting until here to update it, this is functioning as a double lock where even if we don't test any other columns we'll see (I assume) `ReRun` until the job completes and writes the actual persistence.  I like it, I'm just not sure if that was a deliberate decision.

Let's just be aware of this for `ps` and similar live reporting.

src/runtime/database.cpp
```diff
@@ -488,6 +491,14 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
...
   const char *sql_get_incomplete_runs = "select run_id, time from runs where end_time is null";
   const char *sql_clear_run_files = "delete from run_files where run_id = ?";
+  const char *sql_set_starttime = "update jobs set starttime=? where job_id=?";
+  const char *sql_get_open_run_jobs =
+      "select rj.run_id, j.job_id, j.label, j.starttime"
+      " from run_jobs rj"
+      " join jobs j on rj.job_id = j.job_id"
+      " join runs r on rj.run_id = r.run_id"
+      " where r.end_time is null and j.endtime == 0"
+      " order by rj.run_id, j.starttime, j.job_id";
```
ag-eitilt: Is there any order-of-operations optimization needed by filtering `jobs` and `runs` before merging?  Or is SQLite smart enough to realize that `r.end_time` and `j.endtime` are independent and float those tests out before joining the full tables?

ag-eitilt (approved):
To be honest, only reviewed this at a "that looks right", level, but that all does look right.

src/runtime/database.cpp
dtzSiFive: FWIW:
```
sqlite> explain query plan select rj.run_id, j.job_id, j.label, j.starttime from run_jobs rj join jobs j on rj.job_id=j.job_id join runs r on rj.run_id=r.run_id where r.end_time is null and j.endtime == 0 order by rj.run_id, j.starttime, j.job_id;
QUERY PLAN
|--SCAN TABLE run_jobs AS rj USING COVERING INDEX run_jobs_by_run
|--SEARCH TABLE jobs AS j USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE runs AS r USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR RIGHT PART OF ORDER BY
sqlite>
```

I think this is pretty good, with ordering being done "manually" but I think for sizes involved that's totally fine (and something we'd do ourselves if didn't ask sqlite to do!).

In practice it seems very fast, FWIW.  Not tested on pathological.. but I did get a detail wrong early and could tell haha.

Thanks! I'll double-check this before landing.

dtzSiFive: I think we're missing indexes to filter first (efficiently) presently, I'll poke at this tomorrow.

(I don't think it matters "in practice" / worth "cost", will explore 👍 )

dtzSiFive: Partial index on runs does the trick:

```
sqlite> create index active_runs on runs(end_time) where end_time is null;
sqlite> explain query plan   WITH live_runs AS (SELECT run_id FROM runs WHERE end_time IS NULL)
   ...>   SELECT rj.run_id, j.job_id, j.label, j.starttime
   ...>   FROM live_runs r
   ...>   JOIN run_jobs rj ON r.run_id = rj.run_id
   ...>   JOIN jobs j ON rj.job_id = j.job_id
   ...>   WHERE j.endtime = 0
   ...>   ORDER BY rj.run_id, j.starttime, j.job_id;
QUERY PLAN
|--SEARCH TABLE runs USING COVERING INDEX active_runs (end_time=?)
|--SEARCH TABLE run_jobs AS rj USING COVERING INDEX run_jobs_by_run (run_id=?)
|--SEARCH TABLE jobs AS j USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR ORDER BY
sqlite> explain query plan select rj.run_id, j.job_id, j.label, j.starttime from run_jobs rj join jobs j on rj.job_id=j.job_id join runs r on rj.run_id=r.run_id where r.end_time is null and j.endtime == 0 order by rj.run_id, j.starttime, j.job_id;
QUERY PLAN
|--SEARCH TABLE runs AS r USING COVERING INDEX active_runs (end_time=?)
|--SEARCH TABLE run_jobs AS rj USING COVERING INDEX run_jobs_by_run (run_id=?)
|--SEARCH TABLE jobs AS j USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR ORDER BY
```

But we lose `--USE TEMP B-TREE FOR RIGHT PART OF ORDER BY` (previously could rely on run_id's already being in sorted order) -- however given the final result size (compared to possible size of the rest) I think that's less than insignificant.

tools/wake/main.cpp
```diff
@@ -210,6 +210,41 @@ void query_runs(Database &db) {
...
+  }
+  return live;
+}
+
+// Build query for job_ids from live set, checking for "in" set or not.
+// The optional "running" argument controls whether all (nullopt),
+// running (true, starttime!=0), or queued (false, starttime==0).
+std::string job_id_filter(bool in, const std::vector<OpenRunJobReflection> &live,
+                          std::optional<bool> running = std::nullopt) {
+  std::string q = in ? "job_id in (" : "job_id not in (";
```
dtzSiFive: This may be untenable for very large in-flight job sets.

I'll move some of this to `--ps`; instead these filters will use run_id as only need to check originating run for completion or not.

src/runtime/database.cpp
dtzSiFive: This will be part of #1842 .  Thank you!

src/runtime/database.cpp
```diff
@@ -2010,6 +2013,7 @@ std::vector<JobReflection> Database::matching(
                 j.job_id,
                 j.label,
                 j.run_id,
+                j.starttime,
```
dtzSiFive, line 2016: Expose starttime for use in filters.

ngraybeal (approved):
Excited to have this! 

Should we consider creating a separate field entirely for queue start time? Otherwise we lose that insight once the job actually is started. It could be useful for identifying bottlenecks.

dtzSiFive:
> Should we consider creating a separate field entirely for queue start time? Otherwise we lose that insight once the job actually is started. It could be useful for identifying bottlenecks.

Good idea! It'd be nice to show how long things have been queued for, for example.  And recording this for reporting and later inspection certainly has value.

FWIW we're not losing it, we just don't track it.

Before this PR, we only set `starttime` in `finish_job`.  Now we set it when, well, the job starts.
Before it starts, starttime is 0 (the default/sentinel).

AbrarQuazi (approved):
LGTM, thanks!

PR #1842, opened 2026-05-06
wake: add --ps command to show currently running jobs

dtzSiFive:
Lists jobs grouped by run_id, showing elapsed time for running jobs and [queued] for jobs not yet forked. Uses get_live_open_run_jobs() so crashed-but-unreaped runs are excluded.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -104,6 +104,7 @@ struct Database::detail {
   sqlite3_stmt *get_incomplete_runs;
   sqlite3_stmt *clear_run_files;
   sqlite3_stmt *set_starttime;
+  sqlite3_stmt *get_open_run_jobs;
 
   long run_id;
   long gc_watermark;
@@ -161,6 +162,7 @@ struct Database::detail {
         get_incomplete_runs(0),
         clear_run_files(0),
         set_starttime(0),
+        get_open_run_jobs(0),
         run_id(0),
         gc_watermark(0) {}
 };
@@ -490,6 +492,13 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_get_incomplete_runs = "select run_id, time from runs where end_time is null";
   const char *sql_clear_run_files = "delete from run_files where run_id = ?";
   const char *sql_set_starttime = "update jobs set starttime=? where job_id=?";
+  const char *sql_get_open_run_jobs =
+      "select rj.run_id, j.job_id, j.label, j.starttime"
+      " from run_jobs rj"
+      " join jobs j on rj.job_id = j.job_id"
+      " join runs r on rj.run_id = r.run_id"
+      " where r.end_time is null and j.stat_id is null"
+      " order by rj.run_id, j.starttime, j.job_id";
 
 #define PREPARE(sql, member)                                                                     \
   ret = sqlite3_prepare_v2(imp->db, sql, -1, &imp->member, 0);                                   \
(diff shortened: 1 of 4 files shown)


src/runtime/schema.h
```diff
@@ -36,6 +36,7 @@ inline const char *getWakeSchemaSQLTxn() {
          "  time     integer not null,"
          "  cmdline  text    not null,"
          "  end_time integer);"
+         "create index if not exists active_runs on runs(end_time) where end_time is null;"
```
dtzSiFive: See https://github.com/sifiveinc/wake/pull/1840#discussion_r3197745262 for discussion.  Produces improved plan that is expected to scale much better to large databases with many runs, by filtering down to "open" runs first.

h/t to @ag-eitilt for asking the right questions! 👍 

Electing to not bump schema version for this:

1) Not required,
2) We apply schema unconditionally

I could be convinced otherwise.

ngraybeal (approved):
High level, looks great to me. I think `--ps` compliments the other flags (--active) in [PR](https://github.com/sifiveinc/wake/pull/1840) well in how it presents the data in a more concise form. 

It'd be neat to have like a `--top` that could be TUI and have a refresh rate. Not sure at some point we need to be building separate tools.

src/runtime/schema.h
dtzSiFive: Dropped this, modified query to check `stat_id` instead, and `jobstat` index covers this.  New query plan:
```
sqlite>   EXPLAIN QUERY PLAN
   ...>   select rj.run_id, j.job_id, j.label, j.starttime
   ...>    from run_jobs rj
   ...>    join jobs j on rj.job_id = j.job_id
   ...>    join runs r on rj.run_id = r.run_id
   ...>    where r.end_time is null and j.stat_id is null
   ...>    order by rj.run_id, j.starttime, j.job_id;
QUERY PLAN
|--SEARCH TABLE jobs AS j USING INDEX jobstats (stat_id=?)
|--SEARCH TABLE run_jobs AS rj USING COVERING INDEX sqlite_autoindex_run_jobs_1 (job_id=?)
|--SEARCH TABLE runs AS r USING INTEGER PRIMARY KEY (rowid=?)
`--USE TEMP B-TREE FOR ORDER BY
```

Filtering down to "unfinished" jobs immediately seems way better anyway!

dtzSiFive:
Updated to integrate/use the capture filter options, now hidden jobs are hidden by default.

https://github.com/sifiveinc/wake/pull/1842/commits/851436256ee23b39268c5e3ef4a2f8748a206665 .

AbrarQuazi (approved):
LGTM.

Just commenting for the future, if we see value/want to add PID into the DB or not for a run so they show up for `--ps`

dtzSiFive:
Significant performance regression, unfortunately.  Looking into it...

Went from virtually instant to taking over half a second!!!

src/runtime/database.cpp
```diff
@@ -2034,33 +2013,51 @@ static std::string build_matching_id_query(
...
-  // This query wraps the subtable, applies the requested filters, and returns the matching jobs
+  // This query selects from the subtable, applies the requested filters,
+  // and returns matching job_ids.
   std::string id_query =
       "    SELECT core.job_id\n"
-      "    FROM\n"
-      "    (\n" +
-      subtable + "    )";
+      "    FROM\n" +
+      subtable;
```
dtzSiFive, line 2046: Dropped the () here so can reach for `core.job_id` in the hidden filter.

PR #1843, opened 2026-05-06
Minimize use of re2::StringPiece.

dtzSiFive:
This is std::string_view.

Only use it where necessary
to interface with re2::Match.

### DIFF
--- a/src/runtime/value.cpp
+++ b/src/runtime/value.cpp
@@ -303,11 +303,11 @@ MyOpts::MyOpts() {
 
 static MyOpts opts;
 
-RegExp::RegExp(Heap &h, const re2::StringPiece &regexp)
+RegExp::RegExp(Heap &h, std::string_view regexp)
     : Parent(h),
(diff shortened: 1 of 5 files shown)


src/runtime/value.cpp
```diff
@@ -303,11 +303,11 @@ MyOpts::MyOpts() {
 
 static MyOpts opts;
 
-RegExp::RegExp(Heap &h, const re2::StringPiece &regexp)
+RegExp::RegExp(Heap &h, std::string_view regexp)
     : Parent(h),
       exp(std::make_shared<RE2>(has_set_dot_nl<RE2::Options>::value
-                                    ? re2::StringPiece(regexp)
-                                    : re2::StringPiece("(?s)" + regexp.as_string()),
+                                    ? std::string(regexp)
```
dtzSiFive, line 309: RE2's constructor takes non-owning strings (e.g., StringPiece) but stores them locally, so this is safe.

AbrarQuazi (approved):

PR #1850, opened 2026-05-13
database: add index to speed up revtop query.

dtzSiFive:
Should replace filesearch index.

Queue'ing for next schema change, so don't have both indexes.


dtzSiFive:
Closing in favor of #1857 .

PR #1852, opened 2026-05-13
multi-wake: CAS GC implementation

dtzSiFive:
Add machinery to safely remove blobs from the CAS.

Invoke this at end of runs, to keep CAS tidy.

Delete orphaned file records in `Database::clean()` as well.


tools/wake/main.cpp
```diff
@@ -1192,5 +1232,9 @@ int main(int argc, char **argv) {
 
   db.finish_run();
   db.clean();
+
+  if (!gc_dead_cas_blobs(*cas_ctx.get_store(), db))
+    std::cerr << "CAS GC encountered errors (build result unaffected)" << std::endl;
```
ngraybeal: Should this be `run` and not `build`? Or do we use them interchangeably?

src/runtime/database.cpp
```diff
@@ -490,6 +498,28 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_get_incomplete_runs = "select run_id, time from runs where end_time is null";
   const char *sql_clear_run_files = "delete from run_files where run_id = ?";
   const char *sql_set_starttime = "update jobs set starttime=? where job_id=?";
+  const char *sql_get_dead_hashes =
+      "with disk_batch(hash) as ("
+      " VALUES "
+      "   (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?),"
```
ngraybeal, line 504: This is something... 🧐

src/cas/cas.cpp
```diff
@@ -305,4 +305,35 @@ wcl::result<bool, CASError> Cas::materialize_blob(const ContentHash& hash,
...
+  if (fs::is_directory(blobs_dir_, ec)) {
+    for (const auto& prefix_entry : fs::directory_iterator(blobs_dir_, ec)) {
+      if (ec || !prefix_entry.is_directory()) continue;
+      std::string prefix = prefix_entry.path().filename().string();
+      if (prefix.size() != 2) continue;
+
+      for (const auto& blob_entry : fs::directory_iterator(prefix_entry.path(), ec)) {
+        if (ec || !blob_entry.is_regular_file()) continue;
+        std::string suffix = blob_entry.path().filename().string();
+        if (suffix.size() != 62) continue;
```
ngraybeal: What are these hard fixed numbers? I assume how we broke up our CAS hashing?

dtzSiFive: That's right.  They are a bit "magic", don't disagree.

tools/wake/main.cpp
dtzSiFive: Good call, fixed thanks!

src/runtime/database.cpp
AbrarQuazi: Maybe create this string programatically in a loop, and feed it in?

dtzSiFive: We could do this but I kinda hate manipulating query strings and would prefer this as a string literal.

And would like it to be a compile-time constant as well, just because it is.

Is there a problem with this as-is, since we already have it? It's 16x16 if that wasn't clear.

tools/wake/main.cpp
```diff
@@ -453,6 +453,46 @@ static void cleanup_stale_staging(const std::string &staging_dir) {
...
+//   A CAS blob is only written to disk AFTER its `files` row is committed, and
+//   any in-flight file not yet recorded in `filetree` is guarded by a `run_files`
+//   row for the current run. `finish_run()` clears this run's `run_files` entries
+//   only after all `finish_job`s have committed their `filetree` rows, so the
+//   set "reachable from files via filetree or run_files" is never empty for a
+//   live file in the window where another wake could race.
+//
+// From this it follows that: a hash present on disk but absent from `files`
+// is truly dead. Both `delete_orphan_files` (DB side) and this routine
+// (filesystem side) rely on that. Inverting the write order (blob-before-row)
```
AbrarQuazi, line 468: I should have put a similar comment on the job finalization code in `job.wake`. Would you be able to put a similar comment on this PR there?

tools/wake/main.cpp
```diff
@@ -1192,5 +1232,9 @@ int main(int argc, char **argv) {
 
   db.finish_run();
   db.clean();
+
+  if (!gc_dead_cas_blobs(*cas_ctx.get_store(), db))
```
AbrarQuazi, line 1236: We dont need to put a lock or anything on this right, like we need for staging files clean up, because its protected by the single writer lock of the DB?

AbrarQuazi: Ah this has a callback, which is called inside the transaction, nice!

src/runtime/database.cpp
```diff
@@ -2192,3 +2237,37 @@ std::vector<FileDependency> Database::get_file_dependencies() const {
   end_txn();
   return out;
 }
+
+void Database::gc_if_dead(const std::vector<std::string> &hashes,
+                          const std::function<void(std::vector<std::string>)> &callback) {
+  constexpr size_t BATCH_SIZE = 256;            // number of hashes per query, must match statement!
+  constexpr size_t TXN_DEAD_SIZE = 1ULL << 11;  // transaction limit for dead hashes
+  constexpr size_t TXN_HASHES = 1ULL << 13;     // transaction limit for total hashes checked
```
AbrarQuazi, line 2245: Curious to know why these limits are different between the both of them

dtzSiFive: Total hashes: more sql queries.

Dead hashes:
1) More rows returned from a query ("uneven").
2) More IO under the transaction to remove them.  (even if just rename()).

tools/wake/main.cpp
dtzSiFive: Readers can still proceed concurrently.  Ensuring this was confidently safe is the result of a lot of our hard work and careful engineering! 👍

src/cas/cas.cpp
dtzSiFive: Light refactoring of CAS to use constants for these things: https://github.com/sifiveinc/wake/pull/1852/commits/8d60792438402619c119bfeb5ffdaebd37858ed6 .

That should be a separate PR, but for now it's not.

dtzSiFive:
Updated:

1) Use constants throughout CAS code so I can use them in the new method, in response to @ngraybeal 's feedback.  Please spot-check this @AbrarQuazi .
2) Added some unit tests for the new CAS method.  While visiting noticed we don't actually run any of the CAS tests in practice (!!), so fixed that as well.  Thankfully they do pass, which is usually not the case when they have never been run by CI.  See: https://github.com/sifiveinc/wake/pull/1852/commits/f8773b879e09c84466fc101c5d6deb67efef5784

These should both be separate PR's but am considering just landing them here so we can keep moving forward.

dtzSiFive:
Since unit tests are being silly, I plan to pull out enabling unit tests and fight that battle in a separate PR.

EDIT: small bug in my test, see PR.

dtzSiFive:
Unit test and CAS touchup to use constants pulled out to this PR: https://github.com/sifiveinc/wake/pull/1853 .

src/cas/cas.cpp
dtzSiFive: Pulled out to #1853 ; CI stumbled and I want to tackle that separately.

tools/wake/main.cpp
dtzSiFive: Done! Thanks!

AbrarQuazi (approved):
Looking good!

PR #1853, opened 2026-05-14
enable cas unit tests, refactor cas to use constants

dtzSiFive:
Also add unit tests for enumerate_blobs_strings functionality.

~Pulled out because CI unit test failed at first? Not sure what happened there, see: https://github.com/sifiveinc/wake/actions/runs/25867633443/job/76013909717~ (file listing order is unpredictable, whoops, fixed)

### DIFF
--- a/src/cas/cas.cpp
+++ b/src/cas/cas.cpp
@@ -36,19 +36,20 @@ namespace fs = std::filesystem;
 
 namespace cas {
 
-// Global counter for unique temp file names during materialization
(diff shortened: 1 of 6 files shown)


AbrarQuazi (approved):
The new constants look good to me, and the bit calculations make sense. Thanks for activating the CAS tests, was a miss on my part.

dtzSiFive:
Thanks for review @AbrarQuazi !! 💯 ❤️

PR #1856, opened 2026-05-15
wcl: add function_ref; database: use it

dtzSiFive:
Based on LLVM's function_ref[1] and std::function_ref[2]

[1] https://github.com/llvm/llvm-project/blob/70cbd7591abee944151922edecc34093f6a9910e/llvm/include/llvm/ADT/STLFunctionalExtras.h#L37
[2] https://en.cppreference.com/cpp/utility/functional/function_ref (C++26)

Make use of this in database's callback methods.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -1447,7 +1447,7 @@ std::vector<std::string> Database::clear_jobs() {
   return out;
 }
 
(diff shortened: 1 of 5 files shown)


dtzSiFive:
Landing.  I want to use this.

PR #1857, opened 2026-05-15
Optimize critical path update, v15 schema for new (extended) index.

dtzSiFive:
Chunk the update, as even optimized (~60s -> ~12s) that's still much too long for a single transaction.

Otherwise other wake runs cannot make forward progress for the duration.

Replaces #1850 ; this has a single index that works for various filetree queries that today rely on `filesearch`, while also being usable for this work.


dtzSiFive:
This chunking, and that added in #1852 , should probably be moved to adaptively chunk based on a time budget instead of magic numbers.  Magic time budget instead.

Regardless, these "stop the bleeding" and appear to be sufficient in my testing.

PR #1858, opened 2026-05-15
wake: implement --prune feature.

dtzSiFive:
Checks for existance and file type.

Add basic test.


src/runtime/database.cpp
```diff
@@ -481,6 +485,15 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
       "   and substr(cast(j.commandline as varchar), 1, 8) != '<source>'"
       "   and substr(cast(j.commandline as varchar), 1, 7) != '<claim>'"
       " )";
+  const char *sql_get_job_output_paths =
+      "select distinct f.file_id, f.path, f.type"
+      " from filetree ft join files f on f.file_id=ft.file_id join jobs j on ft.job_id=j.job_id"
+      " where ft.access = 2"
```
dtzSiFive: Consideration:

Drop this qualifier (and rename to `get_job_paths` or whatever).

And actually simplify this all the way down to just:

`select distinct f.file_id, f.path, f.type` .

Everything is someone's output.

Anyway hopefully this is easy enough to shape however we'd like.

dtzSiFive: I think that's the right call so went ahead and made the change.

tools/wake/main.cpp
```diff
@@ -414,6 +414,7 @@ void print_help(const char *argv0) {
     << "    --init        DIR  Create or replace a wake.db in the specified directory"     << std::endl
     << "    --list-outputs     List all job outputs"                                       << std::endl
     << "    --clean            Delete all job outputs"                                     << std::endl
+    << "    --prune            Remove DB entries for jobs whose files no longer exist"     << std::endl
```
AbrarQuazi, line 423: NIT: `Remove DB entries for jobs whose files no longer exist _in the workspace_`

src/runtime/database.cpp
```diff
@@ -1482,6 +1494,51 @@ bool Database::clear_jobs_if_safe(wcl::function_ref<void(std::vector<std::string
...
+    const char *path = (const char *)sqlite3_column_text(imp->get_all_file_paths, 1);
+    const char *type = (const char *)sqlite3_column_text(imp->get_all_file_paths, 2);
+    if (path && type && !exists(path, type)) dead_files.push_back(file_id);
+  }
+  finish_stmt(why, imp->get_all_file_paths, imp->debugdb);
+
+  size_t pruned = 0;
+  for (int64_t file_id : dead_files) {
+    bind_integer(why, imp->delete_jobs_by_dead_file, 1, file_id);
+    single_step(why, imp->delete_jobs_by_dead_file, imp->debugdb);
```
AbrarQuazi, line 1575: @ag-eitilt We need to sync up on this PR, since your `--rm` PR adds a new deleted column, we need to converge on implementation on either soft-deleting from DB or hard deleting (which this PR does).

We probably need to update this call, as well as `sql_get_dead_hashes` below if we want to soft-delete

AbrarQuazi (commented):
Doing first review, and will take over this PR for any remaining implementation. Implementation looks good, will need to sync with Sam's `--rm` PR before merging

tests/runtime/prune/pass.sh
```diff
@@ -0,0 +1,53 @@
...
+set -eu
+
+WAKE="${1:+$1/wake}"
+WAKE="${WAKE:-wake}"
+
+cleanup() {
+  rm -rf wake.db* wake.log .wake output.txt
+}
+trap cleanup EXIT
+cleanup
```
ngraybeal, line 13: NIT:
No change needed, but part of our story of needing to improve our testing methodology. Using a trap EXIT will make debugging issues require commenting out this code so we can inspect the generated files. Just a minor inconvenience.

ngraybeal (approved):

src/runtime/database.cpp
ag-eitilt: I'm retaining the previous drop-from-DB behaviour for now, since my (admittedly somewhat superficial) reading of `--prune` is that it's meant to help trim overlarge databases.  If we find that the mechanisms for taming the DB growth work quite well for automatic management of that purpose, and/or once we actually implement something that uses the metadata, we can revisit switching `--prune` to the `deleted=1` soft-deletion.

tools/wake/main.cpp
ag-eitilt: This block is formatted the way it is to keep an 80-column width (failing even that; `--output`'s at 81), so while I liked your addition enough to add it, I added it onto a new line.

PR #1859, opened 2026-05-20
Update RSC to be mw/CAS compliant and fix filetype collision issue

AbrarQuazi:
* Make RSC client upload from CAS
* Make RSC client download to staging, and ingest + materialize into/from CAS
* Fix filetype collision issue in RSC (add filetype and mode to job reuse key)
* Add hash to output symlink target string so RSC client doesn't need to recompute it

### DIFF
--- a/rust/rsc/src/bin/rsc/types.rs
+++ b/rust/rsc/src/bin/rsc/types.rs
@@ -36,14 +36,14 @@ pub trait JobKeyHash {
             hasher.update(file.path.as_bytes());
             hasher.update(&file.hash.len().to_le_bytes());
             hasher.update(file.hash.as_bytes());
-            // Skip when `type` is empty so old-client keys (no type/mode) hash exactly
+            // Skip when `type` is absent so old-client keys (no type/mode) hash exactly
             // as before, preserving their cache. New-client keys mix in type+mode and
             // land in a disjoint namespace, fixing the symlink-vs-file collision.
             //
             // TODO: Remove this check once all clients have been migrated to the newer version.
-            if !file.r#type.is_empty() {
-                hasher.update(&file.r#type.len().to_le_bytes());
-                hasher.update(file.r#type.as_bytes());
+            if let Some(r#type) = &file.r#type {
+                hasher.update(&r#type.len().to_le_bytes());
+                hasher.update(r#type.as_bytes());
                 hasher.update(&file.mode.to_le_bytes());
             }
         }
@@ -55,9 +55,9 @@ pub trait JobKeyHash {
 pub struct VisibleFile {
     pub path: String,
     pub hash: String,
-    // Absent in old client payloads; defaults keep old keys unchanged.
+    // Absent in old client payloads; None keeps old keys unchanged.
     #[serde(rename = "type", default)]
-    pub r#type: String,
+    pub r#type: Option<String>,
     #[serde(default)]
     pub mode: i32,
 }
@@ -81,9 +81,16 @@ pub struct Dir {
 pub struct Symlink {
     pub path: String,
     pub link: String,
-    // Absent in old client payloads; "" means no CAS hash available (legacy entry).
-    #[serde(default)]
     pub hash: String,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct SymlinkPayload {
+    pub path: String,
+    pub link: String,
+    // Absent in old client payloads; None means no CAS hash available (legacy entry).
+    #[serde(default)]
+    pub hash: Option<String>,
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -96,7 +103,7 @@ pub struct AddJobPayload {
     pub hidden_info: String,
     pub visible_files: Vec<VisibleFile>,
     pub output_dirs: Vec<Dir>,
-    pub output_symlinks: Vec<Symlink>,
+    pub output_symlinks: Vec<SymlinkPayload>,
     pub output_files: Vec<File>,
     pub stdout_blob_id: Uuid,
     pub stderr_blob_id: Uuid,
@@ -349,7 +356,7 @@ mod tests {
         VisibleFile {
             path: path.to_string(),
             hash: hash.to_string(),
-            r#type: ty.to_string(),
+            r#type: Some(ty.to_string()),
             mode,
         }
     }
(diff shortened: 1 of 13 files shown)


rust/rsc/src/bin/rsc/types.rs
```diff
@@ -36,6 +36,16 @@ pub trait JobKeyHash {
             hasher.update(file.path.as_bytes());
             hasher.update(&file.hash.len().to_le_bytes());
             hasher.update(file.hash.as_bytes());
+            // Skip when `type` is empty so old-client keys (no type/mode) hash exactly
+            // as before, preserving their cache. New-client keys mix in type+mode and
+            // land in a disjoint namespace, fixing the symlink-vs-file collision.
+            //
+            // TODO: Remove this check once all clients have been migrated to the newer version.
```
ngraybeal, line 43: Should be tracking these in a more central location so we can give some EOL timeline?

rust/rsc/src/database.rs
```diff
@@ -533,7 +533,7 @@ pub async fn create_many_output_files<T: ConnectionTrait>(
 
     let chunked: Vec<Vec<output_file::ActiveModel>> = output_files
         .into_iter()
-        .chunks((MAX_SQLX_PARAMS / 6).into())
+        .chunks((MAX_SQLX_PARAMS / 7).into())
```
ngraybeal, line 536: Magic number. Should we assign this to a var for clarity?

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -693,38 +695,20 @@ export def rscApiGetFileBlob ((CacheSearchBlob blobId uri contentHash): CacheSea
 
     require Pass _ = verifyBlobHash blobId actualHash contentHash
 
-    def fixupScript =
```
ngraybeal, line 696: Happy to see go away for multiple reasons. 😄

ngraybeal (approved):
Happy to see what a CAS system starts affording us.

share/wake/lib/system/cas.wake
```diff
@@ -23,6 +23,14 @@ export def addCasHashEnv = match (getenv "WAKE_CAS")
 export def primCasDir: String =
     prim "cas_dir"
 
+# Returns the absolute CAS filesystem path for the blob with the given content hash.
+# Fails if the blob is not present in the local CAS store.
+export def primCasBlobPath (hash: String): Result String Error =
```
ngraybeal: NIT: Path in the function name, but returns a String.

AbrarQuazi:
I will merge these changes once we are ready to release the MVP, as this requires updating our prod RSC to this newest version

rust/rsc/src/bin/rsc/types.rs
ag-eitilt: I assume incoming RSC queries include the client version for us to build those metrics on top of in the first place?

rust/rsc/src/bin/rsc/types.rs
```diff
@@ -66,6 +81,9 @@ pub struct Dir {
 pub struct Symlink {
     pub path: String,
     pub link: String,
+    // Absent in old client payloads; "" means no CAS hash available (legacy entry).
+    #[serde(default)]
+    pub hash: String,
```
ag-eitilt: One of my long-running code-review traits is that I really dislike meaningful empty strings.  Unless Serde includes the `Some` constructor in how it serializes those cases, could we use `Option<String>` instead?  (Same comment on `VisibleFile`, this one just has the more egregious documentation.)

src/cas/cas.cpp
```diff
@@ -285,16 +285,20 @@ wcl::result<bool, CASError> Cas::materialize_blob(const ContentHash& hash,
...
-  times[0].tv_nsec = UTIME_OMIT;  // Don't change atime
-  times[1].tv_sec = mtime_sec;
-  times[1].tv_nsec = mtime_nsec;
-
-  if (utimensat(AT_FDCWD, temp_path.c_str(), times, 0) != 0) {
-    fs::remove(temp_path, ec);
-    return wcl::make_error<bool, CASError>(CASError::IOError);
+  // Apply timestamp to temp file before rename. (0, 0) is a sentinel meaning
+  // "leave the kernel-set mtime from the reflink/copy" — matches the convention
+  // in cas_prim.cpp::apply_mtime.
```
ag-eitilt, line 290: Similar dislike of this magic-value usage as I do with empty strings, but I do admit that crossing the Wake/C++ boundary with a richer type would be tricky.

share/wake/lib/system/cas.wake
```diff
@@ -23,6 +23,14 @@ export def addCasHashEnv = match (getenv "WAKE_CAS")
 export def primCasDir: String =
     prim "cas_dir"
 
+# Returns the absolute CAS filesystem path for the blob with the given content hash.
+# Fails if the blob is not present in the local CAS store.
+export def primCasBlobPathStr (hash: String): Result String Error =
+    def raw hash = prim "cas_blob_path"
```
ag-eitilt: This needs better signposting, in the name itself.  Even though you're returning a `String` (properly, since `Path`s are/will be relative), the assumption when reading the vast majority of the codebase is that all references to "path" are relative to the workspace.
```suggestion
# Returns the absolute CAS filesystem path for the blob with the given content hash.
# Fails if the blob is not present in the local CAS store.
export def primCasBlobAbsPathStr (hash: String): Result String Error =
    def raw hash = prim "cas_blob_abs_path"
```

src/runtime/cas_prim.cpp
```diff
@@ -524,12 +524,57 @@ static PRIMFN(prim_cas_ingest_staged_item) {
...
+  }
+
+  if (!store->has_blob(hash)) {
+    std::string msg = "Blob not in CAS: " + std::string(hash_str->c_str());
+    runtime.heap.reserve(reserve_result() + String::reserve(msg.size()));
+    auto err = String::claim(runtime.heap, msg);
+    RETURN(claim_result(runtime.heap, false, err));
+  }
+
+  std::string path = store->blob_path(hash);
```
ag-eitilt, line 566: For example, looking at the rest of this function I don't see anywhere the system-level path to the CAS gets added so I have to assume `blob_path` is *also* returning an absolute path... but that's absolutely *not* what I'd expect if I was reading this code without the doc comments.

rust/rsc/src/bin/rsc/types.rs
AbrarQuazi: Fixed. I added a seperate SymlinkPayload struct to differentiate between RSC's response (a request with a symlink payload will have an optional hash since client versions can differ, while the RSC's response will always send out the hash where the new client will parse it out while the old will just skip).

src/cas/cas.cpp
AbrarQuazi: Yeah this is something I will have to revisit and potentially cleanup

AbrarQuazi:
Verified this is working with staging RSC

PR #1862, opened 2026-05-27
Remove artifacts created in the course of testing

ag-eitilt:
A number of our tests create files to test visibility, collisions, etc.  These have been sticking around after the test is complete and cluttering `git status`.  This PR adds cleanup `rm`s at the end of the relevant tests so that the workspace is left in the same state it started.

Two notes on the implementation:

* None of these files should be critical for debugging errors, if we do wind up regressing one or more of these tests.  (If they were, the better solution would be adding them to `.gitignore` -- as we do for the `wake.db` and similar which *are* kept around.)
* We still want to keep the removals of these same files at the beginning of the scripts, in case one of them gets interrupted halfway through and needs cleanup before the next run.

### DIFF
--- a/tests/runtime/mtime-observe/pass.sh
+++ b/tests/runtime/mtime-observe/pass.sh
@@ -11,3 +11,8 @@ export WAKE_CAS=1
 rm -rf .build .fuse wake.db* wake.log output.txt
 
 "${WAKE}" -x "go Unit"
+
+# Clean up
+err=$?
+rm -f output.txt
+exit $err
--- a/tests/pending/negative-input/pass.sh
+++ b/tests/pending/negative-input/pass.sh
@@ -16,3 +16,8 @@ echo "bar" > bar
 # 3. Run job testing for bar with bar visible
 echo "=== Run 2: bar exists ==="
 "${WAKE}" --no-tty -x 'test_bar Unit'
+
+# Clean up
+err=$?
+rm -f bar
+exit $err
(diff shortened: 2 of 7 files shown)


ag-eitilt:
The previous test failure (`inspection/canceled`) was apparently just flakiness.  Though since it was reported as "Early termination requested", it seems to have been a timeout rather than flakiness in the code anywhere.  Weird, and something we'll want to keep an eye out for, but probably not anything we need to track down until and unless we start seeing it more often.

tests/runtime/mtime-observe/pass.sh
```diff
@@ -11,3 +11,8 @@ export WAKE_CAS=1
 rm -rf .build .fuse wake.db* wake.log output.txt
 
 "${WAKE}" -x "go Unit"
+
+# Clean up
+err=$?
+rm -f output.txt
+exit $err
```
ngraybeal, line 18: lines 16 and 18 aren't required since `set -e` is set above. 

Perhaps we should just make all scripts `set -e` and then only `rm -f` at the end is necessary.

AbrarQuazi (approved):
LGTM

ag-eitilt:
I updated this to handle the tests merged since the initial commit, and added a new pattern of identifying the artifacts *once* in a variable rather than trying to keep the list synced between the cleanup-from-abort at the start and the cleanup-after-success at the end (the former is still the only one that cleans `wake.db`, etc. to allow for debugging/inspection).

There's still definitely a good bit of work left to standardize these -- such as whether we do have `set -e` across the board, whether we have `trap cleanup EXIT` or `fail` with a custom message, whether we look into an existing library or create our own for shared functions, ... -- but I want to keep this particular PR a bit more targetted.

ngraybeal (approved):

PR #1864, opened 2026-06-04
Implement a sqrt heap growth curve

ngraybeal:
Wake Heap allocation is based on a simple geomtric growth curve: 
`max (Heap_factor * last_live_heap, current_heap_size)` 

This is a Stop-and-copy GC with 2 semi-spaces, we pay double the cost for the heap. We also have a blind 1.5x multiplier when we actually `malloc/realloc` to help avoid having to pay extra malloc costs. 

This does not scale well or reflect most use cases once it approaches GB heap size. Kilobytes are free, megabytes are cheap, gigabytes of memory gets expensive or unattainable. 

This PR updates the growth algorithm to include a Pivot point (`heap_pivot`) where the growth goes from an aggressive linear growth to a sqrt growth. 
`last_live_heap + heap_factor* sqrt(last_live_heap * heap_pivot))`

`Heap-factor` is still used to impact the growth rate, but it's impact is tied to the sqrt growth. This doesn't address anything about always keeping 2x heap around or switch our GC approach. I decided not to touch the 1.5x overhead malloc in this PR or anything about the heap downsizing (1/3) but I believe it should also follow some growth curve. 

Performance impacts:
There's a lot of variables in play I tried to consider (fast/slow growth, long/short heap objects, how often GC gets triggered, etc.), but don't have any strong mathematical backing or data to share. 

In any real testing example I have, I did not see anything of concern with GC being invoked too aggressively. This mostly helped ensure large heaps don't hit OOM issues unnecessarily. 

Here are a couple graphs showing the growth curve before and after change.
image

image

Here's some literature i referenced when creating this:
https://dl.acm.org/doi/epdf/10.1145/3563323

### DIFF
--- a/src/runtime/config.h
+++ b/src/runtime/config.h
@@ -261,6 +261,9 @@ struct InterpreterRuntimeWarningPolicy {
   }
 };
 
+// Controls how aggressively the GC grows the heap after each collection cycle.
+// Floor limit: must be > 1.0 — values at or below 1.0 leave no post-collection
+// headroom and cause GC to loop.
 struct HeapFactorPolicy {
   using type = double;
   using input_type = type;
@@ -279,6 +282,10 @@ struct HeapFactorPolicy {
   static void set_env_var(HeapFactorPolicy& p, const char*) {}
 };
 
+// Sets the live-data size (in MB) at which the GC heap growth switches from linear
+// to sqrt scaling. Also controls the growth rate of the sqrt formula — a larger
+// pivot means larger sqrt growth.
+// Floor limit: must be > 1MB
 struct HeapPivotPolicy {
   using type = double;
   using input_type = type;
--- a/src/runtime/config.h
+++ b/src/runtime/config.h
@@ -53,6 +53,10 @@ struct WakeConfigOverrides {
 
   // Lets you specify an alternative user config
   std::optional<std::string> user_config;
+
+  // GC heap tuning
+  std::optional<double> heap_factor;
+  std::optional<double> heap_pivot_mb;
 };
 
 template <class T>
@@ -257,6 +261,42 @@ struct InterpreterRuntimeWarningPolicy {
   }
 };
 
+struct HeapFactorPolicy {
+  using type = double;
+  using input_type = type;
+  static constexpr const char* key = "heap_factor";
+  static constexpr bool allowed_in_wakeroot = true;
+  static constexpr bool allowed_in_userconfig = false;
+  type heap_factor = 4.0;
+  static constexpr type HeapFactorPolicy::*value = &HeapFactorPolicy::heap_factor;
+  static constexpr Override<input_type> override_value = &WakeConfigOverrides::heap_factor;
+  static constexpr const char* env_var = nullptr;
+
+  HeapFactorPolicy() = default;
+  static void set(HeapFactorPolicy& p, const JAST& json);
+  static void set_input(HeapFactorPolicy& p, const input_type& v) { p.*value = v; }
+  static void emit(const HeapFactorPolicy& p, std::ostream& os) { os << p.*value; }
+  static void set_env_var(HeapFactorPolicy& p, const char*) {}
+};
+
+struct HeapPivotPolicy {
+  using type = double;
+  using input_type = type;
+  static constexpr const char* key = "heap_pivot_mb";
+  static constexpr bool allowed_in_wakeroot = true;
+  static constexpr bool allowed_in_userconfig = false;
+  type heap_pivot_mb = 64.0;
+  static constexpr type HeapPivotPolicy::*value = &HeapPivotPolicy::heap_pivot_mb;
+  static constexpr Override<input_type> override_value = &WakeConfigOverrides::heap_pivot_mb;
+  static constexpr const char* env_var = nullptr;
+
+  HeapPivotPolicy() = default;
+  static void set(HeapPivotPolicy& p, const JAST& json);
+  static void set_input(HeapPivotPolicy& p, const input_type& v) { p.*value = v; }
+  static void emit(const HeapPivotPolicy& p, std::ostream& os) { os << p.*value; }
+  static void set_env_var(HeapPivotPolicy& p, const char*) {}
+};
+
 /********************************************************************
  * Generic WakeConfig implementation
  *********************************************************************/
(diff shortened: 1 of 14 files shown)


src/runtime/config.h
```diff
@@ -257,6 +261,42 @@ struct InterpreterRuntimeWarningPolicy {
   }
 };
 
+struct HeapFactorPolicy {
+  using type = double;
+  using input_type = type;
```
ag-eitilt, line 269: I can understand the type aliasing to make the purpose more clear than just having a random `double` in the signature, but why the double aliasing through `type` first?  And do we want a more indicative name than "input"?

tools/wake/main.cpp
```diff
@@ -545,6 +536,27 @@ int main(int argc, char **argv) {
...
+      return 1;
+    }
+    config_override.heap_factor = v;
+  }
+
+  if (clo.heappivot) {
+    char *tail;
+    double v = strtod(clo.heappivot, &tail);
+    if (*tail || v < 1.0) {
+      std::cerr << "Cannot run with " << clo.heappivot << " heap-pivot (must be >= 1.0 MB)!"
```
ag-eitilt, line 553: Doesn't need to be output to stderr, but maybe add a comment each describing the math that leads to these restrictions?

src/runtime/config.h
```diff
@@ -257,6 +261,42 @@ struct InterpreterRuntimeWarningPolicy {
...
+  static constexpr const char* env_var = nullptr;
+
+  HeapFactorPolicy() = default;
+  static void set(HeapFactorPolicy& p, const JAST& json);
+  static void set_input(HeapFactorPolicy& p, const input_type& v) { p.*value = v; }
+  static void emit(const HeapFactorPolicy& p, std::ostream& os) { os << p.*value; }
+  static void set_env_var(HeapFactorPolicy& p, const char*) {}
+};
+
+struct HeapPivotPolicy {
```
ag-eitilt, line 289: These structs themselves could also be good places to add comments describing their role in the algorithm.  At this point, we all know how the heap factor is used in the growth algorithm, but we only know that due to a good while of people picking through the code and explaining to the rest of us; it's still not superficially obvious to anyone coming in fresh.  And the pivot is a new concept entirely.

ag-eitilt (approved):
Looks good to me!  I would have expected a more dramatic graph of the savings after the crossover point, but I think that's down to the log scale understating the benefits this provides.

tools/wake/main.cpp
```diff
@@ -266,7 +266,8 @@ void print_help(const char *argv0) {
     << "    --no-wait          Do not wait to obtain database lock; fail immediately"      << std::endl
     << "    --no-workspace     Do not open a database or scan for sources files"           << std::endl
     << "    --fatal-warnings   Do not execute if there are any warnings"                   << std::endl
-    << "    --heap-factor X    Heap-size is X * live data after the last GC (default 4.0)" << std::endl
+    << "    --heap-factor X    GC headroom scale factor (default 4.0)"                     << std::endl
+    << "    --heap-pivot  X    Pivot point in MB where GC growth curve transitions (default 64)" << std::endl
```
AbrarQuazi, line 270: 64 MB is not a lot as a default, I had the assumption this pivot would happen we go into the GB territory , which makes me wonder whats the point of giving a pivot, and not having everything as a sqrt growth? Whats the benefit of having a pivot (the normal growth rate) at all?

AbrarQuazi:
I would recommend testing against our internal repo CI, to see if we see much of a timing difference for the rest of pipelines with these changes

src/runtime/config.h
ngraybeal: Mostly following the pattern established in the file. We have a `using type = bool;` up above.

tools/wake/main.cpp
ngraybeal: The pivot also influences the sqrt rate growth. It helps bend it per say. So lower pivot point, less aggressive growth upper heap sizes.


pivot 64MB, heap factor 4x
```
   Live (MB)      Old heap    Old sys (2×)      New heap    New sys (2×)
----------------------------------------------------------------------
       200.0         800.0          1600.0         652.5          1305.1
       400.0        1600.0          3200.0        1040.0          2080.0
       800.0        3200.0          6400.0        1705.1          3410.2
      1600.0        6400.0         12800.0        2880.0          5760.0
      3200.0       12800.0         25600.0        5010.2         10020.4
      6400.0       25600.0         51200.0        8960.0         17920.0
     12800.0       51200.0        102400.0       16420.4         32840.8
     25600.0      102400.0        204800.0       30720.0         61440.0
     51200.0      204800.0        409600.0       58440.8        116881.5
    102400.0      409600.0        819200.0      112640.0        225280.0
    ```
    
    Pivot 256MB, heap factor 4x
    ```
       Live (MB)      Old heap    Old sys (2×)      New heap    New sys (2×)
----------------------------------------------------------------------
       200.0         800.0          1600.0         800.0          1600.0
       400.0        1600.0          3200.0        1600.0          3200.0
       800.0        3200.0          6400.0        2610.2          5220.4
      1600.0        6400.0         12800.0        4160.0          8320.0
      3200.0       12800.0         25600.0        6820.4         13640.8
      6400.0       25600.0         51200.0       11520.0         23040.0
     12800.0       51200.0        102400.0       20040.8         40081.5
     25600.0      102400.0        204800.0       35840.0         71680.0
     51200.0      204800.0        409600.0       65681.5        131363.1
    102400.0      409600.0        819200.0      122880.0        245760.0
    ```

File: src/runtime/gc.cpp [hunk 5]
```diff
@@ -146,10 +149,23 @@ struct Heap::Imp {
         finalize(nullptr),
         gc_count(0),
         total_gc_time(0) {}
+
+  size_t desired_size(size_t live_pads, size_t requested_pads) const {
+    size_t linear = static_cast<size_t>(heap_factor * live_pads);
+    size_t sqrt_desired =
+        live_pads +
+        static_cast<size_t>(heap_factor * std::sqrt(static_cast<double>(live_pads) *
+                                                    static_cast<double>(heap_pivot_pads)));
```
AbrarQuazi, line 158: Maybe not for this PR, but I think its worth following the equation in the paper in the future: https://dl.acm.org/doi/epdf/10.1145/3563323

where the tunable heap_factor and heap_pivot gets reduced into an adaptive runtime measured constant C
`M = L + sqrt(L · C)`

where the constant `C = g / cs` (GC duration / bytes allocated between GC's divided by mutator time)

benefit here being automatic adaption instead of tuning the builds all the time. I think we should have access to most of the pieces in the equation

AbrarQuazi (approved):

src/runtime/gc.cpp
ngraybeal: Yeah, definitely something worth considering. I do plan on more tweaks to this (dealing with the 1.5x blind growth) so can revisit then.

PR #1865, opened 2026-06-05
Add `--rm` for deleting files from both the workspace and the CAS

ag-eitilt:
This is part of the general collection of mechanisms by which people will be able to reduce their disk usage despite the CAS backing.  Specifically, this introduces `wake --rm PATH...` which (A) removes whatever file is at the target path in the workspace, and (B) garbage-collects any blobs which had been written to that path.  In performing (B), this is written to handle a number of edge cases:

* The calculation of which blobs should be deleted is independent from what files are currently in the workspace (i.e. the user can `rm file.txt && wake --rm file.txt` to clean remnants in the CAS which have already been deleted in the workspace).
* If the blob is shared with anything which isn't being deleted (i.e. two files at different locations with the same content), the blob is kept.
* In a multi-wake environment, if a parallel invocation has referenced the blob through the `run_files` table, the blob is kept.
* If the blob is shared but all paths referencing it are given as part of the same `--rm` invocation and none of them are in active use elsewhere, the blob is *deleted*.[^1]

[^1]: Known bug: doesn't yet handle deleting all the paths referencing the blob in multiple invocations.  I'm fixing that and will push the commit before merging.

This also adds a new column to the `files` table which allows us to retain the metadata in the database (eventual use case: being able to weigh the size of each blob against the number of places it's linked to).  Per discussion, we'll want to pull that into a separate `blobs` table which is separate from the materialization path, timestamp, etc., but to keep this PR simple I'm wanting to handle that in a followup.


ag-eitilt:
Still trying to figure out how to handle directory targets (currently leaning toward just rejecting them and aborting), but I think that's the main remaining bug/edge case.  Quotes are what Sonnet found, my evaluation is unquoted:

> **1. Transaction Failure After CAS Deletion (Lines 2034-2045)**
> This is already acknowledged with a TODO, but it's a serious issue:
>
> * CAS blobs are deleted **before** the transaction
> * If the UPDATE fails, we can't rollback the CAS deletions
> * Database and CAS are now inconsistent
>
> **Severity:** High
> **Recommendation:** Either:
>
> * Use SQLite's `BEGIN IMMEDIATE` earlier (before CAS operations)
> * Or implement a two-phase commit with rollback capability for CAS

This is high *potential* severity, but low in actual use.  Since we know the syntax of what the first query returned, and we know the template we're using for the query, we'd be very surprised if the `sqlite3_prepare_v2` failed.

> **2. No Validation That Paths Are Within Workspace**
> In `tools/wake/clean.cpp`, after path normalization, there's no check that the resulting path doesn't escape the workspace (e.g., `../../etc/passwd`).
>
> **Test case:** What happens if user runs `wake --rm ../../outside-workspace.txt`?
>
> Let me verify:
> Good! Looking at line 335-337 in `filepath.h`, `relative_to` can indeed return paths with `../` if the absolute path is outside the relative base. This is a **security** issue.

Pointing to paths outside the workspace is certainly undesirable, but I'm not sure it's a security issue.  We don't generate any files to be deleted ourselves (yet, we'll want a `--recursive` eventually) and even if we did, it would just be a potential hole for user *surprise*.  We don't make any changes to the user or group, so it's not any different from the user calling `rm` (the shell tool) directly.

> **3. Duplicate Paths in Input**
> If the user provides the same path multiple times (`wake --rm file.txt file.txt`), the code will:
>
> * Process it twice in normalization
> * Query it twice in the SQL (though SQL will deduplicate)
> * Try to unlink it twice (second will fail with ENOENT, which is ignored)
>
> **Severity:** Low (just inefficient, but harmless)

Like you said, this is potentially inefficient, but harmless.  I'd expect that any attempts we make to filter out duplicates would wind up adding more overhead in the non-duplicate case than they save if they do filter something out.

> **4. Empty String After Normalization**
> What if normalization produces an empty string or just "."? The SQL would then try to delete everything or the workspace root.

This is definitely a potential issue.  Looking at the implementation of those functions, I don't think it's ever going to be *empty*, but it could definitely wind up pointing to the workspace root (which would then fall into difficulties removing directories).

> **5. SQL Injection via Path Placeholders**
> Lines 1951-1954 and 2029-2030 build SQL with string concatenation of placeholders. While the placeholders themselves are `?`, the number is determined by `paths.size()`. If somehow `paths` is manipulated... actually, this looks safe since it's just `(?), (?), ...`.

Yeah, the number isn't going to be manipulatable (unless a malicious script tries to allocate more placeholders than there's stack space for), and we're careful to construct the placeholders with known `?` values.  The paths which are later *bound* to those placeholders aren't under our control, but `bind_string` is supposed to be protecting us there.

> **6. Race Condition: File Modified Between Query and Deletion**
> Timeline:
>
> 1. Query finds hash for `file.txt`
> 2. User modifies `file.txt` (different hash now)
> 3. CAS blob for old hash is deleted
> 4. Database marks file as deleted
> 5. New content is orphaned in workspace
>
> **Severity:** Medium - could lead to confusion

Definitely a potential source of confusion, but we're obtaining an exclusive lock on the database to protect against exactly this, and if the user modifies the workspace file outside of wake it'll just be removed when we get to that part of the code, as expected -- it won't affect the blobs that get removed from the CAS (which would be what causes the biggest confusion) since we don't look at any properties of the workspace when deciding what blobs to remove.

> **7. No Check for Directories**
> What happens if user runs `wake --rm somedir/`? The code will try to `unlink()` a directory, which will fail. But will it try to remove directory entries from the database?

This one is a legitimate bug.  We're going to need to handle attempts to remove directories more cleanly.  I actually think that it *wouldn't* fail and would blindly remove the directory... but not everything contained within it.  Either way, we'll need to address it.

ag-eitilt:
As the commit name suggests, this PR doesn't currently handle directories.  That *is* a critical capability we'll need to add soon after this, but I expect it to be more cleanly implemented as a separate PR -- unless we forego following `rm` and `git rm`'s model and instead recurse into directories by default.

src/runtime/database.cpp
```diff
@@ -1935,6 +1940,174 @@ std::vector<std::string> Database::get_outputs() const {
...
+  for (size_t i = 0; i < paths.size(); ++i) {
+    bind_string(why_paths, paths_stmt, i + 1, paths[i]);
+  }
+
+  std::vector<std::string> paths_to_remove;
+  while (sqlite3_step(paths_stmt) == SQLITE_ROW) {
+    std::string path = rip_column(paths_stmt, 0);
+    std::string type = rip_column(paths_stmt, 1);
+    if (type == "directory") {
+      // TODO: This should recurse to all children of the directory, but to keep this first PR
```
ngraybeal: Nit: probably don't need the `but` in the codebase. something more for the PR comment.

ag-eitilt: Fair enough; I've got loosened standards with regards to the code/PR division for TODO notes, but this particular one is indeed leaning pretty heavily to the PR side.

tools/wake/main.cpp
```diff
@@ -414,6 +420,7 @@ void print_help(const char *argv0) {
     << "    --init        DIR  Create or replace a wake.db in the specified directory"     << std::endl
     << "    --list-outputs     List all job outputs"                                       << std::endl
     << "    --clean            Delete all job outputs"                                     << std::endl
+    << "    --rm               Remove files/directories from database (paths as arguments)" << std::endl
```
AbrarQuazi: Probably just include files for now in the description? Until we have the `-rf` feature rolled out?

tools/wake/clean.cpp
```diff
@@ -0,0 +1,94 @@
...
+  // Normalize all paths to be relative to the workspace root, allowing this to be called from
+  // directories other than the root.
+  std::string workspace_root = get_cwd();
+  std::vector<std::string> normalized_paths;
+  normalized_paths.reserve(paths.size());
+
+  for (const auto &path : paths) {
+    std::string normalized;
+
+    if (wcl::is_relative(path)) {
```
AbrarQuazi: so since we upgraded to c++17 we do have access to `std::filesystem` functions that we have already started using in the code base. Might be worthwile trying to consolidate even more (I had only done a first pass at some functions when i upgraded to c++17) and use the standard library functions (unless there is reason to keep our own wcl versions)

tools/wake/main.cpp
```diff
@@ -902,6 +913,23 @@ int main(int argc, char **argv) {
     return delete_error ? 1 : 0;
   }
 
+  if (clo.rm) {
+    std::string fail =
+        db.open(/*wait=*/false, /*memory=*/false, /*tty=*/isatty(0), /*readonly=*/false);
```
AbrarQuazi, line 918: Should we fail when another wake process is running? No calling `rm` during an active wake innvocation?

tools/wake/clean.cpp
```diff
@@ -0,0 +1,94 @@
...
+      return EXIT_FAILURE;
+    }
+    if (normalized == ".build/cas" || normalized.find(".build/cas/") == 0) {
+      std::cerr << "wake --rm: remove materialized paths, not the CAS blobs directly: '" << path
+                << "'" << std::endl;
+      return EXIT_FAILURE;
+    }
+
+    normalized_paths.push_back(std::move(normalized));
+  }
```
AbrarQuazi, line 80: Nice I like these checks. What will happen with absolute paths, how do we deal with them?

src/runtime/database.cpp
```diff
@@ -1935,6 +1940,174 @@ std::vector<std::string> Database::get_outputs() const {
   return out;
 }
 
+std::pair<std::vector<std::string>, std::vector<std::string>> Database::remove_blobs(
+    cas::Cas *cas, const std::vector<std::string> &paths) {
```
AbrarQuazi, line 1979: I am trying to remember and refamiliarize myself again with the GC code that has been implemented already. Does this function already have similarity with `gc_if_dead` that we can reuse? https://github.com/sifiveinc/wake/pull/1852/files#diff-a29a0517c6af9d9210367bea778b442a7c027638b8b9c504fbc70f02189a29b0

src/runtime/database.cpp
```diff
@@ -1935,6 +1940,174 @@ std::vector<std::string> Database::get_outputs() const {
...
+  std::string hashes_query =
+      // Use a CTE to define the removal list once for both select operations.
+      "with paths_to_remove(path) as (values " + hashes_input_placeholders +
+      ")"
+      " select distinct f1.hash from files f1"
+      " where f1.path in paths_to_remove"
+      " and f1.deleted = 0"
+      // Only include hashes where *all* files with that hash are in our removal list
+      // (i.e., exclude hashes that are shared with files we're *not* removing).
+      " and not exists ("
```
AbrarQuazi, line 2046: So if a hash is shared by multiple file paths, we dont delete it if all those file paths are not mentioned?

I can understand this is for safety reasons ( so we dont accidently delete the underlying CAS blob thats utilized by other paths) but I feel this might become a bit annoying with users trying to delete a file and not having knowledge of what other reflinked sibling files there are.

We probably need to address this problem somehow, I remember in discussions we floated the idea of a `wake --stat <path>` command that we can potentially use that shows important details like computation time, hash, as well as potentially reflinked siblings?

Or we need some sort of subcommand for `rm` that forcefully deletes that path and all its sibling reflinks as well as the underlying parent cas blob

src/runtime/schema.h
```diff
@@ -41,7 +41,8 @@ inline const char *getWakeSchemaSQLTxn() {
          "  path      text    not null,"
          "  hash      text    not null,"
          "  type      text    not null,"
-         "  mode      integer not null);"
+         "  mode      integer not null,"
+         "  deleted   integer not null default 0);"
```
AbrarQuazi, line 45: Should `gc_if_dead` also mark deleted in the DB?

AbrarQuazi (commented):
Did my first pass review, looks pretty good so far! Just have a few questions

src/runtime/database.cpp
ag-eitilt: That function is definitely going to need to be adapted to be aware of the `deleted` flag, as you pointed out in the other comment, but they're solving different things.  `gc_if_dead()` deletes blobs which don't have an entry in the `files` DB table (in the future: or if their entry is marked `deleted`), `remove_blobs()` deletes files which *do* still have a concrete reference and updates the `files` table with that mark.  The two definitely resemble each other, but the similarity in broad scope doesn't get reflected in the implementation -- the most they share is a loop of "for all of these hashes, call `cas.remove_blob(*hash)`", and trying to reuse the loop in addition to the already-shared per-hash removal logic would wind up being pretty marginal gains.

src/runtime/database.cpp
ag-eitilt: I've been thinking of `wake --rm` as more of a workspace-management tool (which incidentally avoids surprising data retention), but the follow-the-links is definitely a valid use case.  Implementing that as a behavioral flag on top of the primary command does feel like the better way to go about it, rather than making it the default.  I'll look into that as a followup as well, though if you don't mind, unlike the directories I think that's something which could be pushed to the other side of MVP.

src/runtime/schema.h
ag-eitilt: Described above, but no, because that function is about collecting loose blobs which don't have *any* DB representation.  It should definitely *read* the `deleted` flag, though.

tools/wake/clean.cpp
ag-eitilt: Good point.  My rule of thumb is to use locally-defined functions -- unless they're known to be out-of-date -- in preference to standard-library ones, but in this case these *are* likely falling into that out-of--date exception.

tools/wake/clean.cpp
ag-eitilt: The original plan was for `wake --rm` to be equivalent to a wrapper around `rm` which just happened to delete associated CAS blobs if any exist.  Under that model, absolute/non-workspace paths would just be unlinked like everything else.

When I started filtering out directories, `wake --rm` became a lot pickier and only deletes files which were written out (at least once) by the current `wake.db`.  Non-workspace paths are still handled the same as everything else, they're just ignored a lot of the time since Wake will only very rarely write outside the workspace.  I'm still not sure I like that model and have been thinking about a few solutions to get back to allowing users with no or little knowledge of the workspace history to get intuitive results, but it's a more complex problem than I want to tackle in this PR.

tools/wake/main.cpp
ag-eitilt: It definitely requires a good bit more safety checks this way, but it's not *too* much overhead to allow people to remove blobs which are irrelevant to concurrent runs.  Between the read/write exclusive transaction and filtering out anything active in the `run_files` table, this should be concurrency-safe.

src/runtime/database.cpp
```diff
@@ -519,7 +521,7 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
       "   (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?),"
       "   (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?), (?))"
       "select d.hash from disk_batch d where d.hash is not null and not exists ("
-      "  select 1 from files where hash=d.hash"
+      "  select 1 from files where hash=d.hash and deleted=0"
```
AbrarQuazi, line 530: Is there ever a situation where there's an orphaned CAS blob where the hash is not in the DB at all (not in files table and not referenced by deleted)?

AbrarQuazi (approved):
Part 1 of `--rm` looks good to me! Thanks!

ag-eitilt:
Last couple bugfixes after that approval -- the one in https://github.com/sifiveinc/wake/pull/1865/commits/520778584be17efe52fed4c03e5060e757a2c7ae changes how the `run_files` table is treated, but I'm not sure we meant for it to be "files *created* in this run" (before the change) instead of "files *used* by this run" (after).  I personally expected "used", for what it's worth.

AbrarQuazi (approved):
Reviewed the new changes, its awesome that we can reuse the same blob now. Thanks!

PR #1867, opened 2026-06-08
wakebox prim

ngraybeal:
* Create wakebox write_wakebox_spec prim
* Create wakebox runner input tuple
* update makeJSONRunner to use new prim and tuple

Creation of the wakebox json in wakeLang has observed to be very compute expensive when scaling up, especially with large Visible file lists. The foldr in the [doFormat](https://github.com/sifiveinc/wake/blob/66c3aa1feafe388cf1d6b181ba5286a83308142d/share/wake/lib/core/json.wake#L82) looks to be the source of a lot of this. 

Basically our current flow requires taking our tuples -> converting them to Jvalue -> Converting them to a string -> writing to a file. Since the write is a job, wake will add this to the work to heap. This will remain there until the runtime interpreter finishes and we can run the job scheduler to launch the jobs. 

This is an enhancement to avoid the problem by just taking an explicit tuple and doing all that work in a single prim so we can dump it out disk ASAP.

### DIFF
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -366,20 +366,17 @@ export def defaultRunner: Runner =
 
             fuseRunner
 
-export data MountOpType =
-    MountBind
-    MountCreateDir
-    MountCreateFile
-    MountSquashfs
-    MountTmpfs
-    MountWorkspace
-
-# A mount operation for the wakebox spec.
-export tuple WakeboxMountOp =
-    export Type: MountOpType
-    export Source: String
-    export Dest: String
-    export ReadOnly: Boolean
+# Mount operations for the wakebox spec.
+# These constructors match validate_mount() and do_mounts() in wakebox.
+# See: src/wakefs/namespace.cpp
+# wake-format off
+export data WakeboxMountOp =
+    WakeboxBindOp          (source: String) (dest: String) (readonly: Boolean)
+    WakeboxSquashfsOp      (source: String) (dest: String)
+    WakeboxTmpfsOp                          (dest: String)
+    WakeboxCreateDirOp                      (dest: String)
+    WakeboxCreateFileOp                     (dest: String)
+    WakeboxFuseWorkspaceOp                  (dest: String)
 
 # Fields consumed by the wakebox/fuse sandbox when reading the spec JSON.
 export tuple WakeboxSpec =
@@ -454,7 +451,7 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
             primCasDir
             False
             False
-            (WakeboxMountOp MountWorkspace "" "." False, Nil)
+            (WakeboxFuseWorkspaceOp ".", Nil)
             ""
             ""
             None
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -366,6 +366,40 @@ export def defaultRunner: Runner =
 
             fuseRunner
 
+# A mount operation for the wakebox spec.
+export tuple WakeboxMountOp =
+    export Type: String
+    export Source: String
+    export Dest: String
+    export ReadOnly: Boolean
+
+# Fields consumed by the wakebox/fuse sandbox when reading the spec JSON.
+export tuple WakeboxSpec =
+    export Command: List String
+    export Environment: List String
+    export Visible: List Path
+    export Directory: String
+    export Stdin: String
+    export CasDir: String
+    export IsolateNetwork: Boolean
+    export IsolatePids: Boolean
+    export MountOps: List WakeboxMountOp
+    export Hostname: String
+    export Domainname: String
+    export UserId: Option Integer
+    export GroupId: Option Integer
+    export CommandTimeout: Option Integer
+    export Runner: String
+
+# Write a WakeboxSpec to a JSON file at the given path.
+# indent: spaces per level (0 = compact).
+# The prim correctly awaits any unresolved field promises before writing.
+export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: String): Result Unit Error =
+    def f spec ind path = prim "write_wakebox_spec"
+
+    f spec indent filePath
+    | rmapError makeError
+
 # A plan describing how to construct a JSONRunner
 # RawScript: the path to the script to run jobs with
 # ExtraArgs: extra arguments to pass to ``RawScript``
@@ -388,7 +422,7 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
     def script = which (simplify rawScript)
     def executeOk = access script xOK
 
-    def run (job: Job) ((RunnerInput label command visible environment directory stdin res prefix record isatty fnInputs fnOutputs): RunnerInput): RunnerOutput =
+    def run (job: Job) ((RunnerInput label command visible environment directory stdin _ prefix record isatty fnInputs fnOutputs): RunnerInput): RunnerOutput =
         require True = executeOk
         else
             def runnerError = "Runner {script} is not executable".makeError
@@ -396,76 +430,31 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
 
             RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some runnerError)
 
-        def Usage jobStatus runtime cputime membytes inbytes outbytes = record
-
-        def mkVisibleJson (p: Path): Result JValue Error =
-            require True = match p.getPathType
-                PathTypeRegularFile -> True
-                PathTypeDirectory -> True
-                PathTypeSymlink -> True
-                _ -> False
-            else
-                failWithError "Visible path {p.getPathName} has unsupported transport type {pathTypeToString p.getPathType}"
-
-            Pass (
-                JObject (
-                    "path" :-> JString p.getPathName,
-                    "type" :-> JString (pathTypeToString p.getPathType),
-                    "hash" :-> JString p.getPathHash,
-                    "mode" :-> JInteger p.getPathMode,
-                    "mtime" :-> JInteger p.getPathMtimeNs,
-                    Nil
-                )
-            )
-
-        def visibleJsonResult =
-            visible
-            | map mkVisibleJson
-            | findFail
-
-        require Pass visibleJson = visibleJsonResult
-        else
-            def runnerError =
-                visibleJsonResult
-                | getWhenPass "Failed to encode visible inputs with cached modes".makeError
-
-            def _ = markJobSetupFailure job runnerError
-
-            RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some runnerError)
-
-        def json =
-            JObject (
-                "label" :-> JString label,
-                "command" :-> command | map JString | JArray,
-                "environment" :-> environment | map JString | JArray,
-                "visible" :-> JArray visibleJson,
-                "directory" :-> JString directory,
-                "stdin" :-> JString stdin,
-                "resources" :-> res | map JString | JArray,
-                "version" :-> JString version,
-                "isolate-network" :-> JBoolean False,
-                "isolate-pids" :-> JBoolean False,
-                "mount-ops" :-> JArray (JObject ("type" :-> JString "workspace", "destination" :-> JString ".", Nil), Nil),
-                "cas-dir" :-> JString primCasDir,
-                "usage" :-> JObject (
-                    "status" :-> JInteger jobStatus,
-                    "runtime" :-> JDouble runtime,
-                    "cputime" :-> JDouble cputime,
-                    "membytes" :-> JInteger membytes,
-                    "inbytes" :-> JInteger inbytes,
-                    "outbytes" :-> JInteger outbytes,
-                    Nil
-                ),
-                Nil
-            )
-
         def buildDirName = ".build"
         def specFile = "{buildDirName}/spec-{prefix}.json"
         def resultFile = "{buildDirName}/result-{prefix}.json"
 
+        def spec =
+            WakeboxSpec
+            command
+            environment
+            visible
+            directory
+            stdin
+            primCasDir
+            False
+            False
+            (WakeboxMountOp "workspace" "" "." False, Nil)
+            ""
+            ""
+            None
+            None
+            None
+            script
+
         def writeResult =
-            write specFile (prettyJSON json)
-            | addErrorContext "Failed to 'write {specFile}: '"
+            writeWakeboxSpec spec 2 specFile
+            | addErrorContext "Failed to write {specFile}: "
             | rmapFail (markJobSetupFailure job)
 
         require Pass _ = writeResult
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -367,8 +367,6 @@ export def defaultRunner: Runner =
             fuseRunner
 
 # Mount operations for the wakebox spec.
-# These constructors match validate_mount() and do_mounts() in wakebox.
-# See: src/wakefs/namespace.cpp
 # wake-format off
 export data WakeboxMountOp =
     WakeboxBindOp          (source: String) (dest: String) (readonly: Boolean)
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -383,6 +383,7 @@ export tuple WakeboxMountOp =
 
 # Fields consumed by the wakebox/fuse sandbox when reading the spec JSON.
 export tuple WakeboxSpec =
+    export Label: String
     export Command: List String
     export Environment: List String
     export Visible: List Path
@@ -444,6 +445,7 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
 
         def spec =
             WakeboxSpec
+            label
             command
             environment
             visible
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -380,7 +380,7 @@ export data WakeboxMountOp =
 
 # Fields consumed by the wakebox/fuse sandbox when reading the spec JSON.
 export tuple WakeboxSpec =
-    export Label: String
+    export Label: Option String
     export Command: List String
     export Environment: List String
     export Visible: List Path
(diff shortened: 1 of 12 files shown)


src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+  if (broken) {
+    // Suspend until this Promise is fulfilled, then retry from scratch.
+    next = nullptr;
+    broken->await(runtime, this);
+    return;
+  }
+
+  // Propagate validation errors (e.g. unsupported PathType) as Fail results.
+  std::string err =
+      data.error.empty() ? write_spec_json(data, filepath->c_str(), indent) : data.error;
```
AbrarQuazi: From the looks of it, we are storing 3x heap usage for one spec.json file: the Wake tuple, `SpecData`, and the `JAST` tree representation that you buildup in `write_spec_json`

I think we can make this better by getting rid of the `SpecData` and `JAST` representations by converting `collect_spec` into Promise resolve checker like:
```
  // 1. First pass: just check that everything is fulfilled.
  if (Promise *broken = check_spec_ready(spec.get())) {
    next = nullptr; broken->await(runtime, this); return;
  }
```

then after the promises are resolved, we stream each JSON syntax inline to file as we walk the tree, instead of building up the JAST structure

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+// ---------------------------------------------------------------------------
+
+namespace {
+
+// ---------------------------------------------------------------------------
+// Collected plain-C++ representation of a WakeboxSpec
+// ---------------------------------------------------------------------------
+
+struct VisEntry {
+  std::string path;
```
AbrarQuazi: Should these (and structs below) be `string_view`?

AbrarQuazi: I think we would benefit a lot from it, zero copy allocations plus if we integrate streaming to a file - especially since we store a lot of strings

share/wake/lib/system/runner.wake
```diff
@@ -366,6 +366,46 @@ export def defaultRunner: Runner =
...
+
+# Fields consumed by the wakebox/fuse sandbox when reading the spec JSON.
+export tuple WakeboxSpec =
+    export Label: String
+    export Command: List String
+    export Environment: List String
+    export Visible: List Path
+    export Directory: String
+    export Stdin: String
+    export CasDir: String
```
ag-eitilt: Does this need to be `Option String` until we finish rolling the CAS out?

share/wake/lib/system/runner.wake
```diff
@@ -366,6 +366,46 @@ export def defaultRunner: Runner =
...
+    export IsolateNetwork: Boolean
+    export IsolatePids: Boolean
+    export MountOps: List WakeboxMountOp
+    export Hostname: String
+    export Domainname: String
+    export UserId: Option Integer
+    export GroupId: Option Integer
+    export CommandTimeout: Option Integer
+    export Runner: String
```
ag-eitilt: There's a good chance I'll find the answer later in the code, but why does wakebox care about the `Runner` name?  It shouldn't affect the sandboxing behaviour -- different runners should either set up the rest of the parameters differently, or change the script being run within the sandbox.  Having the sandbox implementation hardcoding behaviour based on what's invoking it (beyond the rest of the tuple parameters) feels a bit dirty.

share/wake/lib/system/runner.wake
```diff
@@ -366,6 +366,46 @@ export def defaultRunner: Runner =
...
+    export Environment: List String
+    export Visible: List Path
+    export Directory: String
+    export Stdin: String
+    export CasDir: String
+    export IsolateNetwork: Boolean
+    export IsolatePids: Boolean
+    export MountOps: List WakeboxMountOp
+    export Hostname: String
+    export Domainname: String
```
ag-eitilt: Nit:  I know these are one word in the system configuration scope, but they look better camel-cased within Wake.
```suggestion
    export HostName: String
    export DomainName: String
```

share/wake/lib/system/runner.wake
```diff
@@ -388,84 +428,40 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
...
+            stdin
+            primCasDir
+            False
+            False
+            (WakeboxFuseWorkspaceOp ".", Nil)
+            ""
+            ""
+            None
+            None
+            None
```
ag-eitilt: Probably add comments to (at least) these "zero" elements to make it clear what fields they're corresponding to.

Separately, let's not be defaulting to empty strings.  Either make the host/domain `Option String`, or (even better) assign reasonable but meaningless static defaults that superficially have the proper form for the respective names.

src/runtime/record_helpers.h
```diff
@@ -0,0 +1,104 @@
...
+//   TRY(rh::integer(rec, 1, data.count));
+//   #undef TRY
+// ---------------------------------------------------------------------------
+
+namespace rh {
+
+inline Promise *string(Record *rec, int idx, std::string &out) {
+  Promise *p = rec->at(idx);
+  if (!*p) return p;
+  out = p->coerce<String>()->as_str();
```
ag-eitilt, line 49: To carry over what Abrar's looking at to here as well, these getters in particular feel like we're copying the contents into the reference object rather than returning a zero-cost pointer.  This is definitely *safer* since we aren't risking dangling pointers, but like was mentioned in the meeting that may or may not be a problem in practice depending on how quickly we use those contents.

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+  std::string type;
+  std::string hash;
+  long mode;
+  long mtime;
+};
+
+struct MountEntry {
+  std::string type;
+  std::string source;  // empty = omit from JSON
```
ag-eitilt: I know this is C++ rather than anything with Option types, but it's still disappointing whenever I see special behaviour assigned to empty strings.

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+};
+
+// Collect all fields from a WakeboxMountOp ADT value into a MountEntry.
+// WakeboxMountOp constructor indices (runner.wake declaration order):
+//   0=WakeboxBindOp(source,dest,readonly), 1=WakeboxSquashfsOp(source,dest),
+//   2=WakeboxTmpfsOp(dest), 3=WakeboxCreateDirOp(dest),
+//   4=WakeboxCreateFileOp(dest), 5=WakeboxFuseWorkspaceOp(dest)
+static Promise *collect_mount_entry(Record *entry, MountEntry &me) {
+  static const char *strings[] = {"bind",       "squashfs",    "tmpfs",
+                                  "create-dir", "create-file", "workspace"};
```
ag-eitilt: Since the number of columns in the docs don't match the number of columns in the declaration, it would probably be more clear to have each of the elements on a separate line and the comments to their right.

src/runtime/wakebox_prim.cpp
AbrarQuazi: Allocation | Region | Size (per spec, N paths) | Lifetime
-- | -- | -- | --
WriteWakeboxJson continuation | Wake GC heap | ~64 B | Until next GC after resume
SpecData outer struct | C++ stack | ~400 B | One execute() call
std::string payloads in SpecData | C++ heap | ~50 small + ~2N path strings | One execute() call
std::vector buffers in SpecData | C++ heap | O(N) | One execute() call
JAST root struct | C++ stack | ~32 B | One write_spec_json()
JAST children vectors and key/value strings | C++ heap | O(N) — duplicated copy of above | One write_spec_json()
std::ofstream object | C++ stack | ~256 B | One call
std::ofstream internal write buffer | C++ heap | ~4–8 KiB | Until ofstream closes
Streaming-version Pretty + locals (proposed) | C++ stack only | ~24 B per recursion frame | Per call frame

Even though the string payloads will be in the c++ transient heap, and not Wake GC's heap, what could potentially happen is if the spec has unfullfilled promises, it would keep doing redundant work as mentioned in your comment `The caller must discard and reinitialise data on every retry` , and reallocate and recopy strings each time, which string_view + streaming can make cheap

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+static Promise *get_path_type_string(Record *path_rec, std::string &out, std::string &error) {
+  // PathType is at field index 1 of the Path tuple.
+  Promise *p = path_rec->at(1);
+  if (!*p) return p;
+  Record *pt = p->coerce<Record>();
+  // PathType constructor indices (io.wake declaration order):
+  //   0=RegularFile, 1=Directory, 2=CharDevice, 3=BlockDevice,
+  //   4=Fifo,        5=Symlink,   6=Socket
+  static const char *strings[] = {"file",  "directory", nullptr, nullptr,
+                                  nullptr, "symlink",   nullptr};
```
ag-eitilt: This one *does* match the layout, but if you reorganize the last one this should follow suit for consistency.

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+// The caller must discard and reinitialise `data` on every retry since a
+// partial traversal may have partially filled vectors.
+static Promise *collect_spec(Record *spec, SpecData &data) {
+  Promise *broken;
+
+#define TRY(expr)              \
+  do {                         \
+    broken = (expr);           \
+    if (broken) return broken; \
+  } while (0)
```
ag-eitilt: I'm sure this is a valid C++ pattern, but it feels particularly indirect.  Why bother with the `do...while` if it's never used to loop?

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+static void write_jast_pretty(std::ostream &os, const JAST &jast, int indent, int depth) {
+  switch (jast.kind) {
+    case JSON_OBJECT: {
+      if (jast.children.empty()) {
+        os << "{}";
+        return;
+      }
+      std::string inner(static_cast<size_t>((depth + 1) * indent), ' ');
+      std::string outer(static_cast<size_t>(depth * indent), ' ');
+      os << "{\n";
```
ag-eitilt: Do we want to turn off the newlines if `indent==0`?  And we might also want to prevent negative indents just in case we wind up exposing this function: `indent = max(0, indent);`

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+      for (size_t i = 0; i < jast.children.size(); ++i) {
+        os << inner << '"' << json_escape(jast.children[i].first) << "\": ";
+        write_jast_pretty(os, jast.children[i].second, indent, depth + 1);
+        if (i + 1 < jast.children.size()) os << ',';
```
ag-eitilt: A more straightforward test would be to move the comma+newline to the start of the loop and then
```suggestion
        if (i) os << ',';
```
Or more explicitly
```suggestion
        if (i > 0) os << ',';
```
(And the same for arrays.)

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+        os << inner;
+        write_jast_pretty(os, jast.children[i].second, indent, depth + 1);
+        if (i + 1 < jast.children.size()) os << ',';
+        os << '\n';
+      }
+      os << outer << ']';
+      break;
+    }
+    default:
+      os << jast;
```
ag-eitilt: How come we can happily use the default stringifier for most types, but not for arrays/objects?

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+  for (const auto &me : d.mounts) {
+    JAST &obj = mops.add(JSON_OBJECT);
+    obj.add("type", me.type);
+    if (!me.source.empty()) obj.add("source", me.source);
+    obj.add("destination", me.dest);
+    if (me.has_read_only) obj.add_bool("read_only", me.read_only);
+  }
+
+  if (!d.hostname.empty()) root.add("hostname", d.hostname);
+  if (!d.domainname.empty()) root.add("domainname", d.domainname);
```
ag-eitilt: Ah, this is why there's the empty strings. 😞

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,500 @@
...
+  if (d.has_user_id) root.add("user-id", d.user_id);
+  if (d.has_group_id) root.add("group-id", d.group_id);
+  if (d.has_command_timeout) root.add("command-timeout", d.command_timeout);
+  if (!d.runner.empty()) root.add("runner", d.runner);
+
+  // Unlink any existing file first so we don't fail on a stale copy.
+  (void)unlink(filepath);
```
ag-eitilt: Thought it through to double-check and we shouldn't be running into race conditions or concurrency issues here in practice.  There's nothing in the function itself to prevent unlinking a file some other wakebox is just about to read (nor, potentially worse, partially writing a new file at that location when another wakebox opens it halfway through), but even in multi-wake environments we should be being fed unique paths based on the (shared) database assigning unique IDs to each job.  And adding logic to this function to *be sure* it's safe would just bog it down for no practical benefit.

ag-eitilt (approved):
A lot of comments, but nothing critical, and the general quality is high enough that so long as the testing looks good I don't see any reason I'd object to it getting merged.

For the record: we've discussed live whether we should be tying wakebox more tightly to wake in this way, and come to the conclusion that the performance benefits outweigh purity/modularity concerns, particularly given how integral a pillar wakebox has become.  We might still change how the sandbox is managed *within* wakebox (e.g. fuse->overlayfs) or present even lighter-weight runners which sidestep wakebox limits/overhead (e.g. `sandboxDirRunner`), but integrating wakebox like this allows for a good degree of optimization and we don't expect structural changes to the degree that the integration would become a liability.

share/wake/lib/system/runner.wake
ngraybeal: It doesn't, same with Label. I left them purely out of convenience.

share/wake/lib/system/runner.wake
ngraybeal: It doesn't appear to trigger anything and retains 1:1 behavior of previous implementation.

share/wake/lib/system/runner.wake
ngraybeal: Well I could be convinced of DomainName, but maybe not Hostname.

share/wake/lib/system/runner.wake
ngraybeal: Agree on the option string.

src/runtime/record_helpers.h
ngraybeal: I don't know the full risk here, so I probably want to circle back to that optimization.

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,550 @@
...
+static const char *k_mount_type_strings[] = {
+    "bind",         // 0 WakeboxBindOp
+    "squashfs",     // 1 WakeboxSquashfsOp
+    "tmpfs",        // 2 WakeboxTmpfsOp
+    "create-dir",   // 3 WakeboxCreateDirOp
+    "create-file",  // 4 WakeboxCreateFileOp
+    "workspace",    // 5 WakeboxFuseWorkspaceOp
+};
+
+// Stream spec JSON directly to filepath without an intermediate SpecData or
```
AbrarQuazi: We can clean up this comment as well as below mentioning SpecData and JAST tree

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,550 @@
...
+// JAST tree. All Promises must be fulfilled (call check_spec_ready first).
+// Returns an empty string on success, or an error message on failure.
+static std::string stream_spec_json(Record *spec, const char *filepath, int indent) {
+  (void)unlink(filepath);
+  std::ofstream out(filepath, std::ios_base::trunc);
```
AbrarQuazi, line 264: I am a bit worried about this function and its compatibility with multi-wake. Although I think master might have this issue as well, where 2 multi-wake instances can write to the same spec JSON? The `spec-{getJobId job}` prefix is not unique enough right?

Maybe we should either add the pid into the prefix as well, or stream to a temp file and autonomically swap

ngraybeal: Currently filepath location/name is all set by the caller of the prim. Since we dont' return a path or string, I wouldn't want to change it as the runner relies on the file.

AbrarQuazi: I still think atleast we would want the atomic rename change here, especially if we can get partial writes

src/runtime/wakebox_prim.cpp
```diff
@@ -0,0 +1,549 @@
...
+enum FieldKind {
+  FK_SCALAR,      // String / Boolean / Integer — check the top-level Promise only
+  FK_OPTION,      // Option T — check outer Promise; if Some, check the inner Promise
+  FK_LIST_STR,    // List String — walk nodes, check head and tail Promises
+  FK_LIST_PATH,   // List Path — walk nodes, check head + all 5 Path field Promises
+  FK_LIST_MOUNT,  // List WakeboxMountOp — walk nodes, check head + variable field count
+};
+
+static Promise *check_spec_ready(Record *spec) {
+  static const FieldKind kFields[] = {
```
AbrarQuazi: Not exactly loving this, hard to see what each field represents, and comments can always go out of order (a change to the middle of the list will mean changing all of the ones after)

what if we had a list of tuples instead:
```
// Order MUST match `tuple WakeboxSpec` in share/wake/lib/system/runner.wake.
// Field index = position in this table.
static const SpecField kSpec[] = {
  { "label",           FieldKind::Option_String  },
  { "command",         FieldKind::List_String    },
  { "environment",     FieldKind::List_String    },
  { "stdin",           FieldKind::Scalar_String  },
...
};

```
Or some better form of ordering?

src/runtime/wakebox_prim.cpp
ag-eitilt: > The `spec-{getJobId job}` prefix is not unique enough right?

My understanding is that the Job ID is reserved in the database before invoking the Runner, but after determining whether or not to load cached results or rerun the job.  In other works, even if the "same" job is launched twice by different multi-wake instances, they'd be assigned different IDs, and thus have different prefixes.  An atomic rename could certainly be *nice*, but so far as I know the only time we'd actually be risking two things writing to the same file is if the database got deleted between the invocations -- which I don't think is a use case we want to optimize for.

share/wake/lib/system/runner.wake
```diff
@@ -366,6 +366,45 @@ export def defaultRunner: Runner =
...
+# wake-format off
+export data WakeboxMountOp =
+    WakeboxBindOp          (source: String) (dest: String) (readonly: Boolean)
+    WakeboxSquashfsOp      (source: String) (dest: String)
+    WakeboxTmpfsOp                          (dest: String)
+    WakeboxCreateDirOp                      (dest: String)
+    WakeboxCreateFileOp                     (dest: String)
+    WakeboxFuseWorkspaceOp                  (dest: String)
+
+# Fields consumed by the wakebox/fuse sandbox when reading the spec JSON.
```
AbrarQuazi, line 380: We should mention how any changes in ordering here will need to change the C++ layer/file 

same with `WakeboxMountOp` above and `PathType` in io.wake

AbrarQuazi (approved):

PR #1869, opened 2026-06-09
Staging dir runner

AbrarQuazi:
Adds stagingDirNsRunner, a non-FUSE runner that executes each job directly inside a private per-job staging directory, then feeds its outputs through the CAS pipeline.

- Allocates a fresh per-job staging directory under {cas_dir}/staging/.
- Materializes the job's visible inputs into that directory at their workspace-relative paths.
- Bind-mounts the directory read-write as . inside a wakebox user/mount namespace and runs the job there 
- After the job exits, harvests the declared outputs (fnOutputs) directly from the staging tree and hands them to job finalization for hashing + CAS ingest, then materializes them into the workspace.
- Cleans up the staging directory via a PostFinalize hook (after CAS ingest completes).

### DIFF
--- a/share/wake/lib/system/cas.wake
+++ b/share/wake/lib/system/cas.wake
@@ -27,6 +27,20 @@ export def primCasBlobAbsPathStr (hash: String): Result String Error =
     raw hash
     | rmapError makeError
 
+# Allocate a private per-job directory beneath the CAS staging root.
+export def casAllocStagingDir (prefix: String): Result String Error =
+    def raw prefix = prim "cas_alloc_staging_dir"
+
+    raw prefix
+    | rmapError makeError
+
+# Remove a directory previously returned by casAllocStagingDir.
+export def casRemoveStagingDir (path: String): Result Unit Error =
+    def raw path = prim "cas_remove_staging_dir"
+
+    raw path
+    | rmapError makeError
+
 # Materialize an item from CAS to workspace.
 def primCasMaterializeItem (destPath: String) (itemType: String) (hashOrTarget: String) (mode: Integer) (mtimeSec: Integer) (mtimeNsec: Integer): Result Unit Error =
     def raw destPath itemType hashOrTarget mode mtimeSec mtimeNsec = prim "cas_materialize_item"
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -432,25 +432,32 @@ export def stagingDirNsRunner: Runner =
     def fuse = "{wakePath}/wakebox"
 
     def run (job: Job) (input: RunnerInput): RunnerOutput =
-        def stagingDirResult = casAllocStagingDir "job.{input.getRunnerInputPrefix}"
+        def stagingDirResult = casAllocStagingDir "job-{input.getRunnerInputPrefix}"
 
         require Pass stagingDir = stagingDirResult
         else
-            def err = stagingDirResult | getWhenPass "Failed to allocate staging directory".makeError
+            def err = match stagingDirResult
+                Fail e -> e
+                Pass _ -> unreachable "require Pass else guarantees Fail here"
             def _ = markJobSetupFailure job err
 
             RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some err)
 
         def cleanup _ =
-            def _ = casRemoveStagingDir stagingDir
+            match (casRemoveStagingDir stagingDir)
+                Pass _ -> Unit
+                Fail err ->
+                    def _ = printlnLevel logWarning "Failed to remove staging dir {stagingDir}: {err.getErrorCause}"
 
-            Unit
+                    Unit
 
         def stageResult = stageVisibleInputs stagingDir input.getRunnerInputVisible
 
         require Pass _ = stageResult
         else
-            def err = stageResult | getWhenPass "Failed to stage visible inputs".makeError
+            def err = match stageResult
+                Fail e -> e
+                Pass _ -> unreachable "require Pass else guarantees Fail here"
             def _ = cleanup Unit
             def _ = markJobSetupFailure job err
 
@@ -487,7 +494,9 @@ export def stagingDirNsRunner: Runner =
 
         require Pass (Pair stagedOutputs cleanableOutputs) = stagingResult
         else
-            def err = stagingResult | getWhenPass "Failed to stage declared outputs".makeError
+            def err = match stagingResult
+                Fail e -> e
+                Pass _ -> unreachable "require Pass else guarantees Fail here"
             def _ = cleanup Unit
 
             innerOutput
@@ -545,7 +554,9 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
 
         require Pass mountOps = mountOpsResult
         else
-            def runnerError = mountOpsResult | getWhenPass "Failed to create wakebox mount operations".makeError
+            def runnerError = match mountOpsResult
+                Fail e -> e
+                Pass _ -> unreachable "require Pass else guarantees Fail here"
             def _ = markJobSetupFailure job runnerError
 
             RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some runnerError)
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -423,6 +423,11 @@ def stageVisibleInputs (stagingDir: String) (visibles: List Path): Result Unit E
 
 # Runs a job in a private copy of its declared inputs, then stages declared outputs directly
 # from that private tree. This avoids a FUSE daemon and keeps concurrent Wake runs isolated.
+#
+# Only explicit/constant `FnOutputs` (e.g. `setPlanFnOutputs`) is supported. `setPlanFilterOutputs`
+# silently returns no outputs here (`filter f Nil = Nil`), and since the staging root is deleted
+# after finalization, undeclared outputs are lost, not just unpublished. Use `workspaceRunner`/
+# `fuseRunner` for filter-based `FnOutputs`.
 export def stagingDirNsRunner: Runner =
     def fuse = "{wakePath}/wakebox"
 
@@ -451,13 +456,24 @@ export def stagingDirNsRunner: Runner =
 
             RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some err)
 
+        # This runner disables FUSE tracking, so `jsonInputs` is always empty; conservatively
+        # report every visible as an input instead (like `localRunner`) to avoid unconditional
+        # cache hits. Pass `identity` inward as `FnInputs` (no-op) and call the caller's real
+        # `FnInputs` exactly once out here against the full visible set.
+        def visiblePaths = input.getRunnerInputVisible | map getPathName
         def declaredOutputs = input.getRunnerInputFnOutputs Nil
-        def innerInput = input | setRunnerInputFnOutputs (\_ Nil)
+        def innerInput =
+            input
+            | setRunnerInputFnOutputs (\_ Nil)
+            | setRunnerInputFnInputs identity
         def plan =
             makeJSONRunnerPlan fuse
             | setJSONRunnerPlanMkMountOps (stagingMountOps stagingDir)
         def (Runner _ innerRun) = makeJSONRunner plan
-        def innerOutput = innerRun job innerInput
+        def reportedInputs = input.getRunnerInputFnInputs visiblePaths
+        def innerOutput =
+            innerRun job innerInput
+            | setRunnerOutputInputs reportedInputs
 
         require None = innerOutput.getRunnerOutputRunnerError
         else
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -413,10 +413,14 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
     f spec indent filePath
     | rmapError makeError
 
-def stagingMountOps (stagingDir: String) (_input: RunnerInput): Result (List WakeboxMountOp) Error =
+# Produces the sandbox mount ops for a staging-dir based runner: just a single
+# bind mount of `stagingDir` onto the sandbox root (".")
+export def stagingMountOps (stagingDir: String) (_input: RunnerInput): Result (List WakeboxMountOp) Error =
     Pass (WakeboxBindOp stagingDir "." False, Nil)
 
-def stageVisibleInputs (stagingDir: String) (visibles: List Path): Result Unit Error =
+# Copies each visible input `Path` into `stagingDir`, preserving its relative
+# workspace path, so a job can run against a private snapshot of its inputs.
+export def stageVisibleInputs (stagingDir: String) (visibles: List Path): Result Unit Error =
     def raw workspace stagingDir relPaths = prim "stage_visible_inputs"
 
     raw workspace stagingDir (visibles | map getPathName | catWith "\n")
--- a/share/wake/lib/system/runner.wake
+++ b/share/wake/lib/system/runner.wake
@@ -238,15 +238,11 @@ def stageWorkspaceOutputsImpl (inPlace: Boolean) (stageFromDir: String) (allOutp
         require True = subset scmp normAllOutputPaths stagedOutputPaths
         else failWithError "Missing staged metadata for workspace outputs"
 
-        def cleanableOutputs =
-            if inPlace then
-                (baseCleanableOutputs ++ stagedOutputPaths)
-                | distinctBy scmp
-            else
-                (baseCleanableOutputs ++ stagedOutputPaths ++ (stagingRoot, Nil))
-                | distinctBy scmp
-
         def stagingRootOpt = if inPlace then None else Some stagingRoot
+
+        def cleanableOutputs =
+            (baseCleanableOutputs ++ stagedOutputPaths ++ optionToList stagingRootOpt)
+            | distinctBy scmp
 
         Pass (
             Pair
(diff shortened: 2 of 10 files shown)


share/wake/lib/system/cas.wake
```diff
@@ -27,6 +27,23 @@ export def addCasHashEnv = match (getenv "WAKE_CAS")
 export def primCasDir: String =
     prim "cas_dir"
 
+# Allocate a fresh per-job staging directory under {cas_root}/staging/.
+# The directory name is "{sanitized_prefix}.{pid}.{counter}" so concurrent
```
ngraybeal: Would you be open to using `-` instead of `.` for name separator?

share/wake/lib/system/runner.wake
ngraybeal, line 1: This will conflict with https://github.com/sifiveinc/wake/pull/1867.

src/cas/cas.cpp
```diff
@@ -40,6 +40,27 @@ namespace cas {
...
+static std::atomic<uint64_t> g_alloc_job_staging_counter{0};
+
+// Sanitize a staging-dir prefix: drop any characters that could let the
+// caller escape the staging root, and cap length so directory names stay sane.
+static std::string sanitize_staging_prefix(const std::string& prefix) {
+  static constexpr size_t kMaxPrefixLen = 64;
+  std::string out;
+  out.reserve(prefix.size());
+  for (char c : prefix) {
+    if (c == '/' || c == '\\' || c == '\0' || c == '.') {
```
ngraybeal, line 51: do we want to allow `./`?

src/cas/cas.cpp
```diff
@@ -305,4 +326,49 @@ wcl::result<bool, CASError> Cas::materialize_blob(const ContentHash& hash,
 
   return wcl::make_result<bool, CASError>(true);
 }
+
+wcl::result<std::string, CASError> Cas::alloc_job_staging_dir(const std::string& prefix) const {
+  std::string sanitized = sanitize_staging_prefix(prefix);
+  std::string name = sanitized + "." + std::to_string(getpid()) + "." +
```
ngraybeal: As noted above, can we change separator away from `.` to `-`? or have we already used `.` elsewhere?

share/wake/lib/system/runner.wake
```diff
@@ -433,6 +623,34 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
...
+
+        require Pass mountOpsJson = mountOpsResult
+        else
+            def runnerError =
+                mountOpsResult
+                | getWhenPass "mkMountOps failed".makeError
+
+            def _ = markJobSetupFailure job runnerError
+
+            RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some runnerError)
```
ngraybeal: Perhaps use a match. otherwise you have an `unreachable` coded in the else.

share/wake/lib/system/runner.wake
```diff
@@ -346,6 +391,140 @@ export def fuseRunner: Runner =
...
+
+# Materialize the entire visible-input set into stagingDir at workspace-relative paths.
+# Files become CoW reflinks; symlinks are recreated with the same target string;
+# directories are mkdir'd as empty entries (children directories are not auto-populated,
+# they must be declared as a visible input).
+def stageVisibleInputs (stagingDir: String) (visibles: List Path): Result Unit Error =
+    def relPaths = visibles | map getPathName | catWith "\n"
+    def raw ws sd rp = prim "stage_visible_inputs"
+
+    raw workspace stagingDir relPaths
```
ngraybeal: Can we just not pass Paths directly?  Do we need `relPaths`?

share/wake/lib/system/runner.wake
```diff
@@ -346,6 +391,140 @@ export def fuseRunner: Runner =
...
+
+    def run (job: Job) (input: RunnerInput): RunnerOutput =
+        def prefix = input.getRunnerInputPrefix
+        def stagingDirResult = casAllocJobStagingDir "job.{prefix}"
+
+        require Pass stagingDir = stagingDirResult
+        else
+            def err =
+                stagingDirResult
+                | getWhenPass "stagingDirNs: failed to allocate staging dir".makeError
```
ngraybeal: you have an `unreachable`. perhaps use a match?

share/wake/lib/system/runner.wake
```diff
@@ -346,6 +391,140 @@ export def fuseRunner: Runner =
...
+
+            def _ = markJobSetupFailure job err
+
+            RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some err)
+
+        def cleanupStaging _ =
```
ngraybeal: is both `_` and `Unit` acceptable?

share/wake/lib/system/runner.wake
```diff
@@ -346,6 +391,140 @@ export def fuseRunner: Runner =
...
+            def _ = casRemoveJobStagingDir stagingDir
+
+            Unit
+        def stageResult = stageVisibleInputs stagingDir input.getRunnerInputVisible
+
+        require Pass _ = stageResult
+        else
+            def err =
+                stageResult
+                | getWhenPass "stagingDirNs: visible pre-materialization failed".makeError
```
ngraybeal: another unreachable.

share/wake/lib/system/runner.wake
```diff
@@ -346,6 +391,140 @@ export def fuseRunner: Runner =
...
+        # PostFinalize anyway, so renaming sources away is a no-op for cleanup.
+        def declaredOutputs = origFnOutputs Nil
+        def innerCleanable = innerOutput.getRunnerOutputCleanableOutputs
+
+        require Pass (Pair stagedOutputs cleanableOutputs) =
+            stageWorkspaceOutputsInPlace stagingDir declaredOutputs declaredOutputs innerCleanable
+        else
+            def err =
+                stageWorkspaceOutputsInPlace stagingDir declaredOutputs declaredOutputs innerCleanable
+                | getWhenPass "stagingDirNs: post-exec output staging failed".makeError
```
ngraybeal: unreachable

share/wake/lib/system/runner.wake
```diff
@@ -346,6 +391,140 @@ export def fuseRunner: Runner =
...
+            def _ = cleanupStaging Unit
+            def _ = markJobSetupFailure job err
+
+            RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some err)
+
+        # Capture the original fnOutputs so we can call it exactly once after the
+        # inner runner returns (the `must be called exactly once` invariant).
+        def origFnOutputs = input.getRunnerInputFnOutputs
+
+        # The inner makeJSONRunner expects a *filter* over wakebox-reported outputs. Because
```
ngraybeal: This doesn't quite make sense to me in first read. Seems like we are working around something using something that already exist instead of handling it ourselves.

share/wake/lib/system/runner.wake
AbrarQuazi: In the c++ layer, we dont have a representation of `List Path` (we only have runtime data types of string, int, double )

share/wake/lib/system/runner.wake
AbrarQuazi: `_ => Unit` is an acceptable convention, but I realized I should log a warning here at the very least if it fails to cleanup

share/wake/lib/system/runner.wake
```diff
@@ -247,15 +239,28 @@ export def stageWorkspaceOutputs (allOutputPaths: List String) (filteredOutputPa
...
-            (baseCleanableOutputs ++ stagedOutputPaths ++ (stagingRoot, Nil))
-            | distinctBy scmp
+            if inPlace then
+                (baseCleanableOutputs ++ stagedOutputPaths)
+                | distinctBy scmp
+            else
+                (baseCleanableOutputs ++ stagedOutputPaths ++ (stagingRoot, Nil))
+                | distinctBy scmp
+
+        def stagingRootOpt = if inPlace then None else Some stagingRoot
```
ag-eitilt: *Pretty* sure this is a stdlib utility function.
```suggestion
        def stagingRootOpt = if inPlace then None else Some stagingRoot

        def cleanableOutputs =
            (baseCleanableOutputs ++ stagedOutputPaths ++ optionToList stagingRootOpt)
            | distinctBy scmp
```

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
     f spec indent filePath
     | rmapError makeError
 
+# Produces the sandbox mount ops for a staging-dir based runner: just a single
+# bind mount of `stagingDir` onto the sandbox root (".")
+export def stagingMountOps (stagingDir: String) (_input: RunnerInput): Result (List WakeboxMountOp) Error =
+    Pass (WakeboxBindOp stagingDir "." False, Nil)
```
ag-eitilt, line 415: Is this what applies the virtual `/workspace` mount?

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+# workspace path, so a job can run against a private snapshot of its inputs.
+export def stageVisibleInputs (stagingDir: String) (visibles: List Path): Result Unit Error =
+    def raw workspace stagingDir relPaths = prim "stage_visible_inputs"
+
+    raw workspace stagingDir (visibles | map getPathName | catWith "\n")
+    | rmapError makeError
+    | addErrorContext "Failed to stage visible inputs into {stagingDir}: "
+
+# Runs a job in a private copy of its declared inputs, then stages declared outputs directly
+# from that private tree. This avoids a FUSE daemon and keeps concurrent Wake runs isolated.
```
ag-eitilt: Wording nit: "directly" makes me think it would bypass the CAS, not (as I assume you meant it) bypass a secondary copy for safety during hashing.  I'd explicitly point the latter out rather than assuming we'll remember this is "directly" in relation to `workspaceRunner`, not in relation to `fuseRunner`.

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+#
+# Only explicit/constant `FnOutputs` (e.g. `setPlanFnOutputs`) is supported. `setPlanFilterOutputs`
+# silently returns no outputs here (`filter f Nil = Nil`), and since the staging root is deleted
+# after finalization, undeclared outputs are lost, not just unpublished. Use `workspaceRunner`/
+# `fuseRunner` for filter-based `FnOutputs`.
```
ag-eitilt: This paragraph is *ultimately* conveying what you want, but that meaning still feels obscured under several layers of indirection.  The most important thing to document is simply that it doesn't offer output file detection, so the outputs list is empty without manual work on the part of the user.  That's not *exactly* the same as requiring "explicit/constant `FnOutputs`" -- calling code could still very easily use `setPlanFilterOutputs` so long as something earlier populated `FnOutputs`.  And the undeclared outputs handling is something that most users will forget about or are unaware of, so it should be phrased in terms of deliberately using that two-step `FnOutputs`+`FilterOutputs` construction if there are side-effect files which should still be copied to the workspace despite skipping the cache, rather than leaving the comparison to `fuseRunner` implicit (and so far as I'm aware, we've been calling those "unhashed" outputs rather than "unpublished" ones).

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+export def stagingDirNsRunner: Runner =
```
ag-eitilt: I assume the `*Ns*` is indicative of "namespace".  I'd consider that *either* an internal detail rather than something which needs to be appended to the primary name, *or* something that's core enough to its implementation that it should be promoted to the primary name.  Either way, I'd get rid of the abbreviation since it's only meaningful to people who are already very aware of the low-level details.
```suggestion
export def stagingDirRunner: Runner =
```
```suggestion
export def namespaceRunner: Runner =
```
```suggestion
export def namespaceDirRunner: Runner =
```

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+        def declaredOutputs = input.getRunnerInputFnOutputs Nil
+
+        def innerInput =
+            input
+            | setRunnerInputFnOutputs (\_ Nil)
+            | setRunnerInputFnInputs identity
+
+        def plan =
+            makeJSONRunnerPlan fuse
+            | setJSONRunnerPlanMkMountOps (stagingMountOps stagingDir)
```
ag-eitilt, line 486: Since this isn't fuse-based, what is the purpose of the `fuse` argument?  Can we rework the function call to make it unnecessary?  At the very least it should be renamed.  And does `wakebox` even have a non-fuse mode?  Some of my confusion is likely coming from how I imagined this would be built on `localRunner`, not on `jsonRunner`.

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+
+        require Pass stagingDir = stagingDirResult
+        else
+            def err = match stagingDirResult
+                Fail e -> e
+                Pass _ -> unreachable "require Pass else guarantees Fail here"
+
+            def _ = markJobSetupFailure job err
+
+            RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some err)
```
ag-eitilt: This is a large and largely-repeated `else` block.  I'd strongly recommend defining it once as a (nested) function and then calling that rather than inlining everything.  At the very least, do so for the `def err = ...` -- as it is, the subtle differences between the blocks are lost in relation to the size of the `match`.

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+
+        def cleanup _ = match (casRemoveStagingDir stagingDir)
+            Pass _ -> Unit
+            Fail err ->
+                def _ =
+                    printlnLevel logWarning "Failed to remove staging dir {stagingDir}: {err.getErrorCause}"
+
+                Unit
```
ag-eitilt:
```suggestion
                printlnLevel logWarning "Failed to remove staging dir {stagingDir}: {err.getErrorCause}"
```

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+
+            def _ = cleanup Unit
+            def _ = markJobSetupFailure job err
+
+            RunnerOutput Nil emptyStagedOutputs Nil emptyUsage (Some err)
+
+        # This runner disables FUSE tracking, so `jsonInputs` is always empty; conservatively
+        # report every visible as an input instead (like `localRunner`) to avoid unconditional
+        # cache hits. Pass `identity` inward as `FnInputs` (no-op) and call the caller's real
+        # `FnInputs` exactly once out here against the full visible set.
```
ag-eitilt, line 472: It shouldn't just disable fuse tracking, it should be circumventing fuse entirely.  Otherwise, this seems accurate, but will probably be made moot once we remove the visible/input distinction.

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+        def visiblePaths =
+            input.getRunnerInputVisible
+            | map getPathName
+
+        def declaredOutputs = input.getRunnerInputFnOutputs Nil
+
+        def innerInput =
+            input
+            | setRunnerInputFnOutputs (\_ Nil)
+            | setRunnerInputFnInputs identity
```
ag-eitilt, line 482: Why are we intercepting these knobs here instead of passing them through for the inner runner to handle?  I'm not seeing any special behaviour in this runner which would prevent the file lists to be generated at that innermost level, and even if this *did* wind up having its own modifications to apply, the modifications should ideally be implemented as *editing* these functions/results rather than *overwriting* them.

share/wake/lib/system/runner.wake
```diff
@@ -408,6 +413,129 @@ export def writeWakeboxSpec (spec: WakeboxSpec) (indent: Integer) (filePath: Str
...
+        def existingPostFinalize = stagedOutputs.getStagedOutputsPostFinalize
+
+        def postFinalize preparedItems =
+            def _ = existingPostFinalize preparedItems
+            def _ = cleanup Unit
+
+            Unit
+
+        innerOutput
+        | setRunnerOutputOutputs (stagedOutputs | setStagedOutputsPostFinalize postFinalize)
```
ag-eitilt: Leaving aside the question of overwriting the `*Outputs`, this can still be made a bit cleaner with an `edit*` function.
```suggestion
        def postFinalize existingPostFinalize preparedItems =
            def _ = existingPostFinalize preparedItems
            def _ = cleanup Unit

            Unit

        innerOutput
        | setRunnerOutputOutputs (stagedOutputs | editStagedOutputsPostFinalize postFinalize)
```

ag-eitilt (dismissed):
The biggest question have is just basing this on the JSON runner (which I'm still fine with approving so long as the architectural reasoning is explained), but there are a few other cleanup/wording changes I'd want to see made as well.

share/wake/lib/system/runner.wake
AbrarQuazi: So it needs to be able to use wakebox, which currently only the jsonRunner is built on top of, so that it can be used remotely and the remote machines have a visible input json list and calls wakebox - so it can't be built on `localRunner`

I will see if I can go about changing the plan to have different options other than fuse

share/wake/lib/system/runner.wake
AbrarQuazi: There is no mount to ` /workspace`, which is for fuse only. It creates a private per-job filesystem tree, copies declared visible inputs into it, makes that tree the process root (/), and runs the command from there.

So the command being run is mount --bind --recursive <stagingDir> .

share/wake/lib/system/runner.wake
AbrarQuazi: One thing to note however, currently the stagingDirRunner still starts up the fuse daemon- even if it does not use it. I didn't include changes in this PR, because it would mean I have to add some restructuring of the wakebox code that does this - which I think we can do later (I'll create a ticket for this as a follow up) - so thats why I am not changing the comment at the moment

share/wake/lib/system/runner.wake
AbrarQuazi: `setRunnerInputFnOutputs (\_ Nil)` replaces the inner runner's output-selection function with one that always returns no outputs, preventing it from trying to stage private-tree outputs from the real workspace.

`setRunnerInputFnInputs identity` replaces its input-selection function with a no-op, preventing it from applying the callers input filter to wakebox's empty tracking list.

`makeJSONRunner` currently both runs wakebox and handles input/output tracking. This staging runner uses an untracked bind mount, so letting the inner makeJSON runner handle those lists would make it report no inputs and stage outputs from the wrong location. The staging runner applies the callers real functions itself, against the correct lists.

A better design would split `makeJSONRunne`r into a wakebox execution layer and a separate tracking/output-staging layer, so this runner can use the execution layer directly.

share/wake/lib/system/runner.wake
ag-eitilt: It would be nice if we could still get a `/workspace` (which would just be a bindmount of the staging directory, including the visible files and associated directories materialized into it) for consistency, but that's of less importance.  Thanks for the explanation!

share/wake/lib/system/runner.wake
ag-eitilt: Makes sense!  So long as we avoid fuse syscall interception for files within the sandbox (including visible files), I'm *very* happy that we can reuse wakebox here.  If I remember the conversation correctly, we're still launching the daemon but not interacting with it in the mode used by this runner -- I'd consider that perfectly fine for an initial version of this code if avoiding it entirely is tricky; we can optimize things to avoid the unused branch later.  That was certainly a very core part of wakebox originally, so I wouldn't be surprised if there's a lot of small places where the two are tightly integrated, making fuse-circumvention a messy change to add onto this tangential PR.

share/wake/lib/system/runner.wake
ag-eitilt: Ah, yeah, what I just said myself. 😁

share/wake/lib/system/runner.wake
ag-eitilt: I hadn't thought about `makeJSONRunner` potentially muddling those lists, good catch.  In that case, I agree this interception is a good way to go about it -- just make sure you have a test to ensure the output list[^input] does indeed only contain what we'd expect it to and, more importantly, that it remains consistent once we do make that split and can pass these functions on through.  After all, that change isn't going to result in any type-signature changes, so there's nothing structural to prevent potential breaks in the semantics from bubbling up to user code.

[^input]: The input list is probably something to verify as well, but since we'll be removing `*FnInputs` eventually it would be less concerning for that to subtly break.

ag-eitilt (approved):
Thanks for the fixes!

PR #1872, opened 2026-06-12
switch doFormat implementation to foldL

AbrarQuazi:
Currently doFormat uses foldR as its implementation, which has an expensive cost in the heap as it is not tail recursive (the memory is not reclaimable until the recursion bottoms out). Our JSON objects can have extremely long lists of strings (as seen by the compounding memory allocations for the wakebox visible list issue), which makes this ineffective. 

Changing doFormat to use foldL will solve this heap allocation issue (since it is tail recursive), at some cost of time complexity as it needs to reverse the list. This will more so benefit for wide arrays in the JSON (similar to the visible list length), rather than super deeply nested JSON's (which I don't believe we have a use case for)

### DIFF
--- a/share/wake/lib/core/json.wake
+++ b/share/wake/lib/core/json.wake
@@ -108,18 +108,22 @@ def doFormat (fmt: JSONFormat) (lhs: JValue): List String =
             JArray list ->
                 def helper acc value = ",{tabbeder}", rec acc deeper value
 
-                if list.empty then
-                    '[]', rhs
-                else
-                    "[{tabbeder}", foldl helper ("{tabbed}]", rhs) list.reverse | tail
+                # For 0- or 1-element lists the fold direction is irrelevant, so skip
+                # the reverse entirely — this keeps deep, narrow structures (single-element
+                # arrays nested many levels) cheap.
+                match list
+                    Nil -> '[]', rhs
+                    single, Nil -> "[{tabbeder}", rec ("{tabbed}]", rhs) deeper single
+                    _ -> "[{tabbeder}", foldl helper ("{tabbed}]", rhs) list.reverse | tail
             JObject list ->
                 def helper acc (Pair key value) =
                     ",{tabbeder}", '"', jsonEscape key, ("\":{space}", rec acc deeper value)
 
-                if list.empty then
-                    '{}', rhs
-                else
-                    "\{{tabbeder}", foldl helper ("{tabbed}\}", rhs) list.reverse | tail
+                match list
+                    Nil -> '{}', rhs
+                    (Pair key value), Nil ->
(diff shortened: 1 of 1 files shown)


AbrarQuazi:
Some preliminary results for a simple test case for wide arrays:

Shape | Metric | foldR | foldL | Δ
-- | -- | -- | -- | --
FlatArray (50k wide) | Peak live heap | 20.68 MB | 16.11 MB | −22%
  | Peak system alloc | 207 MB | 144 MB | −30%

For deeply nested and narrow list worst case (a single JSON element with 50k nested `[[[[[ ... [null] ... ]]]]]`) we see that foldL suffers here for the extra recursions it has to do (but I account for this case in my recent commit for doFormat)

Metric | foldR | foldL | Δ
-- | -- | -- | --
Peak live heap | 5,966,856 B (5.69 MB) | 6,336,768 B (6.04 MB) | +6% (L slightly worse)
Peak system alloc | 27,872,536 B (26.6 MB) | 43,048,072 B (41.0 MB) | +54% (L worse)

I believe most of the JSON's we deal with wont be super deeply nested + narrow lists, so I do not think we have to worry about this case particularly, there is a tax of reversing the list but I dont believe it to be too bad in terms of time complexity

ngraybeal:
Do you have impact on wall/cpu time?

share/wake/lib/core/json.wake
```diff
@@ -106,20 +106,24 @@ def doFormat (fmt: JSONFormat) (lhs: JValue): List String =
...
-                    "[{tabbeder}", foldr helper ("{tabbed}]", rhs) list | tail
+                def helper acc value = ",{tabbeder}", rec acc deeper value
+
+                # For 0- or 1-element lists the fold direction is irrelevant, so skip
+                # the reverse entirely — this keeps deep, narrow structures (single-element
+                # arrays nested many levels) cheap.
+                match list
+                    Nil -> '[]', rhs
+                    single, Nil -> "[{tabbeder}", rec ("{tabbed}]", rhs) deeper single
+                    _ -> "[{tabbeder}", foldl helper ("{tabbed}]", rhs) list.reverse | tail
```
ag-eitilt, line 117: Do we actually have to reverse here, even for longer lists?  I do agree that we'd need to reverse it to maintain the order of operations, but I don't think that maintaining order of operations is critical.  With a `foldr` and your reversed `foldl`, we're adding each piece of rendered JSON to the left side of the accumulated string; with a non-reversed `foldl` we'd be adding each piece to the right side.

It's a subtly different algorithm, yeah, but it's getting back to the commutative/associative impacts I was talking about the other day.  Given the operations we're performing here, the final string is going to be the same whether it grows right to left or left to right.

EDIT: Another consideration, unless I'm severely misunderstanding how we store the data underlying Strings, we shouldn't be seeing any penalty for adding to the end of the string -- that's pretty isolated to single-linked lists.

PR #1873, opened 2026-06-15
wakebox: fix pidns_init stack overflow

ngraybeal:
Fix sizeof(child_stack) evaluating to 8 bytes (pointer size) instead of the intended 4096, causing a stack overflow in the cloned child.

Replace std::cerr with dprintf(STDERR_FILENO) in pidns_init to avoid inheriting locked C++ locale/stream mutexes across clone() and reduce potential stack depth.


AbrarQuazi (approved):
LGTM thanks!

PR #1874, opened 2026-06-18
Loop invariant lifting pass

AbrarQuazi:
### Problem
Wake currently does not evaluate regular expressions (or any pure expressions) inside a lambda body call efficiently, with these expressions being re-evaluated every call to the lambda which can be extremely expensive and blow up heap memory usage (`RegExp` GC object, and allocating that in each recursive call can blow up live heap).

Wake's compiler should be able to be smarter, and detect what expressions can be 'hoisted' or 'lifted' into the parent enclosing scope so these computations are evaluated once. Hence this PR implements the `Loop Invariant Lifting` (LVL) (or also known as Loop-Invariant Code Motion) Pass. 

Three pieces work together:
1. **The LVL pass (src/optimizer/lvl.cpp)** — analyzes each lambda body and hoists only provably-total terms (literals, tuple alloc/projection, and pure+total prims), never RApp/RDes, never recursive self-references.
2. **A new PRIM_PARTIAL flag** (src/types/primfn.h) — marks pure primitives that can still abort (e.g. div/mod/vget out of bounds, as well as regex match/extract/replace/tokenize). These are excluded from hoisting: a lambda may run zero times, so hoisting a partial op could turn a never-taken abort into a real one.
3. **quote made directly hoistable** (src/runtime/regexp.cpp) — previously quote round-tripped through a fallible stringToRegExp, producing an RDes that blocked hoisting. It now builds the RegExp directly as a pure prim, so the compiled regex hoists out of the loop and is built once and shared — the core fix for the memory blowup.

### Example SSA of the issue that causes memory blow up
```
# The same regex is rebuilt from a string for every element of the list.
export def filterMatching (pattern: String) (lines: List String): List String =
    filter (\line matches (quote pattern) line) lines
```
    
**Before LVL — quote is recompiled inside the loop body**
```
  305 [-,1] = "x"
  306 (filter@wake.anon.loop) [R,2] = Fun(.../list.wake:628):
    returns: 310
    307 (_ 1) [E,1] = <arg>
    308 (filter@wake.anon.loop) [-,1] = Fun(.../list.wake:628):
      returns: 7
      309 (_ tuple_case) [E,0] = <arg>
    309 (filter@wake.anon.loop) [-,1] = Fun(.../list.wake:627):
      returns: 318
      310 (_ tuple_case) [E,2] = <arg>
      311 (_ a1) [-,2] = Get:0(310)
      312 (_ a2) [-,1] = Get:1(310)
      313 (sub) [-,2] = App(306 312)
      314 [-,1] = quote(305)            <-- rebuilt on every element
      315 (_ a0) [-,1] = match(314 311)
      ...
```
quote(305) (slot 314) lives inside the recursive loop body, so the regex is recompiled and re-allocated once per element.

**After LVL — quote is hoisted out; only match stays in the loop**
```
  372 [-,1] = "x"
  373 [-,1] = quote(372)                <-- hoisted out; compiled once, shared
  374 (filter@wake.anon.loop) [R,2] = Fun(.../list.wake:628):
    returns: 378
    375 (_ 1) [E,1] = <arg>
    376 (filter@wake.anon.loop) [-,1] = Fun(.../list.wake:628):
      returns: 7
      377 (_ tuple_case) [E,0] = <arg>
    377 (filter@wake.anon.loop) [-,1] = Fun(.../list.wake:627):
      returns: 386
      378 (_ tuple_case) [E,2] = <arg>
      379 (_ a1) [-,2] = Get:0(378)
      380 (_ a2) [-,1] = Get:1(378)
      381 (sub) [-,2] = App(374 380)
      382 (_ a0) [-,1] = match(373 379) <-- partial match stays in the loop
      ...
```
quote(372) (slot 373) is now lifted into the enclosing scope and computed once, with the loop body referencing it.

### DIFF
--- a/share/wake/lib/core/regexp.wake
+++ b/share/wake/lib/core/regexp.wake
@@ -20,12 +20,7 @@ package wake
 # Create RegExp that only matches str, by escaping special characters.
 # quote "a.b" = `a\.b`
 # quote "hello[world]" = `hello\[world\]`
-export def quote (str: String): RegExp =
-    def res str = prim "quote"
-
-    match (res str).stringToRegExp
-        Pass x -> x
-        Fail error -> unreachable "quote primitive bug: {error.getErrorCause}"
+export def quote (str: String): RegExp = (\_ prim "quote") str
 
 # Concatenate a list of regular expressions.
 # The resulting regular expression must match the elements sequentially.
--- a/share/wake/lib/core/regexp.wake
+++ b/share/wake/lib/core/regexp.wake
@@ -20,7 +20,8 @@ package wake
 # Create RegExp that only matches str, by escaping special characters.
 # quote "a.b" = `a\.b`
 # quote "hello[world]" = `hello\[world\]`
-export def quote (str: String): RegExp = (\_ prim "quote") str
+export def quote (str: String): RegExp =
+    (\_ prim "quote") str
 
 # Concatenate a list of regular expressions.
 # The resulting regular expression must match the elements sequentially.
--- a/src/optimizer/lvl.cpp
+++ b/src/optimizer/lvl.cpp
@@ -23,46 +23,27 @@
 
 #include "ssa.h"
 
-// LVL -- Loop-invariant Lifting (the optimization the PRIM_* flag docs in types/primfn.h
-// name "LVL").  wake has no loops; the analogue is a lambda body, so this is more precisely
-// lambda-invariant code motion, but we keep the codebase's established "LVL" name.
+// LVL -- Loop-invariant Lifting.
 //
-// A function body's terms are re-evaluated on every call (runtime.cpp:128 iterates and
-// interprets every body term).  Any body term that is pure and independent of the function's
-// argument(s) produces the same value on every call, so we hoist it into the enclosing scope
-// where it is computed once instead of once per call.  CSE (which runs right after this pass)
-// then dedups identical hoisted terms originating from different lambdas.
+// WHAT: hoist a lambda body's pure, argument-independent terms into the enclosing scope, so
+//   they run once instead of once-per-call (runtime.cpp re-interprets every body term on every
+//   call).  CSE, the next pass, then dedups identical hoisted terms from different lambdas.
+//   (wake has no loops; a lambda body is the analogue, but we keep the established "LVL" name.)
 //
-// SAFETY: wake evaluates eagerly, but a lambda's body only runs when the lambda is *called*,
-// which may be zero times or conditional (e.g. an `if`/match arm, or `filter pred []`).
-// Hoisting a term out of such a lambda makes it run when the *parent* runs regardless, so the
-// term must be TOTAL -- it must not diverge or abort.  Purity is not enough: a pure prim like
-// `vget` aborts out of range, and `RApp(mid, mid)` (the mutual-recursion fixpoint, tossa.cpp)
-// is pure yet diverges if evaluated eagerly.  We therefore hoist only provably-total terms:
-//   * RLit, RCon, RGet            -- a literal, a tuple allocation, a field projection
-//   * RPrim that is pure, total (!PRIM_PARTIAL) and does not invoke a fn arg (!PRIM_FNARG)
-// and never RApp/RDes (callee/handler totality is unknown; this also excludes recursive
-// calls).  Hoisting a total term that the lambda might not have run only ever wastes one
-// evaluation -- it can never change a result, diverge, or abort.
-//
-// Coordinate system: during the optimizer a body term's args are source-coordinate flat
-// indices.  For an RFun at source position P, let S = P + 1 be the source index of its first
-// body term and N = terms.size().  Because TargetScope::unwind reuses positions, a body
-// term's arg x is always either:
-//   x <  S - 1      a capture from an enclosing scope (already placed in the SourceMap)
-//   x == S - 1      the RFun itself (a recursive self-reference, tossa.cpp:48)
-//   S <= x < S + N  the body term at index x - S
-//   x >= S + N      a later sibling (mutual recursion); desugared, so guarded defensively
-// Only the first and third forms are safe inside a hoisted term: hoisting a term that
-// references the function itself or a later sibling would create a forward reference.
+// WHY "TOTAL", not just "pure": a lambda body runs only when the lambda is called, which may be
+//   zero times or conditional (an `if`/match arm, `filter pred []`).  Hoisting makes a term run
+//   whenever the PARENT runs, so a non-total term could turn a never-taken failure into a real
+//   one.  Purity is necessary but not sufficient -- see lvl_total() for exactly what we hoist.
 
 struct PassLVL {
   TermStream stream;
   PassLVL(TargetScope &scope) : stream(scope) {}
 };
 
-// A term is safe to speculatively evaluate (total) iff it is a literal, a tuple
-// allocation/projection, or a total pure primitive.  RApp/RDes/RArg/RFun are never total here.
+// A term is TOTAL (safe to speculatively evaluate) iff it is a literal, a tuple
+// allocation/projection, or a pure+total primitive.  RApp/RDes can call code of unknown
+// totality (and RApp covers recursive calls, which diverge if eagerly evaluated), and
+// RArg/RFun are never hoistable, so all of those return false.
 static bool lvl_total(Term *t) {
   const std::type_info &id = t->id();
   if (id == typeid(RLit) || id == typeid(RCon) || id == typeid(RGet)) return true;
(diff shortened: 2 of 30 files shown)


share/wake/lib/core/regexp.wake
```diff
@@ -20,12 +20,7 @@ package wake
 # Create RegExp that only matches str, by escaping special characters.
 # quote "a.b" = `a\.b`
 # quote "hello[world]" = `hello\[world\]`
-export def quote (str: String): RegExp =
-    def res str = prim "quote"
-
-    match (res str).stringToRegExp
-        Pass x -> x
-        Fail error -> unreachable "quote primitive bug: {error.getErrorCause}"
```
ag-eitilt, line 28: *Theoretically* the compiler is supposed to be able to detect `unreachable` branches and optimize as if they would never be taken (which is why we have both that and `panic` as separate aborts).  I think we do still benefit from embedding the parse within the `prim` here -- ~~particularly but not solely since this *isn't* unreachable and *should* have been a `panic`~~ EDIT: less convinced of this, but the point still stands -- but you needing to do so in order to get the let-floating to work with `quote` means that some part of the stack is still incomplete.  Not sure if it's incomplete in the implementation here (i.e. it needs to manually ignore `unreachable` branches) or if it's incomplete in an earlier stage of the parser (i.e. it needs better marking/AST for `unreachable` branches), but either way even if this *is* incomplete it's still a massive optimization which will be wonderful to roll out.

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <vector>
+
+#include "ssa.h"
+
+// LVL -- Loop-invariant Lifting.
+//
+// WHAT: hoist a lambda body's pure, argument-independent terms into the enclosing scope, so
+//   they run once instead of once-per-call (runtime.cpp re-interprets every body term on every
+//   call).  CSE, the next pass, then dedups identical hoisted terms from different lambdas.
+//   (wake has no loops; a lambda body is the analogue, but we keep the established "LVL" name.)
```
ag-eitilt, line 31: I'd probably call these `def`s, "inner `def`s", "blocks", "scopes" (less used but more correct), or something else along those lines rather than `lambda`s since I'm pretty sure you're not (just) looking at the proper `(\x fn x)` lambdas.

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+struct PassLVL {
+  TermStream stream;
+  PassLVL(TargetScope &scope) : stream(scope) {}
+};
+
+// A term is TOTAL (safe to speculatively evaluate) iff it is a literal, a tuple
+// allocation/projection, or a pure+total primitive.  RApp/RDes can call code of unknown
+// totality (and RApp covers recursive calls, which diverge if eagerly evaluated), and
+// RArg/RFun are never hoistable, so all of those return false.
```
ag-eitilt, line 46: For my own education, I assume `RArg` isn't hoistable since the arguments are specifically connected to the function they're defined for, but what's the limitation for the entire `RFun`?  (This is probably me not knowing exactly what each constructor corresponds to.)

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+  if (x >= S + N) return ArgKind::LaterSibling;
+  if (x >= S) return ArgKind::BodyTerm;
+  if (x == S - 1) return ArgKind::SelfRef;
+  return ArgKind::Capture;
+}
+
+// Phase 1 -- analysis.  Fill liftable[k] for each body term, in topological (forward) order so
+// that a term's body dependencies have already been decided when we reach it.
+static std::vector<char> markLiftable(const std::vector<std::unique_ptr<Term>> &terms, size_t S) {
+  size_t N = terms.size();
```
ag-eitilt, line 78: Is this per-file or per-repo?  It's a maximum size on the code either way, but while I wouldn't expect individual files to ever reach `size_t` terms, I'm less sure that a gigantic repo in aggregate is never going to reach it.

Not that that's a problem for this PR -- I assume the `terms` vector is already established (and already imposing that limit) somewhere outside this optimization, and any repo which runs into that limit is probably one which severely violates the "if something can be easily implemented as a Job, don't implement it in Wake" guidance, but just for my own knowledge of the theoretical bounds of Wake.

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+  std::vector<char> liftable(N, 0);
```
ag-eitilt, line 79: If you're just assigning `true`/`false` here, why `vector<char>` instead of `vector<bool>`?  (Especially since it looks like `vector<bool>` gets compacted to a bit array rather than being byte-per.)

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Phase 2 -- emit liftable terms into the PARENT scope (before transferring self).  Returns
+// hoisted[k] = the parent slot a hoisted term now lives at, or Term::invalid if not hoisted.
+static std::vector<size_t> hoistLiftable(std::vector<std::unique_ptr<Term>> &terms,
+                                         const std::vector<char> &liftable, size_t S, PassLVL &p) {
+  size_t N = terms.size();
+  std::vector<size_t> hoisted(N, Term::invalid);
+  for (size_t k = 0; k < N; ++k) {
+    if (!liftable[k]) continue;
```
ag-eitilt, line 111: I wanted to suggest a potential micro-optimization at the expense of more difficulty statically verifying correctness -- claim the memory ahead of time but only zero the unhoistable terms right before the `continue` -- but it looks like there's no `std::vector<size_t>(count)` constructor overload.

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+  size_t S = p.stream.map().end() + 1;
+  size_t N = terms.size();
+
+  // 1. Decide which body terms are liftable, then 2. move them into the parent scope.
+  std::vector<char> liftable = markLiftable(terms, S);
+  std::vector<size_t> hoisted = hoistLiftable(terms, liftable, S, p);
+
+  // 3. Emit the RFun node, then its retained body (recursing into nested lambdas).  A hoisted
+  //    term's body slot is mapped onto its new parent slot; everything else is re-emitted.
+  p.stream.transfer(std::move(self));
```
ag-eitilt, line 166: I didn't notice the `p` argument at first, and was confused how the hoisted elements were getting emitted before the `RFun`.  Maybe unify the verbs between "move" in 2 and "emit" in 3?

src/optimizer/lvl.cpp
```diff
@@ -0,0 +1,195 @@
...
+// dispatches one override per Term subtype.  Only RFun does real work.
+
+void RArg::pass_lvl(PassLVL &p, std::unique_ptr<Term> self) { p.stream.transfer(std::move(self)); }
+
+void RLit::pass_lvl(PassLVL &p, std::unique_ptr<Term> self) { p.stream.transfer(std::move(self)); }
+
+void RApp::pass_lvl(PassLVL &p, std::unique_ptr<Term> self) {
+  update(p.stream.map());
+  p.stream.transfer(std::move(self));
+}
```
ag-eitilt, line 133: To double-check that this won't break code expectations:
```
def fn Unit =
    def x =
        def y = factorial 1000
        y
    x

def req guard =
    require True = guard # This gets desugared into a `match`, remember.
    def y = factorial 1000
    y
```
We use the "enclose in a thunk" method to gate function calls which shouldn't be run at a given level, and while it's not as frequent might carefully order `require`s for a similar purpose.  A lot of the time those shouldn't run because they perform some side effect/launch a Job, which as I understand it you've (properly) excluded from being lifted.  However, even a pure function application might be gated in those ways due to something like time to evaluate.  In the example above, I'd expect the factorial call to be lifted outside the `x` block, but *not* be lifted outside the `fn` -- if `fn` is at the top level, doing so would try to compute the factorial even when just running `wake -x Unit` -- and similarly I wouldn't expect it to be lifted before the `require` even if it did still stay within the `req`.

How does the algorithm currently handle those examples?  Unfortunately, they both seem like the kind of lifting which outside of their specific use cases very closely resembles things we *do* want to lift for performance, so finding the right programmatic definitions might get tricky.

tests/optimizer/lvl-capture-chain/test.wake
```diff
@@ -0,0 +1,24 @@
+package lvltest
+
+from wake import _
+
+# Edge case: hoisting a term that depends on a value CAPTURED from an enclosing scope (rather
+# than a plain literal).  A capture is still in scope at the hoist site, so a capture-dependent
+# invariant is liftable; an argument-dependent term is not.
+#
+#   captureMarker  -- depends on the captured `c` (not the element `x`) -> MUST be hoisted
+#   retainedMarker -- depends on the element `x`                         -> MUST stay in body
```
ag-eitilt, line 10: Alignment nit:
```suggestion
#   captureMarker  -- depends on the captured `c` (not the element `x`) -> MUST be hoisted
#   retainedMarker -- depends on the element `x`                        -> MUST stay in body
```
Additionally, you've got different test descriptions between the `pass.sh`s and the `test.wake`s -- you should probably use the same text for both or (my preference) only include the long description in one and restrict the other to a one-line summary.

tests/optimizer/lvl-hoist-total/test.wake
```diff
@@ -0,0 +1,22 @@
...
+from wake import _
+
+# LVL (Loop-invariant Lifting) must hoist a pure, total, argument-independent expression out of
+# a lambda body into the enclosing scope, where it runs once instead of once per call.
+#
+# Each invariant is given a uniquely-named `def` so the test can locate the exact term in the
+# `--stop-after-ssa` dump by its label (def names survive into the SSA; comments do not).
+#
+#   hoistedMarker  -- pure + TOTAL + argument-independent -> MUST be hoisted (run shallower)
+#   retainedMarker -- references the argument `x`          -> MUST stay in the lambda body
```
ag-eitilt, line 12:
```suggestion
#   hoistedMarker  -- pure + TOTAL + argument-independent -> MUST be hoisted (run shallower)
#   retainedMarker -- references the argument `x`         -> MUST stay in the lambda body
```

tests/optimizer/lvl-hoist-total/pass.sh
```diff
@@ -0,0 +1,36 @@
+#! /bin/sh
+
+# LVL hoists a pure, total, argument-independent invariant out of a lambda body.
+#
+# test.wake names two invariants with `def`s so we can find them in the SSA dump by label:
+#   hoistedMarker  -- total + argument-independent -> lifted to an outer (shallower) scope
+#   retainedMarker -- depends on the argument       -> stays in the per-call lambda body
```
ag-eitilt, line 7:
```suggestion
#   hoistedMarker  -- total + argument-independent -> lifted to an outer (shallower) scope
#   retainedMarker -- depends on the argument      -> stays in the per-call lambda body
```

tests/optimizer/lvl-recursion/test.wake
```diff
@@ -0,0 +1,21 @@
+package lvltest
+
+from wake import _
+
+# A recursive call is pure but NOT total: evaluating it eagerly diverges (and, when the
+# recursion builds data, blows up memory).  LVL must never hoist an RApp out of a lambda, both
```
ag-eitilt, line 6: Interesting.  This isn't the definition of "total" I'm familiar with from Haskell (recursive functions are still considered total there so long as every argument successfully produces a valid, finite output), but maybe that's a subtlety enabled by the laziness.

tests/optimizer/lvl-recursion/test.wake
```diff
@@ -0,0 +1,21 @@
...
+# the recursive application stayed inside the conditional branch.
+export def countdown (n: Integer): Integer =
+    if n <= 0 then 0
+    else countdown (n - 1)
+
+# A larger, allocation-shaped recursion: building a list. If the recursive build were eagerly
+# hoisted it would attempt to construct an infinite list and exhaust memory.
+export def buildList (n: Integer): List Integer =
+    if n <= 0 then Nil
+    else n, buildList (n - 1)
```
ag-eitilt, line 21: As illustration:
```haskell
partial x =
    -- infinite list: never terminates, so isn't total
    repeat x

total n x =
    take n $ partial x
```

ag-eitilt (approved):
Got a couple suggested cleanup/future development notes, but so long as the one "double-check this" shows the expected result, I'm (very!) happy with it.

PR #1875, opened 2026-06-23
Remove more artifacts from (multi-wake) testing

ag-eitilt:
This follows up from #1862 in allowing the new tests on the `make test` branch to be run without leaving any random files littering the test directories, and thus the `git status`.  See that PR for notes on the design/process.


tests/runtime/overlap-detection/fail.sh
```diff
@@ -6,3 +6,8 @@ WAKE="${WAKE:-wake}"
 
 rm -f overlap.txt wake.db wake.log
 "${WAKE}" -v test
+
+# Clean up
+err=$?
+rm -f .wake-build-lock .wake/locks/* overlap.txt
```
ag-eitilt, line 12: Do we still have to have the `.wake-build-lock` if we've moved to `.wake/locks/`?  Either way, this will be a merge conflict with #1862 -- this version should be used instead of that one in the context of multi-wake.

tests/runtime/multi-wake-output-same-file/pass.sh
```diff
@@ -13,3 +13,5 @@ rm -f wake.db* wake.log shared.txt err.txt
 ! "${WAKE}" -q -x 'test_diff Unit' > err.txt
 
 sed 's/job \([0-9]\+\)/job -/' err.txt
+
+rm .wake/locks/* shared.txt err.txt
```
ag-eitilt, line 17: This test doesn't have a flaky *return*, but the `.wake/locks/` file is only sometimes present, which might indicate a flaky *testing procedure*.

ag-eitilt:
I've run into a fair bit of (known) flakiness with `standard-library/runner-outputs` in the CI for this PR; nothing I've done here should have affected that directly.  We should probably figure out what's going on there and if there's some way to stabilize it.

ag-eitilt:
Closing this now that the multi-wake branch has been merged to master; all cleanup will be handled by the primary PR.

PR #1876, opened 2026-06-23
Add `-r`/`--recursive` flag to `wake --rm`

ag-eitilt:
This completes the primary file-removal features from #1865 by allowing directories (and all their contents known to Wake) to be removed in addition to individual files.  Following the example of `rm` and `git rm`, doing so requires `-r` or `--recursive` to be passed, and otherwise a warning message is printed and the directories retained.  There are still two boxes to check for future development, but they're lesser priority and can be put off to focus on more critical features for the MVP release:

* In order to properly handle directories, the database has been made the source of truth for what paths should be deleted.  This has the undesirable side effect that if the user specifies paths which *aren't* created by Wake, those paths are (silently!) dropped.  A second pass needs to be added to verify that everything the user specifies is handled in one way or another; whether it's better to remove the path anyway (system `stat` used to determine directory/file status) or simply warn that the path is unknown is still an open question.[^unknown]
* If the user wants to clean up their view of the workspace, the current model provides good behaviour.  If they want to aggressively clean up *disk space*, the fact that shared blobs get dropped is undesirable.  We'll want to introduce another flag such that, when we trace the mentioned blobs back to unmentioned paths, we add other files using those blobs to the `paths_to_remove` instead of dropping the blob from `removed_blobs`.  Even when the user asks for it, this expansion still seems like something we'd want to notify them about rather than silently removing the newly-discovered files.

[^unknown]: I personally lean toward removing the path under the model of treating `wake --rm` as a filesystem management tool which happens to be aware of the shared storage, but that opens the further question of how to handle the user pointing `wake --rm` to a *different* workspace/CAS.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -2163,7 +2163,7 @@ Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
 
   // Third stage: Remove the resulting blobs from the CAS (if any need deletion).
   std::vector<std::string> deleted_blobs;
-  for (const auto hash : blobs_to_delete) {
+  for (const auto& hash : blobs_to_delete) {
     // Parse and remove the CAS blob if CAS is available.
     if (cas) {
       auto hash_result = cas::ContentHash::from_hex(hash);
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -1985,10 +1985,11 @@ std::vector<std::string> Database::get_outputs() const {
   return out;
 }
 
-std::pair<std::vector<std::string>, std::vector<std::string>> Database::remove_blobs(
-    cas::Cas *cas, const std::vector<std::string> &paths) {
+Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
+                                                 const std::vector<std::string> &paths,
+                                                 bool recursive) {
   if (paths.empty()) {
-    return {{}, {}};
+    return {};
   }
 
   // Build a dynamic query for handling all files at once.  This allows defining query logic based
@@ -2012,19 +2013,23 @@ std::pair<std::vector<std::string>, std::vector<std::string>> Database::remove_b
   if (ret != SQLITE_OK) {
     std::cerr << why_paths << "; sqlite3_prepare_v2: " << sqlite3_errmsg(imp->db) << std::endl;
     end_txn();
-    return {{}, {}};
+    return {};
   }
   for (size_t i = 0; i < paths.size(); ++i) {
     bind_string(why_paths, paths_stmt, i + 1, paths[i]);
   }
 
   std::vector<std::string> paths_to_remove;
+  std::vector<std::string> directories_to_remove;
   while (sqlite3_step(paths_stmt) == SQLITE_ROW) {
     std::string path = rip_column(paths_stmt, 0);
     std::string type = rip_column(paths_stmt, 1);
     if (type == "directory") {
-      // TODO: This should recurse to all children of the directory.
-      std::cerr << "wake --rm: skipping directory: '" << path << "'" << std::endl;
+      if (recursive) {
+        directories_to_remove.emplace_back(path);
+      } else {
+        std::cerr << "wake --rm: skipping directory: '" << path << "'" << std::endl;
+      }
     } else {
       paths_to_remove.emplace_back(path);
     }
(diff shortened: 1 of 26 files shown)


src/runtime/database.cpp
```diff
@@ -2149,7 +2203,7 @@ std::pair<std::vector<std::string>, std::vector<std::string>> Database::remove_b
   // Release the database lock now that both the CAS and the DB have been updated.
   end_txn();
```
AbrarQuazi, line 2233: We are doing quite a bit in this transaction now, will this be an issue when using multi-wake? How long does it take typically when trying to delete a deeply nested directory?

ag-eitilt:
A bug in this set I've discovered is that we rely on the database file/directory tracking so if someone replaces a file with a directory (or vice versa) outside of Wake then the wrong syscall gets used and the removal fails; I'm not sure how proactive we are about throwing out old paths within wake, but there's probably a similar trigger for that if someone uses `mkdir` and `write` calls within similar runs.  This fails loudly and with a reasonable error message, so while fixing it would be *nice*, it's not critical.

More problematic is that if a workspace removal fails, the path (which remains in the workspace) is still marked as `deleted` in the database, so subsequent attempts to run the `--rm` ignore it.  That would be partially resolved by the second pass to catch non-Wake files/directories which had gotten filtered out, but it's still something which should be handled more gracefully.  I'll address that in a followup PR since it's not specific to `--recursive`.

What *is* directly applicable to this PR is that I'd forgotten to plug `directories_to_remove` into the `set deleted=1` logic, so even proper directories are never marked.  That's a pretty straightforward fix, and I'll push a commit for it this afternoon.

src/runtime/database.cpp
AbrarQuazi: we are copying strings here, maybe should be `for (const auto& hash : blobs_to_delete) {`

src/runtime/database.cpp
```diff
@@ -2033,9 +2038,58 @@ std::pair<std::vector<std::string>, std::vector<std::string>> Database::remove_b
   finish_stmt(why_paths, paths_stmt, imp->debugdb);
   sqlite3_finalize(paths_stmt);
 
+  // If recursive (i.e. directories were marked for further processing), query for
+  // all files with paths that start with any of those directories' paths.
+  if (!directories_to_remove.empty()) {
+    std::string dir_input_placeholders = "?";
+    for (size_t i = 1; i < directories_to_remove.size(); ++i) {
+      dir_input_placeholders += ", ?";
```
AbrarQuazi: I see we are doing a few place holders here. I know there is a limit to how many placeholders we can put in sqlite, I think specified by `SQLITE_MAX_VARIABLE_NUMBER` . For super large nested directories and files, would it be worth batching them?

tests/clean/rm/directory-with-untracked-files/pass.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+# Test: --rm -r handles directories with untracked files correctly
+# Wake should remove tracked files but leave the directory and untracked files
+
+set -eu
+
+WAKE="${1:+$1/wake}"
+WAKE="${WAKE:-wake}"
+
+export WAKE_CAS=1
```
AbrarQuazi, line 10: My PR will need to update these in the future to remove

AbrarQuazi (commented):
Looks good to me so far, just a couple of questions

src/runtime/database.cpp
ag-eitilt: As discussed yesterday, good catch and I think I can split these apart into the appropriate batches.

src/runtime/database.cpp
ag-eitilt: I've not timed it, but I don't think it's a transaction that we could break apart -- the "this is unused elsewhere" logic relies on the determination of what *is* used holding through to the end of the removals (we don't want a new job adding a new file that shares a blob that's about to be deleted).  I also don't expect it's going to be used too often in parallel with other multi-wake jobs, and if it ever is I think it's something people would understand the other executions pausing for.  It's not going to be appropriate for the theoretical delete-on-pass handling, but it should be fine for manual use.

Once we know the CAS/DB GC is robust and can handle live cleanup I wouldn't be averse to refactoring this on top of that structure, but (so far as I know) we've been focusing on the GC being about cleaning up after jobs get dropped entirely, and even if there *have* been other cases we've looked at its integration with the `deleted` column is -- obviously -- entirely new.  So, yes, improvements are possible, but I'd rather focus on provable correctness in this first MVP than optimizing this particular case.

ngraybeal (approved):

PR #1877, opened 2026-06-24
Make cas default

AbrarQuazi:
Make CAS logic the default, and stop reading from the workspace in Fuse

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,9 +8,6 @@ on:
     tags:
       - "v*"
 
-env:
-  WAKE_CAS: "1"
-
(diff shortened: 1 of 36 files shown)


ag-eitilt (approved):
It's hard to know whether a removal like this catches *everything* that should be removed, but from what I can tell this seems comprehensive enough.

One known point: this handles most but not all of the `wake --rm` tests I added `WAKE_CAS` to -- e.g. the `unregistered-paths` from #1879, after this had been pushed -- but even if we miss reverting one or two, it'll just be an extra environment variable getting set; nothing's going to break so far as my recent PRs go.

PR #1884, opened 2026-07-13
Unlink Stranded Orphaned Files Left Over

AbrarQuazi:
This PR fixes a bug that happens if job creates a non-reproducible file name, and that job is re-run (with keep persistence set) `delete_prior` will remove that file from the database replacing it with a new non-reproducible file, without unlinking the original first file from the workspace. When `wake --clean` is then subsequently run, the database has no knowledge of the original file, making this orphaned file persist in the workspace.


src/runtime/database.cpp
ag-eitilt, line 468: This seems like a very similar query on the surface, even given the additional clauses from bringing in the `path` consideration.  We should probably have a comment somewhere pointing out the differences and explaining why this one is able to be simpler.

ag-eitilt (approved):
This will work in the short term if we need a solution quickly, but I think we should take another look at the lifespans within the database to properly solve the issue.  Simply rerunning some job should *not* delete the old materialized files from disk.

The proper solution would probably be to leave those paths in the `files` table indefinitely, until a manual `--clean` or family, even though there's no longer any jobs pointing to them.  We can even go ahead and clean the blobs from the CAS if we think doing so would be worth it, but we should be much more conservative when deleting anything user-visible.  Given that (I assume) this problem is currently only exposed with files intended for internal use I'm marking this as "approved" for now, but we should be treating it as relatively-high-priority tech debt as we address the bigger issue of decoupling the `files` table from the `jobs` -- they should wind up as equal citizens of the database rather than either being subordinate to the other.

AbrarQuazi:
> The proper solution would probably be to leave those paths in the `files` table indefinitely

Yes, I was thinking about this, but thought it would be a little bit too complex to implement right now for a quick solution, since I think there would be some changes to the materialization logic (we can't have all paths be materialized by the job)

> treating it as relatively-high-priority tech debt as we address the bigger issue of decoupling the files table from the jobs -- they should wind up as equal citizens of the database rather than either being subordinate to the other.

This seems like a good idea. Do you know what other benefits we would get from doing this?

ag-eitilt:
> > The proper solution would probably be to leave those paths in the files table indefinitely
>
> Yes, I was thinking about this, but thought it would be a little bit too complex to implement right now for a quick solution, since I think there would be some changes to the materialization logic (we can't have all paths be materialized by the job)

👍🏻 

> > treating it as relatively-high-priority tech debt as we address the bigger issue of decoupling the files table from the jobs -- they should wind up as equal citizens of the database rather than either being subordinate to the other.
>
> This seems like a good idea. Do you know what other benefits we would get from doing this?

I've not given much thought to side benefits yet, no.  The most I have so far is that we'd want `--checkout`/`--build-dir` roots to be stored either as another column in `files` (with an additional row per checked-out path) or as something akin to `filetree` in order to clean those alternate roots up (or list, or ...); in the first case the lifespan of the `files` entry would be more closely tied to the lifespan of the checkout than to that of the job, and in both cases we wind up breaking the one-to-one association between (sets of) files and (outputs of) jobs.

We should probably write up an architecture for the database at some point, though -- most of it would just be formalizing the discussions we've all had over the past several months, but having it laid out will allow us to identify places that haven't been worked through (like the lifespans) as well as items which don't yet have the scope we'd want them to (`files` describing *both* materialized paths and CAS blobs).

PR #1885, opened 2026-07-13
Verify blobs before materialize in rehydration

AbrarQuazi:
This PR fixes a bug in the RSC client's cache-hit rehydration. Previously, rehydration materialized a job's output blobs to the workspace one at a time; if a later blob failed hash verification (e.g. a BLAKE2b/BLAKE3 mismatch), rehydration aborted after some files were already written to disk — but never recorded in the database. Wake then fell back to running the job locally, which recorded a different (non-deterministically-named) output set, so `wake --clean` removed only the locally-recorded files and left the already-materialized cache files stranded in the workspace with no DB entry.

The fix splits rehydration into two phases: first download, verify, and ingest every blob into CAS without touching the workspace; only if all blobs verify does it materialize them to the workspace. A verification failure now leaves the workspace untouched, so there is nothing left stranded

### DIFF
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -673,12 +673,17 @@ export def rscApiGetStringBlob ((CacheSearchBlob blobId uri contentHash): CacheS
 
     Pass content
 
-# rscApiGetFileBlob: Downloads a blob to *path* with *mode* permisssions and return *path*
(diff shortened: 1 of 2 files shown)


ag-eitilt (approved):
I didn't trace the logic in fine detail, but the shape of the changes looks correct.  This will also be a good step toward reducing curl calls by downloading multiple blobs at once, even if we then handle the resulting local files individually.

PR #1886, opened 2026-07-14
Enable mknod to create fifo/sockets/block devices in FUSE

AbrarQuazi:
Allow jobs to create fifo/sockets/block devices in FUSE. These special files are not content addressable, so they don't go through the normal CAS ingestion and materialization flow. Rather these are created in a unique (pid specific) path on disk which the respective job is able to utilize in a multi-wake safe manner.

### DIFF
--- a/tests/wakebox/mknod-fifo/fifo_roundtrip.py
+++ b/tests/wakebox/mknod-fifo/fifo_roundtrip.py
@@ -1,0 +1,31 @@
+import os, stat, sys
+
+# Creates a FIFO via mkfifo (a mknod(S_IFIFO) through FUSE), verifies it is visible as a fifo
+# through the virtualized FUSE view, and round-trips data through it.
+FIFO_PATH = "myfifo"
+
+os.mkfifo(FIFO_PATH)
+
+if not stat.S_ISFIFO(os.stat(FIFO_PATH).st_mode):
+    print("FAIL: stat() does not report a fifo")
+    sys.exit(1)
+
+pid = os.fork()
+if pid == 0:
+    # Child: writer.
+    fd = os.open(FIFO_PATH, os.O_WRONLY)
+    os.write(fd, b"fifo-ok")
+    os.close(fd)
+    os._exit(0)
+
+# Parent: reader.
+fd = os.open(FIFO_PATH, os.O_RDONLY)
+data = os.read(fd, 64)
+os.close(fd)
+os.waitpid(pid, 0)
+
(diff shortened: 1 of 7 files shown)


ag-eitilt (approved):
Looks good to me.  Since the only constructor added is on the C++ side, we seem pretty safe from these special node being added manually within Wake runners.  Just to check, though:

1. What happens if a user adds a socket/etc. path to `setPlanFnOutputs`?
2. What happens if a runner passes a socket/etc. path within a `StagingFileOutput` or other improper `StagingItem`?
3. What happens if a job creates a socket/etc. and then creates a hardlink to it?
4. Same as 3, but a symlink?

My ideal answers would be:

1. It's treated as any other non-existent file -- previously that would have blindly created a `Path` without any backing content, but hopefully there's a proper error generated now that the CAS expects to consume it.  Bonus points if the error message points out that it's an invalid path type rather than simply being missing, half points and tech debt if it's silently dropped, quarter points and heavy tech debt if we still return a broken `Path`.
2. It's treated as any other path with an improper type -- e.g. a similar message as if a directory path were shoved into that `StagingFileOutput`.  Half points if the error message implies it doesn't exist.
3. It's treated as if both were created with `mknod` -- i.e. not returned as Fuse output paths, and both successfully cleaned up.
4. It's treated as any other broken symlink whose target was deleted in the course of the job.

AbrarQuazi:
> 1. What happens if a user adds a socket/etc. path to `setPlanFnOutputs`?

So for runnlers like `localRunner` and `virtualRunner` I believe we error out since the staging helper only supports regular files, directory, and symlinks. For this `fuseRunner` we drop outputting these special nodes from being in the JSON, so I believe when we do a lookup later in `partitionStagingItems`, they silently get dropped. So there might be some tech debt here if we want to error here too.

> 2. What happens if a runner passes a socket/etc. path within a `StagingFileOutput` or other improper `StagingItem`?

I think this is not a problem currently with our Runners, as they will error out

> 3. What happens if a job creates a socket/etc. and then creates a hardlink to it?
> 4. Same as 3, but a symlink?

So we are allowed to symlink to the socket, but since sockets only exist in the lifespan of a job, the symlink will become dangling at the end of it. For hardlinks, currently we refuse it and fail, it would be a bit more tricky to implement because we would need to have the hardlink have the same kernel inode backing of the socket, and there maybe some settings in FUSE to set inorder for this to happen: https://github.com/libfuse/libfuse/issues/199 . We can address this later if this ever is needed

PR #1888, opened 2026-07-20
Rsc batch blob downloads

AbrarQuazi:
This PR significantly improves the RSC client's performance in the RSC hit job path. Previously, the client would spin up a wake process to handle a curl download per job output file, which would lead to timeouts on builds for machines with smaller amount of cores due to resource exhaustion of so many concurrent curl jobs (especially since `setPlanUsage` was set to 0.1).

This PR solves this issue by batching the curl GET requests from the nfs blob store for each job, so that we now have fewer (up to a ~100x less total) curl processes. This PR also adds support for the `--parallel` curl flag for machines that support it (version ≥ 7.66) which allows for many parallel concurrent transfers within a single process. 

The next thing to implement (not on this PR) would be to batch job uploads.

### DIFF
--- a/share/wake/lib/system/http.wake
+++ b/share/wake/lib/system/http.wake
@@ -510,29 +510,6 @@ export def makeBinaryRequests (specs: List BatchDownloadSpec): Result (List Path
 
     require Pass configPath = writeTempFile "curl.batch.config" configBody
 
-    # Force-remove then touch each destination before downloading so that a failed transfer still
-    # yields a file (wake-hash can't hash a non-existent file). Mirrors makeBinaryRequest's cleanup semantics.
-    def cleanupScript =
-        destinations
-        | map (\d "rm -f \"{d}\"\ntouch \"{d}\"")
-        | catWith "\n"
-
-    def cleanupJob =
-        makeShellPlan cleanupScript (httpDirPath, Nil)
-        | setPlanLabel "http: batch cleanup ({str specs.len} files)"
-        | setPlanStdout logNever
-        | setPlanStderr logNever
-        | setPlanPersistence Once
-        | setPlanFnOutputs (\_ Nil)
-        | setPlanUsage (Usage 0 0.0 0.002 0 0 0)
-        | setPlanEcho logDebug
-        | runJobWith localRunner
-        | setJobTag "http.cleanup" "batch"
-        | setJobInspectVisibilityHidden
-
-    require True = cleanupJob.isJobOk
-    else failWithError "Unable to remove binary files before redownloading them"
-
     # Only pass --parallel on curl that supports it (>= 7.66.0). On older curl (el8's 7.61.1) the
     # batch runs sequentially, which is still one process + one wake job + connection reuse.
     def parallelFlags =
@@ -541,21 +518,34 @@ export def makeBinaryRequests (specs: List BatchDownloadSpec): Result (List Path
         else
             Nil
 
-    def cmd =
-        "curl",
-        (
-            parallelFlags ++
-            (
-                # Exit non-zero on any HTTP error so the batch fails all-or-nothing.
-                "--fail",
-                "--config",
-                configPath.getPathName,
-                Nil
-            )
-        )
+    # All destinations quoted for a single-pass shell arg list.
+    def quotedDests =
+        destinations
+        | map (\d "\"{d}\"")
+        | catWith " "
+
+    # The curl invocation. `--create-dirs` makes any missing parent directories so we don't need a
+    # separate mkdir fan-out; `--fail` exits non-zero on any HTTP error (all-or-nothing batch).
+    def curlLine =
+        def flags = parallelFlags ++ ("--create-dirs", "--fail", Nil)
+
+        "curl {catWith " " flags} --config \"{configPath.getPathName}\""
+
+    # Do the cleanup in the same job as curl. One `rm -f` clears any write-protected leftovers,
+    # curl downloads, then one `touch` ensures every declared output exists even if its transfer
+    # failed (wake-hash can't hash a missing file). curl's exit status is preserved so the job
+    # still fails as a whole when any transfer fails.
+    def batchScript =
+        """
+        rm -f %{quotedDests}
+        %{curlLine}
+        curl_status=$?
+        touch %{quotedDests}
+        exit $curl_status
+        """
 
     def job =
-        makeExecPlan cmd (configPath, Nil)
+        makeShellPlan batchScript (configPath, httpDirPath, Nil)
         | setPlanLabel "http: batch GET ({str specs.len} files)"
         | setPlanStdout logNever
         # Curl dumps download stats to stderr
--- a/share/wake/lib/system/http.wake
+++ b/share/wake/lib/system/http.wake
@@ -425,9 +425,12 @@ export tuple BatchDownloadSpec =
     # The local path the downloaded bytes should be written to
     export Output: String
 
-# The default maximum number of concurrent transfers curl performs within a single batch job.
-# This is the real concurrency knob: it bounds process/connection churn independent of how many
-# transfers are in the batch.
+export def downloadUsageCpu: Double =
+    0.25
+
+# The maximum number of concurrent transfers curl performs within a single batch job, used ONLY as
+# the `--parallel-max` value on curl >= 7.66.0 (feature-detected). Older curl (e.g. el8's 7.61.1)
+# runs the batch sequentially and ignores this.
 export def downloadParallelMax: Integer =
     8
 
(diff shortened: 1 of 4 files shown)


share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+
+    # --parallel requires curl >= 7.66.0
+    major > 7 || (major == 7 && minor >= 66)
+
+# makeBinaryRequests: Downloads many urls to their destination paths in a single curl process,
+# returning all destination paths on success or a single Fail for the whole batch.
+#
+# This is the batched analogue of makeBinaryRequest. Instead of one curl process per transfer it
+# writes a curl `--config` file listing one url/output stanza per input and runs a single curl job
+# that transfers them **sequentially** over a reused connection. On curl >= 7.66.0 (feature-detected)
```
ag-eitilt, line 508: Minor nit: I wouldn't call this "feature-detected" since to me that implies directly checking the feature itself, whether through API introspection, program-generated support lists, or running a small test of the expected behaviour.  Manually looking up what version introduced the feature and just testing the numbers is equally valid, but not *feature* detection.  That might just be my own idiosyncratic definition, though.

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+    def httpDir = ".build/wake-{str wakePid}/stdlib/http"
+
+    require Pass httpDirPath = mkdir httpDir
+
+    def destinations =
+        specs
+        | map getBatchDownloadSpecOutput
+
+    # Build the curl config file. Each stanza is a url/output pair. Quoting each value avoids
+    # command-line length limits and shell-escaping issues, mirroring the long-body path above.
```
ag-eitilt: Not sure there is an "above", or I'm misunderstanding what it's referencing.

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+    def configBody =
+        specs
+        | map (\(BatchDownloadSpec url output) "url = \"{url}\"\noutput = \"{output}\"")
```
ag-eitilt: It might be worth escaping any quote chars within `url` and `output`; those should be pretty uncommon, but this is a generic function.  Even more of an edge case, how does curl handle it if either of them contain `\n` or `\0`?

If we thought there was a chance people would invoke this manually, it might also be a case where having an explicit `BatchDownloadSpec` type is slightly worse than the generic, since if it were just `Pair` the list could be written with `:->` operators.  That's probably even less likely than `"` characters, though.

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+    def parallelFlags =
+        if curlSupportsParallel Unit then
+            "--parallel", "--parallel-max", str downloadParallelMax, Nil
+        else
+            Nil
+
+    # All destinations quoted for a single-pass shell arg list.
+    def quotedDests =
+        destinations
+        | map (\d "\"{d}\"")
```
ag-eitilt: Same escaping of `d` should be done here.  There *might* be something along the lines of a `shellEscape` function/prim already written?

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+        | catWith " "
+
+    def curlLine =
+        def flags = parallelFlags ++ ("--create-dirs", "--fail", Nil)
+
+        "curl {catWith " " flags} --config \"{configPath.getPathName}\""
+
+    # Do the cleanup in the same job as curl. One `rm -f` clears any write-protected leftovers,
```
ag-eitilt: We're either writing these to separate directories/unique paths in which case the `rm` doesn't do anything, or we're risking breaking multi-wake if we remove something that's happening in parallel.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -647,31 +630,198 @@ def verifyBlobHash (blobId: String) (actualHash: String) (contentHash: String):
...
+    # Dedupe by blob id: identical ids resolve to identical bytes, so fetch each once.
+    def uniqueBlobs =
+        blobs
+        | distinctBy (\a \b scmp a.getCacheSearchBlobId b.getCacheSearchBlobId)
+
+    # Parse + validate each blob's URI exactly once, pairing the blob with its (scheme, path).
+    # Fails loudly on an unsupported scheme or malformed URI, mirroring the old resolveUriResponse.
+    def parseBlob (blob: CacheSearchBlob): Result (Pair CacheSearchBlob (Pair String String)) Error =
+        require (scheme, path, Nil) =
+            extract `([a-zA-Z][a-zA-Z0-9+.-]*)://(.*)` blob.getCacheSearchBlobUri
```
ag-eitilt, line 659: Made this comment on the original code, but *technically* the `//` is part of the path, typically indicating an authority component according to however the scheme wants to handle authorities (e.g. HTTP looking up a domain name).  Nothing's really actionable based on that info, just the naming being *slightly* off.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -647,31 +630,198 @@ def verifyBlobHash (blobId: String) (actualHash: String) (contentHash: String):
...
+
+    require Pass parsedBlobs =
+        uniqueBlobs
+        | map parseBlob
+        | findFail
+
+    # db-scheme blobs stay on the inline path; file/http(s) blobs are curled.
+    def Pair networkParsed dbParsed =
+        parsedBlobs
+        | splitBy (\(Pair _ (Pair scheme _)) scheme !=* "db")
```
ag-eitilt: Nit: Avoid the mental overhead of the negation?  Also matches the order of the comment.
```suggestion
    # db-scheme blobs stay on the inline path; file/http(s) blobs are curled.
    def Pair dbParsed networkParsed =
        parsedBlobs
        | splitBy (\(Pair _ (Pair scheme _)) scheme ==* "db")
```

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -647,31 +630,198 @@ def verifyBlobHash (blobId: String) (actualHash: String) (contentHash: String):
...
+    # Batched download of the network blobs.
+    def wakePid: Integer = prim "pid"
+    def httpDir = ".build/wake-{str wakePid}/stdlib/http"
+
+    def downloadChunk (chunk: List CacheSearchBlob): Result (List (Pair String Path)) Error =
+        # Sort the chunk deterministically by blob id so that two batches with the same blob set
+        # produce byte-identical curl jobs (which wake then dedups to a single job / single writer).
+        def sortedChunk =
+            chunk
+            | sortBy (\a \b scmp a.getCacheSearchBlobId b.getCacheSearchBlobId)
```
ag-eitilt, line 706: Might be worth the optimization of adding a `distinctBy`/`groupBy` stage, to ensure we only download each blob once and then `cp` the blob to other destinations after the fact?  I don't know how this interacts with the CAS so we might not need that level of deduplication for the primary use case, but this wouldn't be a bad place to put that logic anyway.

share/wake/lib/system/remote_cache_runner.wake
```diff
@@ -106,9 +105,10 @@ export def mkRemoteCacheRunner (rscApi: RemoteCacheApi) (hashFn: RunnerInput =>
             outputDirs
             | map makeDirectoryPreparedItem
 
-        # Files: download + verify + ingest into CAS. Returns the item to materialize.
-        def doDownloadFileToCas (CacheSearchOutputFile path mode blob) =
-            rscApiGetFileBlobToCas blob path mode
+        # Files: download + verify + ingest into CAS in a single (few) batched curl job.
+        def fileDownloadEntries =
+            outputFiles
+            | map (\(CacheSearchOutputFile path mode blob) Pair blob (Pair path mode))
```
ag-eitilt: The naming makes me think we filter out the blobs which are already in the CAS later on, but before we download the ones we don't have.  If we don't do that, it could be a good followup optimization; if we do, we might want to update the comment to make it clear we're not necessarily downloading *everything* in `outputFiles`.

ag-eitilt (approved):
Got a number of comments, but I think the thing I'd most want changed would be the `rm`, and even that's not overly critical.

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+# Feature-detects whether the curl on this image supports `--parallel` (added in curl 7.66.0).
+export target curlSupportsParallel Unit: Boolean =
+    def job =
+        makeExecPlan ("curl", "--version", Nil) Nil
+        | setPlanLabel "http: detect curl --parallel support"
+        | setPlanStdout logNever
+        | setPlanStderr logNever
+        | setPlanPersistence ReRun
+        | setPlanFnOutputs (\_ Nil)
+        | setPlanUsage (Usage 0 0.0 0.002 0 0 0)
```
ngraybeal: Wake doesn't allow usage below .01. https://github.com/sifiveinc/wake/blob/682978ac8f19a9f90432df890532aa8de867add5/src/runtime/job.cpp#L235

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+        | setPlanEcho logDebug
+        | runJobWith localRunner
+        | setJobInspectVisibilityHidden
+
+    require True = job.isJobOk
+    else False
+
+    require Pass stdout = job.getJobStdout
+    else False
```
ngraybeal: Don't need both of them.

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+# The maximum number of concurrent transfers curl performs within a single batch job, used ONLY as
+# the `--parallel-max` value on curl >= 7.66.0 (feature-detected). Older curl versions
+# runs the batch sequentially and ignores this.
+export def downloadParallelMax: Integer =
+    8
+
+# The default maximum number of transfers to place in a single batch curl job. This bounds the
+# curl config file size and per-job output count, mirroring the upload side's blobUploadBatchSize.
+export def downloadBatchSize: Integer =
+    256
```
ngraybeal, line 440: Should we place these closer to the code using them?

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+
+# Feature-detects whether the curl on this image supports `--parallel` (added in curl 7.66.0).
+export target curlSupportsParallel Unit: Boolean =
```
ngraybeal, line 443: Should we just create a prim to get tool version? (ie which)

share/wake/lib/system/http.wake
```diff
@@ -416,3 +416,147 @@ export def makeBinaryRequest (request: HttpRequest): Result Path Error =
...
+        | setPlanEcho logDebug
+        | runJobWith workspaceRunner
+        | setJobTag "http.method" "GET"
+        | setJobTag "http.batch" (str specs.len)
+        | setJobInspectVisibilityHidden
+
+    require True = job.isJobOk
+    else failWithError "Failed attempting to batch download {str specs.len} binary files"
+
+    job.getJobOutputs
```
ngraybeal: Do we need the isJobOK if we return the result status of getJobOutputs? Should it just be addErrorContext instead?

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -647,31 +630,198 @@ def verifyBlobHash (blobId: String) (actualHash: String) (contentHash: String):
...
+        | map resolveDbBlob
+        | findFail
+
+    def networkBlobs =
+        networkParsed
+        | map getPairFirst
+
+    # Batched download of the network blobs.
+    def wakePid: Integer = prim "pid"
+    def httpDir = ".build/wake-{str wakePid}/stdlib/http"
```
ngraybeal, line 699: Should we start centralizing these output dirs in some library/file so it's globally visible?

ngraybeal (approved):
Overall LGTM. Just a few comments

share/wake/lib/system/http.wake
AbrarQuazi: We currently don't have a prim to handle curl escapes, like we do for shell escapes, so I'll just currently create an internal helper for this now in the file

share/wake/lib/system/remote_cache_runner.wake
AbrarQuazi: This would be a good follow up optimization, Ill add a todo for now for a follow up

share/wake/lib/system/http.wake
AbrarQuazi: seems a bit overkill for now, but I can see the need of a generic tool version getter that would be helpful

share/wake/lib/system/http.wake
AbrarQuazi: yep you are right on this, will add

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: I think its a good idea in general, we can maybe do that as part of the clean up and restructuring of the .build/ directory

share/wake/lib/system/http.wake
AbrarQuazi: so we have a prim version of this, but we dont expose a function that calls the prim. I am exposing this function as `shellQuote` so it doesn't interfere with an allready defined `shellEscape` function in our internal repo (which doesnt use the prim), and which we can replace later

PR #1889, opened 2026-07-23
Wrap the database blob removal in a `prim`

ag-eitilt:
It can *occasionally* be helpful to remove files that Wake generates from within Wakelang (e.g. if they're very large and no longer needed, or to prepare the environment for something that requires `localRunner`).  We had previously just used `rm` or similar shell tools for that purpose, in a `localRunner` job, but with the CAS that no longer frees the disk space.  Instead, we decided that promoting the `wake --rm` behaviour to a `prim` allows us to reuse the logic and safety checks developed for that tool, while only requiring the single additional expansion of being able to exclude source files from the removal.

A future companion to this would be a hash-based (rather than path-based) removal function; that serves the "very large file" case better, but as it would require writing more novel code than the simple wrapper here for an only-minimally-better solution, we went with this to be able to better get it merged in time for the release.


share/wake/lib/system/cas.wake
```diff
@@ -99,3 +99,17 @@ export def materializeCachedItemFromCas (pathInfo: PathInfo): Result Unit Error
...
+
+# Remove files from the workspace and CAS, excluding any files which were not generated by Wake.
+#
+# Note that this will break any `Path` objects that point to the removed files, and so should
+# *only* be used when those paths are known to be dropping out of scope, and will not be recreated
+# in the same build (e.g. rehydrated from a `target`).
+#
+# Additionally, this will not remove any files which are currently in use by *other* active runs
+# in a multi-wake context, to avoid breaking `Path` objects in those runs.
+export def unsafe_removeFiles (recursive: Boolean) (paths: List String): Result Unit Error =
```
AbrarQuazi, line 111: Any specific reason why these should be `List String` instead of `List Path`?

src/runtime/database.cpp
```diff
@@ -2376,6 +2398,78 @@ Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
...
+  std::unordered_set<std::string> paths;
+  for (Record *node = path_list; node->cons == &List->members[1];
+       node = static_cast<Record *>(node->at(1)->coerce<HeapObject>())) {
+    HeapObject *head = node->at(0)->coerce<HeapObject>();
+    REQUIRE(typeid(*head) == typeid(String));
+    paths.insert(static_cast<String *>(head)->as_str());
+  }
+
+  // Build exclusion set from the source files; it makes little sense to remove them within wakelang.
+  std::unordered_set<std::string> exclude_paths;
```
AbrarQuazi, line 2458: Is there a reason why we are doing exclusion of source files in the prim, and not in the SQL? I believe there are other queries that exclude the source files by doing `!= '<source>'"`

src/runtime/database.cpp
```diff
@@ -2376,6 +2398,78 @@ Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
...
+
+  // Unlink files and directories (in reverse order, deepest first) from the workspace.
+  for (const auto &file : manifest.files) {
+    if (unlink(file.c_str()) != 0 && errno != ENOENT) {
+      std::string error_msg = "Failed to remove file '" + file + "': " + strerror(errno);
+      runtime.heap.reserve(reserve_result() + String::reserve(error_msg.size()));
+      RETURN(claim_result(runtime.heap, false, String::claim(runtime.heap, error_msg)));
+    }
+  }
+  for (auto it = manifest.directories.rbegin(); it != manifest.directories.rend(); ++it) {
```
AbrarQuazi, line 2481: Don't have to follow this, but for me its easier to understand this loop by starting from .end() iterator and subtrating the iterator till then .begin(), when going reverse order of the loop

src/runtime/database.cpp
```diff
@@ -2376,6 +2398,78 @@ Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
...
+      }
+    }
+  }
+
+  // Remove files from database and CAS within a single transaction.
+  cas::Cas *cas = ctx->second ? ctx->second->get_store() : nullptr;
+  Database::RemovalManifest manifest = ctx->first->remove_blobs(cas, paths, exclude_paths, recursive);
+
+  // Unlink files and directories (in reverse order, deepest first) from the workspace.
+  for (const auto &file : manifest.files) {
```
AbrarQuazi, line 2476: Shouldn't the unlinking happen in the transaction itself, so we avoid any race?

AbrarQuazi (commented):
Some initial review questions

share/wake/lib/system/cas.wake
ag-eitilt: No *strong* reason, just that it (A) emphasizes that any `Path`s that would have gotten passed are no longer going to be valid, and (B) means we don't have to very briefly `claim` files created by previous invocations if we don't have a `Path` handy already.

src/runtime/database.cpp
ag-eitilt: Ooh, that could be a good way of doing it.  I'd just be concerned about the aforementioned paths-that-aren't-`Path`s, which might indeed be source files but happen to not have gone through a `source` job.  This loop is the C++ equivalent of `subscribe sources | ...` rather than ``sources `.*` "." | ...``

src/runtime/database.cpp
ag-eitilt: Didn't put much thought into that, just copied the structure I had for the `wake --rm` CLI.  I'd say that I'm not *very* concerned about races here, though, since these are just cleaning up the workspace which will only mechanically affect the (already unsafe) `workspaceRunner` jobs.

src/runtime/database.cpp
ag-eitilt: That'll work in a `while`, but I'm not sure it works with a `for` -- the order would have to be subtracting *before* the test+body on every step, and so far as I know `for` enforces subtracting *after* the test+body.  I agree it's easier to reason about `.end()`/`.begin()` than it is `.rbegin()`/`.rend()`, but my own code-parsing finds that outweighed by how much easier an explicit `for` is to reason about than a `while` that's emulating a `for`.

share/wake/lib/system/cas.wake
```diff
@@ -99,3 +99,17 @@ export def materializeCachedItemFromCas (pathInfo: PathInfo): Result Unit Error
...
     | addErrorContext "Materializing cached output {pathInfo.getPathInfoName} from CAS"
+
+# Remove files from the workspace and CAS, excluding any files which were not generated by Wake.
+#
+# Note that this will break any `Path` objects that point to the removed files, and so should
+# *only* be used when those paths are known to be dropping out of scope, and will not be recreated
+# in the same build (e.g. rehydrated from a `target`).
+#
+# Additionally, this will not remove any files which are currently in use by *other* active runs
+# in a multi-wake context, to avoid breaking `Path` objects in those runs.
```
ngraybeal, line 110: Does that return an error?

src/runtime/database.cpp
```diff
@@ -2376,6 +2398,78 @@ Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
...
+    if (unlink(file.c_str()) != 0 && errno != ENOENT) {
+      std::string error_msg = "Failed to remove file '" + file + "': " + strerror(errno);
+      runtime.heap.reserve(reserve_result() + String::reserve(error_msg.size()));
+      RETURN(claim_result(runtime.heap, false, String::claim(runtime.heap, error_msg)));
+    }
```
ngraybeal: DO we want to escape early or should we continue on with all and report at the end?

src/runtime/database.cpp
```diff
@@ -2376,6 +2398,78 @@ Database::RemovalManifest Database::remove_blobs(cas::Cas *cas,
...
+  }
+  for (auto it = manifest.directories.rbegin(); it != manifest.directories.rend(); ++it) {
+    if (rmdir(it->c_str()) != 0 && errno != ENOENT) {
+      std::string error_msg = "Failed to remove directory '" + *it + "': " + strerror(errno);
+      runtime.heap.reserve(reserve_result() + String::reserve(error_msg.size()));
+      RETURN(claim_result(runtime.heap, false, String::claim(runtime.heap, error_msg)));
+    }
```
ngraybeal: Same as above? Should we return early?

AbrarQuazi (approved):
Looks good to me, thanks Sam!

share/wake/lib/system/cas.wake
ag-eitilt: It does not, and I've been going back and forth on whether I want it to.  If we see this as being intended as a "shed what you can" cleanup marker, then not deleting shared things would be the expected behaviour and it wouldn't be an error.  If we see this as an explicit deletion command, then an error would be appropriate -- but we'd want to distinguish it from *hard* failures in cases where it's being used as opportunistic cleanup anyway.

My current thought is that it might be best to return `Result (List String) Error` instead, where the `List String` is either everything that got removed (more intuitive and more informational) or everything that got skipped (more relevant to the currently-known use case).  I think that's something that can be expanded upon in a followup PR, though, once we know concretely how we're plugging it into things.

EDIT: Meant to say, the biggest difficulty is just that this main logic is shared between the prim (where it does make some sense to return early, and where error messages should be propagated to Wakelang) and the CLI (where it's nice to be more resilient, and error messages just have to show up on stderr).  I think there's further work on this to allow sharing the logic while also allowing the error handling to be tuned based on the context, but I still need to figure out what that should look like, and I don't want it to hold up getting the main behavior of the prim rolled out.

src/runtime/database.cpp
ag-eitilt: I wrote this to escape early since the type signature can only return one error at a time (it would certainly be nice to return more, but that's a bit of an architectural puzzle on its own).  You are definitely right, though, that removing everything possible rather than just everything found before the first error is more in keeping with Wake's idempotency.  Pulling out the return logic in one last commit.

PR #1892, opened 2026-07-30
Allow input directories to be accessable (if mode allows)

AbrarQuazi:
Previously, we would deny access to all visible input directories so they would report that they couldn't be written into (even if they could, `access` would just misreport). This is incorrect behavior, and should match for whatever mode is set for that directory. Visible input files are still guarded and are not modifiable.

### DIFF
--- a/tests/wakebox/visible-dir-writable/pass.sh
+++ b/tests/wakebox/visible-dir-writable/pass.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+# Regression test: access(W_OK) on a visible directory must agree with actual writability
+# (pre-fix, it always denied W_OK even though creating files there worked).
+#
+# Verifies that:
+# 1. `test -w foo` reports WRITABLE (access(2) no longer lies)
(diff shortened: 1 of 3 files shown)


tests/wakebox/visible-dir-writable/pass.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+# Regression test: access(W_OK) on a visible directory must agree with actual writability
+# (pre-fix, it always denied W_OK even though creating files there worked).
```
ngraybeal: NIT: probably not needed comment

tests/wakebox/visible-dir-writable/pass.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# Verifies that:
+# 1. `test -w foo` reports WRITABLE (access(2) no longer lies)
```
ngraybeal: NIT: unneeded comment?
```suggestion
# 1. `test -w foo` reports WRITABLE
```

ngraybeal (approved):

PR #1896, opened 2026-08-05
Reduce overhead of cleaning stale files

ag-eitilt:
This is a partial revert of #1884, solving the problem that PR was addressing in a lower-impact manner.  Specifically, the edge case we're running into is that Wake will currently evict jobs from the database when a new job with the same input key is run; no check is made that their output files are the same.  When the output files are *not*, we had been left with files in the workspace, which Wake had indeed written, but which it forgot about having owned.  At that point, `wake --clean` was no longer able to remove everything created by Wake.

The original solution was to remove the files from the workspace when an entry in the `files` table was removed with no other owners at that path.  This proved to be expensive in addition to going against the general practice elsewhere of not removing files from the user's view -- see the original PR for more discussion, where it was deemed acceptable since the only nondeterministic outputs we knew of were internal to Wake.  This took about 45 seconds in a particular benchmark which touches a large number of files.

This new solution instead takes advantage of the `deleted` marker introduced shortly after that PR was written, by recognizing that only a single blob can be associated with any specific path at a time and marking everything else at that location as deleted/stale.  Given that, jobs can be deleted without needing to update the `files` table -- leaving that cleanup for routine garbage collection based on `deleted=1` -- and Wake-generated files with no remaining originating job can remain in the database (with `deleted=0`) until they're recreated by a later job or removed entirely with `wake --clean`.  This takes about 10 to 15 seconds in the same benchmark.

### DIFF
--- a/src/runtime/database.cpp
+++ b/src/runtime/database.cpp
@@ -391,12 +391,10 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
   const char *sql_insert_log =
       "insert into log(job_id, descriptor, seconds, output)"
       " values(?, ?, ?, ?)";
-  // The `where deleted=1` on the upsert's update branch (and on sql_reset_deleted below) is load-
-  // bearing, not just a no-op guard: it makes sqlite3_changes() report 0 whenever the row was
+  // The `where deleted=1` on the upsert's on-conflict branch (and on sql_reset_deleted below) is
+  // load-bearing, not just a no-op guard: it makes sqlite3_changes() report 0 whenever the row was
   // already live at this exact (path, hash, type, mode), which callers use to skip the far more
-  // expensive mark_stale_path_files scan in that (common, steady-state) case -- see add_hash/
-  // reuse_job. Any other row's staleness was already correctly resolved the last time this row's
-  // deleted flag went to 0, so skipping is safe, not just faster.
+  // expensive mark_stale_path_files scan in that idempotent case -- see add_hash/reuse_job.
   const char *sql_insert_file =
       "insert into files(hash, type, mode, path, deleted) values (?, ?, ?, ?, 0)"
       " on conflict(hash, type, mode, path) do update set deleted=0 where deleted=1";
@@ -406,10 +404,9 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
       "insert or ignore into run_files(run_id, file_id)"
       " values(?, (select file_id from files where path=? and hash=? and type=? and mode=?))";
   // When a path is (re)materialized, any other files row for that same path is stale -- the
-  // workspace can only ever hold one file at that path, so a different (hash, type, mode) there
-  // now describes content that is no longer on disk. Mark it deleted rather than removing the
-  // row, so cheap DB-only bookkeeping (e.g. CAS blob liveness) stays correct without needing to
-  // eagerly unlink anything from disk.
+  // filesystem can only ever save one file to that path, so a different (hash, type, mode) at that
+  // path now describes content that is no longer on disk.  Mark it deleted rather than removing
+  // the row, so cheap DB-only bookkeeping (e.g. CAS blob liveness) stays correct.
   const char *sql_mark_stale_path_files =
       "update files set deleted=1 where path=? and not (hash=? and type=? and mode=?)";
   const char *sql_get_log =
@@ -465,8 +462,8 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
       "                  filetree.access=2)"
       "  and (select coalesce(max(run_id), 0) from run_jobs where job_id=jobs.job_id) <= ?1";
   // A deleted=1 row never owns the physical file at its path -- some other (live) row does, or
-  // nothing does -- so it's pure bookkeeping cruft once no filetree entry still references it.
-  // Safe to drop outright (no unlink needed); the filetree join guards the FK from files(file_id).
+  // nothing in Wake does -- and once no `filetree` entry references it the metadata which had been
+  // saved for Job bookkeeping is likewise no longer needed.
   const char *sql_delete_stale_files =
       "delete from files"
       " where deleted = 1"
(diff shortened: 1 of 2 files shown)


ngraybeal (approved):

src/runtime/database.cpp
```diff
@@ -446,33 +455,19 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
...
-  const char *sql_delete_orphan_files =
+  // A deleted=1 row never owns the physical file at its path -- some other (live) row does, or
+  // nothing in Wake does -- and once no `filetree` entry references it the metadata which had been
+  // saved for Job bookkeeping is likewise no longer needed.
+  const char *sql_delete_stale_files =
       "delete from files"
-      "  where file_id not in (select file_id from filetree)"
-      "  and   file_id not in (select file_id from run_files)";
+      " where deleted = 1"
+      " and not exists (select 1 from filetree where filetree.file_id = files.file_id)";
```
AbrarQuazi: Don't we still need the `run_files` guard here to keep it multiwake safe?

Claude is telling me:

```
The old delete_orphan_files had two guards; the new statement (database.cpp:467-470) keeps only the filetree one. Since run_files.file_id is on delete cascade (schema.h:114) with foreign_keys=on, deleting a files row silently deletes another live run's GC protection. I reproduced the interleaving:


Run A: add_hash(P, hashA) + claim → run_files(A)   [no filetree yet: job unfinished]
Run B: add_hash(P, hashB) → mark_stale_path_files  → A's row becomes deleted=1
  → get_dead_hashes says hashA is DEAD while A still holds run_files
Run B: clean() → delete_stale_files → A's row deleted, run_files(A) cascaded away
```

src/runtime/database.cpp
```diff
@@ -506,22 +501,29 @@ std::string Database::open(bool wait, bool memory, bool tty, bool readonly) {
...
+  // the `files` entry can outlive its originating job, so `files` needs to be the outer table.
   const char *sql_get_output_files =
-      "select f.path"
-      " from filetree ft join files f on f.file_id=ft.file_id join jobs j on ft.job_id=j.job_id"
-      " where ft.access = 2"
-      " and f.deleted = 0"
-      " and substr(cast(j.commandline as varchar), 1, 8) != '<source>'"
-      " and substr(cast(j.commandline as varchar), 1, 7) != '<claim>'";
+      "select distinct f.path"
+      " from files f"
```
AbrarQuazi, line 515: Another potential issue:

Inverting this to start FROM files instead of FROM filetree means rows with no filetree entry at all now pass the NOT EXISTS vacuously and get returned as deletable outputs.

That's the intended fix for stranded outputs, but it also catches source files. If a run is interrupted in that window, the source's files row survives with deleted=0 and no filetree entry, so nothing records that it belonged to a <source> job. The next --clean then unlinks a git-tracked user file. The old FROM filetree form couldn't return such a row.

AbrarQuazi (commented):
Some potential issues

src/runtime/database.cpp
ag-eitilt: Ah, good catch.  I was just comparing the edits each commit, and kind of ignored the queries which got removed entirely.  I'll fix that.

src/runtime/database.cpp
ag-eitilt: Yeah, that inversion is deliberate since we shouldn't be relying on the filetree -- that's why your PR needed to delete files as each job was jettisoned.  I do see the edge case you identified, though.  At the moment I don't think it's much of an issue since that's a rather tight race condition to hit, threading exactly between the insertion to the `files` and its use in `filetree` (and the associated insertion to `jobs`).[^prune]

Off the top of my head, one solution would be a `deleted=2` value[^deleted] indicating source files which survives both the `deleted=1` DB pruning and the `--clean` workspace removal, but that'll be a bit larger of a change than I particularly want to fold into this PR.

A more straightforward but maybe less robust solution would just be expanding the job registration transaction to encompass the entire set of changes to all three tables.  I'm less fond of it since (in addition to the complication below) it would introduce more contention for the database and more structural requirements on the code itself, but it *is* at least avoiding adding any additional schema complexity.[^schema]

[^prune]: It would be a larger problem in the future if we start pruning jobs based on things other than key replacement and overlapping outputs, but for now the only way a `source` job would get dropped is if some other job overwrote it, as in the fourth paragraph.

[^deleted]: At which point the column probably needs a better name than "deleted".

[^schema]: To be clear, even `deleted=2` wouldn't require any database migration.  The "schema complexity" is purely on the theoretical/reasoning side in raising a "boolean" integer to an "enum" one.

Another edge case is if some non-`source` job writes to a source file.  `deleted=2` would prevent that "this is a source" metadata from getting lost, but only for files which *did* get pulled in via `source` at some point.  Though that could tie into the early git/source-file registration that's come up in the past, by hashing all source files (after modtime filtering for performance) after or as they get added to the `topic source` and its C++ equivalent.  And *that* would allow us to have `--clean` roll back the changes to source files rather than simply ignoring any changes which might have been made to them.  Trying to preemptively register sources is even more dramatic of a change than the other, though, and I have a feeling it might need multithreading to be done efficiently.

If you don't mind, I'd like to merge this in to at least partially address the immediate v50 performance regression, and then I'll see about either `deleted=2` or a wider transaction in a followup (which may or may not make it into 50.0.5).

ag-eitilt (commented):
All right, multi-wake safety restored via the `run_files` check.  I was apparently also testing this against a Wake v50 where I had disabled *both* the old `collect_stranded_files` behaviour (to serve as a "baseline" for how much this adds to a hypothetical dirty-output run) and cached-job CAS materialization (to minimize the difference when testing against v49)... but had forgotten to disable the materialization in the PR repo.  With both of them brought into alignment, the ~10 seconds longer this PR seemed to add to that no-stale-file-handling baseline has turned into ~5 seconds *savings* -- and much more saved compared to v50 as released.

src/runtime/database.cpp
AbrarQuazi: Ok cool, maybe we create a follow up ticket for this

AbrarQuazi (approved):
Thanks!

PR #1897, opened 2026-08-11
Add fix to stop materializng source and claim during cached run

AbrarQuazi:
This PR fixes an issue where we observed source files getting modified and affecting git: during materialization we would `rename` over the destination which replaces the inode, and any inode replacement updates ctime. Git's index caches `(ctime, ino, mtime, size)` per entry, so a changed ino/ctime invalidates the stat cache for every source file on every wake run — `git status` / `git diff` degrade to full content re-reads across the tree. This problem is addressed by gating materialization for `source` and `claim` jobs.

Our current strategy for output files is to continue materialize all output files during a cached run - as a file equivalence check is expensive (through hashing) and using only an lstat does not work for our purposes (multiple wake runs can modify a workspace path) - the workspace is no longer the source of truth either way in the CAS world and therefore not critical to maintain correctness. This PR does not address the situation where an output files becomes elevated to a source file between runs (whenever a user calls `git add` to an output), and therefore that file will continue to materialize - we can address this situation in the future.

### DIFF
--- a/share/wake/lib/system/job.wake
+++ b/share/wake/lib/system/job.wake
@@ -264,6 +264,11 @@ def finalizeJob job inputs outputs cleanable reality runnerError =
     (\_ primJobFinish job inputs.implode preparedOutputStrings.implode cleanable.implode reality)
     clResult
 
+def isSourceOrClaimJob (cmd: List String): Boolean = match cmd
+    prefix, _ if prefix ==* sourceJobCommandPrefix -> True
+    prefix, _ if prefix ==* claimJobCommandPrefix -> True
+    _ -> False
+
 def runAlways cmd env dir stdin res uusage finputs foutputs vis keep run echo stdout stderr label isatty: Job =
     def hash = jobSignature cmd res finputs foutputs keep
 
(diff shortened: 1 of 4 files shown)


ngraybeal (approved):

share/wake/lib/system/job.wake
```diff
@@ -264,6 +264,11 @@ def finalizeJob job inputs outputs cleanable reality runnerError =
     (\_ primJobFinish job inputs.implode preparedOutputStrings.implode cleanable.implode reality)
     clResult
 
+def isSourceOrClaimJob (cmd: List String): Boolean = match cmd
+    prefix, _ if prefix ==* sourceJobCommandPrefix -> True
+    prefix, _ if prefix ==* claimJobCommandPrefix -> True
+    _ -> False
```
ag-eitilt, line 270: As we've been discussing, this is tech debt pending a deeper rework around more "natively" handling `source`/`claim` in such a way that they don't need to be special-cased based on their faked command line.  Looks good for now, though.

ag-eitilt (approved):

ag-eitilt:
> Git's index caches `(ctime, ino, mtime, size)` per entry, so a changed ino/ctime invalidates the stat cache for every source file on every wake run — `git status` / `git diff` degrade to full content re-reads across the tree.

I had hoped we could sidestep over-materialization issues by ensuring the mtime was always properly set, but this is a good catch of how things might try to be smarter, and in doing so break that sort of partial solution.  Thanks for the fix!

PR #1899, opened 2026-08-25
wake --attach

AbrarQuazi:
This PR adds `wake --attach JOBID`, which opens a shell in a live FUSE-sandboxed Wake job. This lets users inspect the job's in-progress outputs through its virtualized workspace view. It also prints out the live FUSE workspace path, giving options to users to use external tools against that fuse path.

It adds a `live_jobs` table that records the active job's Wake run and wakebox-main PID. Attach uses that PID to find the wakebox payload child, which owns the sandbox's user and mount namespaces. A forked attach child joins those namespaces, restores the job's working directory, and execs the user's shell.

The attached shell and printed out fuse path is read/write, so changes affect the in-progress job and may change the build result.

### DIFF
--- a/share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
+++ b/share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
@@ -1,0 +1,34 @@
+# Inspecting a Live Job
+
+_Part of [Workspace Virtualization and Multiple Wake Invocations](../workspace-virtualization-and-multi-wake.md)._
+
+Use `wake --ps` to find a running job ID:
+
+```
+$ wake --ps
+Run 12: wake -x 'buildEverything Unit'
+  JOB     ELAPSED     LABEL
+  341     [1m12s]     compile core/main.cpp
+  342     [1m04s]     compile core/util.cpp
+  348     [queued]    link core
+```
+
+Then attach a shell to job `341`'s live FUSE-sandboxed workspace:
+
+```
+wake --attach 341
+```
+
+The attached shell is read/write: changes to the workspace affect the running
+job and can change its build result. Its banner also prints the host-visible
+FUSE root for the job's live workspace. Use a host-installed external tool
+against a workspace-relative live file below that root, for example:
+
+```
+fuse_root=/scratch/user/workspace/.fuse/uid.gid/job-pid
+vim -R "$fuse_root/build/output"
+```
+
+The FUSE path presents live staging data under its real workspace-relative
+names. It remains valid only while the job is running. It is the same writable
+view used by the job, so do not modify it: writes can affect the build result.
--- a/share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
+++ b/share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
@@ -13,6 +13,10 @@ Run 12: wake -x 'buildEverything Unit'
   348     [queued]    link core
 ```
 
+When this list is large, combine `--ps` with the other job filters, especially
+`--label`, to narrow it. Other useful filters include `--input`, `--output`,
+`--tag`, and `--job`.
+
 Then attach a shell to job `341`'s live FUSE-sandboxed workspace:
 
 ```
--- a/share/doc/wake/workspace-virtualization/limitations.md
+++ b/share/doc/wake/workspace-virtualization/limitations.md
@@ -33,11 +33,15 @@ _Part of [Workspace Virtualization and Multiple Wake Invocations](../workspace-v
   - `wake --ps` lists jobs currently running across all active builds, grouped
     by run, with each job's id, elapsed time (or `[queued]`), and label. Use it
     to find the id of the job you want to inspect.
-  - `wake --job <jobID> -v` describes a specific job in verbose form, including
-    its command-line, environment, working directory, and any stdout/stderr
-    streamed so far.
+   - `wake --job <jobID> -v` describes a specific job in verbose form, including
+     its command-line, environment, working directory, and any stdout/stderr
+     streamed so far.
+   - `wake --attach <jobID>` opens a read/write shell in a FUSE-sandboxed job's
+     live filesystem view, including its in-progress outputs at their real
+     workspace paths. Its banner prints the host-visible FUSE root for that
+     job, which host-installed tools can use to read a live file by its
+     workspace-relative path.
 
-  These options let you watch a job's command and streamed output, but they do
-  *not* expose the job's not-yet-completed output files, since those are not
-  materialized until the job finishes. Inspecting the in-progress build outputs
-  of a long-running job is something the Wake team is actively looking into.
+    Attach is available only for a live FUSE-sandboxed job run by the same user.
+    Its shell and the host-visible FUSE path both use the job's live writable
+    workspace, so changes through either can affect the build result.
--- a/src/runtime/schema.h
+++ b/src/runtime/schema.h
@@ -1,7 +1,7 @@
 #ifndef WAKE_SCHEMA_H
 #define WAKE_SCHEMA_H
 
-#define SCHEMA_VERSION "16"
+#define SCHEMA_VERSION "17"
 
 // Per-connection settings to always apply.  Do this first!
 inline const char *getCommonPragmaSQL() {
@@ -114,6 +114,11 @@ inline const char *getWakeSchemaSQLTxn() {
          "  file_id integer not null references files(file_id) on delete cascade,"
          "  primary key(file_id, run_id));"
          "create index if not exists run_files_by_run on run_files(run_id, file_id);"
+         "create table if not exists live_jobs("
+         "  job_id integer primary key references jobs(job_id) on delete cascade,"
+         "  run_id integer not null references runs(run_id) on delete cascade,"
+         "  pid    integer not null);"
+         "create index if not exists live_jobs_by_run on live_jobs(run_id, job_id);"
          // clang-format off
          "insert or ignore into schema(version) values(" SCHEMA_VERSION ");"
          "pragma user_version=" SCHEMA_VERSION ";"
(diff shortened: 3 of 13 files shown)


AbrarQuazi:
Need to update documentation

tools/wake/main.cpp
```diff
@@ -432,6 +433,7 @@ void print_help(const char *argv0) {
     << "    --last-executed    Capture all jobs executed by the last build. Skips cache"   << std::endl
     << "    --history          Report the cmndline history of all wake commands recorded"  << std::endl
     << "    --ps               Show jobs currently running in active wake builds"          << std::endl
+    << "    --attach           With --job JOB, attach a shell to its live FUSE sandbox"    << std::endl
```
ngraybeal: Could `--attach` just be `--attach JOB` instead of requiring `--job`?

tools/wake/attach.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    std::cerr << "+------------------------------------------------------------------+\n"
+              << warning_line("Attached to the live sandbox for Wake job " + std::to_string(job_id))
+              << '\n'
+              << warning_line("PID: " + std::to_string(*payload_pid) + "  CWD: " + cwd_display)
+              << '\n'
+              << warning_line("") << '\n'
+              << warning_line("WARNING: This shell is read/write. Any changes you make affect")
+              << '\n'
+              << warning_line("the in-progress job and may change the build result.") << '\n'
```
ngraybeal: Would there be a way to convert a job from Share to Keep if we attach to the job? Not sure if we it's possible to intercept that from being shared.

ngraybeal (commented):
Is there any functionality that should be shared between `--ps` and `--attach`, like the live job table? The tracking of job state seems pretty adjacent.

tools/wake/main.cpp
AbrarQuazi: Yeah we can make it this

tools/wake/attach.cpp
AbrarQuazi: Why would we want this? And no, it would be hard to do and not clean code to be able to do that

AbrarQuazi:
> Is there any functionality that should be shared between `--ps` and `--attach`, like the live job table? The tracking of job state seems pretty adjacent.

Technically the live_job pid in the DB is not the pid of the actual job but rather the pid of the parent wakebox-main of that job - so dont know how useful that would be. `attach` currently does not need the pid, but the job id for the command line. We can maybe add this in a later PR if we feel that is useful.

tools/wake/attach.cpp
```diff
@@ -0,0 +1,238 @@
...
+    std::cerr
+        << "+------------------------------------------------------------------+\n"
+        << warning_line("Attached to the live sandbox for Wake job " + std::to_string(job_id))
+        << '\n'
+        << warning_line("Payload PID: " + std::to_string(*payload_pid) + "  CWD: " + cwd_display)
+        << '\n'
+        << warning_line("Host FUSE live workspace:") << '\n'
+        << warning_line(host_live_workspace) << '\n'
+        << warning_line("Tools unavailable in shell can be run using the host FUSE path") << '\n'
+        << warning_line("above.") << '\n'
```
ngraybeal, line 218: Maybe information overload? While helpful, not sure if its pertinent information.

tools/wake/attach.h
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2019 SiFive, Inc.
```
ngraybeal: Update Copyright

ngraybeal (approved):

share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
```diff
@@ -0,0 +1,34 @@
+# Inspecting a Live Job
+
+_Part of [Workspace Virtualization and Multiple Wake Invocations](../workspace-virtualization-and-multi-wake.md)._
+
+Use `wake --ps` to find a running job ID:
```
ag-eitilt, line 5: Might be worth mentioning that the other filters (especially `--label`) can help narrow down a large live list.

share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
```diff
@@ -0,0 +1,34 @@
...
+Then attach a shell to job `341`'s live FUSE-sandboxed workspace:
+
+```
+wake --attach 341
+```
+
+The attached shell is read/write: changes to the workspace affect the running
+job and can change its build result. Its banner also prints the host-visible
+FUSE root for the job's live workspace. Use a host-installed external tool
+against a workspace-relative live file below that root, for example:
```
ag-eitilt, line 29:
```suggestion
FUSE root for the job's live workspace. You can use a host-installed external
tool against a workspace-relative live file below that root, for example:
```
However, note yesterday's discussion that it might be better to find solutions that don't rely pointing users to implementation details.

tools/wake/attach.cpp
ag-eitilt: So long as the sandbox is read/write (which isn't a bad thing in itself), I do think it's a good idea to downgrade the cacheability of the job -- any modifications the user makes are inherently going to be unreproducible.  There's arguments to be made both ways for whether the resulting state should be `Keep` (if the user is just doing transparent debugging of a live job, it would be nice if they weren't forced to rerun some long compute if they rerun the wake command) or whether it should be `Once` (any changes to the outputs which *are* made wouldn't corrupt the reproducibility of the job in later runs) though I lean toward the latter.

However, in order to be more responsive to whether the user *actually* did anything to invalidate reproducibility, we'd have to include fancy logic in the Runner layer, and likely more in wakebox, so to keep the solution general and straightforward I don't think we should try to do anything fancier than an all-or-nothing toggle based purely on `--attach` itself.

The implementation difficulty is a separate issue, though, particularly given that the `Share` connections to the RSC are based on the state of the Plan before anything gets launched in the first place.  For a basic implementation it might be enough to have `--attach` set the `keep` column of the job to some sentinel value, and then prevent the finalization transaction from setting the column to be cacheable if that sentinel is present.  `Once` (and greater) jobs would still be cached properly since that's implemented as a wakelang-side `target`.  Exposing that status to the Runner (to abort sending to the RSC) would be possible but a whole lot more complex, and would probably come after Nathan's work with Usage/resources.

I think it's acceptable to add the cacheability downgrade as a followup to this PR (even in a separate release), but it is probably something we should look into.

share/doc/wake/workspace-virtualization/limitations.md
```diff
@@ -33,11 +33,15 @@ _Part of [Workspace Virtualization and Multiple Wake Invocations](../workspace-v
...
+   - `wake --job <jobID> -v` describes a specific job in verbose form, including
+     its command-line, environment, working directory, and any stdout/stderr
+     streamed so far.
+   - `wake --attach <jobID>` opens a read/write shell in a FUSE-sandboxed job's
+     live filesystem view, including its in-progress outputs at their real
+     workspace paths. Its banner prints the host-visible FUSE root for that
+     job, which host-installed tools can use to read a live file by its
+     workspace-relative path.
+
+    Attach is available only for a live FUSE-sandboxed job run by the same user.
```
ag-eitilt, line 45: Eventually I'd like to expand this into things that don't rely on FUSE, even if that just means a `cd` into a not-even-namespaced `stagingDirRunner`.  That's low priority, though.

src/runtime/schema.h
```diff
@@ -114,6 +114,11 @@ inline const char *getWakeSchemaSQLTxn() {
          "  file_id integer not null references files(file_id) on delete cascade,"
          "  primary key(file_id, run_id));"
          "create index if not exists run_files_by_run on run_files(run_id, file_id);"
+         "create table if not exists live_jobs("
+         "  job_id integer primary key references jobs(job_id) on delete cascade,"
+         "  run_id integer not null references runs(run_id) on delete cascade,"
+         "  pid    integer not null);"
+         "create index if not exists live_jobs_by_run on live_jobs(run_id, job_id);"
```
ag-eitilt, line 121: I'm not sure I like this solution; I'd much prefer if we could determine "live" status from a column or set of columns in the `jobs` table itself -- and ideally something based on the existing values rather than adding a new flag.[^1]  This seems like it's adding a point of failure that could easily fall out of sync with the main job listing.

Besides, this turns the PR into a breaking schema change, which it would be nice to avoid this soon after the big workspace invalidation.

[^1]: Specifically, something like checking one value which gets set in the process of being initialized/launched, and one which gets set when it's finished/canceled.

ag-eitilt (commented):
I am very hesitant about the new table.  If it's the only way to do this then fair enough, I'm not going to block it, but I do want to see if we can find an alternate solution.  Haven't looked closely into the `*.cpp` changes yet because of that.

share/doc/wake/workspace-virtualization/limitations.md
AbrarQuazi: Agreed

src/runtime/schema.h
AbrarQuazi: As mentioned in our meeting, I think we need a better story of keeping system and runner related into into another database or location, rather than the DB. I'd like to separate this system related PID information away from the Jobs table (as that is primarily used for caching), and have pid in its own table that we can later export away to something else when we decide to do that. 

In terms of the live_jobs falling out of sync, I do realize there is potential for this (for example if wake crashes, but that should be reaped), but I do that whatever uses it currently (wake --attach and possibly --ps in the future) is moreso best effort introspection rather than used for anything authoritative.

share/doc/wake/workspace-virtualization/inspecting-live-jobs.md
AbrarQuazi: Yep, but still will include this for now, until we have a better solution

PR #1901, opened 2026-08-31
Make FUSE jobs see only the mapped uid/gid

AbrarQuazi:
This PR solves a bug in FUSE, where if a host backed file has a supplementary gid set, which is not the users gid which gets mapped into the sandbox namespace, there will be permission errors within that job. Jobs should only see the mapped gid in the sandbox, and not anything else to avoid these permission issues.

### DIFF
--- a/tests/wakebox/setgid-owner-metadata/input.json
+++ b/tests/wakebox/setgid-owner-metadata/input.json
@@ -1,0 +1,23 @@
+{
+  "command": [
+    "/bin/sh",
+    "-c",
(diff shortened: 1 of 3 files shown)


ag-eitilt (changes requested):
I haven't checked that this is *all* the places which need the ownership tweak, but what is here seems reasonable.

However, looks like you accidentally added a lot of files which just happened to be in your working directory...

ag-eitilt (approved):

PR #1902, opened 2026-09-01
RSC decode db-scheme content directly

jackkoenig:
I noticed RSC cache hits that were failing to rehydrate. The added breadcrumb pointed me (and my best agent) to a race condition from identical stdout/stderr sharing a common path and the file being deleted while a later rehydration is trying to read it 

The error messages were of the form:
```
read .build/tmp/writes/rsc.blob...: No such file or directory
```

This is an attempt to avoid the issue by just reading the content directly in memory rather than writing it to a file.

A couple of things I'd like feedback on:
1. Is the performance of this okay using regular expressions? That's what `resolveBlobsToPaths` does.
2. I changed this just for `rscApiGetStringBlobs` (avoiding calling `resolveBlobsToPaths`) but not for the other caller which is `rscApiGetBlobsToCas`. Should that also be changed or should perhaps this logic be inlined into `resolveBlobsToPaths` (and obviously the function be changed to contemplate returning values of either `Path` or `String`)?

### DIFF
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -643,10 +643,16 @@ def chunkList (n: Integer) (xs: List a): List (List a) =
 
     head, chunkList n tail
 
-# resolveBlobsToPaths: Resolves many blobs to local file paths holding their bytes. db-scheme blobs
-# are decoded inline (no network); file/http(s) blobs are downloaded via few batched curl jobs. Blobs
-# are deduped by id so shared content is only fetched once. Returns an assoc list of (blobId -> Path).
-def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String Path)) Error =
+# Either: Holds one of two possible values.
+data Either a b =
+    Left a
+    Right b
+
+# resolveBlobsToPathsOrStrings: Resolves many blobs to local file paths or inline strings. db-scheme
+# blobs are decoded inline (no network); file/http(s) blobs are downloaded via few batched curl jobs.
+# Blobs are deduped by id so shared content is only fetched once. Returns an assoc list of
+# (blobId -> Either Path String).
+def resolveBlobsToPathsOrStrings (blobs: List CacheSearchBlob): Result (List (Pair String (Either Path String))) Error =
     # Dedupe by blob id: identical ids resolve to identical bytes, so fetch each once.
     def uniqueBlobs =
         blobs
@@ -677,13 +683,11 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
         parsedBlobs
         | splitBy (\(Pair _ (Pair scheme _)) scheme ==* "db")
 
-    # Resolve db-scheme blobs inline to temp files (no curl), reusing the already-parsed scheme/path.
-    def resolveDbBlob ((Pair blob (Pair scheme path)): Pair CacheSearchBlob (Pair String String)): Result (Pair String Path) Error =
-        require Pass tmp =
-            resolveDbScheme scheme path
-            |> writeTempFile "rsc.blob.{blob.getCacheSearchBlobId}"
+    # Decode db-scheme blobs inline rather than materializing tracked temporary files.
+    def resolveDbBlob ((Pair blob (Pair scheme path)): Pair CacheSearchBlob (Pair String String)): Result (Pair String (Either Path String)) Error =
+        require Pass content = resolveDbScheme scheme path
 
-        Pass (Pair blob.getCacheSearchBlobId tmp)
+        Pass (Pair blob.getCacheSearchBlobId (Right content))
 
     require Pass dbResolved =
         dbParsed
@@ -698,7 +702,7 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
     def wakePid: Integer = prim "pid"
     def httpDir = ".build/wake-{str wakePid}/stdlib/http"
 
-    def downloadChunk (chunk: List CacheSearchBlob): Result (List (Pair String Path)) Error =
+    def downloadChunk (chunk: List CacheSearchBlob): Result (List (Pair String (Either Path String))) Error =
         # Sort the chunk deterministically by blob id so that two batches with the same blob set
         # produce byte-identical curl jobs (which wake then dedups to a single job / single writer).
         def sortedChunk =
@@ -731,7 +735,7 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
         sortedChunk
         | map (
             \b match (mlookup (destForBlob b) pathByName)
-                Some p -> Pass (Pair b.getCacheSearchBlobId p)
+                Some p -> Pass (Pair b.getCacheSearchBlobId (Left p))
                 None ->
                     failWithError "rsc: batched download did not produce output for blob {b.getCacheSearchBlobId}"
         )
@@ -749,33 +753,20 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
 # rscApiGetStringBlobs: Batched analogue of rscApiGetStringBlob. Downloads many blobs, verifies each
 # blob's content hash, and returns the contents in the same order as the input.
 export def rscApiGetStringBlobs (blobs: List CacheSearchBlob): Result (List String) Error =
-    def isDbSchemeBlob (blob: CacheSearchBlob): Boolean =
-        matches `db://.*` blob.getCacheSearchBlobUri
+    require Pass resolved = resolveBlobsToPathsOrStrings blobs
 
-    # Decode db-scheme blob content directly rather than materializing a tracked temporary file.
-    # Routing them through writeTempFile causes
-    # concurrent cache hits for identical stdout/stderr to share a tracked path,
-    # which can disappear while another rehydration is still reading it.
-    def networkBlobs = filter (! isDbSchemeBlob _) blobs
-
-    require Pass resolved = resolveBlobsToPaths networkBlobs
-
-    def pathById =
+    def contentById =
         resolved
         | listToMap scmp
 
     def resolveOne ((CacheSearchBlob blobId _uri contentHash): CacheSearchBlob): Result String Error =
         require Pass content =
-            require (scheme, path, Nil) = extract `([a-zA-Z][a-zA-Z0-9+.-]*)://(.*)` _uri
-            else failWithError "rsc: uri has unexpected format: '{_uri}'"
+            require Some resolvedContent = mlookup blobId contentById
+            else failWithError "rsc: no downloaded bytes found for blob {blobId}"
 
-            if scheme ==* "db" then
-                resolveDbScheme scheme path
-            else
-                require Some blobPath = mlookup blobId pathById
-                else failWithError "rsc: no downloaded bytes found for blob {blobId}"
-
-                read blobPath
+            match resolvedContent
+                Left blobPath -> read blobPath
+                Right inlineContent -> Pass inlineContent
 
         def actualHash = hashString content
 
@@ -808,15 +799,20 @@ export def rscApiGetBlobsToCas (entries: List (Pair CacheSearchBlob (Pair String
         entries
         | map getPairFirst
 
-    require Pass resolved = resolveBlobsToPaths blobs
+    require Pass resolved = resolveBlobsToPathsOrStrings blobs
 
-    def pathById =
+    def contentById =
         resolved
         | listToMap scmp
 
     def prepareOne ((Pair (CacheSearchBlob blobId _uri contentHash) (Pair path mode)): Pair CacheSearchBlob (Pair String Integer)): Result PreparedStagingItem Error =
-        require Some blobPath = mlookup blobId pathById
+        require Some resolvedContent = mlookup blobId contentById
         else failWithError "rsc: no downloaded bytes found for blob {blobId}"
+
+        require Pass blobPath = match resolvedContent
+            Left downloadedPath -> Pass downloadedPath
+            # CAS ingestion needs a path, unlike the stdout/stderr string consumer above.
+            Right inlineContent -> writeTempFile "rsc.blob.{blobId}" inlineContent
 
         def actualHash = blobPath.getPathHash
 
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -792,7 +792,8 @@ export def rscApiGetStringBlob (blob: CacheSearchBlob): Result String Error =
     Pass content
 
 # rscApiGetBlobsToCas: Batched analogue of rscApiGetFileBlobToCas. Downloads many file blobs,
-# verifies each blob's hash, and ingests each into CAS. Returns the PreparedStagingItems
+# verifies each blob's hash, and ingests each into CAS. Inline blob content is stored directly in
+# CAS rather than being written to a temporary file. Returns the PreparedStagingItems
 # (in input order) so the caller can materialize them later, after every blob for a job has been verified.
 export def rscApiGetBlobsToCas (entries: List (Pair CacheSearchBlob (Pair String Integer))): Result (List PreparedStagingItem) Error =
     def blobs =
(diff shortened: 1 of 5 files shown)


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -749,17 +749,33 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
 # rscApiGetStringBlobs: Batched analogue of rscApiGetStringBlob. Downloads many blobs, verifies each
 # blob's content hash, and returns the contents in the same order as the input.
 export def rscApiGetStringBlobs (blobs: List CacheSearchBlob): Result (List String) Error =
-    require Pass resolved = resolveBlobsToPaths blobs
+    def isDbSchemeBlob (blob: CacheSearchBlob): Boolean =
+        matches `db://.*` blob.getCacheSearchBlobUri
+
+    # Decode db-scheme blob content directly rather than materializing a tracked temporary file.
+    # Routing them through writeTempFile causes
+    # concurrent cache hits for identical stdout/stderr to share a tracked path,
```
ag-eitilt, line 757: Is this an issue with `writeTempFile` itself?  Should we be using `ReRun` rather than caching the job at all?

share/wake/lib/system/remote_cache_api_test.wake
```diff
@@ -2,6 +2,25 @@ package remote_cache
 
 from wake import _
 
+export def testInlineStringBlobDecoding Unit: Result Unit Error =
+    def content = "shared stdout"
+    def encoded =
+        content
+        | unicodeToBytes
+        | map (\byte ('%', strHex byte, Nil) | cat)
```
ag-eitilt, line 10: Not that optimizing a test is crucial, but:
```suggestion
        | map ("%{strHex _}")
```

ag-eitilt (commented):
I can't speak to question 2, but so far as the use of regular expressions goes (throughout this PR, not just `resolveBlobsToPaths`), we've seen them have an outsized impact on memory/compute in the past.  I don't think we need to go out of our way to *avoid* them most of the time, but this is high-frequency code and if we can find an alternate solution in this case I do think we should take it.

share/wake/lib/system/remote_cache_api.wake
jackkoenig: I'm not really sure here so need to you and @AbrarQuazi 

What I do see is that `writeTempFile` calls `writeIn` calls `writeImp` which does [setPlanOnce False](https://github.com/sifiveinc/wake/blob/c4e444af7b3930ef0c389275c8ea61c1bb7bd8ed/share/wake/lib/system/io.wake#L258) which has the comment:
```
# If `Once` is set to `False`, no job deduplication is performed and so it must
# *not* write any files (stdout/stderr are fine) or be guaranteed to only be
# encountered once in any wake execution anyway.
```

I don't fully understand the race condition, but my debugging indicates there is one around these files.

ag-eitilt: That comment shouldn't actually come into play; since `writeImp` is a `target`, it's only ever going to be called once for the same arguments to begin with, and the `setPlanOnce False` is just a micro optimization to avoid the job *also* being stored in its own memoization table (where it would never be hit a second time).

Still looking for what exactly might be causing the race.

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -749,17 +749,33 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
...
+    # which can disappear while another rehydration is still reading it.
+    def networkBlobs = filter (! isDbSchemeBlob _) blobs
```
AbrarQuazi: I am a little bit confused on why this is an issue for only db based blobs and not all blobs

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: Oh so I think I understand what is going on here, so DB files can end up writing to the same path `.build/tmp/writes/rsc.blob.{blobId}.{hash content}`  (not keyed by pid unlike the network files `.build/wake-{pid}/stdlib/http/rsc.blob.{batchKey}.{blobId}`.

So `WriteTempFile` calls `prim_write` which unlinks the destination file before writing to that path - which is what is causing the race

share/wake/lib/system/remote_cache_api.wake
AbrarQuazi: Ah nvm I think I understand now

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -792,16 +799,21 @@ export def rscApiGetBlobsToCas (entries: List (Pair CacheSearchBlob (Pair String
...
 
     def prepareOne ((Pair (CacheSearchBlob blobId _uri contentHash) (Pair path mode)): Pair CacheSearchBlob (Pair String Integer)): Result PreparedStagingItem Error =
-        require Some blobPath = mlookup blobId pathById
+        require Some resolvedContent = mlookup blobId contentById
         else failWithError "rsc: no downloaded bytes found for blob {blobId}"
 
+        require Pass blobPath = match resolvedContent
+            Left downloadedPath -> Pass downloadedPath
+            # CAS ingestion needs a path, unlike the stdout/stderr string consumer above.
+            Right inlineContent -> writeTempFile "rsc.blob.{blobId}" inlineContent
```
jackkoenig: I pulled the `db://` scheme handling into `resolveBlobsToPathsOrStrings` (formerly `resolveBlobsToPaths`), but the CAS path is a lot more Path-oriented so I'm punting on the problem here.

@AbrarQuazi should I pull on this thread some more? I'll probably need to change `StagingItem` to have a `StagingString` alternative or something like that.

AbrarQuazi: Yeah we need a fix to the CAS path as well, since it looks like that is affected as well. I assume we can do this by just making the temp file keyed by pid path like we do for the network blobs right?

AbrarQuazi: Dont think we should edit `StagingItem`

jackkoenig: I was thinking it would be more performant to keep the Strings just in memory rather than writing to a file, or is that just how the CAS works with all of the String-y things in the RSC?

jackkoenig: Or put another way, is there any point to handling the `db://` values as Strings in `rscApiGetStringBlobs` or should we just write temp files there keyed by pid path as well?

AbrarQuazi: Ah yeahh I see what you mean, the DB files are very small < 95 bytes, so the cost of going through the entire staging path and materialization for CAS is pretty costly.

That would mean that we need a new `StagingInlineFile` /  `StagingString` for the `StagingItem` data structure (plus some c++ changes to support that). Up to you if you want to do this in this PR, or just keep it written to a temp file keyed by pid for the temporary fix, and we address this in follow up PRs

jackkoenig: I tried out undoing most of this PR and just including the PID in the temp file path:
```diff
diff --git a/share/wake/lib/system/remote_cache_api.wake b/share/wake/lib/system/remote_cache_api.wake
index 56cae31e..5e6b547f 100644
--- a/share/wake/lib/system/remote_cache_api.wake
+++ b/share/wake/lib/system/remote_cache_api.wake
@@ -677,11 +677,13 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
         parsedBlobs
         | splitBy (\(Pair _ (Pair scheme _)) scheme ==* "db")
 
+    def wakePid: Integer = prim "pid"
+
     # Resolve db-scheme blobs inline to temp files (no curl), reusing the already-parsed scheme/path.
     def resolveDbBlob ((Pair blob (Pair scheme path)): Pair CacheSearchBlob (Pair String String)): Result (Pair String Path) Error =
         require Pass tmp =
             resolveDbScheme scheme path
-            |> writeTempFile "rsc.blob.{blob.getCacheSearchBlobId}"
+            |> writeTempFile "rsc.blob.wake-{str wakePid}.{blob.getCacheSearchBlobId}"
 
         Pass (Pair blob.getCacheSearchBlobId tmp)
 
@@ -695,7 +697,6 @@ def resolveBlobsToPaths (blobs: List CacheSearchBlob): Result (List (Pair String
         | map getPairFirst
 
     # Batched download of the network blobs.
-    def wakePid: Integer = prim "pid"
     def httpDir = ".build/wake-{str wakePid}/stdlib/http"
 
     def downloadChunk (chunk: List CacheSearchBlob): Result (List (Pair String Path)) Error =
```

And when I do that, I still get rehydration failures, e.g.
```
read .build/tmp/writes/rsc.blob.wake-3115200.0944e941-8fd0-4da7-a6fc-e9b0b651f453.af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262: No such file or directory
```

So I don't think keying by PID is enough to resolve the race condition.

AbrarQuazi: Hmmm thats interesting.... not understanding why, when we do that for network blobs.

I guess that makes the case to have the StagingString that you mentioned

jackkoenig: Okay @AbrarQuazi, the most recent commit is an attempt to directly inject inline content into the CAS, avoiding a temporary file.

jackkoenig: It ended up being a lot cleaner to do it directly in `rscApiGetStringBlobs` rather than adding `StagingString` as a type of `StagingItem`

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -785,36 +792,49 @@ export def rscApiGetStringBlob (blob: CacheSearchBlob): Result String Error =
...
+                PreparedStagingItem
+                (StagingFile (StagingFileOutput path downloadedPath.getPathName mode 0 0))
+                contentHash
+                | Pass
+            Right inlineContent ->
+                require Pass _ = ingestStringIntoCas path inlineContent contentHash
+
+                Pass (PreparedStagingItem (StagingFile (StagingFileOutput path path mode 0 0)) contentHash)
+
+        def ingestPrepared = match resolvedContent
```
AbrarQuazi: Any way to consolidate these `match resolvedContent`, or make it cleaner? We are doing it for the third time here

share/wake/lib/system/remote_cache_api.wake
```diff
@@ -785,36 +792,49 @@ export def rscApiGetStringBlob (blob: CacheSearchBlob): Result String Error =
...
+        # Use 0 for mtime; this is set to the current time on materialization. The staging path is
+        # not needed after CAS ingestion, so inline blobs use their destination as a placeholder.
+        require Pass preparedItem = match resolvedContent
+            Left downloadedPath ->
+                PreparedStagingItem
+                (StagingFile (StagingFileOutput path downloadedPath.getPathName mode 0 0))
+                contentHash
+                | Pass
+            Right inlineContent ->
+                require Pass _ = ingestStringIntoCas path inlineContent contentHash
```
AbrarQuazi: hmmm this is a bit awkward, because we are ingesting into CAS first and then making it a PreparedStagingItem.

Typically the flow of the code should be  PreparedStagingItem (Item is in staging directory) -> Ingestion (put it into CAS) . But it looks like we are doing opposite here because of the match. Can we do something like this:

```
def Pair actualHash stagingPath = match resolvedContent
    Left downloadedPath -> Pair downloadedPath.getPathHash downloadedPath.getPathName
    Right inlineContent -> Pair (hashString inlineContent) path

require Pass _ = verifyBlobHash blobId actualHash contentHash

# Use 0 for mtime; set to the current time on materialization. Inline blobs have no staging
# file, so they use their destination as the staging path (ignored by CAS materialization).
def preparedItem =
    PreparedStagingItem (StagingFile (StagingFileOutput path stagingPath mode 0 0)) contentHash

require Pass _ =
    match resolvedContent
        Left _ -> ingestPreparedStagingItemIntoCas preparedItem
        Right inlineContent -> ingestStringIntoCas path inlineContent actualHash
    | addErrorContext "rsc: Failed to ingest blob {blobId} into CAS for {path}"

Pass preparedItem
``

**Review by AbrarQuazi — COMMENTED** [review]
Some nits, but overall looks good

File: share/wake/lib/system/remote_cache_api.wake (hunk 1 above)
Comment (ag-eitilt) [reply share/wake/lib/system/remote_cache_api.wake]: ~~I'm still not convinced that's the case.  If the job was actually being run multiple times, yeah, definitely the issue, and the `setPlanOnce False` does indeed mean that every time it got hit then it would run.  However, it's wrapped in `writeImp` -- which as I said is a `target` so in order to reach the job multiple times, one of the `writeImp` arguments would have to be different.  `path` being different would obviously mean the unlink wouldn't happen, `inputs` only contains the directory in which the file is being written, and `writeTempFile` hardcodes the `mode` so that's not going to be the change.  The only thing that might be different is the `content`, and since that's fed right into the `path` via a `hashString`, we'd only have colliding paths if we actually found a hash collision.~~

...Or, actually, are you seeing this bug in *multi-wake* cases rather than within a single run?  That would certainly run into the exact issue you're seeing.  Apparently still not used to thinking about the category of cross-run bugs, given that that framing never crossed my mind.

Comment (AbrarQuazi) [reply share/wake/lib/system/remote_cache_api.wake]: blobs from different jobs can have the same content, right? So they would write to the same path .build/tmp/writes/rsc.blob.{blobId}.{hash content} ? So its not only the multi-wake case

File: share/wake/lib/system/remote_cache_api.wake (hunk 6 above)
Comment (ag-eitilt) [reply share/wake/lib/system/remote_cache_api.wake]: I've actually run into a case where I want to do something similar: construct a PreparedStagingItem that skips being ingested, because it's constructed from a hash we retrieved from the CAS/DB already (specifically, to optimize overlapping `sources` lists when batching source hashing).  Given that model's popped up in two separate places, is it something we want to look into officially supporting, or are Jack and i both mangling the architecture?

For the record, the solution I was playing with was to add an `Option String` hash field to StagingFileOutput, which would then trigger skipping the `primCasIngestStagedItem` in `ingestPreparedStagingItemIntoCas` if the hash was `Some`.

Comment (AbrarQuazi) [reply share/wake/lib/system/remote_cache_api.wake]: In Jacks case, we were ingesting first before creating `PreparedStagingItem`. In your case, it seems like you need some data structure to reflect something already in CAS, something like `StagingAlreadyInCas` (although the Stage name here doesnt make much sense in this case).

I do think we should have separate datastructure's for the different cases, because it the cas ingestion and materialization logic is already complex as it is - and reusing the same structure for different things complicates readability and understanding even further. I'd like to keep `PreparedStagingItem` in its place in the flow of things , which is pre-materialization. I'm open to discuss if there are better ways to organize and enforce these structures for their specific use cases.

File: share/wake/lib/system/remote_cache_api.wake (hunk 1 above)
Comment (ag-eitilt) [reply share/wake/lib/system/remote_cache_api.wake]: It still shouldn't matter, though?  The only things that factor into whether the file would get overwritten are the path and the content, so either the blobs from different jobs have the same blobId and the same content hash (in which case the `writeImp` would be loaded from the memoization table and the file wouldn't be unlinked) or they have different blobIds (in which case they'd be written to different paths to begin with).

That makes it sound like this might be #549 cropping up again.  @jackkoenig, could you add `println`s (or some similar form of less-noisy debugging) to the `writeImp` so we can compare the values of its arguments on each of the times it's trying to write to the same file?  If they're equivalent, then it's definitely a case of the memory hashes in the `target` table being improperly generated, and we *will* have to solve this via workaround rather than root cause.

File: src/runtime/cas_prim.cpp [hunk 7]
```diff
@@ -429,6 +429,7 @@ static PRIMFN(prim_materialize_staged_workspace_item) {
 // prim "cas_ingest_staged_item" destPath type stagingPathOrTarget hash -> Result Unit Error
 // Stores staged content in CAS under its precomputed hash. Does not write to the workspace.
 // - type="file": stagingPathOrTarget = staging path
+// - type="string": stagingPathOrTarget = inline file contents
 // - type="symlink": stagingPathOrTarget = symlink target
 // - type="directory": no-op (directories have no CAS blob)
```
ag-eitilt, line 434: I'm still hesitant about adding `string` as a logic flag here.  The CAS doesn't store strings, and we aren't adding a new type of storage, we're just using it to indicate a different code path to get to a `file`.  This is still the best workaround we have *so far* so if we do decide it's what we want to merge, then I'll accept that.  Personally, though, I'd prefer if we either find a solution which is unreservedly good architecture, or find why exactly the `No such file or directory` is getting triggered so we know exactly what we're trying to avoid with a workaround.

ag-eitilt (commented):
What's the impact scale of this failure?  If it's causing Wake as a whole to abort, then it's very high priority and I'm happier to accept something subpar but working in the here and now.  If it's just causing a false cache miss and a job rerun, then it's still definitely still reasonably high priority, but I'd encourage spending a bit more time trying to better understand what's going wrong rather than just hacking out a broad swathe around the symptom.

share/wake/lib/system/remote_cache_api.wake
ag-eitilt: Yeah, that's one of the big reasons I've not published that branch yet; it's not a staging item at all in my case, and it's just being forced into that shape because of the Runner tuple types.  Hadn't gotten to an architecture I was happy with the scope of before I left.

jackkoenig: @AbrarQuazi your suggestion is a lot better, applied!

AbrarQuazi (approved):
LGTM

share/wake/lib/system/remote_cache_api.wake
ag-eitilt: Jack pointed out that I'd not shared my findings.  Turns out I'm pretty sure it's *not* that bug.  I can see `writeTempFile` getting hit multiple times with the same path+content, while `writeImp` is correctly only hit once.

So, not only is nothing going wrong with the `target` there, there's also *very* small chance this is due to a job or prim getting hit multiple times, and if it *is* that has to be outside the `writeImp` entirely.  Taking a short break from this issue, but when I pick it up again my next guess is that it'll be something in the ingestion/materialization code itself.

AbrarQuazi: Thanks for looking in to this, I didnt have the chance to fully go into the problem further

PR #1903, opened 2026-09-02
Add splitOnce for splitting a String based on another String

jackkoenig:
Motivated by mild performance concerns over uses of regex in https://github.com/sifiveinc/wake/pull/1902, this API is inspired by [Rust's split_once](https://doc.rust-lang.org/std/string/struct.String.html#method.split_once).

I used GPT 5.6 Terra, so I would appreciate review on the C++ especially.

This primitive can be used to build other things like `split` (which would presumably return a `List String`).


src/runtime/string.cpp
```diff
@@ -78,6 +79,35 @@ static PRIMFN(prim_strlen) {
...
+static PRIMFN(prim_split_once) {
+  EXPECT(2);
+  STRING(separator, 0);
+  STRING(string, 1);
+
+  const char *begin = string->c_str();
+  const char *end = begin + string->size();
+  const char *match =
+      std::search(begin, end, separator->c_str(), separator->c_str() + separator->size());
+  if (separator->size() != 0 && match == end) RETURN(alloc_nil(runtime.heap));
```
ag-eitilt: I don't know how the performance of extracting the `c_str`s for `std::search` would compare to a `std::string`-based function, or even if one's available.  Types aside, this looks reasonable, and certainly memory-efficient.

tests/standard-library/string/test.wake
ag-eitilt, line 1: Mind adding a test or two for separators with more than one character, particularly where there's a partial overlap before the actual match?  e.g. `splitOnce "ab" "aabbcc"` and/or `splitOnce "value1" "key:value0:value1"`.

ag-eitilt (approved):
Looks good to me, and would alleviate a lot of the regex overkill.  Thanks!

> I think the biggest question is of the type, do we want `splitOnce` to be the version that uses a `String` as a separator? What about a version taking a `RegExp` or a future `Character` type if that exists?

There's a valid argument for this taking a hypothetical `Character`, but since we can still build that around this prim but can't build this prim around a single-character type, I do think it's better to have this `String` as the base.  As for `RegExp`, the obvious problem there is that we don't necessarily know what the separator that was found looks like without returning the matched span as well.  I'm happy with this as it is.

tests/standard-library/string/test.wake
jackkoenig: Good suggestion, done.

src/runtime/string.cpp
jackkoenig: I asked GPT-5.6 Sol and it suggested using std::string_view::find which captures the intent better but won't do any allocations like `std::string` would.

PR #1905, opened 2026-09-14
runJobImp: sort visible paths

ngraybeal:
Goal:
Achieve a higher cache hit rate irrespective of visible file list inputs ordering. 

Background:
Job key hash is reliant on deterministic ordering of Visible inputs. While wake database will return a job's file list in order they were inserted into the database, there are no rules enforcing that insertion order. A runner is allowed to return files in different orders.

To achieve maximum cache ability, we sort in the runJobImp before calling runAlways or RunOnce. This ensure the job key hash is pretty deterministic and that target memoization for runOnce gets hit. 

Opens:
This does not help with cases of `target` relying on a deterministic order, but that gets pretty complicated. A user might have a list Path ordered in specific matter that we cannot sort incorrectly. 

Alternatives:
I did explore sorting a runner's output list in finalizeJob  but that solution relies heavily on some assumptions about how that list order is kept. This might be a valuable performance optimization if we end up having to sort visible lists constantly. This would also help with our `target` open a bit since


ag-eitilt (commented):
> This does not help with cases of target relying on a deterministic order **in general**, but that gets pretty complicated.

It does, however, solve the `runOnce` target specifically so that all the job persistences are handled correctly.

Before I approve this, though, could you benchmark some of the more complex flows with and without this change?  Specifically looking at speed and memory usage -- memory because the new visible-files `List` (and intermediate `Tree`) created in the course of sorting should theoretically be short-lived and quickly GCed, but we know that's one common source of problems; speed because we've seen problems with the unwrapping overhead of `Tree` before (though on further thought I'm not sure why it's any different from the singly-linked `List`).

If this does wind up with significant enough increases, it will likely be better to do the sort in the low-level `prim_job_create` and `prim_job_cache` implementations in `job.cpp`.  As discussed on the call, that would leave `setPlanPersistence Once` jobs as the only ones sensitive to visible-files ordering, but I do still believe the sort-on-lookup solution based on the DB `keep` column expansion would work for those.

ag-eitilt (approved):
Sounds like the results in that regard are good.  Thanks for checking, and thanks for the PR!

PR #1906, opened 2026-09-14
Migrate debian to trixie

ngraybeal:
Debian Bullseye has reached EOL of support

This migrates Debian build to Trixie.


ag-eitilt (approved):

ag-eitilt:
It looks like there's something that still needs to be updated wherever the "required to pass before merging" list is set.

ngraybeal:
> It looks like there's something that still needs to be updated wherever the "required to pass before merging" list is set.

Yeah, it's a setting in github I can change.

AbrarQuazi (approved):
LGTM

PR #1907, opened 2026-09-15
Format httpResponseStatusCode

ngraybeal:
Addresses files being produced with spaces and unnecessary Option type (ie `Some 400`).


share/wake/lib/system/remote_cache_api.wake
```diff
@@ -940,6 +940,12 @@ export def guardRemoteCacheDisabled Unit: Result Unit Error =
 def contentTypeFromSize size =
     if size < 95 then Some "blob/small" else None
 
+# Formats HTTP status code
+def formatHttpResponseStatusCode (response: HttpResponse): String =
+    response.getHttpResponseStatusCode
+    | omap format
+    | getOrElse "NA"
```
ag-eitilt, line 947: I'm curious about what cases we'd get a `None` (and why it's only checked after the fact), but given that addressing that would be a larger refactor I'm fine with this.

ag-eitilt (approved):

PR #1908, opened 2026-09-15
Make FUSE persist workspace mappings in a recovery manifest

AbrarQuazi:
Persist each FUSE job's final workspace-output mapping as a versioned recovery manifest under `.build/cas/staging/recovery/`.

The manifest retains enough information to recover staged regular files, symlinks, and directories after the daemon's in-memory job state is gone. This differs from result.json, which is sent to the current Wake/wakebox process as a protocol for normal finalization. The recovery manifest stays in the workspace so a later command can recover outputs if that process is interrupted.


share/wake/lib/system/runner.wake
```diff
@@ -735,9 +735,23 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
                         content // `staging_root`
                         | getJString
 
+                    def recoveryManifest =
```
ngraybeal, line 738: Is it worth considering a single dir hierarchy under `.build/cas/staging` to help consolidate files and then we can dump the manifest in there. Then we could avoid having a separate dir?

tools/fuse-waked/main.cpp
```diff
@@ -438,6 +450,135 @@ void Job::parse() {
...
+  int mkdir_error = mkdir_with_parents(recovery_dir, 0755);
+  if (mkdir_error != 0) {
+    fprintf(stderr, "fuse-waked: create recovery directory '%s': %s\n", recovery_dir.c_str(),
+            strerror(mkdir_error));
+    return false;
+  }
+  struct timespec now;
+  clock_gettime(CLOCK_REALTIME, &now);
+  const std::string name = std::to_string(getpid()) + "-" + job_id + "-" +
+                           std::to_string(now.tv_sec) + "-" + std::to_string(now.tv_nsec) + ".json";
```
ngraybeal, line 510: I'm not sure there's a good reason to keep multiple snapshots of the manifest, so i'd be in favor of dropping timestamps in the file name.

ngraybeal (approved):
Offline discussions have lead to an agreement my questions can be addressed in a follow up as we shape the staging dir.

share/wake/lib/system/runner.wake
ag-eitilt: Agreed with the offline discussion: having per-job subdirs in `.build/cas/staging` containing both the actual staging files and collateral such as these recovery files (or some similar layout) would be nice, but it's not critical to implement in this particular PR stack.  We'll revisit the staging organization in the future.

tools/fuse-waked/main.cpp
ag-eitilt: Yeah, I'd be concerned about clobbering between jobs, but since the `job_id` is part of the name (in the future: part of the `.build/cas/staging` subdirectory) I'd be surprised if we wound up trying to create the same file from the same wakebox at the same time.  And even then handling failures to get an exclusive write lock should be protective enough.

Pretty sure that's what we decided on in yesterday's(?) meeting as well.

ag-eitilt (commented):
I'm not fully sure how the control flow triggers this (likely just didn't look close enough), but the code looks good.  The one question I have before merging is whether we want to remove the file timestamps now -- we agreed that we don't need to rethink the staging dir as a whole until this stack settles, but the recovery files are a new addition and it feels like now's a good time to simplify that out.

share/wake/lib/system/runner.wake
```diff
@@ -397,6 +397,8 @@ export tuple WakeboxSpec =
     export Version: Option String
     export Runner: Option String
     export CasDir: String
+    export WakeRunId: Option Integer
+    export WakeJobId: Option Integer
```
ag-eitilt, line 401: This is an interesting question.  I assume they're `Option` to handle manual invocations outside of Wake's run+job framework, but those sandboxes should still have *some* ID.  Is it just something that we'll never wind up encountering here because this is only concerned about instantiating the Wake pathway?

share/wake/lib/system/runner.wake
```diff
@@ -604,6 +608,8 @@ export def makeJSONRunner ((JSONRunnerPlan rawScript extraArgs extraEnv estimate
             (Some version)
             (Some script)
             primCasDir
+            (Some currentWakeRunId)
+            (Some (getJobId job))
```
ag-eitilt, line 612: If you want to get rid of one level of parentheses:
```suggestion
            (Some job.getJobId)
```

src/runtime/database.cpp
```diff
@@ -943,6 +943,8 @@ void Database::prepare(const std::string &cmdline) {
   end_txn();
 }
 
+long Database::current_run_id() const { return imp->run_id; }
```
ag-eitilt, line 946: I'm only slightly surprised that we haven't needed this in wakelang before, but more surprised that we haven't needed this function in C++ before.  I guess we've just accessed the field directly in the past?

src/runtime/database.h
```diff
@@ -166,8 +166,10 @@ struct Database {
   void entropy(uint64_t *key, int words);
 
   void prepare(const std::string &cmdline);  // prepare for job execution
-  void finish_run();                         // mark run as complete (sets end_time)
-  void clean();                              // finished execution; sweep stale jobs
+  // Returns the current WakeDB run ID after prepare() has created it.
+  long current_run_id() const;
+  void finish_run();  // mark run as complete (sets end_time)
+  void clean();       // finished execution; sweep stale jobs
```
ag-eitilt, line 172: Maybe make the docstring more terse so that the right-aligned formatting doesn't get messed up?  (This is why I don't like trying to align *anything* more than one or two spaces, rather than just using separate lines.)

src/wakefs/fuse.cpp
```diff
@@ -124,6 +124,10 @@ bool json_as_struct(const std::string &json, json_args &result) {
   if (result.cas_dir.empty()) {
     result.cas_dir = ".build/cas";
   }
+  if (result.cas_dir != ".build/cas") {
+    std::cerr << "cas-dir must be exactly .build/cas" << std::endl;
+    return false;
+  }
```
ag-eitilt, line 130: It seems unfortunate to lock this down here, when the rest of the mechanisms that allow it to adapt to other layouts (or remote CASs) are still in place.  What were you running into which needed it to be this path and *only* this path?

src/wakefs/fuse.cpp
```diff
@@ -142,6 +146,25 @@ bool json_as_struct(const std::string &json, json_args &result) {
...
+      result.wake_run_id = std::stol(wake_run_id.value);
+      result.wake_job_id = std::stol(wake_job_id.value);
+    } catch (const std::exception &e) {
+      std::cerr << "wake_run_id and wake_job_id must be integer values: " << e.what() << std::endl;
+      return false;
+    }
+  } else if (wake_run_id.kind != JSON_NULLVAL) {
+    std::cerr << "wake_run_id and wake_job_id must be integer values" << std::endl;
+    return false;
+  }
```
ag-eitilt, line 166: I feel like there's a better way of running through these conditionals.  Maybe:
```suggestion
  if (wake_run_id.kind != JSON_NULLVAL) {
    if (wake_job_id.kind == JSON_NULLVAL) {
      std::cerr << "wake_run_id and wake_job_id must be provided together" << std::endl;
      return false;
    }
    try {
      result.wake_run_id = std::stol(wake_run_id.value);
      result.wake_job_id = std::stol(wake_job_id.value);
    } catch (const std::exception &e) {
      std::cerr << "wake_run_id and wake_job_id must be integer values: " << e.what() << std::endl;
      return false;
    }
  } else if (wake_job_id.kind != JSON_NULLVAL) {
    std::cerr << "wake_run_id and wake_job_id must be provided together" << std::endl;
    return false;
  }
```
...though that still leaves two pathways to the "provided together" message.  At least it doesn't print "must be provided together" if they both occur but one happens to be a string or something.

tests/wakebox/cas-file-write/pass.sh
```diff
@@ -4,12 +4,13 @@
...
 # 2. File content can be read back immediately
 # 3. staging_files output contains the file with type "file" and a staging_path
+# 4. the result hands off an atomically written final recovery manifest
 set -eu
 
 export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
 
 STATS_FILE=$(mktemp)
-trap 'rm -f test_output.txt "$STATS_FILE"' EXIT
+trap 'rm -rf test_output.txt "$STATS_FILE" .build/cas/staging/recovery' EXIT
```
ag-eitilt, line 13: Slightly leery about `$STATS_FILE` being hijacked to remove an entire directory tree unintentionally, but I can't figure out an actual attack that would make that possible without editing the file.  And this is just a test case, anyway, not something we'd expect people to be running as a standard course of action.

tools/fuse-waked/main.cpp
```diff
@@ -331,6 +335,25 @@ void Job::parse() {
     return;
   }
 
+  JAST run_id = jast.get("wake_run_id");
+  JAST job_id = jast.get("wake_job_id");
+  if (run_id.kind != job_id.kind) {
```
ag-eitilt, line 340: Same conditional restructuring here.

tools/fuse-waked/main.cpp
```diff
@@ -438,6 +460,141 @@ void Job::parse() {
...
+
+  const std::string recovery_dir = g_staging_dir + "/recovery";
+  int mkdir_error = mkdir_with_parents(recovery_dir, 0755);
+  if (mkdir_error != 0) {
+    fprintf(stderr, "fuse-waked: create recovery directory '%s': %s\n", recovery_dir.c_str(),
+            strerror(mkdir_error));
+    return false;
+  }
+  struct timespec now;
+  clock_gettime(CLOCK_REALTIME, &now);
```
ag-eitilt, line 518: Maybe a more specific name?  "Now" stops being accurate pretty quickly. 😉

tools/fuse-waked/main.cpp
```diff
@@ -438,6 +460,141 @@ void Job::parse() {
...
+      wake_run_id
+          ? "run-" + std::to_string(*wake_run_id) + "-job-" + std::to_string(*wake_job_id) + ".json"
+          : "wakebox-" + std::to_string(getpid()) + "-" + job_id + ".json";
+  const std::string final_path = recovery_dir + "/" + name;
+  const std::string temporary_path = recovery_dir + "/." + name + ".tmp";
+
+  JAST manifest(JSON_OBJECT);
+  // Increment this for incompatible manifest schema changes; materializers must reject unknown
+  // versions.
+  manifest.add("version", 1);
```
ag-eitilt, line 529: Preferably we wind up extracting this into something mirroring the database `schema.h`, with whatever degree of backwards-compatible read logic we wind up supporting also specified somewhere other than 525 lines down a monolithic file.  Not something that's really going to become relevant until we actually update this version, though.

ag-eitilt (approved):
Some cleanup (and one bug in error message printing) but otherwise this looks good to me.

share/wake/lib/system/runner.wake
AbrarQuazi: In fuse, they do have an ID called `job-key` which is different than the wakeDB's jobID, but I dont believe it is persisted into the result.json or anything. Maybe we should consolidate both, but that can be a clean up PR seperate from here. 

Wakeboxes not supplied a WakeRunId or JobId (such as the interactive case), create their recovery files via: `wakebox-<daemon-pid>-<job-key>.json`

src/runtime/database.cpp
AbrarQuazi: Yeah, I believe we only use it in the GC and --ps logic so far, and both of them directly call the database table

src/wakefs/fuse.cpp
AbrarQuazi: So for materialization for wakebox, I need the workspace to be the current working directory of where `wakebox` is invoked, so it needs to look at a specific locked down path to find the recovery files, unless we make wakebox supply an extra argument telling where its cas staging files and recovery files are, and where it should materialize into. This wouldnt have been needed if the file entries in the recovery files were absolute paths (which I recently changed to relative to work with the remote execution case).

We can revisit this once more when we actually plan to make remote CAS's. For that situation, I think we can split out the `.build/cas/staging` directory seperately away from the cas directory, so we can avoid this.

ag-eitilt: I'm still not entirely clear on what logic requires the paths to be limited (and particularly to be specified in the manifest file), but I agree that it's a relatively minor detail that we can explore in the future.

PR #1909, opened 2026-09-15
Consolidate materialization library into wcl and use fd-based api's

AbrarQuazi:
Add shared materialization helpers in wcl and move existing CAS and staged-output paths to use them.

The new FD-based APIs materialize files using file descriptors that have already been opened and validated, rather than reopening paths during the operation. This keeps the source file and destination directory fixed for the whole copy, preventing a path from being replaced by a symlink or different file between validation and use.

The helpers also centralize atomic file replacement, reflink-or-copy behavior, symlink creation, directory handling, and metadata application. This removes duplicated materialization logic and provides the safe primitives needed by staging recovery.


ngraybeal (approved):
The creation of the new library make sense.

src/runtime/cas_prim.cpp
```diff
@@ -63,47 +58,8 @@ bool parse_hash_string(const std::string& id, cas::ContentHash& out, std::string
 }
 
 // Unique temp path for atomic rename. PID + counter avoids races across concurrent processes.
-static std::string make_temp_path(const std::string& dest) {
-  return dest + "." + std::to_string(getpid()) + "." +
-         std::to_string(g_staged_materialize_counter.fetch_add(1));
-}
-
```
ag-eitilt, line 70: Comment should be deleted as well.

src/wcl/file_ops.cpp
```diff
@@ -41,31 +41,55 @@
...
+result<CopyResult, posix_error_t> reflink_or_copy_fd(int src_fd, int dst_fd, bool attempt_reflink) {
+#ifdef HAS_FICLONE
+  if (attempt_reflink && ioctl(dst_fd, FICLONE, src_fd) == 0)
+    return make_result<CopyResult, posix_error_t>(CopyResult{CopyStrategy::Reflink, 0});
+  if (attempt_reflink && errno != EOPNOTSUPP && errno != ENOTTY && errno != EINVAL &&
+      errno != EXDEV)
+    return make_errno<CopyResult>();
+#else
+  (void)attempt_reflink;
+#endif
```
ag-eitilt, line 55: What's the benefit of casting the bool to a void for a no-op on this branch?

src/wcl/file_ops.cpp
```diff
@@ -41,31 +41,55 @@
...
+    ssize_t offset = 0;
+    while (offset < read_bytes) {
+      ssize_t written = write(dst_fd, buffer + offset, static_cast<size_t>(read_bytes - offset));
+      if (written < 0) {
+        if (errno == EINTR) continue;
+        return make_errno<CopyResult>();
+      }
+      offset += written;
+    }
+    copied += static_cast<size_t>(read_bytes);
```
ag-eitilt, line 74: Can this be simplified by calling your `write_all` from #1908?

src/wcl/file_ops.cpp
```diff
@@ -78,46 +102,31 @@ result<bool, posix_error_t> try_reflink(const std::string& src, const std::strin
...
+    if (reflink)
       return make_result<CopyResult, posix_error_t>(CopyResult{CopyStrategy::Reflink, 0});
-    }
-
-    // Only fall back if reflink is not supported
-    int reflink_err = reflink_result.error();
-    if (reflink_err != EOPNOTSUPP && reflink_err != ENOTTY && reflink_err != EINVAL &&
-        reflink_err != EXDEV) {
-      return make_error<CopyResult, posix_error_t>(reflink_err);
-    }
```
ag-eitilt, line 93: Why'd you remove the comments in this function?  On the whole, I'm noticing a few helpful renames through it and some good `const` additions, but no difference in logic and a number of formatting changes which just make the code denser (including other name changes which *don't* serve to clarify things).

src/wcl/file_ops.h
```diff
@@ -34,14 +34,17 @@ struct CopyResult {
...
+result<CopyResult, posix_error_t> reflink_or_copy_fd(int src_fd, int dst_fd,
+                                                     bool attempt_reflink = true);
+
+// Tries reflink first, then falls back to std::filesystem::copy_file.
 result<CopyResult, posix_error_t> reflink_or_copy_file(const std::string& src,
                                                        const std::string& dst, mode_t mode,
                                                        bool attempt_reflink = true);
 
-// Try to reflink a file (copy-on-write clone)
+// Try to reflink a file (copy-on-write clone) into an exclusive destination.
```
ag-eitilt, line 47: Helpful note, but what's the lifespan on the exclusivity limit?

src/wcl/materialize.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <filesystem>
+
+#include "compat/nofollow.h"
+
+namespace fs = std::filesystem;
+
+namespace wcl {
+namespace {
+
+std::atomic<uint64_t> materialize_counter{0};
```
ag-eitilt, line 32: Do namespace-scoped variables not need to be declared `static`, then?

src/wcl/materialize.cpp
```diff
@@ -0,0 +1,193 @@
...
+  // forcing the file's timestamp to the Unix epoch.
+  if (seconds == 0 && nanoseconds == 0) return true;
+  struct timespec times[2] = {{0, UTIME_OMIT}, {seconds, nanoseconds}};
+  return futimens(fd, times) == 0;
+}
+
+result<int, posix_error_t> open_destination_parent(const std::string& destination,
+                                                   std::string* name) {
+  fs::path path(destination);
+  *name = path.filename().string();
```
ag-eitilt, line 57: Seems like the contract is that this extracts the final part of `destination` and returns it to the caller via `name`, but it could really use a docstring confirming that.  Specifically, that this isn't "open a fd corresponding to the file named `name` within the directory `destination`".  And also that the `int` being returned is a file descriptor and not some random other number.

src/wcl/materialize.cpp
```diff
@@ -0,0 +1,193 @@
...
+  const std::string temporary = temporary_name();
+  int destination = openat(destination_parent_fd, temporary.c_str(),
+                           O_WRONLY | O_CREAT | O_EXCL | O_CLOEXEC, mode & 07777);
+  if (destination < 0) return make_errno<CopyResult>();
+  auto copy = reflink_or_copy_fd(src_fd, destination, attempt_reflink);
+  int failure = copy ? 0 : copy.error();
+  if (copy && fchmod(destination, mode & 07777) != 0) failure = errno;
+  // Apply mtime before replacement so observers never see a new file with stale metadata.
+  if (copy && !set_file_mtime(destination, mtime_sec, mtime_nsec)) failure = errno;
+  if (close(destination) != 0 && failure == 0) failure = errno;
```
ag-eitilt, line 82: Shouldn't each of these also be guarded by the previous failure state `== 0`?

src/wcl/materialize.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+result<bool, posix_error_t> apply_directory_metadata(int directory_fd, mode_t mode,
+                                                     time_t mtime_sec, long mtime_nsec) {
+  if (fchmod(directory_fd, mode & 07777) != 0) return make_errno<bool>();
+  if (mtime_sec == 0 && mtime_nsec == 0) return make_result<bool, posix_error_t>(true);
+  struct timespec times[2] = {{0, UTIME_OMIT}, {mtime_sec, mtime_nsec}};
+  if (futimens(directory_fd, times) != 0) return make_errno<bool>();
+  return make_result<bool, posix_error_t>(true);
+}
```
ag-eitilt, line 178: This doesn't have the `(0, 0)` exclusion that was implemented for files (nor, for that matter, do symlinks).  That should probably be consistent -- I'm not familiar enough with the standard functions to know if we can implement a single polymorphic mtime-setting function that doesn't care about what the underlying type of the fd is, but even if it does have to be duplicated code/branches it's still better than `(0, 0)` only working in one specific case.

ag-eitilt (commented):
Logic largely looks good, but there's enough cleanup here I'm not going to approve it quite yet.

src/wcl/materialize.cpp
```diff
@@ -0,0 +1,193 @@
...
+result<CopyResult, posix_error_t> materialize_regular_file_at(int src_fd, int destination_parent_fd,
+                                                              const std::string& destination_name,
+                                                              mode_t mode, time_t mtime_sec,
+                                                              long mtime_nsec,
+                                                              bool attempt_reflink) {
+  // Copy to a unique sibling first so rename publishes either the complete file
+  // or the previous destination, never a partially materialized file.
+  const std::string temporary = temporary_name();
+  int destination = openat(destination_parent_fd, temporary.c_str(),
+                           O_WRONLY | O_CREAT | O_EXCL | O_CLOEXEC, mode & 07777);
```
ag-eitilt, line 75: I'm not familiar with the intricacies of `O_CLOEXEC`.  Does that remove the temporary file if whatever binary this is linked into receives a sigterm?  I wasn't seeing anything else that would clean this up in that situation.

src/wcl/file_ops.cpp
AbrarQuazi: mainly for compiler warnings, since machines without reflink support will not check `attempt_reflink`, so to avoid the compiler warning of the variable not being used.

src/wcl/file_ops.cpp
AbrarQuazi: Good catch, these changes are not intended. I will changes these back to what it was in master

src/wcl/materialize.cpp
AbrarQuazi: Its mainly to prevent a situation to prevent leaky file descriptors, I dont really see why we need it specifically here, but it doesnt hurt to have and protect against.

I dont think we should add complex logic now to handle this file cleanup during a sigterm. If we see it ends up being an issue, we can revisit

src/wcl/materialize.cpp
AbrarQuazi: its actually there for all three, but Ill add comments to all of them to highlight this fact

ag-eitilt: Ah, I was getting thrown off by the file case having a separate function *just* for the `(0, 0)` checks, while directories roll it in alongside the mode and symlinks have the check inline.

src/wcl/materialize.cpp
ag-eitilt: Fair enough, I'm just a bit concerned given that this has the potential for duplicating (and after termination, losing track of) big files.  You're right that that's probably not a likely enough scenario to be preemptively concerned about.

ag-eitilt (approved):
Thanks!

PR #1911, opened 2026-09-15
Add staging recovery manifest materialization library

AbrarQuazi:
Add the versioned staging-recovery manifest parser, validator, and completed-workspace materializer.

The materializer safely restores a completed FUSE job's outputs into its recorded workspace and consumes staging sources only after a persisted recovery progress is recorded.


src/wakefs/materialize_staging.h
```diff
@@ -0,0 +1,79 @@
...
+struct StagingManifest {
+  std::string workspace_root;
+  std::string cas_staging_root;
+  std::string job_key;
+  int64_t created_at_ns = 0;
+  int64_t daemon_pid = 0;
+  std::vector<StagingEntry> entries;
+};
+
+// Parse and validate the version-1 final-staging manifest schema. This validates
```
ngraybeal, line 47: NIT: Specific reference to version number should be dropped

ngraybeal (approved):

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+
+bool fail(std::string* error, const std::string& message) {
+  if (error) *error = message;
+  return false;
+}
+
+std::string errno_message(const std::string& action) { return action + ": " + strerror(errno); }
+
+bool is_safe_relative_path(const std::string& path) {
+  if (path.empty() || path.front() == '/' || path.find('\0') != std::string::npos) return false;
```
ag-eitilt, line 52: Should we use a `path` type and let it handle some of the processing for us?

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+    size_t end = path.find('/', begin);
+    if (end == std::string::npos) end = path.size();
+    const std::string component = path.substr(begin, end - begin);
+    if (component.empty() || component == "." || component == "..") return false;
+    begin = end + 1;
+  }
+  return path.back() != '/';
+}
+
+bool parse_integer(const JAST& json, int64_t* value) {
```
ag-eitilt, line 64: I know this is wakefs and not wake itself, but surely we can reference the main parser for this?  Or is it still too intertwined with the rest of the runtime?

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+bool required_boolean(const JAST& object, const char* key, bool* value, std::string* error) {
+  auto child = object.get_opt(key);
+  if (!child) return fail(error, std::string("manifest field '") + key + "' must be a boolean");
+  auto parsed = (*child)->expect_boolean();
+  if (!parsed) return fail(error, std::string("manifest field '") + key + "' must be a boolean");
+  *value = *parsed;
+  return true;
+}
+
+bool has_only_fields(const JAST& object, std::initializer_list<const char*> allowed,
```
ag-eitilt, line 103: Is this required?  We're already versioning the manifest; doesn't that imply that if we see a field we don't expect, but for a version number that we do, then it is safely backwards-compatible to ignore the field?

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+    if (!known) return fail(error, "manifest contains an unknown field: " + child.first);
+  }
+  return true;
+}
+
+// Validate the serialized manifest contract before trusting any paths or metadata.
+bool validate_metadata(const StagingManifest& manifest, std::string* error) {
+  if (manifest.workspace_root != ".") return fail(error, "workspace_root must be exactly .");
+  if (manifest.cas_staging_root != ".build/cas/staging")
+    return fail(error, "cas_staging_root must be exactly .build/cas/staging");
```
ag-eitilt, line 122: I've already commented on a previous PR about requiring these fields be set to specific paths.  Among other things, we'd expect the `cas_staging_root` to be set to a subdirectory of this in the near future.  Admittedly not sure what an alternate `workspace_root` would imply though.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+        (entry.type != StagingEntryType::Symlink && !entry.target.empty()))
+      return fail(error, "entry target is invalid");
+    if ((entry.type == StagingEntryType::File || entry.type == StagingEntryType::Directory) &&
+        (entry.mode & ~07777) != 0)
+      return fail(error, "entry mode has unsupported bits");
+  }
+  return true;
+}
+
+bool write_all(int fd, const std::string& data) {
```
ag-eitilt, line 149: This is indeed a better place for this generic helper than `main.cpp`, but do we have any sort of utils file which would be even better?

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+  size_t slash = path.rfind('/');
+  if (slash == std::string::npos) {
+    parent->clear();
+    *leaf = path;
+  } else {
+    *parent = path.substr(0, slash);
+    *leaf = path.substr(slash + 1);
+  }
+  return !leaf->empty();
+}
```
ag-eitilt, line 223: This one *really* needs a docstring.  Is it just finding the final path component and the directory that's in?  If so, another call for a proper `path` type.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+  split_parent(entry.staging_path, &source_parent_path, &source_leaf);
+  split_parent(entry.destination, &target_parent_path, &target_leaf);
+  int source_parent = -1;
+  int target_parent = -1;
+  if (!open_relative_directory(source_rootfd, source_parent_path, false, &source_parent, error))
+    return false;
+  if (!open_relative_directory(target_rootfd, target_parent_path, true, &target_parent, error)) {
+    close(source_parent);
+    return false;
+  }
```
ag-eitilt, line 238: Do these need error messages?

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+    close(source_parent);
+    return fail(error, errno_message("open staging source " + entry.staging_path));
+  }
+  struct stat source_stat;
+  if (fstat(source, &source_stat) != 0 || !S_ISREG(source_stat.st_mode)) {
+    close(source);
+    close(target_parent);
+    close(source_parent);
+    return fail(error, "staging source is not a regular file: " + entry.staging_path);
+  }
```
ag-eitilt, line 251: I'm not sure if the inversion would actually result in an easier read, but it *feels* like there should be a way of nesting all these `if`s to avoid having to carefully ensure everything gets closed properly.  Or using C++ destructors (a finalize-pointer type?) to do so automatically when they fall out of scope.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+    errno = copy.error();
+    return fail(error, errno_message("copy staging source " + entry.staging_path));
+  }
+  close(target_parent);
+  close(source_parent);
+  return true;
+}
+
+// After an interrupted placement, a missing source is safe to accept only when
+// its exact destination is already a regular file below non-symlink parents.
```
ag-eitilt, line 268: Oh, I think I misunderstood this comment.  Is it talking about files listed in the manifest but which no longer exist in the staging directory because a previous materialization was interrupted after it started deleting the now-fully-reflinked materialized outputs (which it would only be doing after *every* staging file had been materialized), but before deleting the manifest file?  In that case, printing a warning seems reasonable so long as the rest of the files are still processed without aborting.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+  int parent = -1;
+  if (!open_relative_directory(target_rootfd, parent_path, false, &parent, error)) return false;
+  struct stat destination_stat;
+  const bool present = fstatat(parent, leaf.c_str(), &destination_stat, AT_SYMLINK_NOFOLLOW) == 0;
+  if (!present) {
+    const int saved = errno;
+    close(parent);
+    if (saved == ENOENT) return false;
+    errno = saved;
+    return fail(error, errno_message("inspect materialized destination " + entry.destination));
```
ag-eitilt, line 281: What does this message actually look like by the time the user encounters it?  I'm not seeing anything in the stack which adds a `"failed to " + ...` prefix, and I'm worried that it might be read as an instruction to the user for how to fix the error.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+    const int saved = errno;
+    close(parent);
+    errno = saved;
+    return fail(error, errno_message("inspect staging source " + entry.staging_path));
+  }
+  if (!S_ISREG(source_stat.st_mode)) {
+    close(parent);
+    return fail(error, "staging source is not a regular file: " + entry.staging_path);
+  }
+  if (unlinkat(parent, leaf.c_str(), 0) != 0 && errno != ENOENT) {
```
ag-eitilt, line 356: We shouldn't be doing this here.  One of the things I was trying to walk through over the call was the order in which the materialization and removals happen.  I *strongly* believe that we should materialize every file in the manifest before we start removing any of them.  That gives us the most consistent behaviour between materialization calls[^1] and makes for the cleanest separation of concerns between "materialization" and "removal".  It's still not perfectly atomic, but it's atomic *enough* so long as the two aren't intermixed.

[^1]: If we delete each staging file opportunistically, then we can get into a situation where one materialization is aborted (perhaps due to Wake timing it out), a second file is materialized or otherwise created at the same path as one of the files from the aborted materialization (perhaps due to CI queuing multiple Wake runs in sequence), and the first sandbox now being impossible to debug without rerunning the whole thing because even if the materialization is manually restarted we've lost the staging file to rematerialize back overtop the one that replaced it.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+  }
+  close(parent);
+  return true;
+}
+
+bool entry_before(const StagingEntry& left, const StagingEntry& right) {
+  return left.destination.size() < right.destination.size() ||
+         (left.destination.size() == right.destination.size() &&
+          left.destination < right.destination);
+}
```
ag-eitilt, line 370: This isn't accurate.  `entry_before("a/long/path.ext", "b/short.ext") == false`  Unless you don't care about bouncing around the directory tree just so long as parent/child order is maintained (in which case this needs a better name and docstring).

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+  if (!canonical_existing_directory(current, &workspace, error)) return false;
+
+  int workspace_fd = open(workspace.c_str(), O_RDONLY | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);
+  if (workspace_fd < 0) return fail(error, errno_message("open current workspace"));
+  int build_fd = -1;
+  int cas_fd = -1;
+  int staging_fd = -1;
+  if (!open_directory_at(workspace_fd, ".build", &build_fd, error) ||
+      !open_directory_at(build_fd, "cas", &cas_fd, error) ||
+      !open_directory_at(cas_fd, "staging", &staging_fd, error)) {
```
ag-eitilt, line 391: Okay, I see why you hard-failed if the path wasn't specifically this.  That doesn't mean hard-coding these directories is a good idea.  This function should be generic over any staging root path (optional, temporary constraint: ...within the workspace).

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+  for (StagingEntrySummary& entry : summary->entries) {
+    if (entry.destination == destination) return &entry;
+  }
+  return nullptr;
+}
+
+}  // namespace
+
+// Parse a versioned manifest and reject schema or lexical-safety violations.
+bool parse_staging_manifest(const std::string& text, StagingManifest* manifest,
```
ag-eitilt, line 415: Same comment as https://github.com/sifiveinc/wake/pull/1908#discussion_r4077502768: this should be being handled by a shared library/file.  Only now you've got the reads and the writes being defined not just in different files but different *tools*, so I'm a lot less enthused about putting off extracting them than I was on that PR.

src/wakefs/materialize_staging.cpp
```diff
@@ -0,0 +1,673 @@
...
+        if (entry.type != StagingEntryType::Directory) continue;
+        std::string metadata_error;
+        if (!materialize_directory(workspace_root, entry, true, &metadata_error))
+          record_failure(entry.destination, metadata_error);
+      }
+    }
+    if (summary->failed == 0) {
+      manifest.materialization_complete = true;
+      std::string checkpoint_error;
+      if (!write_staging_manifest_atomic(manifest_path, manifest, &checkpoint_error))
```
ag-eitilt, line 643: What's the purpose of this file?  If we succeeded, we succeeded; the only thing I'd expect us to do at this point is to go on and delete the staging files (see above).  For that matter, the only thing I'd expect us to do if we failed to materialize something -- except for "staging file is already gone" errors -- is to *not* delete the staging files.

ag-eitilt (changes requested):
The majority of the flow seems fine, but I have enough architectural hesitations that I'm marking this as "changes requested" until we talk them through.

src/wakefs/materialize_staging.cpp
AbrarQuazi: Workspace root is the location where you will be materializing from, where you can expect your root of your build outputs to be (in this case it is where wakebox is called from), so that we can know exactly which path to materialize your output since the `.build/cas/staging/recovery` is relative paths

src/wakefs/materialize_staging.cpp
AbrarQuazi: so this function is only called when all of materialization succeeds (not for an individual staged file, but the entire recovery manifest) below: 
https://github.com/sifiveinc/wake/pull/1911/changes#diff-3d29f0113308ef9aba70ff71cfc2f180dd32cf8f0e4f5424abb6733e615658fcR648-R653


```
  if (summary->failed == 0 && manifest.materialization_complete) {
    for (const StagingEntry& entry : manifest.entries) {
      if (entry.type != StagingEntryType::File) continue;
      std::string cleanup_error;
      if (!consume_regular_source(source_root, entry, &cleanup_error)) {
        record_failure(entry.destination, cleanup_error);
```

So what happens with the current logic is that after materialization succeeds, we atomically update the boolean flag `materialization_complete` in the JSON. If Wake/wakebox terminates before making this atomic update, that means that materialization is incomplete, and the next try will go through the loop of materializing everything again. If does happen to atomically update this bool before being terminated, wakebox knows it can just clean up the staging files and the recovery file

src/wakefs/materialize_staging.cpp
AbrarQuazi: The point of this file is to update the boolean marker `materialization_complete` in the JSON file. Otherwise we don't really know what state the recovery files or staging files is on. Without this, Wake does not know if recovery was complete, maybe Wake terminates while deleting staging files - so materialization will have to start all over again without it (which can take a long time for some jobs) and has to stat the workspace to check if the deleted staging file entries still exist and were actually materialized.

With this automically updated flag, wakebox knows it just has to clean the files because we know for a fact everything was materialized previously

src/wakefs/materialize_staging.cpp
AbrarQuazi: I thought about this for a bit, I dont think we are gaining much by creating a new relative path type. It would be helpful if we added a new path type as part of `StagingEntry` instead of using strings, so a lot of the code here could be simplified, but that would mean putting processing in `fuse-waked` just to use the `StagingEntry` class, which seems like extra unecessary work.

Im going to keep these as functions for now, until we need to revisit and come up with a better design.

src/wakefs/materialize_staging.cpp
AbrarQuazi: I can use the JSON libraries version of this

src/wakefs/materialize_staging.cpp
AbrarQuazi: the wcl library has this now, so deleting this version here

src/wakefs/materialize_staging.cpp
AbrarQuazi: just using std::filesystem path here to make implentation cleaner

src/wakefs/materialize_staging.cpp
AbrarQuazi: the underlying functions report out the error

src/wakefs/materialize_staging.cpp
AbrarQuazi: luckily we have  `wcl::unique_fd` just for this :)

src/wakefs/materialize_staging.cpp
AbrarQuazi: ah yeah let me add the `failed to`

src/wakefs/materialize_staging.cpp
AbrarQuazi: moving everything to wcl::unique_fd cleaned up a lot of things

src/wakefs/materialize_staging.cpp
AbrarQuazi: I think I will defer this change till later, when we have a better idea the next time around when we have to change the versions of this file

src/wakefs/materialize_staging.cpp
ag-eitilt: Yeah, definitely agree rolling our own would be overkill.  I was mostly thinking about [`std::filesystem::path`](https://en.cppreference.com/cpp/filesystem/path) but kept the language open in case there was some more preferred type among the dependencies we're already pulling in.

src/wakefs/materialize_staging.cpp
ag-eitilt: I guess I was thinking about this as a manifest for the sandbox as a whole rather than for a specific materialization out of that sandbox.  The latter understanding definitely works, but I'm not sure how important it is to capture each materialization individually.  Like, if we did have a hypothetical `--materialize-to`, would we really want to create a whole series of manifest files which only differ in their `workspace_root`?

Separate question: I was thinking about the `cas_staging_root`, and specifically whether it's actually helpful to encode at this level.  Theoretically this file should always live in a known location relative to its staging files -- for now it's in a "nephew" directory; in the future it will be the same per-sandbox subdirectory.  Is there any realistic use case where the logic *wouldn't* be able to just start with the path of this file itself and take a known step or two from there?

Neither of these are blockers for this PR, but I do think we'll want to give serious thought to stripping both fields out.

src/wakefs/materialize_staging.cpp
ag-eitilt: This might be another case where the file is a bit overdesigned.  Yeah, there's a chance the termination happens in the narrow window after materialization completes and before everything has been removed.  Is that really something we expect to have happen often enough with large enough jobs that we don't just eat the cost of rematerializing the subset of files which weren't deleted?

In my mind, the better argument for this flag is not as "materialization is complete" but as "removal has started", to avoid materializing a known-partial view of what had been in the sandbox.  And that *is* a good enough argument that I've talked myself into thinking we keep it.  (Though I'm still not sure it'll be necessary once each sandbox has its own staging subdirectory and we can mark "removal has started" by removing the manifest *before* the rest of the staging files.)

src/wcl/unique_fd.h
```diff
@@ -45,8 +45,11 @@ class unique_fd {
   explicit unique_fd(int fd) : fd(fd) {}
 
   unique_fd& operator=(unique_fd&& f) noexcept {
-    fd = f.fd;
-    f.fd = -1;
+    if (this != &f) {
+      close();
```
ag-eitilt: Good catch.

src/wakefs/materialize_staging.cpp
```diff
@@ -228,39 +207,31 @@ bool materialize_file(int source_rootfd, int target_rootfd, const StagingEntry&
...
-  if (source < 0) {
-    close(target_parent);
-    close(source_parent);
-    return fail(error, errno_message("open staging source " + entry.staging_path));
-  }
+
+  int source_fd =
+      openat(source_parent.get(), source_leaf.c_str(), O_RDONLY | O_NOFOLLOW | O_CLOEXEC);
+  if (source_fd < 0)
+    return fail(error, errno_message("failed to open staging source " + entry.staging_path));
```
ag-eitilt: Come on, clang-format, do these two cases *really* need to have different depths of indentation? 🙄

src/wakefs/materialize_staging.cpp
ag-eitilt: Still wondering about this.

ag-eitilt (approved):
Still have a couple questions about the use-case-level architecture, but my main concerns were misunderstandings or have been addressed.  Thanks!

src/wakefs/materialize_staging.cpp
AbrarQuazi: Yeah sounds good to me, I can make this into filesystem::path too

src/wakefs/materialize_staging.cpp
AbrarQuazi: Yeah sorry, glanced over this. I will go ahead and remove this, dont think its needed

src/wakefs/materialize_staging.cpp
AbrarQuazi: You have good points, I think theres no value adding `workspace root` nor `cas_staging_root` the in the manifest, because currently the materializer will just default call `cwd` and find the `.build/cas/staging` path (so the caller has no choice currently to call from the above parent directory of that) and use `cwd` to materialize from, ignoring the fields in the manifest.

For the question about the future of being in a known location relative to the staging files, and using logic based on that, sure we could do this, but I still think it is cleaner to start at where the user invoked `wakebox` at `cwd`, and start your search from there, rather than going up levels from the staging dir location.

src/wakefs/materialize_staging.cpp
AbrarQuazi: we can revisit this during work for consolidating of the staging directory and cleanup in general.

Issue #427, opened 2020-04-22
Bug in caching strategy with respect to visible files and inputs


richardxia:
The current strategy for determining whether a previous run of a Job can be reused is flawed. Let me start by presenting a demonstration of behavior I find confusing and that I hope we all agree is a "bug".

I created a small Git repository which demonstrates the bug: https://github.com/richardxia/wake-caching-bug (use commit 0678e353d81bfa61fa66911275cb7bc0ebc265bb in case if I add more commits to it in the future). I have confirmed that I see this behavior under both Wake 0.17.2 and Wake 0.18.1.

# Demonstration of bug

We have a simple build that consists of a Makefile, a single input file (`in.txt`), and single build target in the Makefile which copies `in2.txt` to `out.txt`. The `in2.txt` is not a typo and is part of the example, so this Makefile is intentionally broken. This Makefile target is also forced to exit with a status code of 0, which is contrived, but is representative of real life scenarios where you may have faulty false successes. There is a single Wake function defined that runs a `make` command with both `Makefile` and `in.*\.txt` passed in as Visible files.

When you run that Wake function, it "succeeds" but does not create or return the desired output file (`out.txt`).

If you then "fix" the build by renaming `in.txt` to `in2.txt`, and then rerun the Wake function, it will cache the previous run and neither create nor return `out.txt`.

If you then, however, reset the Wake database with `wake --init .` and then rerun the Wake function, this time it _will_ actually create `out.txt` and return the `Path` object corresponding to it.

# Explanation of bug

I don't know the Wake internals very well, but I think what's happening is that the first time you run the Wake function, both Makefile and in.txt are passed in as Visibles, but because the underlying Makefile rule wants to read in2.txt and not in.txt, then the FUSE runner detects that only Makefile is an input to the Job. I confirmed this with `wake -i`.

However, if you rename in.txt to in2.txt, then Makefile and in2.txt are passed in as Visibles, but because the previous run saw that it only needed Makefile as an actual input, Wake thinks that it can reuse the previous run.

I don't think that this is a valid strategy for determining whether a previous Job can be cached, since the presence of new Visibles may actually have an effect of the behavior of a Job.

# Thoughts on what to do

These are just some raw thoughts I have, but I'm not making any specific recommendation on what we should do here.

I understand that the "naive" approach of treating all Visibles as inputs may reduce our performance in the common case, as the addition of an unused file may cause a cache miss, but it would be "safer" and we should consider this as an option.

[...]

PR #1913, opened 2026-09-16
Stop distinguishing job input files from all visible

ag-eitilt:
While it is a very *intriguing* feature, the way the ability to mark certain visible files as read (and thus others as superfluous) had previously been implemented has lead to a number of subtle bugs (e.g. #427), and in our internal runners we've even disabled this entirely via a somewhat-ugly hack.  The general conversations on the team for the past couple years have assumed this feature has been on the way out; I'm just finally getting around to acting on that assumption.

* The `FnInputs` field has been removed from `Plan` and `RunnerInput`, and the processed `Inputs` field from `RunnerOutput`.[^fn-inputs]
* Dropped the `fninputs` from the `jobSignature` hash.  This means that no jobs before this PR will be recognized as a cache hit, and vice versa.  Due to the way functions contribute to the hash, there's no way of retaining cache compatibility unless the `Plan` field *and everything it gets set to* are left in as dead code.
* `getJobInputs` and `getJobFailedInputs` are renamed to `*Visible` to mirror the field name from `Plan` (and obviously now return visible files rather than read inputs).
* Removed inputs from various C++ functions, so that they are now no longer written to the database anywhere while everywhere they were read is collapsed to reference the visible files instead.
  * The exception to this is the `--metadata` query output, where the inputs were removed from the rendering entirely (`-v --metadata` has always been distinguished by printing the visible list (and the stdout/err); `--simple-metadata` remains as not printing out the outputs either).
  * Notably including the `fuse-waked` job JSON dump -- any runners which try to read the `inputs` field will need to be updated.
* The `--input` database filter argument was changed to search through the input files, and something in doing so started triggering SQL ambiguous-column errors; query columns were annotated with `core.` following the existing pattern for explicit names.

[^fn-inputs]: I can see a potential *very long-term* future where we restore the ability for runners to mark which specific files were read, but that's far enough out that it's nothing we should hold onto dead code for.  Even if we do wind up restoring the *behaviour*, there's no telling that we wind up implementing it with the same API.


tools/fuse-waked/main.cpp
```diff
@@ -469,17 +469,7 @@ void Job::dump(const std::string &job_id) {
   bool first;
   std::stringstream s;
 
-  s << "{\"ibytes\":" << ibytes << ",\"obytes\":" << obytes << ",\"inputs\":[";
-
-  for (auto &x : files_wrote) files_read.erase(x);
```
ngraybeal, line 474: Any reason to keep tracking `files_read` at all anymore?

ag-eitilt: I honestly didn't dig further beyond half-remembering that it was used in something like determining writeability/reporting what got written.  But, yeah, that is something I should probably verify (and even if it *is*, see if the logic can be done in some other manner).

src/runtime/database.cpp
```diff
@@ -54,7 +54,7 @@
 #include "wcl/iterator.h"
 
 #define VISIBLE 0
-#define INPUT 1
+// INPUT previously 1; now deprecated
```
ngraybeal, line 57: Any reason not to shift the tree ordering?

If yes, keeping a placeholder comment to identify the skip in sequential ordering make sense.

src/wakefs/fuse.h
```diff
@@ -1,4 +1,4 @@
-/* Wake FUSE launcher to capture inputs/outputs
+/* Wake FUSE launcher to capture file access and outputs
```
ngraybeal, line 1: This one caught my eye since we are only capturing outputs no?

ngraybeal (approved):
Overall LGTM. I'm happy to eject this baggage and revisiting later if needed/wanted.

I'm okay with collapsing things under `visible` instead of `inputs`, even though `inputs` matches parity of `outputs` better.

ag-eitilt:
> I'm okay with collapsing things under visible instead of inputs, even though inputs matches parity of outputs better.

Agreed.  As I said live, I do think it would be nice to start using visible/inputs interchangeably at some point, but I want to let this change settle in people's minds first rather than immediately using a word which has always meant one thing in a way which it was always contrasted against.

src/wakefs/fuse.h
ag-eitilt: In my mind, it's still correct though definitely under a different sense of the word.  After all, Fuse is still redirecting reads (i.e. access in a technical sense) to the underlying CAS path.  Though if you have any synonyms for "capture" and/or "access" that would make it clearer, I'll happily update it.

src/runtime/database.cpp
ag-eitilt: Mostly just ease for us.  `OUTPUT` has always been 2, and shifting it to 1 won't give us any space benefits (SQLite represents boolean columns as integers anyway).  Without any reason *to* make the change, I didn't see a need to require us to retrain our mental models of the DB, nor worry about whether some commit was using the pre- or post-input-dropping numbering.

Plus this way we can avoid a *schema* bump, even if the stored jobs still aren't compatible during retrieval due to the hash change.

share/wake/lib/system/job.wake
```diff
@@ -441,7 +436,7 @@ def stdio job fd =
     prim "job_output" # 1=stdout, 2=stderr; blocks till closed
 
 def tree job typ =
-    prim "job_tree" # 0=visible, 1=input, 2=output; blocks till finished
+    prim "job_tree" # 0=visible, 1=DEPRECATED, 2=output; blocks till finished
```
AbrarQuazi: Why do we have to depricate this instead of just removing it? And when would it be good to go ahead and remove?

src/runtime/job.cpp
```diff
@@ -85,7 +85,7 @@
 #define STATE_RUNNER_OUT 8   // runner_out has been closed and in database
 #define STATE_RUNNER_ERR 16  // runner_err has been closed and in database
 #define STATE_MERGED 32      // exit status in struct
-#define STATE_FINISHED 64    // inputs+outputs+status+runtime in database
+#define STATE_FINISHED 64    // visible+outputs+status+runtime in database
```
AbrarQuazi: I forget how we are coming up with the number here, but I am assuming visible and inputs is one to one, so we dont have to update the number here right?

AbrarQuazi (commented):
The change in itself looks good to me, but I have some questions. Is it possible to keep the fuse layer intact, maybe or maybe not keep the wakelang api's around to extract it, and still have it not affect the jobsignature hash-ing of it for cache hits/misses?

The reason why I say this, is because even though maybe its not used as of now, having a sandbox that tracks reads and writes enables observability. A future use case, that may come sooner than expected, is running AI agents inside the wakebox sandbox - being able to track their observable behavior with both with what was read and wrote into - has been an important use case for sandboxes with AI workflows in general.

Also as a side point, we will need to update our internal repos to reflect these changes as well

PR #1914, opened 2026-09-16
wakebox signal/cancellation handling and materialization logic

AbrarQuazi:
Wakebox now handles external SIGINT and SIGTERM while it owns a FUSE payload.
On the first signal, wakebox:
records cancellation

- -> sends SIGTERM to the payload process group
- -> gives it 5 seconds to exit cleanly
- -> sends SIGKILL on a repeated signal or expired grace period
- -> waits for payload-group writers to exit
- -> disconnects fuse-waked
- -> receives the exact recovery manifest
- -> projects staged outputs into the workspace
- -> retains the manifest and regular staging sources for later materialization
- -> exits nonzero

The payload runs in its own process group, so cancellation reaches its normal descendants without signaling wakebox itself. Wakebox keeps the FUSE daemon live until those writers are gone, preventing the final manifest from being frozen while output is still changing.

Wake's existing child shutdown is adjusted to send one SIGTERM and wait up to 30 seconds before SIGKILL. This gives a wakebox enough time to complete payload shutdown, FUSE finalization, and workspace projection when Wake itself receives Ctrl-C or SIGTERM.

A later explicit:
wakebox --materialize-staging
re-materializes the retained recovery manifests and consumes their staging sources normally.


ngraybeal (approved):
Should we consider adding a print message from wake during a signal event to let users know wake is cleaning up in some fashion? Currently there's no feedback, which might leave users wondering what's going on.

tests/inspection/canceled/pass.sh
AbrarQuazi: Hitting flaky behavior with this test for this PR, so hardening it. There is a couple race conditions with this test, that we often hit in CI with this test. The previous 3 second timeout may kill wake before it processes any job, as well as the fact that immediately calling ` --canceled --simple` after timeout may not make it into the DB yet. 

We are doing a little bit more work during termination now (checking for recovery files), so that might explain the reason for more flakiness with this PR specifically

